PR #38, opened 2019-07-12
Fixing doc gen call

bettinaheim:
A proper unit test will follow shortly - just putting this up in case it's needed today.


anpaz-msft:
these should be merged to `release/v0.8`

anpaz-msft:
hey @bettinaheim, looks like you picked up other changes on master (specifically VS extension) that are not part of 0.8... might need to redo the branch.

PR #39, opened 2019-07-12
Fixing doc gen call

bettinaheim:
Doc gen is looking at all files in the doc gen folder upon generation of the namespace documentation. Hence, running docs generation without cleaning the doc gen folder prior to execution causes a failure. 
Let's have a chat on if/how we want to address that.


alan-geller:
The reason docgen reads the files in the folder is to handle the situation where you have a library namespace that spans multiple projects. Cleaning out the folder is OK for a quick fix, but won't handle this situation in the long term.
My guess is that when docgen goes to write a new file for a specific callable or UDT, it throws if the file already exists. We should probably only delete files when we're going to rewrite them.

bettinaheim:
> The reason docgen reads the files in the folder is to handle the situation where you have a library namespace that spans multiple projects. Cleaning out the folder is OK for a quick fix, but won't handle this situation in the long term.
> My guess is that when docgen goes to write a new file for a specific callable or UDT, it throws if the file already exists. We should probably only delete files when we're going to rewrite them.

For now we are fine, but in terms of usability we might want to be more permissive and just silently replace doc files that already existed. We could make it a warning instead. I'll file a bug, please comment on the bug in terms of requirements (e.g. don't just delete the entire folder prior to execution).

cgranade:
Related: #40.

PR #49, opened 2019-07-16
Rsoiffer/constant propagation

rsoiffer:
Added constant propagation, including evaluation of pure functions at compile time

### DIFF
--- a/src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
+++ b/src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
@@ -37,7 +37,8 @@ type ConstantPropagator(compiledCallables: ImmutableDictionary<QsQualifiedName, 
     member this.undoMarkChanged() =
         changed <- prevChanged
 
-    /// Gets a sorted list of the names of all the constant local variables
+    /// Gets a sorted list of the names of all the constant local variables.
+    /// Used for temporary logging/testing purposes, will be removed.
     member this.getDeclarations =
         declarations |> List.sort |> Seq.ofList
         
--- a/src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
+++ b/src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
@@ -1,33 +1,51 @@
-module CompilerOptimizations
+namespace Microsoft.Quantum.QsCompiler.CompilerOptimization
 
 open System.Collections.Immutable
 open Microsoft.Quantum.QsCompiler.SyntaxExtensions
 open Microsoft.Quantum.QsCompiler.SyntaxTree
 open Microsoft.Quantum.QsCompiler.Transformations.Core
 
-open Utils
-open ExpressionEvaluation
+open Microsoft.Quantum.QsCompiler.CompilerOptimization.Utils
+open Microsoft.Quantum.QsCompiler.CompilerOptimization.ExpressionEvaluation
 
 
-type CP_SyntaxTree(compiledCallables: ImmutableDictionary<QsQualifiedName, QsCallable>) =
+/// The SyntaxTreeTransformation used to evaluate constants
+type ConstantPropagator(compiledCallables: ImmutableDictionary<QsQualifiedName, QsCallable>) =
     inherit SyntaxTreeTransformation()
     let vars = VariablesDict()
+    let cd = {compiledCallables = compiledCallables}
 
     let mutable changed = true
     let mutable prevChanged = false
+    let mutable foundConstants = []
 
+    /// Returns whether the syntax tree has been modified since this function was last called
     member this.checkChanged() =
           let x = changed
           changed <- false
           x
 
+    /// Marks the syntax tree as having changed
     member this.markChanged() =
         prevChanged <- changed
         changed <- true
 
+    /// Ignores the most recent marking of the syntax tree changing
     member this.undoMarkChanged() =
         changed <- prevChanged
 
+    /// Marks the given local variables as constants
+    member this.addFoundConstants(lhs) =
+        match lhs with
+        | VariableName n -> foundConstants <- foundConstants @ [n.Value]
+        | VariableNameTuple t -> t |> Seq.map (fun x -> this.addFoundConstants x) |> List.ofSeq |> ignore
+        | _ -> ()
+
+    /// Gets a sorted list of the names of all the constant local variables
+    member this.getFoundConstants =
+        foundConstants |> List.sort |> Seq.ofList
+        
+    /// The ScopeTransformation used to evaluate constants
     override syntaxTree.Scope = { new ScopeTransformation() with
 
         override scope.Transform x =
@@ -35,13 +53,15 @@ type CP_SyntaxTree(compiledCallables: ImmutableDictionary<QsQualifiedName, QsCal
             let result = base.Transform x
             vars.exitScope()
             result
-
-        override scope.Expression = upcast { new ExpressionEvaluator(vars, compiledCallables, 2) with
+            
+        /// The ExpressionTransformation used to evaluate constant expressions
+        override scope.Expression = upcast { new ExpressionEvaluator(vars, cd, 10) with
             override ee.Transform x =
                 let newX = base.Transform x
                 if x.ToString() <> newX.ToString() then changed <- true
                 newX }
-
+                
+        /// The StatementKindTransformation used to evaluate constants
         override scope.StatementKind = { new StatementKindTransformation() with 
             override statementKind.ExpressionTransformation x = scope.Expression.Transform x
             override statementKind.LocationTransformation x = scope.onLocation x
@@ -52,9 +72,10 @@ type CP_SyntaxTree(compiledCallables: ImmutableDictionary<QsQualifiedName, QsCal
                 let lhs = statementKind.onSymbolTuple stm.Lhs
                 let rhs = statementKind.ExpressionTransformation stm.Rhs
                 if stm.Kind = ImmutableBinding then
-                    if isLiteral rhs.Expression then
+                    if isLiteral rhs.Expression cd then
                         fillVars vars (StringTuple.fromSymbolTuple lhs, rhs.Expression)
-                        printfn "Found constant declaration: %O = %O" (StringTuple.fromSymbolTuple lhs) (prettyPrint rhs.Expression)
+                        syntaxTree.addFoundConstants lhs
+                        // printfn "Found constant declaration: %O = %O" (StringTuple.fromSymbolTuple lhs) (prettyPrint rhs.Expression)
                 QsBinding<TypedExpression>.New stm.Kind (lhs, rhs) |> QsVariableDeclaration
         }
     }
--- a/src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
+++ b/src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
@@ -17,7 +17,7 @@ type ConstantPropagator(compiledCallables: ImmutableDictionary<QsQualifiedName, 
 
     let mutable changed = true
     let mutable prevChanged = false
-    let mutable foundConstants = []
+    let mutable declarations = []
 
     /// Returns whether the syntax tree has been modified since this function was last called
     member this.checkChanged() =
@@ -34,16 +34,9 @@ type ConstantPropagator(compiledCallables: ImmutableDictionary<QsQualifiedName, 
     member this.undoMarkChanged() =
         changed <- prevChanged
 
-    /// Marks the given local variables as constants
-    member this.addFoundConstants(lhs) =
-        match lhs with
-        | VariableName n -> foundConstants <- foundConstants @ [n.Value]
-        | VariableNameTuple t -> t |> Seq.map (fun x -> this.addFoundConstants x) |> List.ofSeq |> ignore
-        | _ -> ()
-
     /// Gets a sorted list of the names of all the constant local variables
-    member this.getFoundConstants =
-        foundConstants |> List.sort |> Seq.ofList
+    member this.getDeclarations =
+        declarations |> List.sort |> Seq.ofList
         
     /// The ScopeTransformation used to evaluate constants
     override syntaxTree.Scope = { new ScopeTransformation() with
@@ -74,7 +67,7 @@ type ConstantPropagator(compiledCallables: ImmutableDictionary<QsQualifiedName, 
                 if stm.Kind = ImmutableBinding then
                     if isLiteral rhs.Expression cd then
                         fillVars vars (StringTuple.fromSymbolTuple lhs, rhs.Expression)
-                        syntaxTree.addFoundConstants lhs
+                        declarations <- declarations @ [sprintf "%O = %O" (StringTuple.fromSymbolTuple lhs) (prettyPrint rhs.Expression)]
                         // printfn "Found constant declaration: %O = %O" (StringTuple.fromSymbolTuple lhs) (prettyPrint rhs.Expression)
                 QsBinding<TypedExpression>.New stm.Kind (lhs, rhs) |> QsVariableDeclaration
         }
--- a/src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
+++ b/src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
@@ -13,10 +13,13 @@ open Microsoft.Quantum.QsCompiler.CompilerOptimization.ExpressionEvaluation
 type ConstantPropagator(compiledCallables: ImmutableDictionary<QsQualifiedName, QsCallable>) =
     inherit SyntaxTreeTransformation()
     let vars = VariablesDict()
-    let cd = {compiledCallables = compiledCallables}
+    let callableDict = {compiledCallables = compiledCallables}
 
-    let mutable changed = true
+    // For determining if constant folding should be rerun
+    let mutable changed = true 
     let mutable prevChanged = false
+    
+    // For logging the constants we found
     let mutable declarations = []
 
     /// Returns whether the syntax tree has been modified since this function was last called
@@ -48,10 +51,10 @@ type ConstantPropagator(compiledCallables: ImmutableDictionary<QsQualifiedName, 
             result
             
         /// The ExpressionTransformation used to evaluate constant expressions
-        override scope.Expression = upcast { new ExpressionEvaluator(vars, cd, 10) with
+        override scope.Expression = upcast { new ExpressionEvaluator(vars, callableDict, 10) with
             override ee.Transform x =
                 let newX = base.Transform x
-                if x.ToString() <> newX.ToString() then changed <- true
+                if x <> newX then changed <- true
                 newX }
                 
         /// The StatementKindTransformation used to evaluate constants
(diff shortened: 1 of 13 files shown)


src/QsCompiler/CompilerOptimizations/Utils.fs
```diff
@@ -0,0 +1,203 @@
...
+
+    override this.ToString(): string =
+        "[" + String.Join("\n", scopeStack |> Seq.map (fun x ->
+            "{" + String.Join(", ", x |> Seq.map (fun y -> y.Key + "=" + y.Value.ToString())) + "}"
+        )) + "]"
+
+        
+/// Represents a (possibly-nested) tuple of strings.
+/// Used as a way to compare QsTuples and SymbolTuples.
+type StringTuple =
```
bettinaheim: You could instead use the ITuple interface, which serves pretty much that purpose. 
Otherwise please make this type private or internal, since making things public means changes to it are breaking changes (i.e. you are implicitly giving at least some stability and maintenance guarantees).
The same goes for a lot of the used data structures here - things that are publicly accessible are things that cannot be changed easily.

src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
```diff
@@ -0,0 +1,74 @@
...
+    member this.markChanged() =
+        prevChanged <- changed
+        changed <- true
+
+    /// Ignores the most recent marking of the syntax tree changing
+    member this.undoMarkChanged() =
+        changed <- prevChanged
+
+    /// Gets a sorted list of the names of all the constant local variables
+    member this.getDeclarations =
```
bettinaheim: Unless I am overlooking something, it seems that this method is mostly used for logging/testing purposes? 
If so, please make it internal, and add the corresponding make internals visible to attribute to the project for testing purposes if needed.

src/QsCompiler/CompilerOptimizations/Utils.fs
```diff
@@ -0,0 +1,203 @@
...
+    | a -> a.ToString()
+
+
+/// Converts a range literal to a sequence of integers
+let rangeLiteralToSeq (r: Expr): seq<int> =
+    match r with
+    | RangeLiteral (a, b) ->
+        match a.Expression, b.Expression with
+        | IntLiteral start, IntLiteral stop ->
+            seq { int start .. int stop }
```
bettinaheim: A blind conversion to int is not a good option here (and the same below) - the language does not behave according to specs otherwise. 
You could either define your own sequence generator (this is e.g. what the simulator runtime does: https://github.com/microsoft/qsharp-runtime/blob/master/src/Simulation/Core/Range.cs), or you could refuse to constant fold if a 32-bit int is not enough. I might prefer the former.

src/QsCompiler/CompilerOptimizations/FunctionEvaluation.fs
```diff
@@ -0,0 +1,145 @@
...
+                        vars.enterScope()
+                        fillVars vars (StringTuple.fromSymbolTuple (fst s.LoopItem), loopValue.Expression)
+                        let result = this.evaluateScope vars s.Body
+                        vars.exitScope()
+                        result) |>
+                    Seq.tryFind isInterrupt |? Normal
+            | _ ->
+                "Unknown IterationValue in for loop: " + (prettyPrint iterValues) |> CouldNotEvaluate
+        | QsWhileStatement s ->
+            Seq.initInfinite (fun _ -> this.evaluateExpression vars s.Condition |> castToBool) |>
```
bettinaheim: Nice!

src/QsCompiler/CompilerOptimizations/Utils.fs
```diff
@@ -0,0 +1,203 @@
...
+
+
+/// Shorthand for a QsExpressionKind
+type Expr = QsExpressionKind<TypedExpression, Identifier, ResolvedType>
+/// Shorthand for a QsTypeKind
+type TypeKind = QsTypeKind<ResolvedType, UserDefinedType, QsTypeParameter, CallableInformation>
+
+
+/// Represents the global dictionary that maps names to callables
+type CallableDict = {
```
bettinaheim: So, what was the intent/purpose of this additional wrapper?

src/QsCompiler/CompilerOptimizations/Utils.fs
```diff
@@ -0,0 +1,203 @@
...
+    | CallLikeExpression ({Expression = Identifier (GlobalCallable qualName, _)}, b)
+        when (cd.getCallable qualName).Kind = TypeConstructor -> isLiteral b.Expression cd
+    | CallLikeExpression (a, b) ->
+        isLiteral a.Expression cd && isLiteral b.Expression cd &&
+            TypedExpression.IsPartialApplication (CallLikeExpression (a, b))
+    | _ -> false
+
+
+/// Helper method to improve expression readability
+let rec prettyPrint (expr: Expr): string =
```
bettinaheim: You could also just generate Q# for the given expression (using the ExpressionKindToQs transformation)?

src/QsCompiler/CompilerOptimizations/Utils.fs
```diff
@@ -0,0 +1,203 @@
...
+        | DiscardedItem -> SingleItem "__discarded__"
+
+    override this.ToString() =
+        match this with
+        | SingleItem item -> item
+        | MultipleItems items -> "(" + String.Join(", ", items) + ")"
+        
+        
+/// Modifies the VariablesDict by setting the given argument tuple to the given values
+let rec fillVars (vars: VariablesDict) (argTuple: StringTuple, arg: Expr): unit =
```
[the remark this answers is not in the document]
bettinaheim: A descriptive error message if the matching fails would also be nice.

src/QsCompiler/CompilerOptimizations/Utils.fs
```diff
@@ -0,0 +1,203 @@
...
+            Seq.fold (fun (cpa, ca) v ->
+                let newPa, newCa = fillPartialArg (v, ca)
+                cpa @ [newPa], newCa
+            ) ([], arg) vt
+        let newPa = wrapExpr (ValueTuple (ImmutableArray.CreateRange newList)) partialArg.ResolvedType.Resolution
+        newPa, newArg
+    | _ -> partialArg, arg
+    
+/// Transforms a partially-application call by replacing missing values with the new arguments
+let partialApplyFunction (baseMethod: TypedExpression) (partialArg: TypedExpression) (arg: TypedExpression): Expr =
```
bettinaheim: This function needs proper and extensive testing!

src/QsCompiler/CompilerOptimizations/Utils.fs
```diff
@@ -0,0 +1,203 @@
...
+    let argsList =
+        if countMissingExprs partialArg = 1 then [arg] else
+        match arg.Expression with
+        | ValueTuple vt ->
+            if countMissingExprs partialArg <> vt.Length
```
bettinaheim: What about nested argument tuples?

src/QsCompiler/CompilerOptimizations/ExpressionEvaluation.fs
```diff
@@ -0,0 +1,244 @@
...
+    override this.onFunctionCall (method, arg) =
+        let method, arg = this.simplify (method, arg)
+        if maxRecursiveDepth > 0 && isLiteral arg.Expression cd then
+            match method.Expression with
+            | Identifier (GlobalCallable qualName, types) ->
+                fe.evaluateFunction qualName arg types |? CallLikeExpression (method, arg)
+            | CallLikeExpression (baseMethod, partialArg) ->
+                this.Transform (partialApplyFunction baseMethod partialArg arg)
+            | _ ->
+                failwithf "Unknown function call: %O" (prettyPrint method.Expression)
```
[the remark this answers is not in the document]
rsoiffer: What do you mean by relying on knowledge of what isLiteral does exactly? It's not in the cope snippet you linked above, and the only thing it's used for in this function is to check whether the argument to a function call is a constant. The only promise isLiteral makes is that if isLiteral(x) is true, then x is a constant value.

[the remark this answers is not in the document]
bettinaheim: You are right. I somehow thought the match was the same as isLiteral - I don't know what code I was reading. So never mind, looks good.

src/QsCompiler/CompilerOptimizations/FunctionEvaluation.fs
```diff
@@ -0,0 +1,145 @@
...
+        | QsValueUpdate s ->
+            match s.Lhs.Expression with
+                | Identifier (LocalVariable name, tArgs) -> vars.setVar(name.Value, this.evaluateExpression vars s.Rhs); Normal
+                | _ -> "Unknown LHS of value update statement: " + (prettyPrint s.Lhs.Expression) |> CouldNotEvaluate
+        | QsConditionalStatement s ->
+            match s.ConditionalBlocks |>
+                Seq.map (fun (ts, block) -> this.evaluateExpression vars ts |> castToBool, block) |>
+                Seq.tryFind (fst >> function ExprValue true -> true | ExprValue false -> false | ExprError _ -> true), s.Default with
+            | Some (ExprError s, _), _ -> CouldNotEvaluate s
+            | Some (ExprValue _, block), _ | None, Value block -> this.evaluateScope vars block.Body
```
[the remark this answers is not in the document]
rsoiffer: I think you're misreading the code. If the evaluator cannot determine whether a case is executed, it will give an ExprError, which the first match case will report as a CoundNotEvaluate. The (None, Value block) represents knowing with certainty that none of the cases are evaluated, so we evaluate the default block instead.

src/QsCompiler/CompilerOptimizations/Utils.fs
```diff
@@ -0,0 +1,203 @@
...
+    | Identifier (GlobalCallable a, _) -> "GlobalCallable " + a.Name.Value
+    | CallLikeExpression (f, x) -> "(" + (prettyPrint f.Expression) + " of " + (prettyPrint x.Expression) + ")"
+    | ValueTuple a -> "(" + String.Join(", ", a |> Seq.map (fun x -> x.Expression) |> Seq.map prettyPrint) + ")"
+    | ValueArray a -> "[" + String.Join(", ", a |> Seq.map (fun x -> x.Expression) |> Seq.map prettyPrint) + "]"
+    | ADD (a, b) -> "(" + (prettyPrint a.Expression) + " + " + (prettyPrint b.Expression) + ")"
+    | a -> a.ToString()
+
+
+/// Converts a range literal to a sequence of integers
+let rangeLiteralToSeq (r: Expr): seq<int> =
```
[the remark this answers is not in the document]
rsoiffer: Why would an active pattern be preferred over a function that converts a Range expression to a sequence?

src/QsCompiler/CompilerOptimizations/Utils.fs
rsoiffer: CallableDict is much easier to write than ImmutableDictionary<QsQualifiedName, QsCallable>. Also, the type specifically represents the global dictionary the compiler uses to map qualified names to callables in the program, and has a helper function to support that.

src/QsCompiler/CompilerOptimizations/Utils.fs
rsoiffer: These are handled by the fillPartialArg function, the partialApplyFunction just uses the match to deal with the special case of a single argument.

src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
```diff
@@ -0,0 +1,74 @@
...
+        declarations |> List.sort |> Seq.ofList
+        
+    /// The ScopeTransformation used to evaluate constants
+    override syntaxTree.Scope = { new ScopeTransformation() with
```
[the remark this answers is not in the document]
rsoiffer: This doesn't seem necessary - if we want to call constant folding on a certain scope, we can use ConstantPropagator.Scope to access the scope transformation.

src/QsCompiler/CompilerOptimizations/FunctionEvaluation.fs
```diff
@@ -0,0 +1,145 @@
...
+            Seq.tryFind isInterrupt |? Normal
+        vars.exitScope()
+        res
+
+    /// Evaluates a Q# function
+    member this.evaluateFunction (name: QsQualifiedName) (arg: TypedExpression) (types: QsNullable<ImmutableArray<ResolvedType>>): Expr option =
+        // TODO: assert compiledCallables contains name
+        let callable = cd.getCallable name
+        // TODO: assert callable.Specializations.Length == 1
+        let impl = callable.Specializations.[0].Implementation
```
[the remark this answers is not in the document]
rsoiffer: When will a function have more than one specialization? I thought that wasn't possible right now.

src/QsCompiler/CompilerOptimizations/ExpressionEvaluation.fs
```diff
@@ -0,0 +1,244 @@
...
+        (this.ExpressionTransformation e1, this.ExpressionTransformation e2, this.ExpressionTransformation e3)
+
+    override this.onIdentifier (sym, tArgs) =
+        match sym with
+        | LocalVariable name -> vars.getVar name.Value |? Identifier (sym, tArgs)
+        | _ -> Identifier (sym, tArgs)
+
+    override this.onFunctionCall (method, arg) =
+        let method, arg = this.simplify (method, arg)
+        if maxRecursiveDepth > 0 && isLiteral arg.Expression cd then
```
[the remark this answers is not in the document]
rsoiffer: It's the other way around - I only evaluate functions, but not type constructors.

src/QsCompiler/CompilerOptimizations/ExpressionEvaluation.fs
```diff
@@ -0,0 +1,244 @@
...
+        match ex.Expression with
+        | CallLikeExpression ({Expression = Identifier (GlobalCallable qualName, types)}, arg)
+            when (cd.getCallable qualName).Kind = TypeConstructor ->
+            arg.Expression
+        | _ -> UnwrapApplication ex
+
+    override this.onArrayItem (arr, idx) =
+        let arr, idx = this.simplify (arr, idx)
+        match arr.Expression, idx.Expression with
+        | ValueArray va, IntLiteral i -> va.[int i].Expression
```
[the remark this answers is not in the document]
rsoiffer: Marked as TODO, I'll implement it later but it seems low priority for now. There aren't any downsides except that it takes time to write the code.

src/QsCompiler/CompilerOptimizations/ExpressionEvaluation.fs
```diff
@@ -0,0 +1,244 @@
...
+    override this.onNewArray (bt, idx) =
+        let idx = this.simplify idx
+        match idx.Expression with
+        | IntLiteral i -> constructNewArray bt.Resolution (int i) |? NewArray (bt, idx)
+        | _ -> NewArray (bt, idx)
+ 
+    override this.onCopyAndUpdateExpression (lhs, accEx, rhs) =
+        let lhs, accEx, rhs = this.simplify (lhs, accEx, rhs)
+        match lhs.Expression, accEx.Expression with
+        | ValueArray va, IntLiteral i -> ValueArray (va.SetItem(int i, rhs))
```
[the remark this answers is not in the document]
rsoiffer: Marked as TODO, I'll implement both cases later but they seem low priority for now.

src/QsCompiler/CompilerOptimizations/FunctionEvaluation.fs
```diff
@@ -0,0 +1,145 @@
...
+        | QsReturnStatement expr ->
+            this.evaluateExpression vars expr |> Returned
+        | QsFailStatement expr ->
+            this.evaluateExpression vars expr |> Failed
+        | QsVariableDeclaration s ->
+            fillVars vars (StringTuple.fromSymbolTuple s.Lhs, this.evaluateExpression vars s.Rhs); Normal
+        | QsValueUpdate s ->
+            match s.Lhs.Expression with
+                | Identifier (LocalVariable name, tArgs) -> vars.setVar(name.Value, this.evaluateExpression vars s.Rhs); Normal
+                | _ -> "Unknown LHS of value update statement: " + (prettyPrint s.Lhs.Expression) |> CouldNotEvaluate
```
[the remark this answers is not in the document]
rsoiffer: Marked as TODO, I'll implement it later but it seems low priority for now.

src/QsCompiler/CompilerOptimizations/Utils.fs
rsoiffer: Yeah... I didn't know about the ExpressionKindToQs transformation. I expanded this function significantly in the Loop Unrolling PR, see the Printer.fs file there. An existing transformation sounds like it would have been much easier.

src/QsCompiler/CompilerOptimizations/Utils.fs
rsoiffer: I fixed this issue in the Loop Unrolling PR. I'll copy the code from that branch to here.

src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
```diff
@@ -0,0 +1,74 @@
...
+        override scope.StatementKind = { new StatementKindTransformation() with 
+            override statementKind.ExpressionTransformation x = scope.Expression.Transform x
+            override statementKind.LocationTransformation x = scope.onLocation x
+            override statementKind.ScopeTransformation x = scope.Transform x
+            override statementKind.TypeTransformation x = scope.Expression.Type.Transform x
+
+            override statementKind.onVariableDeclaration stm =
+                let lhs = statementKind.onSymbolTuple stm.Lhs
+                let rhs = statementKind.ExpressionTransformation stm.Rhs
+                if stm.Kind = ImmutableBinding then
```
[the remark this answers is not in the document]
rsoiffer: No, while constant folding over mutable variables might happen eventually, I think it's lower priority. I imagine that users will declare almost all variables as immutable, and will only use mutable variables if truly needed, and we won't be able to fold in most of those cases.

src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
```diff
@@ -0,0 +1,74 @@
...
+
+    /// Gets a sorted list of the names of all the constant local variables
+    member this.getDeclarations =
+        declarations |> List.sort |> Seq.ofList
+        
+    /// The ScopeTransformation used to evaluate constants
+    override syntaxTree.Scope = { new ScopeTransformation() with
+
+        override scope.Transform x =
+            vars.enterScope()
```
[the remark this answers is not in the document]
rsoiffer: I fixed this issue in the Loop Unrolling PR. I'll copy the code from that branch to here.

src/QsCompiler/CompilerOptimizations/Utils.fs
rsoiffer: Yes, it does. I implemented a much more comprehensive test framework in the next PR. I can't just copy the code from that PR into here, though, as it's extensive enough that doing so would essentially mean merging the two PRs.

bettinaheim: Ok.

PR #70, opened 2019-07-23
Always log exceptions from file queries in debug mode

marshallsa:
This change applies to all of the calls to `CompilationUnitManager.FileQuery` that use `suppressExceptionLogging: true` (which is many of the language server command handlers). In release mode, the behavior is unchanged. In debug mode, the parameter is ignored and exceptions are always logged.

### DIFF
--- a/src/QsCompiler/CompilationManager/CompilationUnitManager.cs
+++ b/src/QsCompiler/CompilationManager/CompilationUnitManager.cs
@@ -580,6 +580,11 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
         // -> these commands need to be responsive and therefore won't wait for any processing to finish
         // -> if the query cannot be processed immediately, they simply return null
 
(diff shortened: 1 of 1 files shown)


bettinaheim:
@marshallsa Didn't you say one project file was also missing the definition of the DEBUG constant?

marshallsa:
@bettinaheim Yes, it looks like `QsLanguageServer` is the only project that defines `DEBUG` for debug builds. I thought that might've been a problem since the file I changed is in `QsCompilationManager` but it looks like it still works. I'm not sure if that's intentional - should I define `DEBUG` in all of the projects?

marshallsa:
Looks like defining `DEBUG` is probably the default in debug mode - even after unchecking the box in the Visual Studio project settings, `DEBUG` is still defined in debug mode. However, the checkbox works in release mode. It might just be a display issue in Visual Studio.

PR #79, opened 2019-07-24
Add install template and new project commands to VS Code extension.

cgranade:
This PR adds two new commands to the VS Code extension:

image

Still TODO:
- [x] Prompt user what kind of Q# project they want to make.

### DIFF
--- a/src/VSCodeExtension/src/commands.ts
+++ b/src/VSCodeExtension/src/commands.ts
@@ -1,0 +1,133 @@
+'use strict';
+import * as vscode from 'vscode';
+import * as cp from 'child_process';
+
+import { DotnetInfo } from './dotnet';
+import { IPackageInfo } from './packageInfo';
+import { sendTelemetryEvent, EventNames } from './telemetry';
+
+export function registerCommand(context: vscode.ExtensionContext, name: string, action: () => void) {
+    context.subscriptions.push(
+        vscode.commands.registerCommand(
+            name,
+            () => {
+                sendTelemetryEvent(
+                    EventNames.commandExecuted,
+                    {
+                        commandName: name
+                    },
+                    {}
+                )
+                action();
+            }
+        )
+    )
+}
+
+export function createNewProject(dotNetSdk: DotnetInfo) {    
+    const projectTypes: {[key: string]: string} = {
+        "Standalone console application": "console",
+        "Quantum library": "classlib",
+        "Unit testing project": "xunit"
+    };
+    let errorMessage = "";
+    vscode.window.showQuickPick(
+        Object.keys(projectTypes)
+    ).then(
+        projectTypeSelection => {
+            if (projectTypeSelection === undefined) {
+                throw undefined;
+            }
+            let projectType = projectTypes[projectTypeSelection];
+            
+            vscode.window.showSaveDialog({
+                saveLabel: "Create Project"
+            }).then(
+                (uri) => {
+                    if (uri !== undefined) {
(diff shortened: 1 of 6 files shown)


bettinaheim:
Is there a way to only show "Create new project" if the project templates have actually been installed first? Or at least give more concrete instructions to install the templates first instead of `.NET Core SDK exited with code ${code} when creating a new project.`?

src/VSCodeExtension/src/extension.ts
```diff
@@ -200,6 +200,94 @@ export function activate(context: vscode.ExtensionContext) {
...
+                                if (uri.scheme !== "file") {
+                                    vscode.window.showErrorMessage(
+                                        "New projects must be saved to the filesystem."
+                                    );
+
+                                }
+                                // TODO: show quick pick between console, library, and tests.
+                                cp.spawn(
+                                    dotNetSdk.path,
+                                    ["new", "console", "-lang", "Q#", "-o", uri.fsPath]
```
bettinaheim: Looks like making the distinction for different kinds of project seems fairly straighforward - do you want to add that or shall I?

cgranade: Just took a stab at it, but I'd appreciate a test to make sure I got the UX right.

cgranade:
> Is there a way to only show "Create new project" if the project templates have actually been installed first? Or at least give more concrete instructions to install the templates first instead of `.NET Core SDK exited with code ${code} when creating a new project.`?

That's a good idea, I agree, but I don't know of a good programmatic way of detecting whether the templates have been installed already...

cgranade:
OK, 7f3682b should do it.

bettinaheim:
Looks awesome! 
It would be great if the error message when creating a project before installing project templates could be a little more descriptive. Currently it is this: 
647be656-64cd-40c1-aa58-513b2c831475

cgranade:
Thanks for the feedback, @bettinaheim! I've modified so it now tries to give a better error message if the project templates are missing. The new message offers the user to install the project templates and then retry.

src/VSCodeExtension/src/commands.ts
```diff
@@ -22,13 +22,76 @@ export function registerCommand(context: vscode.ExtensionContext, name: string,
...
+                                }
+                            );
+                        }
+                    }
+                );
+            } else {
+                // Check if the problem was that the project templates are missing.
+                // If so, we can give a more helpful error message and offer the
+                // user to install templates.
+                if (errorMessage.includes("'Q#' is not a valid value for -lang")) {
```
bettinaheim: If someone has a different language setting, I would not expect that this is in English? I think it should be fine here to just look for Q# only, and modify the text below to say
"... the Q# project templates may not be installed. "
```suggestion
                if (errorMessage.includes("Q#") && errorMessage.includes("-lang")) {
```

src/VSCodeExtension/src/commands.ts
```diff
@@ -22,13 +22,76 @@ export function registerCommand(context: vscode.ExtensionContext, name: string,
...
+                    const installTemplatesItem = "Install project templates and retry";
+                    vscode.window.showErrorMessage(
+                        "Project creation failed, as the Q# project templates not installed.",
```
bettinaheim: I would suggest the edit, but it doesn't let me. How about instead: 
"Project creation failed. The Q# project templates may not be installed.",

bettinaheim: edited.

Issue #81, opened 2019-07-24
Parsing for Q# attributes


jviszlai:
With an end goal of supporting attributes in Q#, the compiler should first be able parse the syntax of attributes and store some information about the identifier and arguments. 
For example:
```
@ExampleAttribute("ExampleArgument")@
operation BitFlip(target : Qubit) : Unit {
    X(target);
}
```

bettinaheim:
@jviszlai Could you please link to the branch where you are working on this?
Let me know how things are going and if you have any questions and/or need help!

jviszlai:
Sure!
Working on this here: https://github.com/jviszlai/qsharp-compiler/tree/features/attributes
I made a PR that addresses most of the parts of this issue I was thinking of, but let me know if there's other parts I should work on: https://github.com/microsoft/qsharp-compiler/pull/82

bettinaheim:
> Sure!
> Working on this here: https://github.com/jviszlai/qsharp-compiler/tree/features/attributes
> I made a PR that addresses most of the parts of this issue I was thinking of, but let me know if there's other parts I should work on: #82

Awesome, thanks! I'll take a look shortly - I will first aggregate all the code action PRs, and then dive into this one.

bettinaheim:
With the PR to master up, I'll close this issue.

PR #82, opened 2019-07-25
Parsing for Q# Attributes

jviszlai:
Adds support for parsing the syntax of attributes. Includes some specific error handling for if the ending tag is missing on an attribute or the arguments are missing for an attribute, as well as context verification so that the attribute is only above an operation, function, or type definition.

Making this a draft pull request until a feature branch exists for attributes.

Targets #81


bettinaheim:
@jviszlai Thanks again for the contribution, and sorry for the delay in processing!
I don't know what the best way is to contact you, but if you reach out and contact me on my Microsoft address we've got a surprise for you. :) Purely optional of course.

bettinaheim:
@jviszlai Perfect, thanks!

Issue #50, opened 2019-07-17
Operations with the same specialization defined twice cause a language server error


marshallsa:
**Describe the bug**

When an operation has the same specialization defined twice, the language server has an error:

```
[Error - 2:28:53 PM] 
=============================
error while running global type checking
QsCompilerException: error while running type checking in background
QsCompilerException: more than one specialization of the same kind exists
InvalidOperationException: Sequence contains more than one element
Stack trace: 
   at System.Linq.Enumerable.Single[TSource](IEnumerable`1 source)
   at Microsoft.Quantum.QsCompiler.QsCompilerError.RaiseOnFailure[T](Func`1 func, String header) in C:\Users\t-samars\source\repos\qsharp-compiler\src\QsCompiler\DataStructures\ErrorHandling.fs:line 59

Stack trace: 
   at Microsoft.Quantum.QsCompiler.QsCompilerError.Raise(String message, Exception innerException) in C:\Users\t-samars\source\repos\qsharp-compiler\src\QsCompiler\DataStructures\ErrorHandling.fs:line 25
   at Microsoft.Quantum.QsCompiler.QsCompilerError.Raise(String message) in C:\Users\t-samars\source\repos\qsharp-compiler\src\QsCompiler\DataStructures\ErrorHandling.fs:line 29
   at <StartupCode$Microsoft-Quantum-QsDataStructures>.$ErrorHandling.RaiseAndReturnNull@58-1.Invoke(String msg) in C:\Users\t-samars\source\repos\qsharp-compiler\src\QsCompiler\DataStructures\ErrorHandling.fs:line 58
   at Microsoft.Quantum.QsCompiler.QsCompilerError.RaiseOnFailure[T](Func`1 func, String header) in C:\Users\t-samars\source\repos\qsharp-compiler\src\QsCompiler\DataStructures\ErrorHandling.fs:line 61
   at Microsoft.Quantum.QsCompiler.CompilationBuilder.TypeChecking.<>c.<BuildSpecializations>b__33_2(IGrouping`2 specs) in C:\Users\t-samars\source\repos\qsharp-compiler\src\QsCompiler\CompilationManager\TypeChecking.cs:line 1153
   at System.Collections.Immutable.ImmutableDictionary.<>c__DisplayClass9_0`3.<ToImmutableDictionary>b__0(TSource element)
   at System.Linq.Enumerable.SelectEnumerableIterator`2.MoveNext()
[...]

PR #85, opened 2019-07-26
Handling specialization redeclarations

bettinaheim:
This PR fixes the issue #50. 
Duplicate declarations will be deleted from the symbol table, and excluded from the built syntax tree.

### DIFF
--- a/src/QsCompiler/CompilationManager/CompilationUnit.cs
+++ b/src/QsCompiler/CompilationManager/CompilationUnit.cs
@@ -408,8 +408,11 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
                     var specializations = this.GlobalSymbols.DefinedSpecializations(header.QualifiedName).Select(defined =>
                     {
                         var specHeader = defined.Item2;
-                        var compiledSpecs = compiled.Specializations.Where(spec => spec.Kind == specHeader.Kind);
-                        QsCompilerError.Verify(compiledSpecs.Count() <= 1, "more than one specialization of the same kind exists"); // currently not supported
+                        var compiledSpecs = compiled.Specializations.Where(spec => 
+                            spec.Kind == specHeader.Kind &&
+                            spec.TypeArguments.Equals(specHeader.TypeArguments)
+                        ); 
+                        QsCompilerError.Verify(compiledSpecs.Count() <= 1, "more than one specialization of the same kind exists");
                         if (!compiledSpecs.Any()) return null; // may happen if a file has been modified during global type checking
 
                         var compiledSpec = compiledSpecs.Single();
--- a/src/QsCompiler/DataStructures/SymbolTable.fs
+++ b/src/QsCompiler/DataStructures/SymbolTable.fs
@@ -170,10 +170,13 @@ type private PartialNamespace private
         | true, specs -> specs.Add spec // it is up to the namespace to verify the type specializations
         | false, _ -> CallableSpecializations.Add(cName, new List<_>([spec]))
 
-    /// Deletes the specialization(s) defined at the specified location for the callable with the given name.
-    /// Throws the standard key does not exist exception if no callable with that name exists.
+    /// Deletes the *explicitly* defined specialization at the specified location for the callable with the given name.
+    /// Does not delete specializations that have been inserted by the compiler, i.e. specializations whose location matches the callable declaration location.
+    /// Throws the standard key does not exist exception if no specialization for the callable with that name exists.
     member internal this.RemoveCallableSpecialization (location : QsLocation) cName = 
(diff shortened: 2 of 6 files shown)


src/QsCompiler/CompilationManager/CompilationUnit.cs
```diff
@@ -408,8 +408,11 @@ internal void UpdateCallables(IEnumerable<QsCallable> updates)
                     var specializations = this.GlobalSymbols.DefinedSpecializations(header.QualifiedName).Select(defined =>
                     {
                         var specHeader = defined.Item2;
-                        var compiledSpecs = compiled.Specializations.Where(spec => spec.Kind == specHeader.Kind);
-                        QsCompilerError.Verify(compiledSpecs.Count() <= 1, "more than one specialization of the same kind exists"); // currently not supported
+                        var compiledSpecs = compiled.Specializations.Where(spec => 
+                            spec.Kind == specHeader.Kind &&
+                            spec.TypeArguments.Equals(specHeader.TypeArguments)
```
marshallsa: Is `spec.TypeArguments.Equals(specHeader.TypeArguments)` necessary for handling redeclared specializations or is this part of something else?

src/QsCompiler/DataStructures/SymbolTable.fs
```diff
@@ -538,6 +546,14 @@ and Namespace private
         let doc = ImmutableArray.Create(sprintf "automatically generated %A specialization for %s.%s" kind this.Name.Value parentName.Value)
         this.TryAddCallableSpecialization kind (source, location) ((parentName, declLocation.Range), generator, doc) 
 
+    /// Deletes the specialization(s) defined at the specified location and source file for the callable with the given name.
+    /// Throws an ArgumentException if the given source file is not listed as a source for (part of) the namespace.
+    /// Throws the standard key does not exist exception if no callable with that name exists.
+    member internal this.RemoveSpecialization (source, location) cName =
```
marshallsa: Same comment from above about maybe adding a note describing the return value.

src/QsCompiler/DataStructures/SymbolTable.fs
```diff
@@ -170,14 +170,22 @@ type private PartialNamespace private
         | true, specs -> specs.Add spec // it is up to the namespace to verify the type specializations
         | false, _ -> CallableSpecializations.Add(cName, new List<_>([spec]))
 
+    /// Deletes the *explicitly* defined specialization at the specified location for the callable with the given name.
+    /// Does not delete specializations that have been inserted by the compiler, i.e. specializations whose location matches the callable declaration location.
+    /// Throws the standard key does not exist exception if no specialization for the callable with that name exists.
+    member internal this.RemoveCallableSpecialization (location : QsLocation) cName = 
+        match CallableDeclarations.TryGetValue cName with 
+        | true, (_, decl) when decl.Position = location.Offset && decl.Range = location.Range -> 0 
+        | _ -> CallableSpecializations.[cName].RemoveAll (fun (_, res) -> location.Offset = res.Position && location.Range = res.Range)
```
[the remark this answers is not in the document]
bettinaheim: The specializations (in the future) could potentially be specified in different files (if we are to support declaring specializations outside the declaration body). The symbol table is set up to handle that. Correspondingly, I could move the assertion to the namespace (opposed to here in the partial namespace, which is the part in a particular file) but I'd rather keep the logic in the namespace manager. I'll add a doc a line to the doc comment though.

src/QsCompiler/CompilationManager/CompilationUnit.cs
bettinaheim: We could potentially have a specializations for particular type arguments. The symbol table is set up to in fact also handle that a "bundle" of specializations (i.e. all specializations for particular type arguments) may have different characteristics. We pretty much want to treat the operation specializied for particular type arguments as if if was it's own declaration.

src/QsCompiler/DataStructures/SymbolTable.fs
bettinaheim: Yup, will do.

src/QsCompiler/DataStructures/SymbolTable.fs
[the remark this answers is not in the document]
marshallsa: OK - just want to make sure I'm understanding this, but even if a callable had specializations defined in multiple files, a given location in a partial namespace or a given (source, location) pair in a namespace uniquely identifies one declared specialization, so the return value could still be at most 1?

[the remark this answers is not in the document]
bettinaheim: Probably. That depends a little bit on what we allow if we have type specializations, and how it's handled in the symbol table (whether one line of code can represent multiple different specializations - probably not, but I haven't sufficiently thought about it yet). Right now it is indeed at most one, since we do not delete auto-generated functor specializations.

Issue #86, opened 2019-07-26
Cannot parse smallest possible int64 value


msoeken:
**Describe the bug**

The smallest possible int64 value `-9223372036854775808` cannot be parsed into a Q# `Int` type.

**To Reproduce**

A test case to trigger the bug is described in this branch:

https://github.com/msoeken/qsharp-compiler/blob/msoeken/minint-bug/src/QsCompiler/Tests.Compiler/SyntaxTests.fs#L171

**Expected behavior**

The parser returns an `Int` value with the smallest integer value.

**System information**

Reproduced on master, other system information seems to be irrelevant for this bug.

**Additional context**

Maybe the sign is parsed separate from the int value, since the largest int64 value is 9223372036854775807.

bettinaheim:
Hehe, yes, indeed, I parse the sign separately... That's one edge case to overlook. Thanks for pointing it out!

PR #87, opened 2019-07-27
Compile time errors for overflowing int and double literals

bettinaheim:
This PR adds a proper Q# diagnostic for everything that would result in a compile time error in C#. 
It also fixes issue #86.


src/VisualStudioExtension/QsharpVSIX/QsharpVSIX.csproj
```diff
@@ -117,6 +117,9 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.VisualStudio.SDK" Version="15.9.3" ExcludeAssets="runtime" />
+    <PackageReference Include="Microsoft.VisualStudio.Telemetry">
```
anpaz-msft: It's probably a good idea to have these changes in a seperate PR.

Issue #44, opened 2019-07-16
Support for Code Completion


bettinaheim:
The Q# integration into Visual Studio via the extension in this repository currently does not provide any code completion suggestions while typing, while the native completion for text documents is available in VS Code. The goal is for the language server to provide completion suggestions informed by the compiler.

bettinaheim:
Since this is a larger endeavor, I create a feature branch to PR against: https://github.com/microsoft/qsharp-compiler/tree/features/codeCompletion

bettinaheim:
To be release with 0.9.

PR #92, opened 2019-07-29
Provide context-aware completions for declarations and open directives

marshallsa:
Adds a custom parser to provide exact completions at each position in an open directive or type, function or operation declaration.

Also moved the code from previous PRs that used to be in EditorSupport.cs into a new file, EditorCompletion.cs. This makes the diff much larger than it should be, sorry. :( The actual new methods are:

* `GetCompletionEnvironment`
* `GetCompletionsForKind`
* `GetFragmentAtOrBefore`
* `GetDelimitingCharPosition`

The other methods have had minor changes.

This is part of issue #44.


src/QsCompiler/CompilationManager/EditorCompletion.cs
```diff
@@ -0,0 +1,686 @@
...
+        /// <exception cref="ArgumentNullException">Thrown when any argument is null.</exception>
+        private static FragmentRelationship GetFragmentAtOrBefore(
+            FileContentManager file,
+            Position position,
+            out CodeFragment.TokenIndex index,
+            out CodeFragment fragment)
+        {
+            if (file == null)
+                throw new ArgumentNullException(nameof(file));
+            if (position == null)
```
bettinaheim: same as the other comment regarding the position null check

src/QsCompiler/CompilationManager/EditorCompletion.cs
```diff
@@ -0,0 +1,686 @@
...
+                throw new ArgumentNullException(nameof(position));
+
+            var lineNumber = position.Line;
+            var tokens = file.GetTokenizedLine(lineNumber);
+
+            // If the current line is empty, find the last non-empty line before it.
```
bettinaheim: I there are a bunch of similar things implemented in the TextProcessor, so you may be able to use one of those if a suitable one already exists (I don't remember if a suitable one exists, but it seems like something that might).

src/QsCompiler/CompilationManager/EditorCompletion.cs
```diff
@@ -0,0 +1,686 @@
...
+                fragment = null;
+                return FragmentRelationship.Missing;
+            }
+
+            index = new CodeFragment.TokenIndex(file, lineNumber, indexNumber);
+            fragment = index.GetFragment();
+            if (fragment.FollowedBy != CodeFragment.MissingDelimiter &&
+                GetDelimitingCharPosition(file, fragment).IsSmallerThan(position))
+            {
+                switch (fragment.FollowedBy)
```
bettinaheim: It may be better to check the indentation level that is assigned to each fragment, to avoid having to hardcode the meaning of '{','}',';' here.

src/QsCompiler/CompilationManager/EditorCompletion.cs
```diff
@@ -0,0 +1,686 @@
...
+                throw new ArgumentNullException(nameof(fragment));
+            if (fragment.FollowedBy == CodeFragment.MissingDelimiter)
+                throw new ArgumentException("Code fragment has a missing delimiter", nameof(fragment));
+
+            var end = fragment.GetRange().End;
+            for (var lineNumber = end.Line; lineNumber < file.NrLines(); lineNumber++)
+            {
+                var start = lineNumber == end.Line ? end.Character : 0;
+                var line = file.GetLine(lineNumber);
+                var index = line.FindInCode(s => s.IndexOf(fragment.FollowedBy), start, line.Text.Length - start);
```
bettinaheim: So, I represent a missing delimiter by the char @, which may occur in the code for valid reasons. You need to ask the fragment is it's delimiter is missing first and special case on that.

src/QsCompiler/CompilationManager/EditorCompletion.cs
```diff
@@ -0,0 +1,686 @@
...
+            if (position == null)
+                throw new ArgumentNullException(nameof(position));
+
+            var relationship = GetFragmentAtOrBefore(file, position, out var index, out _);
+            CodeFragment.TokenIndex parent = null;
+            if (relationship == FragmentRelationship.EnclosingScope)
+                parent = index;
+            else if (relationship == FragmentRelationship.Self || relationship == FragmentRelationship.SameScope)
+                parent = index.GetNonEmptyParent();
+            else if (relationship == FragmentRelationship.InnerScope)
```
bettinaheim: Wouldn't you have to count if you have several closing brackets?

src/QsCompiler/TextProcessor/CompletionParsing.fs
```diff
@@ -0,0 +1,230 @@
...
+    pchar '\u0004'
+
+/// Parses a symbol. Includes reserved keywords if the keyword is followed immediately by EOT without any whitespace
+/// (i.e., a possibly incomplete symbol that happens to start with a keyword). Does not consume whitespace.
+let private symbol =
+    let qsIdentifier =
+        identifier <| IdentifierOptions (isAsciiIdStart = (fun c -> isAsciiLetter c || c = '_'),
+                                            isAsciiIdContinue = isSymbolContinuation)
+    notFollowedBy qsReservedKeyword >>. qsIdentifier <|>
+    (qsIdentifier .>> previousCharSatisfiesNot Char.IsWhiteSpace .>> followedBy eot)
```
bettinaheim: the previousCharSatisfiesNot seems redundant if you are not consuming any whitespace in any case?

src/QsCompiler/TextProcessor/CompletionParsing.fs
```diff
@@ -0,0 +1,230 @@
...
+
+/// Parses the end-of-transmission character.
+let private eot =
+    pchar '\u0004'
+
+/// Parses a symbol. Includes reserved keywords if the keyword is followed immediately by EOT without any whitespace
+/// (i.e., a possibly incomplete symbol that happens to start with a keyword). Does not consume whitespace.
+let private symbol =
+    let qsIdentifier =
+        identifier <| IdentifierOptions (isAsciiIdStart = (fun c -> isAsciiLetter c || c = '_'),
```
bettinaheim: Even though I don't think we will change what constitutes a valid symbol start, it is probably still good to add an isSymbolStart to the Parsing primitives, and using that for both the parser here and the symbolNameLike parser in the SyntaxBuilder.fs

src/QsCompiler/TextProcessor/CompletionParsing.fs
```diff
@@ -0,0 +1,230 @@
...
+                                            isAsciiIdContinue = isSymbolContinuation)
+    notFollowedBy qsReservedKeyword >>. qsIdentifier <|>
+    (qsIdentifier .>> previousCharSatisfiesNot Char.IsWhiteSpace .>> followedBy eot)
+
+/// If `p1` succeeds and consumes all input, returns the result of `p1`. Otherwise, parses `p1` then `p2` and returns
+/// the result of `p2`.
+let private (?>>) p1 p2 =
+    attempt (p1 .>> eof) <|> (p1 >>. p2)
+
+/// Parses an expected identifier of the given kind. The identifier is optional if EOT occurs first.
```
bettinaheim: The formulation for me is a bit confusing, because the kind is irrelevant for the parsing - it is assumed to be what p parses and returned if eot is encountered before or right after (i.e. no whitespace in between) p succeeds.

src/QsCompiler/TextProcessor/CompletionParsing.fs
```diff
@@ -0,0 +1,230 @@
+﻿module Microsoft.Quantum.QsCompiler.TextProcessing.CompletionParsing
```
bettinaheim: A brief description of what that module is for would be good.

src/QsCompiler/Tests.Compiler/CompletionParsingTests.fs
```diff
@@ -0,0 +1,166 @@
...
+    test "newtype" [Keyword "function"; Keyword "operation"; Keyword "newtype"; Keyword "open"]
+    test "open" [Keyword "function"; Keyword "operation"; Keyword "newtype"; Keyword "open"]
+
+[<Fact>]
+let ``Function declaration parser tests`` () =
+    test "function " [Declaration]
+    test "function Foo" [Declaration]
+    test "function Foo " []
+    test "function Foo (" [Declaration]
+    test "function Foo (x" [Declaration]
```
bettinaheim:
```suggestion
    test "function Foo (x" [Declaration]
    test "function Foo (x " []
```

src/QsCompiler/Tests.Compiler/CompletionParsingTests.fs
```diff
@@ -0,0 +1,166 @@
...
+    test "function Foo (x :" [Type]
+    test "function Foo (x : Int" [Type]
```
bettinaheim:
```suggestion
    test "function Foo (x : Int" [Type]
    test "function Foo (x : Int " []
```

src/QsCompiler/Tests.Compiler/CompletionParsingTests.fs
```diff
@@ -0,0 +1,166 @@
...
+    test "function Foo (x : Int)" []
+    test "function Foo (x : Int) :" [Type]
+    test "function Foo (x : Int) : " [Type]
+    test "function Foo (x : Int) : Unit" [Type]
+    test "function Foo (x : Int) : Unit " []
+    test "function Foo (x : Int) : (Int, MyT" [Type]
+    test "function Foo (x : Int) : (Int," [Type]
```
bettinaheim:
```suggestion
    test "function Foo (x : Int) : (Int," [Type]
    test "function Foo (x : Int) : (Int, " [Type]
```

src/QsCompiler/Tests.Compiler/CompletionParsingTests.fs
```diff
@@ -0,0 +1,166 @@
...
+    test "function Foo (x : Int) : (Int, MyT " []
+    test "function Foo (x : (" [Type]
+    test "function Foo (x : (Int" [Type]
+    test "function Foo (x : (Int," [Type]
+    test "function Foo (x : (Int, " [Type]
+    test "function Foo (x : (Int, Int" [Type]
+    test "function Foo (x : (Int, Int " []
+    test "function Foo (x : (Int, Int), " [Declaration]
+    test "function Foo (f : (Int -> " [Type]
```
bettinaheim:
```suggestion
    test "function Foo (f : (Int -> " [Type]
    test "function Foo (f : (Int -> Int" [Type]
    test "function Foo (f : (Int -> Int " []
    test "function Foo (f : (Int -> Int)" []
```

src/QsCompiler/Tests.Compiler/CompletionParsingTests.fs
```diff
@@ -0,0 +1,166 @@
...
+    test "operation Foo (q : Qubit) : Unit " [Keyword "is"]
+    test "operation Foo (q : Qubit) : Unit i" [Keyword "is"]
+    test "operation Foo (q : Qubit) : Unit is" [Keyword "is"]
+    test "operation Foo (q : Qubit) : Unit is " [Characteristic]
+    test "operation Foo (q : Qubit) : Unit is A" [Characteristic]
+    test "operation Foo (q : Qubit) : Unit is Adj" [Characteristic]
+    test "operation Foo (q : Qubit) : Unit is Adj " []
+    test "operation Foo (q : Qubit) : Unit is Adj +" [Characteristic]
+    test "operation Foo (q : Qubit) : Unit is Adj + " [Characteristic]
+    test "operation Foo (q : Qubit) : Unit is Adj + Ctl" [Characteristic]
```
bettinaheim:
```suggestion
    test "operation Foo (q : Qubit) : Unit is Adj + Ctl" [Characteristic]
    test "operation Foo (q : Qubit) : Unit is (" [Characteristic]
    test "operation Foo (q : Qubit) : Unit is ( " [Characteristic]
    test "operation Foo (q : Qubit) : Unit is (Adj" [Characteristic]
    test "operation Foo (q : Qubit) : Unit is (Adj " []
    test "operation Foo (q : Qubit) : Unit is (Adj + " [Characteristic]
    test "operation Foo (q : Qubit) : Unit is (Adj + Ctl" [Characteristic]
    test "operation Foo (q : Qubit) : Unit is (Adj + Ctl)" []    
```

src/QsCompiler/Tests.Compiler/CompletionParsingTests.fs
```diff
@@ -0,0 +1,166 @@
...
+    test "operation Foo (q : (Qubit" [Type]
+    test "operation Foo (q : (Qubit," [Type]
+    test "operation Foo (q : (Qubit, " [Type]
+    test "operation Foo (q : (Qubit, Qubit" [Type]
+    test "operation Foo (q : (Qubit, Qubit " []
+    test "operation Foo (q : (Qubit, Qubit), " [Declaration]
+    test "operation Foo (q : (Qubit => " [Type]
+    test "operation Foo (q : (Qubit => Unit" [Type]
+    test "operation Foo (q : (Qubit => Unit " [Keyword "is"]
+    test "operation Foo (q : (Qubit => Unit is " [Characteristic]
```
bettinaheim:
```suggestion
    test "operation Foo (q : (Qubit => Unit is " [Characteristic]
    test "operation Foo (q : (Qubit => Unit is Adj" [Characteristic]
    test "operation Foo (q : (Qubit => Unit is Adj " []
    test "operation Foo (q : (Qubit => Unit is Adj + " [Characteristic]
    test "operation Foo (q : (Qubit => Unit is Adj)" []
```

src/QsCompiler/Tests.Compiler/CompletionParsingTests.fs
```diff
@@ -0,0 +1,166 @@
...
+    test "operation Foo (q : (Qubit => Unit is Adj)) : Unit" [Type]
```
bettinaheim: Does     test "operation Foo (q : ((Qubit => Unit) is Adj)" []
work as well? If not, no worried - I think it's acceptable to not give completions in that case. I'm just curious.

src/QsCompiler/Tests.Compiler/CompletionParsingTests.fs
```diff
@@ -0,0 +1,166 @@
...
+    test "newtype MyType = (Int, Bool)" []
+    test "newtype MyType = (MyItem :" [Type]
+    test "newtype MyType = (MyItem : " [Type]
+    test "newtype MyType = (MyItem : In" [Type]
+    test "newtype MyType = (MyItem : Int" [Type]
+    test "newtype MyType = (MyItem : Int," [Declaration; Type]
+    test "newtype MyType = (MyItem : Int, " [Declaration; Type]
+    test "newtype MyType = (MyItem : Int, Boo" [Declaration; Type]
+    test "newtype MyType = (MyItem : Int, Bool" [Declaration; Type]
+    test "newtype MyType = (MyItem : Int, Bool)" []
```
bettinaheim:
```suggestion
    test "newtype MyType = (MyItem : Int, Bool)" []
    test "newtype MyType = (MyItem : Int, (" [Declaration; Type]
    test "newtype MyType = (MyItem : Int, (Item2" [Declaration; Type]
    test "newtype MyType = (MyItem : Int, (Item2 " []
    test "newtype MyType = (MyItem : Int, (Item2, " [Declaration; Type]
    test "newtype MyType = (MyItem : Int, (Item2, Item3 :" [Type]
    test "newtype MyType = (MyItem : Int, (Item2, Item3 : Int)" []
    test "newtype MyType = (MyItem : Int, (Item2, Item3 : Int))" []
```

src/QsCompiler/CompilationManager/EditorCompletion.cs
```diff
@@ -0,0 +1,686 @@
...
+using Microsoft.Quantum.QsCompiler.SyntaxTree;
+using Microsoft.Quantum.QsCompiler.TextProcessing;
+using Microsoft.VisualStudio.LanguageServer.Protocol;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.Serialization;
+using static Microsoft.Quantum.QsCompiler.TextProcessing.CompletionParsing;
+
+namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
```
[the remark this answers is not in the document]
marshallsa: I tried this but the EditorSupport namespace conflicts with the EditorSupport class:

> EditorSupport.cs(22,27): error CS0101: The namespace 'Microsoft.Quantum.QsCompiler.CompilationBuilder' already contains a definition for 'EditorSupport' [C:\Users\t-samars\source\repos\qsharp-compiler\src\QsCompiler\CompilationManager\QsCompilationManager.csproj]

src/QsCompiler/CompilationManager/EditorCompletion.cs
```diff
@@ -0,0 +1,686 @@
...
+        /// <summary>
+        /// The qualified name of the completion item.
+        /// </summary>
+        public QsQualifiedName QualifiedName
+        {
+            get =>
+                @namespace == null || name == null
+                ? null
+                : new QsQualifiedName(NonNullable<string>.New(@namespace), NonNullable<string>.New(name));
+            set
```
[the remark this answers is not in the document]
marshallsa: I made it immutable and it looks like the serialization still works.

src/QsCompiler/CompilationManager/EditorCompletion.cs
```diff
@@ -0,0 +1,686 @@
...
+            /// The second fragment is in the enclosing scope of the first fragment.
+            /// </summary>
+            EnclosingScope
+        }
+
+        /// <summary>
+        /// Completion items for built-in type keywords.
+        /// </summary>
+        private static readonly IEnumerable<CompletionItem> typeKeywords =
+            new[]
```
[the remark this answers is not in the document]
marshallsa: I removed this list in a newer branch by having the completion parser return all of the type keywords individually instead of just `Type`. I think it would be possible to bring that change to this PR as well, should I do that?

src/QsCompiler/CompilationManager/EditorCompletion.cs
marshallsa: This method throws an exception if the delimiter is missing (line 668) - I think that should handle it?

src/QsCompiler/CompilationManager/EditorCompletion.cs
marshallsa: If there are multiple closing brackets in a row, e.g.

```
operation Foo () : Unit {
    if (true) {
        Bar();
    }
}
```

Then if the position is after the last closing bracket, the fragment found by `GetFragmentAtOrBefore` seems to be an empty fragment that starts and ends at the same position as the last bracket, and is followed by `}`. I'm not sure why this is the case - maybe I shouldn't rely on it...

src/QsCompiler/CompilationManager/EditorCompletion.cs
```diff
@@ -0,0 +1,686 @@
...
+                NonNullable<string>.New(alias), NonNullable<string>.New(nsName), file.FileName) ?? alias;
+        }
+
+        /// <summary>
+        /// Gets the token index and corresponding code fragment at, or the closest before, the given position in the
+        /// file. Returns the relationship between the found fragment and the fragment that would exist directly at the
+        /// given position, which may be different from <see cref="FragmentRelationship.Self"/> if the given position is
+        /// not part of any fragment.
+        /// <para/>
+        /// If <see cref="FragmentRelationship.Missing"/> is returned, then both the token index and code fragment out
```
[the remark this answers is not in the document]
marshallsa: `GetFragmentAtOrBefore` only returns `Missing` if there is no fragment at the current position and no fragments before (so yes I think this would include position (0, 0)). If the given position is not part of a fragment but there's a fragment before the current position, it doesn't return `Missing`. It's kind of confusing because the two fragments `GetFragmentAtOrBefore` is comparing using `FragmentRelationship` is the fragment that it actually found, and the fragment that *would* be at the given position (even if that fragment doesn't exist)…

I think it might be less confusing to replace `FragmentRelationship` with two pieces of information:

1. Did `GetFragmentAtOrBefore` find a fragment exactly at the given position or did it find one before?
2. If it found a fragment before the position, what is its indentation relative to the indentation at the given position?

src/QsCompiler/Tests.Compiler/CompletionParsingTests.fs
marshallsa: Yes, this works - it can parse the operation input/output type with or without parentheses.

src/QsCompiler/Tests.Compiler/CompletionParsingTests.fs
marshallsa: Some of these tests fail because the parser I'm not using a custom characteristics expression parser in this PR. It ends up working for simple cases but gets confused by missing parentheses and will think that `operation Foo (q : Qubit) : Unit is (Adj ` should be followed by another characteristic.

In my expression parser branch, I have not tried adapting the new expression parser for characteristics yet, but I think it will fix these issues. Is it OK to wait until a later PR?

src/QsCompiler/CompilationManager/EditorCompletion.cs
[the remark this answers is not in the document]
marshallsa: The change would only let me get rid of the list of types - the list of characteristics is still there for now but I think I can remove it when I replace the characteristics expression parser later.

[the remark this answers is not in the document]
bettinaheim: Sounds good.

src/QsCompiler/CompilationManager/EditorCompletion.cs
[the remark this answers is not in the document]
bettinaheim: How about making the EditorSupport static class a partial class and using it here?

src/QsCompiler/CompilationManager/EditorCompletion.cs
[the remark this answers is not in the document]
marshallsa: Resolved by refactoring and removing the `FragmentRelationship` enum.

src/QsCompiler/CompilationManager/EditorCompletion.cs
marshallsa: This code has been generalized to use relative indentation numbers now.

src/QsCompiler/CompilationManager/EditorCompletion.cs
marshallsa: We weren't able to find an existing function for `GetTokenAtOrBefore`, but I was able to replace most of the logic in `GetDelimiterPosition` with `FragmentEnd` from TextProcessor.cs.

PR #99, opened 2019-08-01
Support for conjugations as statements

bettinaheim:
Since this PR is already getting big enough as it is, I will make a separate PR with tests for the resolution to unique variable names and the actually inclining of conjugations. Hence PRing against a feature branch. The support itself should be complete and does not require any additional support from another repo - so ready to be picked up by libraries and give it a spin. :)


src/QsCompiler/CompilationManager/TypeChecking.cs
```diff
@@ -1176,7 +1232,9 @@ private static ImmutableArray<QsSpecialization> BuildSpecializations
                     var (arg, messages) = buildArg(userDefined.Item);
                     foreach (var msg in messages) diagnostics.Add(Diagnostics.Generate(spec.SourceFile.Value, msg, specPos));
 
-                    implementation = BuildUserDefinedImplemenation(root, spec.SourceFile, arg, symbolTracker, diagnostics);
+                    QsGeneratorDirective GetDirective(QsSpecializationKind k) => definedSpecs.TryGetValue(k, out defined) && defined.Item1.IsValue ? defined.Item1.Item : null;
+                    var requiredFunctorSupport = RequiredFunctorSupport(kind, GetDirective).ToImmutableHashSet();
+                    implementation = BuildUserDefinedImplemenation(root, spec.SourceFile, arg, requiredFunctorSupport, symbolTracker, diagnostics);
```
thomashaener: Minor typo: `BuildUserDefinedImplemenation`

bettinaheim: Thanks!

bettinaheim:
> Thanks for working on this, @bettinaheim!
> Besides the minor typo, I was wondering if the conjugation shouldn't apply
> U* V U
> 
> instead of the other way around, unless the comments are using circuit notation.
> In any case, I will give it a try asap :-)

Yeah, the comments are not very clear since they do not specify where the argument goes... The way it's implemented it is consistent with what the library does, I believe, so I'll leave the implementation as is and adapt the comments to be clearer. Thanks for pointing it out!

src/QsCompiler/Transformations/SearchAndReplace.cs
```diff
@@ -243,6 +249,87 @@ private void OnExpression(TypedExpression ex)
...
+
+        private int VariableNr;
+        private Dictionary<NonNullable<string>, NonNullable<string>> UniqueNames;
+
+        /// <summary>
+        /// Will overwrite the dictionary entry mapping a variable name to the corresponding unique name if the key already exists. 
+        /// </summary>
+        internal NonNullable<string> GenerateUniqueName(NonNullable<string> varName)
+        {
+            var unique = NonNullable<string>.New($"__{Prefix}{this.VariableNr++}__{varName.Value}__");
```
rsoiffer: "this.VariableNr++" should probably be moved onto its own line, as it seems like good practice to separate the definition of "unique" from mutating the state of "VariableNr"

bettinaheim:
I forgot to add a check if any of the variables used in the outer block is reassigned in the inner block. Currently this would cause the auto-generated adjoint inserted after the inner block to not be the uncompute of the outer block. The plan is to error in this case. I'll make a separate PR for that. 
@cgranade @apaetz @thomashaener could one of you please approve this, since we seem to be having permissions issues currently... - it's going against a feature branch in any case, so there will be a merge to master before the end of the month.

PR #107, opened 2019-08-07
Initial draft for packaging Q# data structures as dlls.

cgranade:
This PR prototypes a new compilation target that encapsulates a serialized Q# AST as an assembly resource with a well-known name. This approach allows us to reuse infrastructure for finding and loading assemblies for delivering quantum libraries as well.

Still TODO:
- [x] The Roslyn compilation seems to strip out "unused" assembly references, such that the emitted assembly only references `typeof(object).Assembly`, and not the other quantum libraries that were passed as references immediately before `compilation.Emit`.
- [ ] Assembly metadata (name, version, delay sign, etc.) needs to be set.


src/QsCompiler/DllPackaging/Program.cs
```diff
@@ -0,0 +1,203 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.IO;
+using System.Linq;
+using McMaster.Extensions.CommandLineUtils;
```
bettinaheim: We should stick to using one tool for command line parsing. But I suspect that will change anyway, so I suggest to leave it for now, and then I'll refactor this.

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -439,13 +439,21 @@ private string GenerateBinary(MemoryStream ms)
...
+        /// Potentially throws an exception in particular also if the given binary file has been compiled with a different compiler version.
         /// </summary>
-        public static IEnumerable<QsNamespace> ReadBinary(string file)
+        public static IEnumerable<QsNamespace> ReadBinary(string file) =>
+            ReadBinary(
+                new MemoryStream(File.ReadAllBytes(Path.GetFullPath(file)))
+            );
+
+        /// <summary>
+        /// Given a stream for reading a Q# binary file, reads the content of that file and returns the corresponding syntax tree.
```
bettinaheim:
```suggestion
        /// Given a stream with the content of a Q# binary file, returns the corresponding syntax tree.
```

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -439,13 +439,21 @@ private string GenerateBinary(MemoryStream ms)
         /// <summary>
         /// Given the path to a Q# binary file, reads the content of that file and returns the corresponding syntax tree. 
         /// Throws the corresponding exception if the given path does not correspond to a suitable binary file.
-        /// Potentially throws an exception in particular also if the given binary file has been compiled with a different compiler version. 
+        /// Potentially throws an exception in particular also if the given binary file has been compiled with a different compiler version.
```
bettinaheim:
```suggestion
        /// May throw an exception if the given binary file has been compiled with a different compiler version.
```

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -439,13 +439,21 @@ private string GenerateBinary(MemoryStream ms)
...
-        public static IEnumerable<QsNamespace> ReadBinary(string file)
+        public static IEnumerable<QsNamespace> ReadBinary(string file) =>
+            ReadBinary(
+                new MemoryStream(File.ReadAllBytes(Path.GetFullPath(file)))
+            );
+
+        /// <summary>
+        /// Given a stream for reading a Q# binary file, reads the content of that file and returns the corresponding syntax tree.
+        /// Throws the corresponding exception if the given path does not correspond to a suitable binary file.
+        /// Potentially throws an exception in particular also if the given binary file has been compiled with a different compiler version.
```
bettinaheim:
```suggestion
        /// May throw an exception if the given binary file has been compiled with a different compiler version.
```

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -439,13 +439,21 @@ private string GenerateBinary(MemoryStream ms)
...
         /// </summary>
-        public static IEnumerable<QsNamespace> ReadBinary(string file)
+        public static IEnumerable<QsNamespace> ReadBinary(string file) =>
+            ReadBinary(
+                new MemoryStream(File.ReadAllBytes(Path.GetFullPath(file)))
+            );
+
+        /// <summary>
+        /// Given a stream for reading a Q# binary file, reads the content of that file and returns the corresponding syntax tree.
+        /// Throws the corresponding exception if the given path does not correspond to a suitable binary file.
```
bettinaheim:
```suggestion
        /// Throws an ArgumentNullException if the given stream is null.
```

bettinaheim:
@cgranade  I suggest to merge this to a feature branch for further processing. Sounds good?

anpaz-msft:
Can we add some unittests for this please? 
We should at least test that the syntax tree is embedded in an Assembly and that it can be read correctly.

anpaz-msft:
Also, wondering if we should have yet another command line program just for this. 
- What's the value?
- who is the intended audience?
- what's the expected usage workflow?

anpaz-msft:
Finally, what's the plan to integrate with the rest of the QDK? 
The goal is that the dlls that we currently generate will also have the syntax-tree embedded, so we can use it at runtime. Are you envisioning that we will change the QDK to generate binary syntax-tree files, then call another tool to embed those in a dll? 
In the prototype I have, I changed the C# generation app to ask the compiler to generate the syntax-tree into a file, then add this file as a <Reference> (with a well-known name) so it gets picked up automatically by the csc when compiling the .dll.

bettinaheim:
@anpaz-msft, kindly read the comments - this is work in progress, as is usual for work on feature branches.

PR #108, opened 2019-08-09
Rsoiffer/operation inlining

rsoiffer:
Implements a simple form of operation inlining. Is greatly improved by the following PR, which removes QsScopeStatements and ensures that variables have unique names.


src/QsCompiler/CompilerOptimizations/TransformationState.fs
```diff
@@ -0,0 +1,104 @@
...
+    | CallLikeExpression (a, b) ->
+        isLiteral state a.Expression && isLiteral state b.Expression &&
+            TypedExpression.IsPartialApplication (CallLikeExpression (a, b))
+    | _ -> false
+
+
+/// Returns a TransformationState with the given variable defined as the given value
+let internal defineVar state (name, value) =
+    if not (isLiteral state value) then state else
+    // TODO: assert variable is undefined
```
bettinaheim: Please add the corresponding QsCompilerError.Verify.

src/QsCompiler/CompilerOptimizations/TransformationState.fs
```diff
@@ -0,0 +1,104 @@
...
+
+/// Returns a TransformationState with the given variable set to the given value
+let internal setVar state (name, value) =
+    if not (isLiteral state value) then state else
+    // TODO: assert variable is defined, is same type
+    match state.scopeStack |> List.tryFindIndex (Map.containsKey name) with
+    | Some index ->
+        let newScopeStack = state.scopeStack |> List.mapi (fun i -> if i = index then Map.add name value else id)
+        {state with scopeStack = newScopeStack}
+    | None -> failwithf "Variable %s is undefined" name
```
bettinaheim:
```suggestion
    | None -> System.ArgumentException (sprintf "Variable %s is undefined" name) |> raise
```

src/QsCompiler/CompilerOptimizations/TransformationState.fs
```diff
@@ -0,0 +1,104 @@
...
+let internal defineVar state (name, value) =
+    if not (isLiteral state value) then state else
+    // TODO: assert variable is undefined
+    match state.scopeStack with
+    | head :: tail ->
+        let newHead = head.Add (name, value)
+        {state with scopeStack = newHead :: tail}
+    | [] -> failwithf "No scope to define variables in"
+
+/// Returns a TransformationState with the given variable set to the given value
```
bettinaheim:
```suggestion
/// Returns a TransformationState with the given variable set to the given value.
/// Throws an ArgumentException if no variable with the given name is defined in the given state. 
```

src/QsCompiler/CompilerOptimizations/TransformationState.fs
```diff
@@ -0,0 +1,104 @@
...
+/// Gets the QsCallable with the given qualified name
+let internal getCallable state qualName =
+    state.compiledCallables.[qualName]
+
+
+/// Returns a TransformationState inside of a new scope
+let internal enterScope state =
+    {state with scopeStack = Map.empty :: state.scopeStack}
+
+/// Returns a TransformationState outside of the current scope
```
bettinaheim:
```suggestion
/// Returns a TransformationState outside of the current scope.
/// Throws an InvalidOperationException if the scope stack of the given state is empty. 
```

src/QsCompiler/CompilerOptimizations/TransformationState.fs
```diff
@@ -0,0 +1,104 @@
...
+
+/// Returns a TransformationState with the given variable defined as the given value
+let internal defineVar state (name, value) =
+    if not (isLiteral state value) then state else
+    // TODO: assert variable is undefined
+    match state.scopeStack with
+    | head :: tail ->
+        let newHead = head.Add (name, value)
+        {state with scopeStack = newHead :: tail}
+    | [] -> failwithf "No scope to define variables in"
```
bettinaheim:
```suggestion
    | [] -> InvalidOperationException "No scope to define variables in" |> raise
```

src/QsCompiler/CompilerOptimizations/TransformationState.fs
```diff
@@ -0,0 +1,104 @@
...
+let internal setVar state (name, value) =
+    if not (isLiteral state value) then state else
+    // TODO: assert variable is defined, is same type
+    match state.scopeStack |> List.tryFindIndex (Map.containsKey name) with
+    | Some index ->
+        let newScopeStack = state.scopeStack |> List.mapi (fun i -> if i = index then Map.add name value else id)
+        {state with scopeStack = newScopeStack}
+    | None -> failwithf "Variable %s is undefined" name
+
+/// Applies on of the operations above on a SymbolTuple, ValueTuple pair
```
bettinaheim: Can you be more descriptive? I think you mean one of the functions in this module rather than a Q# operation?

```suggestion
/// Applies the given function op on a SymbolTuple, ValueTuple pair
```

src/QsCompiler/CompilerOptimizations/TransformationState.fs
```diff
@@ -0,0 +1,104 @@
...
+let rec private onTuple op state (names, values) =
+    match names, values with
+    | VariableName name, _ ->
+        op state (name.Value, values)
+    | VariableNameTuple namesTuple, ValueTuple valuesTuple ->
+        // TODO: assert items and vt are same length
```
bettinaheim: Please insert the corresponding QsCompilerError.Verify.

src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
```diff
@@ -27,101 +29,129 @@ type ConstantPropagator(compiledCallables: ImmutableDictionary<QsQualifiedName,
...
+        let newDefault = cbListEnd |> Option.map (snd >> Value) |? stm.Default
+
+        let cbList = cbList |> List.map (fun (c, b) -> this.onPositionedBlock (c, b))
+        let newDefault = match newDefault with Value x -> this.onPositionedBlock (None, x) |> snd |> Value | Null -> Null
+
+        match cbList, newDefault with
+        | [], Value x ->
+            x.Body |> QsScopeStatement.New |> QsScopeStatement
+        | [], Null ->
+            QsScope.New ([], LocalDeclarations.New []) |> QsScopeStatement.New |> QsScopeStatement
```
bettinaheim: What is your progress in getting rid of the Scope statement you introduced?

src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
```diff
@@ -27,101 +29,129 @@ type ConstantPropagator(compiledCallables: ImmutableDictionary<QsQualifiedName,
...
+        let ex = this.ExpressionTransformation ex
+        maybe {
+            let! qualName, newScope = tryInline !stateRef ex
+            let mySet = HashSet()
+            findAllCalls !stateRef newScope mySet
+            let! currentCallable = (!stateRef).currentCallable
+            do! not (mySet.Contains currentCallable.FullName) |> check
+            let s1 = printNamespaceElem (QsCallable currentCallable)
+            let s2 = printScope 0 newScope
+            let newScope2 = this.ScopeTransformation newScope
```
bettinaheim: Can you find a better naming? If you have to start numbering your variables, the names are not descriptive enough.

src/QsCompiler/CompilerOptimizations/Utils.fs
```diff
@@ -238,3 +130,82 @@ let rec internal takeWhilePlus1 (f: 'A -> bool) (l : list<'A>) =
...
+/// Tries to inline the given expression, checking whether it's valid to do so.
+/// Returns the name of the callable and the modified scope after inlining.
+let internal tryInline (state: TransformationState) (ex: TypedExpression) =
+    maybe {
+        let! expr, arg =
+            match ex.Expression with
+            | CallLikeExpression ({Expression = expr}, arg) -> Some (expr, arg)
+            | _ -> None
+        let! qualName, specKind =
+            match expr with
```
bettinaheim: You should probably make it recursive, so it handles an arbitrary list of preceding modifiers (functors).

src/QsCompiler/CompilerOptimizations/Utils.fs
```diff
@@ -238,3 +130,82 @@ let rec internal takeWhilePlus1 (f: 'A -> bool) (l : list<'A>) =
...
+            | AdjointApplication {Expression = ControlledApplication {Expression = Identifier (GlobalCallable qualName, _)}} -> Some (qualName, QsControlledAdjoint)
+            | ControlledApplication {Expression = AdjointApplication {Expression = Identifier (GlobalCallable qualName, _)}} -> Some (qualName, QsControlledAdjoint)
+            | _ -> None
+        let callable = getCallable state qualName
+        let! impl = callable.Specializations |> Seq.tryFind (fun s -> s.Kind = specKind)
+        let! specArgs, scope =
+            match impl with
+            | {Implementation = Provided (specArgs, scope)} -> Some (specArgs, scope)
+            | {Implementation = Generated SelfInverse} ->
+                let newKind = match specKind with QsAdjoint -> Some QsBody | QsControlledAdjoint -> Some QsControlled | _ -> None
```
bettinaheim: Why do you only match two of the four cases?

src/QsCompiler/CompilerOptimizations/Utils.fs
```diff
@@ -238,3 +130,82 @@ let rec internal takeWhilePlus1 (f: 'A -> bool) (l : list<'A>) =
...
+                match callable.Specializations |> Seq.tryFind (fun s -> Some s.Kind = newKind) with
+                | Some {Implementation = Provided (specArgs, scope)} -> Some (specArgs, scope)
+                | _ -> None
+            | _ -> None
+        do! not (hasReturnStatement scope) |> check
+        let newBinding = QsBinding.New ImmutableBinding (toSymbolTuple callable.ArgumentTuple, arg)
```
bettinaheim: And what about the specArgs?

src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
```diff
@@ -27,101 +29,129 @@ type ConstantPropagator(compiledCallables: ImmutableDictionary<QsQualifiedName,
...
+
+/// The StatementKindTransformation used to simplify statements
+and [<AbstractClass>] private StatementOptimizer(stateRef: TransformationState ref) =
+    inherit StatementKindTransformation()
+
+
+    override this.onExpressionStatement ex =
+        let ex = this.ExpressionTransformation ex
+        maybe {
+            let! qualName, newScope = tryInline !stateRef ex
```
bettinaheim: I think you need to rename this file - the various functionalities are not as modular as they should be.

src/QsCompiler/CompilerOptimizations/Utils.fs
```diff
@@ -69,88 +31,34 @@ let internal rangeLiteralToSeq (r: Expr): seq<int64> =
...
+        VariableNameTuple ((Seq.map toSymbolTuple items).ToImmutableArray())
+
+/// Wraps a QsExpressionType in a basic TypedExpression
+/// The returned TypedExpression has no type param / inferred info / range information,
+/// and it should not be used for any code step that requires this information.
+let internal wrapExpr (expr: Expr) (bt: TypeKind): TypedExpression =
+    let ii = {IsMutable=false; HasLocalQuantumDependency=false}
+    TypedExpression.New (expr, ImmutableDictionary.Empty, ResolvedType.New bt, ii, Null)
+
+/// Wraps a QsStatementKind in a basic QsStatement
```
bettinaheim: What does "basic" mean?

src/QsCompiler/CompilerOptimizations/Utils.fs
```diff
@@ -69,88 +31,34 @@ let internal rangeLiteralToSeq (r: Expr): seq<int64> =
...
+let internal wrapStmt (stmt: QsStatementKind): QsStatement =
```
bettinaheim:
```suggestion
let private wrapStmt (stmt: QsStatementKind): QsStatement =
```

bettinaheim: You are using this in one place, and one place only, and there the usage of this is not correct. This needs fixing.

src/QsCompiler/CompilerOptimizations/Utils.fs
```diff
@@ -166,50 +74,34 @@ and private defaultValue (bt: TypeKind): Expr option =
...
-/// and it should not be used for any code step that requires this information.
-and internal wrapExpr (expr: Expr) (bt: TypeKind): TypedExpression =
-    let ii = {IsMutable=false; HasLocalQuantumDependency=false}
-    TypedExpression.New (expr, ImmutableDictionary.Empty, ResolvedType.New bt, ii, Null)
-
 
-/// Counts the number of MissingExprs in the given tuple
-let rec private countMissingExprs (expr: TypedExpression): int =
+/// Returns true if the given expression contains any MissingExprs
+let rec internal hasMissingExprs (expr: TypedExpression): bool =
```
bettinaheim: How about splitting out containsMissing and IsPartialApplication (SyntaxExtensions.fs) into it's own method and using that instead.

src/QsCompiler/CompilerOptimizations/Utils.fs
```diff
@@ -69,88 +31,34 @@ let internal rangeLiteralToSeq (r: Expr): seq<int64> =
...
+    QsStatement.New QsComments.Empty Null (stmt, [])
 
 
 /// Returns a new array of the given type and length.
 /// Returns None if the type doesn't have a default value.
 let rec internal constructNewArray (bt: TypeKind) (length: int): Expr option =
-    match defaultValue bt with
-    | Some x -> ImmutableArray.CreateRange (List.replicate length (wrapExpr x bt)) |> ValueArray |> Some
-    | None -> None
+    defaultValue bt |> Option.map (fun x -> ImmutableArray.CreateRange (List.replicate length (wrapExpr x bt)) |> ValueArray)
```
bettinaheim: This seems to be the only place where defaultValue is used, and contructNewArray in turn is also used in one place and one place only. Please move both of these routines to teh onNewArray override in ExpressionEvaluation.fs

src/QsCompiler/CompilerOptimizations/Utils.fs
```diff
@@ -69,88 +31,34 @@ let internal rangeLiteralToSeq (r: Expr): seq<int64> =
...
 
 /// Returns the default value for a given type (from Q# documentation)
 and private defaultValue (bt: TypeKind): Expr option =
```
bettinaheim: What happens in the None case?

src/QsCompiler/CompilerOptimizations/FunctionEvaluation.fs
```diff
@@ -6,153 +6,147 @@ open Microsoft.Quantum.QsCompiler.DataTypes
...
-    /// Returns whether a FunctionState is an interrupt
-    let isInterrupt = function Normal -> false | _ -> true
+        | BoolLiteral b -> Ok b
+        | _ -> "Not a BoolLiteral: " + (printExpr x) |> CouldNotEvaluate |> Error
 
     /// The callback for the subroutine that evaluates and simplifies an expression
-    abstract member evaluateExpression: VariablesDict -> TypedExpression -> Expr
+    abstract member evaluateExpression: TransformationState -> TypedExpression -> Expr
 
     /// Evaluates a single Q# statement
```
bettinaheim: That does not give me any information about why e.g. return and fail statements are tagged as Errors?

src/QsCompiler/CompilerOptimizations/FunctionEvaluation.fs
```diff
@@ -6,153 +6,147 @@ open Microsoft.Quantum.QsCompiler.DataTypes
...
         | QsVariableDeclaration s ->
-            fillVars vars (StringTuple.fromSymbolTuple s.Lhs, this.evaluateExpression vars s.Rhs); Normal
+            defineVarTuple state (s.Lhs, this.evaluateExpression state s.Rhs) |> Ok
         | QsValueUpdate s ->
             match s.Lhs.Expression with
-                | Identifier (LocalVariable name, tArgs) -> vars.setVar(name.Value, this.evaluateExpression vars s.Rhs); Normal
-                // TODO - allow any symbol tuple on the LHS
-                | _ -> "Unknown LHS of value update statement: " + (printExpr s.Lhs.Expression) |> CouldNotEvaluate
+            | Identifier (LocalVariable name, tArgs) -> setVar state (name.Value, this.evaluateExpression state s.Rhs) |> Ok
+            // TODO - allow any symbol tuple on the LHS
```
bettinaheim: How about implementing this TODO?

src/QsCompiler/CompilerOptimizations/FunctionEvaluation.fs
```diff
@@ -6,153 +6,147 @@ open Microsoft.Quantum.QsCompiler.DataTypes
...
+            | Some iterSeq ->
+                iterSeq |> Seq.fold (fun result loopValue ->
+                    match result with
+                    | Ok state ->
+                        let state = enterScope state
+                        let state = defineVarTuple state (fst s.LoopItem, loopValue)
+                        let result = this.evaluateScope state true s.Body
+                        match result with
+                        | Ok state -> exitScope state |> Ok
+                        | x -> x
```
bettinaheim: The assumption here is that x will never wrap a value that contains state? That seems like something that might easily change, and it would thus be better to exit the scope in any case.

src/QsCompiler/CompilerOptimizations/FunctionEvaluation.fs
```diff
@@ -6,153 +6,147 @@ open Microsoft.Quantum.QsCompiler.DataTypes
...
-                | Failed expr ->
-                    // printfn "Function %O failed with %O" name.Name.Value (prettyPrint expr)
-                    None
-                | CouldNotEvaluate reason ->
-                    // printfn "Could not evaluate function %O on arg %O for reason %O" name.Name.Value (prettyPrint arg.Expression) reason
-                    None
-            vars.exitScope()
-            res
+            let state = enterScope state
+            let state = defineVarTuple state (toSymbolTuple callable.ArgumentTuple, arg.Expression)
```
bettinaheim: Here, too, you seem to be missing the specArgs?

src/QsCompiler/CompilerOptimizations/TransformationState.fs
```diff
@@ -0,0 +1,104 @@
+﻿module Microsoft.Quantum.QsCompiler.CompilerOptimization.TransformationState
+
+open System
+open System.Collections.Generic
```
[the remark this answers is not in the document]
rsoiffer: Fixed in future PR

src/QsCompiler/CompilerOptimizations/TransformationState.fs
```diff
@@ -0,0 +1,104 @@
...
+open System.Collections.Immutable
```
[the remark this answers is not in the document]
rsoiffer: Fixed in future PR

src/QsCompiler/CompilerOptimizations/TransformationState.fs
```diff
@@ -0,0 +1,104 @@
...
+open System.Numerics
```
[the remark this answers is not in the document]
rsoiffer: Fixed in future PR

src/QsCompiler/CompilerOptimizations/TransformationState.fs
```diff
@@ -0,0 +1,104 @@
...
+open Microsoft.Quantum.QsCompiler.DataTypes
```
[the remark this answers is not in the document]
rsoiffer: Fixed in future PR

src/QsCompiler/CompilerOptimizations/TransformationState.fs
rsoiffer: Fixed in future PR

src/QsCompiler/CompilerOptimizations/ExpressionEvaluation.fs
```diff
@@ -63,32 +65,43 @@ and [<AbstractClass>] internal ExpressionKindEvaluator(vars: VariablesDict, cd:
...
             | Identifier (GlobalCallable qualName, types) ->
-                fe.evaluateFunction qualName arg types |? CallLikeExpression (method, arg)
+                do! check (maxRecursiveDepth > 0 && isLiteral !stateRef arg.Expression)
+                return! fe.evaluateFunction !stateRef qualName arg types
             | CallLikeExpression (baseMethod, partialArg) ->
-                this.Transform (partialApplyFunction baseMethod partialArg arg)
-            | _ ->
-                failwithf "Unknown function call: %O" (printExpr method.Expression)
-        else CallLikeExpression (method, arg)
+                do! hasMissingExprs partialArg |> check
```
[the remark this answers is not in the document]
rsoiffer: The line "do! check (hasMissingExprs partialArg)" should have the same effect as checking TypedExpression.IsPartialApplication. When should we prefer one over the other?

src/QsCompiler/CompilerOptimizations/Printer.fs
```diff
@@ -5,13 +5,13 @@ open Microsoft.Quantum.QsCompiler.DataTypes
 open Microsoft.Quantum.QsCompiler.SyntaxTokens
 open Microsoft.Quantum.QsCompiler.SyntaxTree
 
-open Utils
+open TransformationState
 
 
 // Pretty-print various syntax elements
```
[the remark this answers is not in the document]
rsoiffer: Will eventually be fixed in a future PR

src/QsCompiler/Tests.Compiler/TestCases/Optimizer/Inlining_input.qs
```diff
@@ -0,0 +1,24 @@
+﻿/// This namespace contains test cases for operation inlining
+namespace Microsoft.Quantum.Testing.Optimization.Inlining {
+
+    operation Test (q : Qubit) : Unit {
+		f(q, 5);
```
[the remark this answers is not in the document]
rsoiffer: Fixed in future PR

src/QsCompiler/Tests.Compiler/TestCases/Optimizer/Inlining_input.qs
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    function X (q : Qubit) : Unit {
+        body intrinsic;
+    }
+
+	function f (q : Qubit, x : Int) : Unit {
```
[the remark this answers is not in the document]
rsoiffer: Fixed in future PR

src/QsCompiler/Tests.Compiler/TestCases/Optimizer/Inlining_input.qs
```diff
@@ -0,0 +1,24 @@
+﻿/// This namespace contains test cases for operation inlining
+namespace Microsoft.Quantum.Testing.Optimization.Inlining {
```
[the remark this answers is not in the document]
rsoiffer: Fixed in future PR

src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
```diff
@@ -27,101 +29,129 @@ type ConstantPropagator(compiledCallables: ImmutableDictionary<QsQualifiedName,
...
           x
 
-    /// Marks the syntax tree as having changed
-    member this.markChanged() =
-        changed <- true
+    /// Checks whether the syntax tree changed at all
+    override this.Transform x =
+        let newX = base.Transform x
+        if x <> newX then
+            let s1 = printNamespace x
```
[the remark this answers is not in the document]
rsoiffer: Yes, we should. For some reason the first check isn't enough, and it claims that two syntax trees are unequal even when I haven't made any changes. I'm still trying to find the cause of this bug.

src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
rsoiffer: Fixed in future PR

src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
rsoiffer: Fixed in future PR

src/QsCompiler/CompilerOptimizations/Utils.fs
rsoiffer: Yeah, I'll fix that in a future PR

src/QsCompiler/CompilerOptimizations/Utils.fs
rsoiffer: Because these are the only cases for which I can find the implementation - either it's Provided, or it's self-inverse and the inverse is Provided

src/QsCompiler/CompilerOptimizations/Utils.fs
rsoiffer: Fixed in future PR

src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
rsoiffer: Fixed in future PR

src/QsCompiler/CompilerOptimizations/Utils.fs
```diff
@@ -238,3 +130,82 @@ let rec internal takeWhilePlus1 (f: 'A -> bool) (l : list<'A>) =
...
+            Seq.concat [findAllBaseStatements scope1.Body; findAllBaseStatements scope2.Body]
+        | x -> Seq.singleton x
+    )
+
+/// Returns whether this scope contains any return statement
+let rec internal hasReturnStatement (scope: QsScope): bool =
+    scope |> findAllBaseStatements |> Seq.exists (function QsReturnStatement _ -> true | _ -> false)
+
+/// Returns the number of "bottom-level" statements in this scope
+let rec internal scopeLength (scope: QsScope): int =
```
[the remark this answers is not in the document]
rsoiffer: Fixed in future PR

src/QsCompiler/CompilerOptimizations/Utils.fs
```diff
@@ -238,3 +130,82 @@ let rec internal takeWhilePlus1 (f: 'A -> bool) (l : list<'A>) =
...
+        do! not (hasReturnStatement scope) |> check
+        let newBinding = QsBinding.New ImmutableBinding (toSymbolTuple callable.ArgumentTuple, arg)
+        let newStatements = scope.Statements.Insert (0, newBinding |> QsVariableDeclaration |> wrapStmt)
+        return qualName, {scope with Statements = newStatements}
+    }
+
+/// Finds all the callables that could appear below this scope in the call stack.
+/// Mutates the given HashSet by adding all the found callables to the set.
+/// Is used to prevent inlining recursive functions into themselves forever.
+let rec internal findAllCalls (state: TransformationState) (scope: QsScope) (found: HashSet<QsQualifiedName>): unit =
```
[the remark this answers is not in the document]
rsoiffer: Fixed in future PR

src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
```diff
@@ -27,101 +29,129 @@ type ConstantPropagator(compiledCallables: ImmutableDictionary<QsQualifiedName,
...
+
+    override this.onExpressionStatement ex =
+        let ex = this.ExpressionTransformation ex
+        maybe {
+            let! qualName, newScope = tryInline !stateRef ex
+            let mySet = HashSet()
+            findAllCalls !stateRef newScope mySet
+            let! currentCallable = (!stateRef).currentCallable
+            do! not (mySet.Contains currentCallable.FullName) |> check
+            let s1 = printNamespaceElem (QsCallable currentCallable)
```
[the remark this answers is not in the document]
rsoiffer: Fixed in future PR

src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
```diff
@@ -27,101 +29,129 @@ type ConstantPropagator(compiledCallables: ImmutableDictionary<QsQualifiedName,
...
+            let s2 = printScope 0 newScope
```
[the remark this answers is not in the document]
rsoiffer: Fixed in future PR

src/QsCompiler/CompilerOptimizations/Utils.fs
```diff
@@ -238,3 +130,82 @@ let rec internal takeWhilePlus1 (f: 'A -> bool) (l : list<'A>) =
...
+            | {Implementation = Generated SelfInverse} ->
+                let newKind = match specKind with QsAdjoint -> Some QsBody | QsControlledAdjoint -> Some QsControlled | _ -> None
+                match callable.Specializations |> Seq.tryFind (fun s -> Some s.Kind = newKind) with
+                | Some {Implementation = Provided (specArgs, scope)} -> Some (specArgs, scope)
+                | _ -> None
+            | _ -> None
+        do! not (hasReturnStatement scope) |> check
+        let newBinding = QsBinding.New ImmutableBinding (toSymbolTuple callable.ArgumentTuple, arg)
+        let newStatements = scope.Statements.Insert (0, newBinding |> QsVariableDeclaration |> wrapStmt)
+        return qualName, {scope with Statements = newStatements}
```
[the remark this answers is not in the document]
rsoiffer: Fixed in future PR

src/QsCompiler/CompilerOptimizations/Utils.fs
```diff
@@ -238,3 +130,82 @@ let rec internal takeWhilePlus1 (f: 'A -> bool) (l : list<'A>) =
         else [], Some first
     | [] -> [], None
 
+
+/// Returns all the "bottom-level" statements in the current scope.
+/// Bottom-level statements are defined as those that don't contain their own scopes.
+/// This function recurses through all the subscopes of the current scope.
+let rec private findAllBaseStatements (scope: QsScope): seq<QsStatementKind> =
```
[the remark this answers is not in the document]
rsoiffer: Fixed in future PR

src/QsCompiler/CompilerOptimizations/Utils.fs
rsoiffer: It means that I'm leaving out all the optional information, so that I'm returning close to a bare wrapper around a StatementKind. I'll add more comments in a future PR.

src/QsCompiler/CompilerOptimizations/Utils.fs
rsoiffer: Fixed in future PR

src/QsCompiler/CompilerOptimizations/Utils.fs
rsoiffer: Yeah, this would be a good change

src/QsCompiler/CompilerOptimizations/Utils.fs
rsoiffer: Will be fixed in future PR

src/QsCompiler/CompilerOptimizations/Utils.fs
rsoiffer: It returns None, because it failed to create the array

src/QsCompiler/CompilerOptimizations/FunctionEvaluation.fs
rsoiffer: Error is the case of FunctionState that corresponds to a computation not proceeding normally, but instead getting interrupted, eg. by a return or fail statement. I'll add more documentation in a future PR.

src/QsCompiler/CompilerOptimizations/FunctionEvaluation.fs
rsoiffer: I will eventually, but I had some questions - why is the LHS an expression, and not a SymbolTuple as the LHS of a variable declaration? Could I handle these in the same way?

src/QsCompiler/CompilerOptimizations/FunctionEvaluation.fs
rsoiffer: We only care about the state of the function evaluation while the function is being evaluated normally. As soon as we reach an interrupt, we want to propagate this interrupt back to the top of the evaluation, and the state of the evaluation is unimportant. The interrupt doesn't even store a state.

src/QsCompiler/CompilerOptimizations/FunctionEvaluation.fs
rsoiffer: In the case of functions, is it safe to assume that specArgs is the same as callable.ArgumentTuple?

src/QsCompiler/CompilerOptimizations/TransformationState.fs
```diff
@@ -0,0 +1,104 @@
...
+/// Returns whether a given expression is a literal (and thus a constant)
+let rec internal isLiteral (state: TransformationState) (expr: Expr): bool =
+    match expr with
+    | UnitValue | IntLiteral _ | BigIntLiteral _ | DoubleLiteral _ | BoolLiteral _ | ResultLiteral _ | PauliLiteral _ -> true
+    | ValueTuple a | StringLiteral (_, a) | ValueArray a -> Seq.forall (fun x -> isLiteral state x.Expression) a
+    | RangeLiteral (a, b) -> isLiteral state a.Expression && isLiteral state b.Expression
+    | NewArray (_, a) -> isLiteral state a.Expression
+    | Identifier (GlobalCallable _, _) | MissingExpr -> true
+    | CallLikeExpression ({Expression = Identifier (GlobalCallable qualName, _)}, b)
+        when (getCallable state qualName).Kind = TypeConstructor -> isLiteral state b.Expression
```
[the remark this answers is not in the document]
bettinaheim: Did you incorporate these into the future code?

src/QsCompiler/CompilerOptimizations/TransformationState.fs
```diff
@@ -0,0 +1,104 @@
...
+    | CallLikeExpression (a, b) ->
+        isLiteral state a.Expression && isLiteral state b.Expression &&
+            TypedExpression.IsPartialApplication (CallLikeExpression (a, b))
+    | _ -> false
+
+
+/// Returns a TransformationState with the given variable defined as the given value
```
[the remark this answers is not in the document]
bettinaheim: What about this one?

src/QsCompiler/CompilerOptimizations/ExpressionEvaluation.fs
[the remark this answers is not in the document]
bettinaheim: I know it has the same effect, but my suggestion is to pipe it through the proper override to make sure code only ever needs to be adapted in one place instead of two.

src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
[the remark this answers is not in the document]
bettinaheim: It would be good to find the cause.

src/QsCompiler/CompilerOptimizations/Utils.fs
bettinaheim: Sorry, I meant why do you return None if spedKind is not QsAdjoint or QsControlledAdjoint?

src/QsCompiler/CompilerOptimizations/Utils.fs
bettinaheim: Yes, the adding more to the doc comment was where I was going. :)

src/QsCompiler/CompilerOptimizations/Utils.fs
bettinaheim: But why? You should be able to always create an array of known length, independent on the item type.

src/QsCompiler/CompilerOptimizations/FunctionEvaluation.fs
bettinaheim: Keeping the lhs an expression seemed more future proove. Right now, the expression on the lhs necessarily can only contain (mutable) symbols and tuples thereof - and missing expressions.

src/QsCompiler/CompilerOptimizations/FunctionEvaluation.fs
bettinaheim: Sure, but code is easily modified. Maybe someone at some point feels like attaching the state to interrupts.

src/QsCompiler/CompilerOptimizations/FunctionEvaluation.fs
bettinaheim: Hm, you should rather not. In the past we have discussed a couple of scenarios where this may not be the case.

src/QsCompiler/CompilerOptimizations/ExpressionEvaluation.fs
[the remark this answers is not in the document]
rsoiffer: Fixed in future PR

src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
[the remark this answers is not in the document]
rsoiffer: Yeah, I'll work on that after the next PR

src/QsCompiler/CompilerOptimizations/Utils.fs
rsoiffer: For simplicity, I don't attempt to create arrays that contain invalid expressions, such as an uninitialized function or qubit

src/QsCompiler/CompilerOptimizations/FunctionEvaluation.fs
rsoiffer: Fixed in future PR

src/QsCompiler/CompilerOptimizations/FunctionEvaluation.fs
rsoiffer: Fixed in future PR

PR #125, opened 2019-08-19
Rsoiffer/pure circuit finding

rsoiffer:
Implements pure circuit finding and operation - Q# finally automatically cancels adjacent X gates!


src/QsCompiler/Tests.Compiler/TestCases/Optimizer/LoopUnrolling_output.txt
```diff
@@ -1,22 +1,9 @@
...
-                        let r = IntLiteral 3L;
-                    }
-                }
-                {
-                    let i = IntLiteral 4L;
-                    { EmptyScope }
-                }
-            }
+            mutable r = 0;
+            set LocalVariable r = 3;
```
bettinaheim: Nice!

bettinaheim: It would be good to also have some test cases where no unrolling is/can be done.

src/QsCompiler/Tests.Compiler/TestCases/Optimizer/Reordering_output.txt
```diff
@@ -0,0 +1,14 @@
+namespace Microsoft.Quantum.Testing.Optimization.Reordering {
```
bettinaheim: Here, too, having a test where reordering is not done/possible would be good. 
According to what are you reordering? Are you pulling up all "classical" statements to the top such that you can make operation calls be directly after one another? Are you reordering the operation calls as well?

src/QsCompiler/CompilerOptimizations/CallableInlining.fs
```diff
@@ -0,0 +1,183 @@
...
+    /// Inline an expression as a ScopeStatement, with many checks to ensure correctness.
+    /// Returns None if the expression cannot be safely inlined.
+    let safeInline expr =
+        maybe {
+            let! ii = tryGetInliningInfo callables expr
+            
+            do! check (countReturnStatements ii.body = 0)
+            let! current = currentCallable
+            do! check (cannotReachCallable callables ii.body current.FullName)
+            do! check (ii.functors.controlled < 2)
```
bettinaheim: What is the reason for this restriction?

src/QsCompiler/CompilerOptimizations/CallableInlining.fs
```diff
@@ -0,0 +1,183 @@
...
+            
+            let newBinding = QsBinding.New ImmutableBinding (toSymbolTuple ii.specArgs, ii.arg)
```
bettinaheim: What are you doing to ensure that variable names do not clash?

src/QsCompiler/CompilerOptimizations/CallableInlining.fs
```diff
@@ -0,0 +1,183 @@
...
+
+                override exprKind.onOperationCall (method, arg) =
+                    let expr = CallLikeExpression (exprKind.ExpressionTransformation method, exprKind.ExpressionTransformation arg)
+                    maybe {
+                        let! scopeStatement = safeInline expr
+                        statementsToAdd <- statementsToAdd @ [wrapStmt scopeStatement]
+                        return UnitValue
+                    } |? expr
+
+                // The functions before are overriden to prevent inlining of lazily evaluated operations
```
bettinaheim: I am not sure I understand; the things you are skipping the transformation for are expressions - I was under then impression that this inlining routine would only consider inlining for expression statements?

src/QsCompiler/CompilerOptimizations/CallableInlining.fs
```diff
@@ -0,0 +1,183 @@
...
+            override stmtKind.LocationTransformation x = scope.onLocation x
+            override stmtKind.ScopeTransformation x = scope.Transform x
+            override stmtKind.TypeTransformation x = scope.Expression.Type.Transform x
+
+            override this.onExpressionStatement ex =
+                match safeInline ex.Expression with
+                | Some result -> result
+                | None -> QsExpressionStatement (this.ExpressionTransformation ex)
+        }
+    }
```
bettinaheim: Nice job refactoring this!

src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
```diff
@@ -69,89 +68,54 @@ type ConstantPropagator(compiledCallables: ImmutableDictionary<QsQualifiedName,
...
+            override this.onQubitScope (stm : QsQubitScope) =
+                let kind = stm.Kind
+                let lhs = this.onSymbolTuple stm.Binding.Lhs
+                let rhs = this.onQubitInitializer stm.Binding.Rhs
+
+                jointFlatten (lhs, rhs) |> Seq.iter (fun (l, r) ->
+                    match l, r.Resolution with
+                    | VariableName name, QubitRegisterAllocation {Expression = IntLiteral num} ->
+                        let arrayIden = Identifier (LocalVariable name, Null) |> wrapExpr (ArrayType (ResolvedType.New Qubit))
+                        let elemI = fun i -> ArrayItem (arrayIden, IntLiteral (int64 i) |> wrapExpr Int)
```
bettinaheim: Not that I think any executable program will allocate more qubits that can be describe by a 32-bit int... but technically this is not according to spec. Even for cases that do not seem like reasonable use cases it's always good to also think about exploits.

src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
```diff
@@ -69,89 +68,54 @@ type ConstantPropagator(compiledCallables: ImmutableDictionary<QsQualifiedName,
...
+                        let expr = List.init (int num) (elemI >> wrapExpr Qubit) |> ImmutableArray.CreateRange |> ValueArray
```
bettinaheim:
```suggestion
                        let expr = Seq.init (int num) (elemI >> wrapExpr Qubit) |> ImmutableArray.CreateRange |> ValueArray
```

src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
```diff
@@ -69,89 +68,54 @@ type ConstantPropagator(compiledCallables: ImmutableDictionary<QsQualifiedName,
...
+                result
+
+            override this.onQubitScope (stm : QsQubitScope) =
+                let kind = stm.Kind
+                let lhs = this.onSymbolTuple stm.Binding.Lhs
+                let rhs = this.onQubitInitializer stm.Binding.Rhs
+
+                jointFlatten (lhs, rhs) |> Seq.iter (fun (l, r) ->
+                    match l, r.Resolution with
+                    | VariableName name, QubitRegisterAllocation {Expression = IntLiteral num} ->
```
bettinaheim: Do you do a similar thing for all new arrays?

src/QsCompiler/CompilerOptimizations/Evaluation.fs
```diff
@@ -0,0 +1,366 @@
...
+    override this.onMultiplication (lhs, rhs) =
+        this.arithNumBinaryOp MUL (*) (*) (*) lhs rhs
+
+    override this.onDivision (lhs, rhs) =
+        this.arithNumBinaryOp DIV (/) (/) (/) lhs rhs
+
+    override this.onExponentiate (lhs, rhs) =
+        let lhs, rhs = this.simplify (lhs, rhs)
+        match lhs.Expression, rhs.Expression with
+        | BigIntLiteral a, IntLiteral b -> BigIntLiteral (BigInteger.Pow(a, Convert.ToInt32(b)))
```
bettinaheim: The silent conversion to int 32 still needs to be replaced - it will fail according to spec, but should not fail silently.

src/QsCompiler/CompilerOptimizations/Evaluation.fs
```diff
@@ -0,0 +1,366 @@
...
+        | ValueArray va, IntLiteral i -> va.[int i].Expression
+        | ValueArray va, RangeLiteral _ when isLiteral callables idx.Expression ->
+            rangeLiteralToSeq idx.Expression |> Seq.map (fun i -> va.[int i]) |> ImmutableArray.CreateRange |> ValueArray
+        | _ -> ArrayItem (arr, idx)
+
+    override this.onNewArray (bt, idx) =
+        let idx = this.simplify idx
+        match idx.Expression with
+        | IntLiteral i -> constructNewArray bt.Resolution (int i) |? NewArray (bt, idx)
+        // TODO - handle array slicing
```
bettinaheim: How about adding that? :)

src/QsCompiler/CompilerOptimizations/Evaluation.fs
```diff
@@ -0,0 +1,366 @@
...
+        | _ -> NewArray (bt, idx)
+ 
+    override this.onCopyAndUpdateExpression (lhs, accEx, rhs) =
+        let lhs, accEx, rhs = this.simplify (lhs, accEx, rhs)
+        match lhs.Expression, accEx.Expression with
+        | ValueArray va, IntLiteral i -> ValueArray (va.SetItem(int i, rhs))
+        // TODO - handle array slicing
```
bettinaheim: same for the two todos here

src/QsCompiler/CompilerOptimizations/Utils.fs
```diff
@@ -134,7 +167,7 @@ let rec internal takeWhilePlus1 (f: 'A -> bool) (l : list<'A>) =
 /// Returns all the "bottom-level" statements in the current scope.
 /// Bottom-level statements are defined as those that don't contain their own scopes.
 /// This function recurses through all the subscopes of the current scope.
-let rec private findAllBaseStatements (scope: QsScope): seq<QsStatementKind> =
+let rec internal findAllBaseStatements (scope: QsScope): seq<QsStatementKind> =
```
bettinaheim: As mentioned on a previous review, this needs to be replaced by a suitable transformation.

src/QsCompiler/CompilerOptimizations/Utils.fs
```diff
@@ -41,12 +67,25 @@ let rec internal toSymbolTuple (x: QsTuple<LocalVariableDeclaration<QsLocalSymbo
     | QsTuple items when items.Length = 1 ->
         toSymbolTuple items.[0]
     | QsTuple items ->
-        VariableNameTuple ((Seq.map toSymbolTuple items).ToImmutableArray())
+        Seq.map toSymbolTuple items |> ImmutableArray.CreateRange |> VariableNameTuple
+
+/// Matches a TypedExpression as a SymbolTuple
```
bettinaheim: It's probably worth mentioning what the default case is.

bettinaheim:
@rsoiffer One other thing before I forget: you also need to add the copyright header with the MIT license to all files: 

// Copyright (c) Microsoft Corporation. All rights reserved.

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -220,12 +220,7 @@ public CompilationLoader(SourceLoader loadSources, ReferenceLoader loadReference
 
             // executing the specified optimization steps
 
-            var constantPropagator = new ConstantPropagator(VerifiedCompilation.Callables);
-            while (constantPropagator.checkChanged())
-            {
-                this.GeneratedSyntaxTree = this.GeneratedSyntaxTree.Select(p => constantPropagator.Transform(p)).ToList();
-            }
-
+            this.GeneratedSyntaxTree = Optimizations.optimize(this.GeneratedSyntaxTree);
```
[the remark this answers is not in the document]
rsoiffer: I assume that the syntax tree passed to the optimizer is valid. If this is the case, the optimizer should not throw any exceptions. If this is not the case, the optimizer will throw an exception describing how the input is invalid. The optimizer does not catch any exceptions.

src/QsCompiler/CompilerOptimizations/CallableInlining.fs
```diff
@@ -0,0 +1,183 @@
...
+        | false, 0 -> QsBody
+        | true, 0 -> QsAdjoint
+        | false, _ -> QsControlled
+        | true, _ -> QsControlledAdjoint
+    member this.withAdjoint = { this with adjoint = not this.adjoint }
+    member this.withControlled = { this with controlled = this.controlled + 1 }
+
+
+/// Tries to decompose a method expression into the method name and the functors applied.
+/// Returns None if the input is not a valid method expression.
```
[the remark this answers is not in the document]
rsoiffer: Comment clarified in new commit

src/QsCompiler/CompilerOptimizations/CallableInlining.fs
```diff
@@ -0,0 +1,183 @@
...
+    | Identifier (GlobalCallable qualName, _) -> Some (qualName, Functors.None)
+    | AdjointApplication ex ->
+        tryGetQualNameAndFunctors ex |> Option.map (fun (qualName, functors) -> qualName, functors.withAdjoint)
+    | ControlledApplication ex ->
+        tryGetQualNameAndFunctors ex |> Option.map (fun (qualName, functors) -> qualName, functors.withControlled)
+    | _ -> None
+
+
+/// Tries to split a callable invocation into the functors applied, the callable, and the argument.
+/// Returns None if the input is not a valid callable invocation.
```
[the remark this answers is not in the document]
rsoiffer: Comment clarified in new commit

src/QsCompiler/CompilerOptimizations/CallableInlining.fs
```diff
@@ -0,0 +1,183 @@
...
+type private InliningInfo = {
+    functors: Functors
+    callable: QsCallable
+    arg: TypedExpression
+    specArgs: QsArgumentTuple
+    body: QsScope
+}
+
+/// Tries to construct an InliningInfo from the given expression.
+/// Returns None if the expression is not a callable invocation that can be inlining.
```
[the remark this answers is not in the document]
rsoiffer: Comment clarified in new commit

src/QsCompiler/CompilerOptimizations/CallableInlining.fs
```diff
@@ -0,0 +1,183 @@
...
+let private tryGetInliningInfo callables expr =
+    maybe {
+        let! functors, callable, arg = trySplitCall callables expr
+        let! specArgs, body = tryGetProvidedImpl callable functors
+        return { functors = functors; callable = callable; arg = arg; specArgs = specArgs; body = body }
+    }
+
+
+/// Recursively finds all the callables that could be inlined into the given scope.
+/// Includes callables that could be indirectly called or inlined into this scope.
```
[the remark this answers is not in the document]
rsoiffer: Comment clarified in new commit

src/QsCompiler/CompilerOptimizations/CallableInlining.fs
```diff
@@ -0,0 +1,183 @@
...
+    not (mySet.Contains cannotReach)
+
+    
+/// The SyntaxTreeTransformation used to inline callables
+type internal CallableInliner(callables) =
+    inherit OptimizingTransformation()
+
+    // The current callable we're in the process of transforming
+    let mutable currentCallable: QsCallable option = None
+    // The list of new statements to add
```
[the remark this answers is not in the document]
rsoiffer: This code is currently unfinished, and is unneeded if we only care about inlining callables that are invoked as ExpressionStatements. If we do care about inlining callables that are used as part of other expressions, then I will have to finish implementing this code. Otherwise, I can remove this code. How much do we care about inlining non-expression-statements?

src/QsCompiler/DataStructures/SyntaxExtensions.fs
```diff
@@ -155,16 +155,20 @@ type TypedExpression with
 
     // utils for walking the data structure
 
+    /// Returns true if the expression contains any missing expression.
+    /// Returns false otherwise.
+    static member public containsMissing (ex : TypedExpression) =
```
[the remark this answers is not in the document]
rsoiffer: Fixed in latest commit

src/QsCompiler/DataStructures/SyntaxExtensions.fs
```diff
@@ -155,16 +155,20 @@ type TypedExpression with
...
+        // we only need to check for top-level missing items
+        match ex.TupleItems with
+        | Some items when items.Length > 1 -> items |> List.exists TypedExpression.containsMissing
```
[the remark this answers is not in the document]
rsoiffer: Fixed in latest commit

src/QsCompiler/Tests.Compiler/OptimizationTests.fs
```diff
@@ -77,3 +75,8 @@ let ``miscellaneous`` () =
 let ``partial evaluation`` () =
     // createTestCase "TestCases/Optimizer/PartialEval"
     assertOptimization "TestCases/Optimizer/PartialEval"
+
+[<Fact>]
+let ``reordering`` () =
```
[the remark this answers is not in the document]
rsoiffer: It's exchanging a quantum expression statement with a classical let statement.

src/QsCompiler/Tests.Compiler/TestCases/Optimizer/Arithmetic_input.qs
```diff
@@ -1,10 +1,11 @@
 ﻿/// This namespace contains test cases for arithmetic optimization
 namespace Microsoft.Quantum.Testing.Optimization.Arithmetic {
-    operation Test () : Unit {
+    operation Test () : Int {
 	    let x = 5;
 	    let y = x + 3;
 	    let z = y * 5 % 4;
 	    let w = z > 1;
 	    let a = w ? 2 | y / 2;
+		return a;
```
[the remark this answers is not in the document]
rsoiffer: Fixed in latest commit

src/QsCompiler/CompilerOptimizations/CallableInlining.fs
rsoiffer: I check there is at most one Controlled functor because handling multiple Controlled functors means that I have to pull out all the controls from the nested argument tuple and combine them into a single array, which I haven't gotten around to implementing yet.

src/QsCompiler/CompilerOptimizations/CallableInlining.fs
rsoiffer: The VariableRenaming transformation is called by another optimization step after this one, and it transforms variable names to be unique.

src/QsCompiler/CompilerOptimizations/CallableInlining.fs
rsoiffer: This is currently dead code that is there to support inlining callables inside expressions in the future (see earlier comment).

src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
```diff
@@ -1,65 +1,64 @@
...
-        stateRef := {!stateRef with currentCallable = Some c}
-        let result = base.onCallableImplementation c
-        stateRef := {!stateRef with currentCallable = prev}
-        result
+type internal ConstantPropagator(callables) =
+    inherit OptimizingTransformation()
+    
+    /// The current dictionary of values to substitute for variables
+    let mutable constants = Constants []
+    /// Whether we should skip entering the next scope we encounter
```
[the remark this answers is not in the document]
rsoiffer: Yeah, having a counter is a better solution - implemented in latest commit.

src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
```diff
@@ -1,65 +1,64 @@
...
-    override this.onCallableImplementation c =
-        let prev = (!stateRef).currentCallable
-        stateRef := {!stateRef with currentCallable = Some c}
-        let result = base.onCallableImplementation c
-        stateRef := {!stateRef with currentCallable = prev}
-        result
+type internal ConstantPropagator(callables) =
+    inherit OptimizingTransformation()
+    
+    /// The current dictionary of values to substitute for variables
```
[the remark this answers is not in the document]
rsoiffer: Comment clarified in latest commit

src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
rsoiffer: I made it throw an exception if given an integer that requires more than 32 bits

src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
rsoiffer: I don't think it really matters whether we use List or Seq in this case, but fixed in latest commit.

src/QsCompiler/CompilerOptimizations/ConstantPropagation.fs
rsoiffer: Yes, the code for that is in ExpressionKindEvaluator

src/QsCompiler/CompilerOptimizations/Evaluation.fs
rsoiffer: I made it throw an ArgumentException if the exponent is too big for 32 bits

src/QsCompiler/CompilerOptimizations/Utils.fs
rsoiffer: What do you mean? I don't think this match statement has a default case.

rsoiffer: The default case is just anything that's not a SymbolTuple

src/QsCompiler/CompilerOptimizations/StatementRemoving.fs
```diff
@@ -0,0 +1,129 @@
...
+/// Reorders a list of statements such that the pure classical statements occur before the pure quantum statements
+let rec private reorderStatements (stmtList: QsStatement list): QsStatement list =
+    match stmtList with
+    | [] -> []
+    | head :: tail ->
+        let tail = reorderStatements tail
+        match head :: tail with
+        | head1 :: head2 :: tail
+            when isPureQuantum head1.Statement && isPureClassical head2.Statement ->
+            head2 :: reorderStatements (head1 :: tail)
```
bettinaheim: I'm sure you can tell me the complexity of this routine. :) 
Can you do better?

src/QsCompiler/CompilerOptimizations/StatementRemoving.fs
```diff
@@ -0,0 +1,129 @@
...
+    inherit OptimizingTransformation()
+
+    /// The VariableRenamed used to ensure unique variable names
+    let mutable renamer = None
+
+    /// Returns whether the given statement is needed.
+    /// If this returns false, the given statement can be removed without changing the code.
+    let isStatementNeeded stmt =
+        match stmt.Statement with
+        | QsVariableDeclaration s when isAllDiscarded s.Lhs && not (s.Rhs.InferredInformation.HasLocalQuantumDependency) -> false
```
bettinaheim: Please add a comment that this is correct as long as any classical computation is side effect free.

bettinaheim: And actually the same is true for value updates (they are not needed if all all expression on the left hand side are missing and the rhs has no quantum dependencies).

src/QsCompiler/CompilerOptimizations/StatementRemoving.fs
```diff
@@ -0,0 +1,129 @@
...
+        | QsScopeStatement s when s.Body.Statements.IsEmpty -> false
+        | QsForStatement s when s.Body.Statements.IsEmpty -> false
+        | QsWhileStatement s when s.Body.Statements.IsEmpty -> false
```
bettinaheim: Please verify that the condition does not have a quantum dependency such that you are not relying on the assumption that the condition cannot have a quantum dependency. 

Also, an if statement is not needed if all blocks are empty and none of the conditions have any quantum dependencies. Similarly for a repeat statement, and in particular also for a qubit scope.

src/QsCompiler/CompilerOptimizations/StatementRemoving.fs
```diff
@@ -0,0 +1,129 @@
...
+    /// The VariableRenamed used to ensure unique variable names
+    let mutable renamer = None
+
+    /// Returns whether the given statement is needed.
+    /// If this returns false, the given statement can be removed without changing the code.
+    let isStatementNeeded stmt =
+        match stmt.Statement with
+        | QsVariableDeclaration s when isAllDiscarded s.Lhs && not (s.Rhs.InferredInformation.HasLocalQuantumDependency) -> false
+        | QsScopeStatement s when s.Body.Statements.IsEmpty -> false
+        | QsForStatement s when s.Body.Statements.IsEmpty -> false
```
bettinaheim: You need to check if the value to iterate over has a quantum dependency.

src/QsCompiler/CompilerOptimizations/StatementRemoving.fs
```diff
@@ -0,0 +1,129 @@
...
+    /// Used to split variable declarations across lines, and to remove scope statements.
+    /// If a statement should not be split, returns a sequence with a single element.
+    let splitStatement stmt =
+        match stmt.Statement with
+        | QsVariableDeclaration s ->
+            match s.Lhs, s.Rhs with
+            | Tuple l, Tuple r ->
+                if l.Length <> r.Length then
+                    ArgumentException "Tuple lenghts do not match" |> raise
+                Seq.zip l r |> Seq.map (QsBinding.New s.Kind >> QsVariableDeclaration >> wrapStmt)
```
bettinaheim: It seems weird that this is not recursive - why only split the top level?

src/QsCompiler/CompilerOptimizations/StatementRemoving.fs
```diff
@@ -0,0 +1,129 @@
...
+            | _ -> Seq.singleton stmt
+        | QsScopeStatement s -> s.Body.Statements |> Seq.cast
```
bettinaheim: Documenting that this function should only be called after making sure all variable names are unique would be good.

src/QsCompiler/CompilerOptimizations/StatementRemoving.fs
```diff
@@ -0,0 +1,129 @@
...
+            when isPureQuantum head1.Statement && isPureClassical head2.Statement ->
+            head2 :: reorderStatements (head1 :: tail)
+        | a -> a
+
+
+/// The SyntaxTreeTransformation used to remove useless statements
+type internal StatementRemover() =
+    inherit OptimizingTransformation()
+
+    /// The VariableRenamed used to ensure unique variable names
```
bettinaheim:
```suggestion
    /// The VariableRenamer used to ensure unique variable names
```

src/QsCompiler/CompilerOptimizations/StatementRemoving.fs
```diff
@@ -0,0 +1,129 @@
...
+    | head :: tail ->
+        let tail = reorderStatements tail
+        match head :: tail with
+        | head1 :: head2 :: tail
+            when isPureQuantum head1.Statement && isPureClassical head2.Statement ->
+            head2 :: reorderStatements (head1 :: tail)
+        | a -> a
+
+
+/// The SyntaxTreeTransformation used to remove useless statements
```
bettinaheim: Looks like this is doing a bit more than just removing unnecessary statements. Consider whether splitting the functionality into individual transformations makes sense, and update the doc comment in any case.

src/QsCompiler/Compiler/CompilationLoader.cs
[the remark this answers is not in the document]
bettinaheim: In that case I suggest to catch all exceptions there and log them properly, indicating that this compilation step failed.

src/QsCompiler/CompilerOptimizations/CallableInlining.fs
bettinaheim: Ok. How about a adding todo then?

src/QsCompiler/CompilerOptimizations/CallableInlining.fs
bettinaheim: Seems rather problematic 1.) that you inline first and then only resolve to unique names and 2) that you couple these two steps together this way.

src/QsCompiler/CompilerOptimizations/CallableInlining.fs
bettinaheim: Then I suggest to remove it before merging. Code should be useful or removed for the sake of maintainability.

src/QsCompiler/CompilerOptimizations/Utils.fs
bettinaheim: Yes, but it would be nice to point out in the doc comment.

src/QsCompiler/CompilerOptimizations/VariableRenaming.fs
```diff
@@ -0,0 +1,124 @@
...
+open Microsoft.Quantum.QsCompiler.SyntaxTree
+open Microsoft.Quantum.QsCompiler.Transformations.Core
+
+open ComputationExpressions
+open Types
+open Utils
+
+
+/// The ScopeTransformation used to ensure unique variable names
+type VariableRenamer(argTuple: QsArgumentTuple) =
```
bettinaheim: Are there any tests for this? It would be good to add an extensive set of tests.

src/QsCompiler/CompilerOptimizations/VariableRenaming.fs
```diff
@@ -0,0 +1,124 @@
...
+        override stmtKind.ScopeTransformation x = scope.Transform x
+        override stmtKind.TypeTransformation x = scope.Expression.Type.Transform x
+
+        override this.onSymbolTuple syms =
+            match syms with
+            | VariableName item -> VariableName (NonNullable<_>.New (generateUniqueName item.Value))
+            | VariableNameTuple items -> Seq.map this.onSymbolTuple items |> ImmutableArray.CreateRange |> VariableNameTuple
+            | InvalidItem | DiscardedItem -> syms
+
+        override this.onVariableDeclaration stm =
```
bettinaheim: What is the reason for this override? The only thing you changed compared to the default implementation is the order in which the lhs and rhs are processed - which should be irrelevant here.

src/QsCompiler/CompilerOptimizations/VariableRenaming.fs
```diff
@@ -0,0 +1,124 @@
...
+            let rhs = this.ExpressionTransformation stm.Rhs
+            let lhs = this.onSymbolTuple stm.Lhs
+            QsBinding<TypedExpression>.New stm.Kind (lhs, rhs) |> QsVariableDeclaration
+
+        override this.onForStatement stm =
```
bettinaheim: same question here

src/QsCompiler/CompilerOptimizations/VariableRenaming.fs
```diff
@@ -0,0 +1,124 @@
...
+            QsForStatement.New ((loopVar, loopVarType), iterVals, body) |> QsForStatement
+
+        override this.onRepeatStatement stm =
+            constants <- enterScope constants
+            skipScope <- true
+            let result = base.onRepeatStatement stm
+            constants <- exitScope constants
+            result
+
+        override this.onQubitScope (stm : QsQubitScope) =
```
bettinaheim: and same question here

bettinaheim: Ok, I think I get it: you have written this basically specifically to remove shadowing - i.e. this transformation only does something if you actually have shadowed variables due to a prior inlining. It seems a bit of a risk that 1) you actually have to overload all statements that do any kind of binding and 2) you are duplicating the logic regarding the life times of variables. I am not sure it's worth trying to minimize the renaming in this way. What are your thoughts here.

src/QsCompiler/CompilerOptimizations/VariableRenaming.fs
```diff
@@ -0,0 +1,124 @@
...
+open Microsoft.Quantum.QsCompiler.SyntaxTokens
+open Microsoft.Quantum.QsCompiler.SyntaxTree
+open Microsoft.Quantum.QsCompiler.Transformations.Core
+
+open ComputationExpressions
+open Types
+open Utils
+
+
+/// The ScopeTransformation used to ensure unique variable names
```
bettinaheim: You need to be more specific regarding what the transformation does - see my last comment on this file.

src/QsCompiler/CompilerOptimizations/PureCircuitAPI.fs
```diff
@@ -0,0 +1,211 @@
...
+
+/// A pure, parameter-free quantum circuit
+type Circuit = {
+    numQubitRefs: int
+    gates: GateCall list
+}
+
+/// Metadata used for constructing circuits
+type private CircuitContext = {
+    qubitReferences: string list
```
bettinaheim: Could you describe in more detail what this is used for?

src/QsCompiler/CompilerOptimizations/PureCircuitAPI.fs
```diff
@@ -0,0 +1,211 @@
...
+}
+
+
+/// Converts a TypedExpression to a Const, mutating the given CircuitContext as needed.
+/// Returns None if the given expression cannot be converted to a constant.
+let rec private exprToConst (cc: CircuitContext ref) (expr: TypedExpression): Const option =
```
bettinaheim: What is the difference between this routine here and the constant folding one? Is there a way to leverage the same logic?

src/QsCompiler/CompilerOptimizations/PureCircuitAPI.fs
```diff
@@ -0,0 +1,211 @@
...
+
+
+/// Given a pure circuit, performs basic optimizations and returns the new circuit
+let private optimizeCircuit (circuit: Circuit): Circuit =
+    let mutable circuit = circuit
+    let mutable i = 0
+    while i < circuit.gates.Length - 1 do
+        if circuit.gates.[i] = { circuit.gates.[i+1] with adjoint = not circuit.gates.[i+1].adjoint } then
+            circuit <- { circuit with gates = removeIndices [i; i+1] circuit.gates}
+        elif circuit.gates.[i] = circuit.gates.[i+1] && List.contains circuit.gates.[i].gate.Name.Value ["X"; "Z"; "H"; "CNOT"] then
```
bettinaheim: Ideally you could leverage the operation metadata tracked in the Q# compilation here rather than hardcode this list.

src/QsCompiler/CompilerOptimizations/Printer.fs
```diff
@@ -5,12 +5,19 @@ open Microsoft.Quantum.QsCompiler.DataTypes
...
 
 
+let internal printIdentifier (iden: Identifier): string =
+    match iden with
+    | LocalVariable a -> a.Value
+    | GlobalCallable a -> a.Name.Value
+    | InvalidIdentifier -> "__invalid__"
+
+
 let rec internal printExpr (expr: Expr): string =
```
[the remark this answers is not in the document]
rsoiffer: I'll remove the Printer class once I fix the issues in QsCodeOutput that make the order of the generated code non-deterministic.

src/QsCompiler/CompilerOptimizations/StatementRemoving.fs
```diff
@@ -0,0 +1,129 @@
...
+| _ -> false
+
+/// Returns whether a statement is purely quantum.
+/// The statement must have no classical side effects, but can have quantum side effects.
+let private isPureQuantum = function
+| QsVariableDeclaration s when isAllDiscarded s.Lhs -> true
+| QsExpressionStatement _ -> true
+| _ -> false
+
+/// Reorders a list of statements such that the pure classical statements occur before the pure quantum statements
```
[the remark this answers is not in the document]
rsoiffer: Added to todo list

src/QsCompiler/CompilerOptimizations/StatementRemoving.fs
rsoiffer: Added to todo list

src/QsCompiler/CompilerOptimizations/StatementRemoving.fs
rsoiffer: Added to todo list

src/QsCompiler/CompilerOptimizations/PureCircuitFinding.fs
```diff
@@ -0,0 +1,51 @@
...
+                        printfn "Removed %d gates" (circuit.Length - newCircuit.Length)
+                        printfn "Old: %O" (List.map (fun x -> printExpr x.Expression) circuit)
+                        printfn "New: %O" (List.map (fun x -> printExpr x.Expression) newCircuit)
+                        printfn ""*)
+                    newStatements <- newStatements @ List.map (QsExpressionStatement >> wrapStmt) newCircuit
+                    circuit <- []
+
+            for stmt in x.Statements do
+                match stmt.Statement with
+                | QsExpressionStatement expr -> circuit <- circuit @ [expr]
```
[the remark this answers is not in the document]
rsoiffer: Added to todo list

src/QsCompiler/CompilerOptimizations/PureCircuitFinding.fs
```diff
@@ -0,0 +1,51 @@
...
+                | _ ->
+                    finishCircuit()
+                    newStatements <- newStatements @ [scope.onStatement stmt]
```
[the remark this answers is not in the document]
rsoiffer: Added to todo list

src/QsCompiler/CompilerOptimizations/Evaluation.fs
```diff
@@ -0,0 +1,368 @@
...
+                    constantsRef := enterScope !constantsRef
+                    constantsRef := defineVarTuple (isLiteral callables) !constantsRef (fst stmt.LoopItem, loopValue)
+                    let! s = this.evaluateScope !constantsRef true stmt.Body
+                    constantsRef := exitScope s
+                return constantsRef.Value
+            }
+        | QsWhileStatement stmt ->
+            result {
+                let constantsRef = ref constants
+                while this.evaluateExpression !constantsRef stmt.Condition |> castToBool do
```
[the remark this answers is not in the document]
rsoiffer: Added to todo list

src/QsCompiler/CompilerOptimizations/Evaluation.fs
```diff
@@ -0,0 +1,368 @@
...
+                return constantsRef.Value
+            }
+        | QsRepeatStatement stmt ->
+            result {
+                let constantsRef = ref constants
+                constantsRef := enterScope !constantsRef
+                let! s = this.evaluateScope !constantsRef false stmt.RepeatBlock.Body
+                constantsRef := s
+                while this.evaluateExpression !constantsRef stmt.SuccessCondition |> castToBool do
+                    let! s = this.evaluateScope !constantsRef false stmt.FixupBlock.Body
```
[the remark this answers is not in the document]
rsoiffer: I could, but I think separating them is more explicit about how they behave

src/QsCompiler/CompilerOptimizations/VariableRenaming.fs
```diff
@@ -0,0 +1,124 @@
...
+type VariableRenamer(argTuple: QsArgumentTuple) =
+    inherit ScopeTransformation()
+
+    /// The number of times a variable is referenced
+    let mutable numUses = Map.empty
+    /// The current dictionary of new names to substitute for variables
+    let mutable constants = enterScope (Constants [])
+    /// Whether we should skip entering the next scope we encounter
+    let mutable skipScope = false
```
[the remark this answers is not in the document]
rsoiffer: Added to todo list

src/QsCompiler/CompilerOptimizations/VariableRenaming.fs
```diff
@@ -0,0 +1,124 @@
...
+
+    /// Given a possibly-mangled variable name, returns the original variable name
+    let demangle varName =
+        let m = Regex.Match (varName, "^__qsVar\d+__(.+)__$")
```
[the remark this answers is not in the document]
rsoiffer: Added to todo list

src/QsCompiler/CompilerOptimizations/VariableRenaming.fs
rsoiffer: Added to todo list

src/QsCompiler/CompilerOptimizations/PureCircuitAPI.fs
rsoiffer: No, the code is almost entirely different, and I don't think I could reuse the logic

src/QsCompiler/CompilerOptimizations/PureCircuitAPI.fs
rsoiffer: Added to todo list

src/QsCompiler/CompilerOptimizations/StatementRemoving.fs
bettinaheim: Yes, splitting would be good.

src/QsCompiler/CompilerOptimizations/Evaluation.fs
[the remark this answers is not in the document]
bettinaheim: Having something like a maximum count makes sense.

src/QsCompiler/CompilerOptimizations/Evaluation.fs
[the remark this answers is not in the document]
bettinaheim: Duplicating code that servers the same functionality and always will is something I recommend avoiding. It's a matter of maintainablilty.

src/QsCompiler/CompilerOptimizations/Evaluation.fs
```diff
@@ -0,0 +1,368 @@
...
+                constantsRef := exitScope constantsRef.Value
+            return !constantsRef
+        }
+
+    /// Evaluates a Q# function
+    member internal this.evaluateFunction (name: QsQualifiedName) (arg: TypedExpression) (types: QsNullable<ImmutableArray<ResolvedType>>): Expr option =
+        let callable = getCallable callables name
+        if callable.Specializations.Length <> 1 then
+            Exception "Functions must have exactly one specialization" |> raise
+        let impl = callable.Specializations.[0].Implementation
```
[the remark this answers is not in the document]
bettinaheim: IEnumerable does.

PR #128, opened 2019-08-21
Support for Conjugations as Statements

bettinaheim:
This PR adds support for defining statement that generate a patterns of the for U*VU (applied from right to left) - the same functionality as provided by a call to ApplyWith. 
U and V are defined via two statement blocks. The usual restrictions that hold for autogeneration of the adjoint apply to the outer block U. The only restriction on the inner block V is that it cannot reassign mutable variable that are used within the outer block U, thus guaranteeing that the generated block is indeed "undoing" the quantum computation. The statement is valid within operations only.


src/QsCompiler/CommandLineTool/Commands/Diagnose.cs
```diff
@@ -56,6 +56,10 @@ public static IEnumerable<Example> UsageExamples
             [Option("code", Required = false, Default = false,
             HelpText = "Specifies whether to print the Q# code generated based on the built syntax tree.")]
             public bool PrintCompiledCode { get; set; }
+
+            [Option("trim", Required = false, Default = false,
+            HelpText = "Specifies whether to trim the syntax tree to eliminate selective abstractions.")]
```
cgranade: I'm not sure I understand what this option does from the `HelpText`, sorry. As a side question, is it possible to expose this through an MSBuild property?

src/QsCompiler/Tests.Compiler/TestCases/GlobalVerification.qs
```diff
@@ -365,6 +364,20 @@ namespace Microsoft.Quantum.Testing.GlobalVerification {
         }
     }
 
+    operation AllPathsReturnValue13 () : Int {
+        within {
+            return 1;
```
cgranade: What are the semantics for returning from within a `within` block? How does that inline if the `apply` block isn't empty?

src/QsCompiler/Tests.Compiler/TestCases/GlobalVerification.qs
```diff
@@ -365,6 +364,20 @@ namespace Microsoft.Quantum.Testing.GlobalVerification {
...
+        }
+        apply {}
+    }
+
+    operation AllPathsReturnValue14 () : Int {
+        within {}
+        apply {
+            return 1;
```
cgranade: Similarly here, what are the semantics for when `within` isn't empty?

src/VSCodeExtension/syntaxes/qsharp.tmLanguage.json.v.template
```diff
@@ -46,11 +46,11 @@
       "patterns": [
         {
           "name": "keyword.control.qsharp",
-          "match": "\\b(if|elif|else|repeat|until|fixup|for|while|in|\\.\\.|w\/|return|fail)\\b"
+          "match": "\\b(using|borrowing|mutable|let|set|if|elif|else|repeat|until|fixup|for|in|while|return|fail|within|apply)\\b"
```
cgranade: We'll likely need to update the IQ# and LaTeX syntax highlighting as well.

src/QsCompiler/CommandLineTool/Commands/Diagnose.cs
bettinaheim: The trimming currently inlines all conjugations. This will need to be revised and properly integrated with the whole setup for targeting later, so this is mostly temporary and the handle only exists for the "diagnose" verb (that is specifically for debugging the compiler really). For building, then trimming is currently always done.

src/QsCompiler/Tests.Compiler/TestCases/GlobalVerification.qs
bettinaheim: This specific case actually results in an error : ErrorCode.ReturnStatementWithinAutoInversion.

src/QsCompiler/Tests.Compiler/TestCases/GlobalVerification.qs
bettinaheim: This is valid code currently, since it seemed that the behavior seems fairly clear (it actually returns and does not execute further - same as if you return somehere else in a statement). But let's have a quick chat on whether that makes sense to keep, or if there are any reasons why we would want to give an error in this case.

src/VSCodeExtension/syntaxes/qsharp.tmLanguage.json.v.template
bettinaheim: There are separate IQ# highlights? Should I file an issue for that?

src/QsCompiler/Tests.Compiler/TestCases/GlobalVerification.qs
bettinaheim: After some thinking: I think the right thing to do here is to limit the use of return statements to be at the end of an execution path in the apply block - same as for returns from within using and borrowing - and guarantee that the adjointed within block is executed before returning (treat is as a try/finally, same as deallocating qubits). However, to limit the burden on a runtime, I would like to not support it until we have the proper translation. I will thus disallow it for now.

Issue #80, opened 2019-07-24
Code Action for looping over IndexRange


bettinaheim:
For a for loop over the indices of an array 'arr' 
for (i in 0 .. Length(arr)-1) {
    // code
}
it would be nice if there was a code action (i.e. the little light-bulbs popping up) that a) makes the edit to
for (i in IndexRange(arr)) {
    // code
}
and b) inserts the corresponding open directive for IndexRange if needed. 

**Describe the solution you'd like**
A code action following the same pattern as other code actions should be implemented. 
The relevant code is in [EditorSupport.cs](https://github.com/microsoft/qsharp-compiler/blob/master/src/QsCompiler/CompilationManager/EditorSupport.cs). 

**Additional context**
Please comment on the issue if you would like to pick it up and would like more details/help.

ScottCarda:
I can take a look at this.

ScottCarda:
The branch I am using to work on this issue is:
https://github.com/ScottCarda/qsharp-compiler/commits/sccarda/LoopRangeSimplifyCodeAction

ScottCarda:
The corresponding PR is https://github.com/microsoft/qsharp-compiler/pull/84.

bettinaheim:
Merged into master with PR #130.

Issue #78, opened 2019-07-24
Code action suggesting replacement for array item updates


sw23:
Arrays in Q# are immutable, meaning array items cannot be updated once defined.  See [Local Variables](https://docs.microsoft.com/en-us/quantum/techniques/local-variables?view=qsharp-preview#mutability) for more details.

As a result, the following code results in a compilation error:
```
set arr[i] = 1;
```
Instead, a [Copy-and-Update Expression](https://docs.microsoft.com/en-us/quantum/language/expressions?view=qsharp-preview#copy-and-update-expressions) should be used to create a new array with the modified value:
```
set arr w/= i <- 1;
```

Since array item updates are a common thing for users to want to do, it'd be nice to have a code action suggesting the copy-and-update expression corresponding to the array item update specified.  This can obviously get fairly complicated, but handling the common/easy cases would go a long way.

vadym-kl:
This looks like a great code action, maybe you could also support it for `arr[i] = 1` in addition to `set arr[i] = 1`. It could be quite common ( this is my subjective opinion) for people who are just learning Q# to just want to write `arr[i] = 1` for array assignment if they are already familiar with C#, Java, C or other similar languages. 

If there were a code action expanding `arr[i][j] = 1` into 
`set arr w/= i <- ( set arr[j] w/= j <- 1 );` that also could be very helpful.

bettinaheim:
The corresponding PR is https://github.com/microsoft/qsharp-compiler/pull/91.

bettinaheim:
The code action as filed above is merged into master with PR #130.
I'll close this issue for now. @sw23 , @vadym-kl  let me know if you are interested in picking up implementing additional support for the cases that Vadym mentioned.

sw23:
> The code action as filed above is merged into master with PR #130.
> I'll close this issue for now. @sw23 , @vadym-kl let me know if you are interested in picking up implementing additional support for the cases that Vadym mentioned.

@bettinaheim - Cool! Yeah I can update the code action to support the additional cases that @vadym-kl mentioned now that the initial one has been reviewed and almost pulled in.  Should I open a follow-on issue?

bettinaheim:
@sw23 Awesome!
Yes, opening a new issue would be good - I'd rather close issues as the planned things get done, and open new issues for addtional capabilities. I am looking forward to those! :) Let me know if you need/want help!

Issue #75, opened 2019-07-24
Code Action - Remove Unreachable Code


ScottCarda:
We want to add a code action that removes unreachable code.

ScottCarda:
The branch I am working on this issue is:
https://github.com/ScottCarda/qsharp-compiler/tree/sccarda/RemoveUnexecutedCode

bettinaheim:
The corresponding PR is https://github.com/microsoft/qsharp-compiler/pull/76.

bettinaheim:
Merged into master with PR #130.

Issue #65, opened 2019-07-22
Code action for deprecated "not", "and", and "or"


ryan-moreno:

bettinaheim:
Merged into master with PR #130.

PR #130, opened 2019-08-21
Additional Code Actions

bettinaheim:
This PR adds support for the given code actions for the following cases: 
- deprecated syntax for UnitType, AND, OR, NOT, as well as for operation characteristics
- suggestions for update-and-reassign statements
- suggestions for ranges over indices
- unreachable code
- documenting comments

in addition to the two existing cases for: 
- ambiguous identifiers 
- unknown identifiers


bettinaheim:
This PR implements the following tracked issues: 
#80 #78 #75 #65 #64

PR #138, opened 2019-08-22
Update CI version, and add prerelease feed configuration

cgranade:
This PR updates the CI version to 0.9 and configures the repo to use packages from the prerelease feed.


bettinaheim:
@cgranade I will close this PR - I added the NuGet.config file to add the alpha-feed for this repo to the PR that also adds a setup to test whether the compiler output runs on various targets. I think that should do.

PR #140, opened 2019-08-24
Adding a folder with target examples for easy testing

bettinaheim:
Adds an target example that allows to compile Q# code with the compiler built from the source code, and executes the code on the QuantumSimulator for testing. 
In the QsCompiler.sln under TestTargets/Simulation set the Example as start up project and simply hit Ctrl+F5. For testing the latest compiler version and whether the produced compilation can be executed on the simulation framework, edit the implementation of Main.qs in that project.

See https://github.com/microsoft/qsharp-compiler/issues/141.


anpaz-msft:
Sorry I didn’t get a chance to review this, but it all happened quickly during the weekend.
I suspect this is going to be problematic in Solid, as now you have circular dependencies among repositories.
Even if it doesn’t get build during the CI build, having this circular dependencies is not a good practice and we should try to avoid it.
An option could be to create these cross-repo projects in Solid; Solid has all the repos as submodules so it would be fairly clean to do it there. Similar to the sanity tests that do the end-to-end testing, we could a X-Dev folder that let’s you have everything from source.
Another option is to create a script that let’s you replace nuget with project references, similar to the old solidify…
The current approach is convenient, but can add a lot of headache in the future…

-a.

________________________________

PR #151, opened 2019-08-30
Publishing outdated header and semantic diagnostics until the updates have been computed

bettinaheim:


src/QsCompiler/CompilationManager/FileContentManager.cs
```diff
@@ -190,6 +196,25 @@ private void InvalidateOrUpdateBySyntaxCheckDelimeters(ManagedList<Diagnostic> d
             finally { diagnostics.SyncRoot.ExitWriteLock(); }
         }
 
+        /// <summary>
+        /// Replaces the given list of latest diagnostics with a copy of the current ones, 
+        /// Updates the line numbers for both the current and latest diagnostics that start after the syntax check end delimiter, 
+        /// and removes all diagnostics that overlap with the given range for teh syntax check update in the latest diagnostics only.
```
marshallsa:
```suggestion
        /// and removes all diagnostics that overlap with the given range for the syntax check update in the latest diagnostics only. 
```

src/QsCompiler/CompilationManager/FileContentManager.cs
```diff
@@ -190,6 +196,25 @@ private void InvalidateOrUpdateBySyntaxCheckDelimeters(ManagedList<Diagnostic> d
...
+        /// Throws an ArgumentNullException if the given current and/or latest diagnostics are null, or if the syntax check delimiters are null. 
+        /// Throws an ArgumentException if the given start and end position do not denote a valid range.
+        /// </summary>
+        private static void DelayInvalidateOrUpdate(ManagedList<Diagnostic> current, ManagedList<Diagnostic> latest,
+            Range syntaxCheckDelimiters, int lineNrChange)
+        {
+            if (current == null) throw new ArgumentNullException(nameof(current));
+            if (latest == null) throw new ArgumentNullException(nameof(latest));
+
+            latest.ReplaceAll(current);
```
marshallsa: What is current and latest here? Is latest newer than current - but it's being replaced by current?

bettinaheim: Yes, the latest is used to store partially computed diagnostics and thus always newer than latest. The replacement was indeed an artifact of an early draft - I removed it.

marshallsa: Another name for "latest" like "next" might help. It could just be me but I am a little confused by the difference between "current" and "latest". :)

src/QsCompiler/CompilationManager/FileContentManager.cs
```diff
@@ -305,20 +330,26 @@ private void InvalidateOrUpdateContextDiagnostics(int start, int count, int line
         /// Throws an ArgumentException if the given start and end position do not denote a valid range.
         /// </summary>
         private void InvalidateOrUpdateHeaderDiagnostics(Range syntaxCheckDelimiters, int lineNrChange) =>
-            this.InvalidateOrUpdateBySyntaxCheckDelimeters(this.HeaderDiagnostics, syntaxCheckDelimiters, lineNrChange);
+            DelayInvalidateOrUpdate(this.HeaderDiagnostics, this.UpdatedHeaderDiagnostics, syntaxCheckDelimiters, lineNrChange);
 
 
         /// <summary>
-        /// Replaces the current semantic diagnostics with the given sequence.
+        /// Replaces the current semantic diagnostics with the given sequence and pushes the updated semantic diagnostics.
```
marshallsa:
```suggestion
        /// Replaces both the current and updated semantic diagnostics with the given sequence.
```

It looks like this method is replacing both lists (unlike `AddAndFinalizeSemanticDiagnostics` which appends to the updated diagnostics).

src/QsCompiler/CompilationManager/FileContentManager.cs
bettinaheim: renamed to diagnostics and updated

src/QsCompiler/CompilationManager/FileContentManager.cs
bettinaheim: reworded

PR #159, opened 2019-09-04
Delay publishing diagnostics while typing a symbol

marshallsa:
Using the change from #151, delay publishing diagnostics for changes that match the `ValidAsSymbol` regex, but still update the file immediately so code completion is accurate.

Part of issue #44.

### DIFF
--- a/src/QsCompiler/CompilationManager/FileContentManager.cs
+++ b/src/QsCompiler/CompilationManager/FileContentManager.cs
@@ -857,25 +857,26 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
             publishDiagnostics = true;
             if (count == 1 && line == start)
             {
+                this.UnprocessedUpdates.Enqueue(change);
+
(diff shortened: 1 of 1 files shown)


src/QsCompiler/CompilationManager/FileContentManager.cs
```diff
@@ -857,22 +857,25 @@ internal void PushChange(TextDocumentContentChangeEvent change, out bool publish
...
                     trimmedText == "\\" || trimmedText == "/") // ... and the same here
                 {
-                    this.Timer.Start(); // we can simply queue this update - no need to actually execute it
+                    // we can simply queue this update - no need to actually execute it
+                    this.UnprocessedUpdates.Enqueue(change);
+                    this.Timer.Start();
                     publishDiagnostics = false;
                     return;
                 }
-                this.Update(); // update only the currently queued changes
```
bettinaheim: A tiny difference, but if you can call Update() instead of Update(change), you are saving one iteration of the scope tracking update.

PR #162, opened 2019-09-05
Hide completions for "private" (underscore-prefixed) global symbols

marshallsa:
An underscore prefix is sometimes used to indicate private operations/types (e.g., in the `Microsoft.Quantum.Arithmetic` namespace). This PR hides those symbols from code completion unless they belong to the same namespace that code completion is activated from.

Depends on PR #161. Part of issue #44.

### DIFF
--- a/src/QsCompiler/CompilationManager/EditorSupport/EditorCompletion.cs
+++ b/src/QsCompiler/CompilationManager/EditorSupport/EditorCompletion.cs
@@ -200,13 +200,14 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
                 case CompletionKind.Keyword keyword:
                     return new[] { new CompletionItem { Label = keyword.Item, Kind = CompletionItemKind.Keyword } };
             }
-            var namespaces =
+            var currentNamespace = file.TryGetNamespaceAt(position);
+            var openNamespaces =
                 namespacePrefix == "" ? GetOpenNamespaces(file, compilation, position) : new[] { namespacePrefix };
             switch (kind.Tag)
             {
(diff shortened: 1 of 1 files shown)


src/QsCompiler/CompilationManager/EditorSupport/EditorCompletion.cs
```diff
@@ -661,5 +676,17 @@ private static Position GetDelimiterPosition(FileContentManager file, CodeFragme
...
+        /// Returns true if the qualified name is visible given the current namespace and a list of open namespaces.
+        /// <para/>
+        /// Names that start with "_" are treated as "private"; they are only visible from the namespace in which they
+        /// are declared.
+        /// </summary>
+        private static bool IsVisible(QsQualifiedName qualifiedName,
+                                      string currentNamespace,
+                                      IEnumerable<string> openNamespaces) =>
+            openNamespaces.Contains(qualifiedName.Namespace.Value) &&
+            (!qualifiedName.Name.Value.StartsWith("_") || qualifiedName.Namespace.Value == currentNamespace);
```
bettinaheim: To sum up things: any types or callables starting with an underscore are suggested for completions "anywhere" (depending on context) within the same namespace as they are declared in, possibly across compilation units, correct? And not in any other namespace, independent on file or compilation unit. I think once we actually properly handle that with access modifiers in the language we will probably restrict it to visible within the compilation unit, but at that point the completion here will automatically pick it up. So, looks good to me. :)

marshallsa: Yes, that's correct, it's only based on having the same parent namespace regardless of file or compilation unit.

PR #163, opened 2019-09-05
Show completions for named type items

marshallsa:
This PR adds a simple kind of named item completion that shows all named items for every type whenever a named item is expected.

Depends on #161. Part of issue #44.


src/QsCompiler/CompilationManager/EditorSupport/EditorCompletion.cs
```diff
@@ -661,5 +684,21 @@ private static Position GetDelimiterPosition(FileContentManager file, CodeFragme
                 IsIncomplete = isIncomplete,
                 Items = items?.ToArray()
             };
+
+        /// <summary>
+        /// Returns a flattened enumerable of all items in the <see cref="QsTuple{Item}"/>.
+        /// </summary>
+        private static IEnumerable<T> FlattenQsTuple<T>(QsTuple<T> tuple)
```
bettinaheim: You can replace that with a call to SyntaxGenerator.ExtractItems (admittedly probably not the best place to put that routine...).

PR #164, opened 2019-09-05
Rsoiffer/circuit api

rsoiffer:
Rewrote the pure circuit API, and fixed issues from previous PRs


src/QsCompiler/CompilerOptimizations/Printer.fs
```diff
@@ -98,7 +98,7 @@ let rec internal printQsTuple (qt: QsTuple<LocalVariableDeclaration<QsLocalSymbo
 
 let rec internal printStm (indent: int) (stm: QsStatementKind): string =
```
bettinaheim: Is something blocking replacing this file with the standard Q# generation? There seems to be an old branch that I think started to do that but hasn't been updated in a while?

src/QsCompiler/CompilerOptimizations/PureCircuitFinding.fs
```diff
@@ -13,8 +14,18 @@ open PureCircuitAPI
 open OptimizingTransformation
 
 
+/// Returns whether an expression is an operation call
+let private isOperation expr =
```
bettinaheim: Seems like this would break if we at some point decided to allow that returned values are ignored instead of requiring that they are always assigned. Since this is easy to avoid I suggest to adapt this to not having to rely on this implicit assumption.

src/QsCompiler/CompilerOptimizations/SideEffectChecking.fs
```diff
@@ -8,24 +8,36 @@ open Microsoft.Quantum.QsCompiler.Transformations.Core
 open Utils
 
 
+/// A ScopeTransformation that tracks what side effects the transformed code could cause
```
bettinaheim: Could you be more specific regarding what you mean by side effects? Thanks!

src/QsCompiler/CompilerOptimizations/SideEffectChecking.fs
```diff
@@ -8,24 +8,36 @@ open Microsoft.Quantum.QsCompiler.Transformations.Core
...
+    /// Whether the transformed code might output any messages to the console
+    member __.hasOutput = anyOutput
 
     override __.Expression = { new ExpressionTransformation() with
         override expr.Kind = { new ExpressionKindTransformation() with
             override __.ExpressionTransformation ex = expr.Transform ex
             override __.TypeTransformation t = expr.Type.Transform t
 
+            override __.onFunctionCall (method, arg) =
+                anyOutput <- true
```
bettinaheim: It seems you are indicating the transformation might output messages if there is any call whatsoever - that seems rather pessimistic. Why is that?

src/QsCompiler/CompilerOptimizations/StatementRemoving.fs
```diff
@@ -21,8 +23,10 @@ type internal StatementRemover() =
     /// The VariableRenamer used to ensure unique variable names
     let mutable renamer = None
 
-
-
+    /// Given a statement, returns a sequence of statements to replace this statement with.
+    /// Removes useless statements, such as variable declarations with discarded values.
+    /// Simplifies statements, such as a replacing ScopeStatements with their contents.
```
bettinaheim:
```suggestion
    /// Simplifies statements, such as replacing ScopeStatements with their contents.
```

src/QsCompiler/CompilerOptimizations/StatementRemoving.fs
```diff
@@ -63,6 +90,12 @@ type internal StatementRemover() =
                 |> Seq.map wrapStmt
             QsScope.New (statements, parentSymbols)
 
+        override __.onLocation _ = Null
```
bettinaheim: Why is the logic for stripping range information embedded into this particular transformation? What in the code relies on there not being any range information?

src/QsCompiler/CompilerOptimizations/Printer.fs
rsoiffer: The standard Q# generation is not deterministic. The namespace elements are compiled in a random order, so they are printed in a random order, breaking the test cases. I started working on a fix in the Q# code generator, but I've been working on other code recently.

src/QsCompiler/CompilerOptimizations/SideEffectChecking.fs
rsoiffer: The transformation is meant to track any kind of side effect we might have. Documentation on the functions in the class specify what side effects it currently looks for.

src/QsCompiler/CompilerOptimizations/SideEffectChecking.fs
rsoiffer: I don't know for sure what other functions a given function might call, so I make the pessimistic assumption that it might eventually call a Message function. While this means that I'll almost never remove useless function calls, this doesn't seem like a very important optimization, as in almost all cases it will only affect classical performance. It also means that Message functions aren't removed, which is desirable in most cases. It'd be good to have a compiler option later that determines whether we should remove all functions anyways.

src/QsCompiler/CompilerOptimizations/PureCircuitFinding.fs
rsoiffer: What do you mean? I don't see what returning values has to do with checking whether an expression is an operation call.

src/QsCompiler/CompilerOptimizations/Utils.fs
```diff
@@ -200,12 +255,30 @@ let rec internal isAllDiscarded = function
 | _ -> false
 
 
+/// Casts an int64 to an int, throwing an ArgumentException if outside the allowed range
```
bettinaheim: Both for this and the conversion function below it seems like you are not covering one edge case similar to this https://github.com/microsoft/qsharp-compiler/issues/86. :)

src/QsCompiler/Tests.Compiler/ExecutionTests.fs
```diff
@@ -41,8 +41,7 @@ type ExecutionTests (output:ITestOutputHelper) =
 
 
     [<Fact>]
-    member this.``Specialization Generation for Conjugations`` () = 
-
+    member this.``Specialization Generation for Conjugations in Body`` () =
```
bettinaheim: I understand the urge to separate this into different functions, but could you keep the four pieces as one test case? You could make four (non-test) functions, and then one test funcion invoking each one.

src/QsCompiler/Tests.Compiler/TestCases/Optimizer/Miscellaneous_output.txt
```diff
@@ -91,43 +91,31 @@ namespace Microsoft.Quantum.Testing {
                 ((Adjoint GlobalCallable CNOT) of ((Unwrap LocalVariable q), (Unwrap LocalVariable b2)[0]));
                 for item in (Unwrap LocalVariable b2) {
                     (GlobalCallable CNOT of ((Unwrap LocalVariable q), LocalVariable item));
-                    (GlobalCallable CNOT of ((Unwrap LocalVariable q), LocalVariable item));
```
bettinaheim: Why did this change? The test case didn't seem to have changed?

src/QsCompiler/CompilerOptimizations/Evaluation.fs
```diff
@@ -13,170 +13,181 @@ open Microsoft.Quantum.QsCompiler.SyntaxTree
...
+
 /// Represents any interrupt to the normal control flow of a function evaluation.
 /// Includes return statements, errors, and (if they were added) break/continue statements.
 type private FunctionInterrupt =
 /// Represents the function invoking a return statement
 | Returned of TypedExpression
 /// Represents the function invoking a fail statement
 | Failed of TypedExpression
+/// We reached the limit of statement executions
+| OutOfStatements
```
bettinaheim: Pretty sure this is the opposite - you have more statements but reached the maximum evaluation count. ;)

src/QsCompiler/CompilerOptimizations/Evaluation.fs
```diff
@@ -13,170 +13,181 @@ open Microsoft.Quantum.QsCompiler.SyntaxTree
...
+    let castToBool x: Imp<bool> = imperative {
         match x.Expression with
-        | BoolLiteral b -> Ok (f b)
-        | _ -> "Not a BoolLiteral: " + (printExpr x.Expression) |> CouldNotEvaluate |> Error
-
-    /// The callback for the subroutine that evaluates and simplifies an expression
-    member internal this.evaluateExpression constants expr =
-        ExpressionEvaluator(callables, constants, maxRecursiveDepth - 1).Transform expr
+        | BoolLiteral b -> return b
+        | _ -> yield CouldNotEvaluate ("Not a BoolLiteral: " + (printExpr x.Expression))
```
bettinaheim: Is this a case that should be reached? If so, how about renaming the function to tryAsBoolLiteral or tryEvaluateBool or something?

src/QsCompiler/CompilerOptimizations/Evaluation.fs
```diff
@@ -13,170 +13,181 @@ open Microsoft.Quantum.QsCompiler.SyntaxTree
...
+    }
+
+    /// Evaluates and simplifies a single Q# expression
+    member internal __.evaluateExpression expr: Imp<TypedExpression> = imperative {
+        let! vars, counter = getState
+        let result = ExpressionEvaluator(callables, vars, counter / 2).Transform expr
```
bettinaheim: What's the reason for the counter / 2 ?

src/QsCompiler/CompilerOptimizations/Evaluation.fs
```diff
@@ -13,170 +13,181 @@ open Microsoft.Quantum.QsCompiler.SyntaxTree
...
-            let! s = this.evaluateScope !constantsRef newScope scope
-            constantsRef := s }
+    member private this.evaluateStatement (statement: QsStatement): Imp<Unit> = imperative {
+        do! incrementState
 
         match statement.Statement with
-        | QsExpressionStatement expr ->
-            this.evaluateExpression constants expr |> ignore; constants |> Ok
+        | QsExpressionStatement _ ->
+            ()
```
bettinaheim: Where are expression statements evaluated now / why did this change?

src/QsCompiler/CompilerOptimizations/Evaluation.fs
```diff
@@ -13,170 +13,181 @@ open Microsoft.Quantum.QsCompiler.SyntaxTree
...
-                    | _ -> "Unknown IterationValue in for loop: " + (printExpr iterValues.Expression) |> CouldNotEvaluate |> Error
-                let constantsRef = ref constants
-                for loopValue in iterSeq do
-                    constantsRef := enterScope !constantsRef
-                    constantsRef := defineVarTuple (isLiteral callables) !constantsRef (fst stmt.LoopItem, loopValue)
-                    do! eval constantsRef true stmt.Body
-                    constantsRef := exitScope !constantsRef
-                return !constantsRef
+            let! iterExpr = this.evaluateExpression stmt.IterationValues
+            let! iterSeq = imperative {
```
bettinaheim: So, the construction and let! here is for the sole purpose of counting?

src/QsCompiler/CompilerOptimizations/Evaluation.fs
```diff
@@ -13,170 +13,181 @@ open Microsoft.Quantum.QsCompiler.SyntaxTree
 open Microsoft.Quantum.QsCompiler.Transformations.Core
```
bettinaheim: this file has cleaned up nicely!

src/QsCompiler/CompilerOptimizations/CallableInlining.fs
```diff
@@ -81,12 +81,30 @@ type private InliningInfo = {
...
 }
 
+/// A scope transformation that substitutes type parameters according to the given dictionary
+type private ReplaceTypeParams(typeParams: ImmutableDictionary<QsTypeParameter, ResolvedType>) =
+    inherit ScopeTransformation()
+
+    override __.Expression = { new ExpressionTransformation() with
+        override __.Type = { new ExpressionTypeTransformation() with
+            override this.onTypeParameter tp =
+                if typeParams.ContainsKey tp then
```
bettinaheim: You probably want to explicitly ignore range information here and only compare origin and name.

src/QsCompiler/CompilerOptimizations/CallableInlining.fs
```diff
@@ -81,12 +81,30 @@ type private InliningInfo = {
...
+                    typeParams.[tp].Resolution
+                else
+                    let origin = tp.Origin
+                    let name = tp.TypeName
+                    let range = this.onRangeInformation tp.Range
```
bettinaheim: Are you explicitly rebuilding the type parameter because you need the onRangeInformation for something?

src/QsCompiler/CompilerOptimizations/PureCircuitAPI.fs
```diff
@@ -4,197 +4,175 @@
...
-| ValueArray            of Const list
-| ArrayItem             of Const * Const
-| QubitReference        of QubitReference
+| PauliArray            of QsPauli list
+
+/// Any expression
+type Expression =
+| Literal               of Literal
+| Tuple                 of Expression list
+| Qubit                 of int
```
bettinaheim: I know I am being nitpicky, but having spent days on hunting down bugs due to similar things: Are you checking that the indices fit into a 32-bit integer?

src/QsCompiler/CompilerOptimizations/PureCircuitAPI.fs
```diff
@@ -4,197 +4,175 @@
...
+| QubitArray            of int list
+| UnknownValue          of int
 
 /// A call to a quantum gate, with the given functors and arguments
 type GateCall = {
     gate:     QsQualifiedName
     adjoint:  bool
-    controls: Const list
-    arg:      Const
+    controls: Expression list
```
bettinaheim: What Expressions exactly are you allowing here? Why is this an Expression list opposed to just an Expression?
A doc comment would be helpful.

src/QsCompiler/CompilerOptimizations/PureCircuitAPI.fs
```diff
@@ -4,197 +4,175 @@
...
 /// A pure, parameter-free quantum circuit
 type Circuit = {
-    numQubitRefs: int
+    numQubits: int
+    numUnknownValues: int
     gates: GateCall list
 }
 
 /// Metadata used for constructing circuits.
 /// Currently just stores the variable names corresponding to each qubit reference.
```
bettinaheim: I think the doc comment here needs to be slightly adapted.

src/QsCompiler/DataStructures/SyntaxExtensions.fs
```diff
@@ -311,10 +311,9 @@ type QsStatement with
                     (match s.Default with Null -> Seq.empty | Value v -> upcast v.Body.Statements))
             | QsForStatement s -> upcast s.Body.Statements
             | QsWhileStatement s -> upcast s.Body.Statements
-            | QsConjugation s -> Seq.append s.OuterTransformation.Body.Statements s.InnerTransformation.Body.Statements // FIXME: not sure this folder makes sense...
+            | QsConjugation s -> Seq.append s.OuterTransformation.Body.Statements s.InnerTransformation.Body.Statements
```
bettinaheim: The thing with the folder here that at the very least needs to be documented in the doc comment is that it is not clear to me how it should behave: In the case of the conjugation, the adjoint of the outer block is implicit. If you have something like a statement counter or similar, should it count all three blocks or just two?

src/QsCompiler/CompilerOptimizations/PureCircuitAPI.fs
bettinaheim: I think I saw that the statement removing should take care of that. Is it also handled if that step is not run?

src/QsCompiler/CompilerOptimizations/StatementRemoving.fs
```diff
@@ -21,8 +23,10 @@ type internal StatementRemover() =
     /// The VariableRenamer used to ensure unique variable names
     let mutable renamer = None
 
-
-
+    /// Given a statement, returns a sequence of statements to replace this statement with.
+    /// Removes useless statements, such as variable declarations with discarded values.
+    /// Simplifies statements, such as a replacing ScopeStatements with their contents.
+    /// Splits QsQubitScopes by replacing register allocations with single qubit allocations.
```
bettinaheim: I like it because it is making it easier to move around allocations.

src/QsCompiler/CompilerOptimizations/StatementRemoving.fs
rsoiffer: The PureCircuitAPI file relies on equality comparisons between expressions to determine whether two expressions should be replaced with the same symbol or with different symbols. Range information messes up these equality comparisons.

src/QsCompiler/CompilerOptimizations/Utils.fs
rsoiffer: Resolved this function in latest commit. The function below should be correct without any changes.

src/QsCompiler/CompilerOptimizations/StatementRemoving.fs
```diff
@@ -36,12 +40,35 @@ type internal StatementRemover() =
...
+                | _ -> Seq.singleton (l, r)) |> List.ofSeq
+            match myList with
+            | [] -> newScopeStatement s.Body |> Seq.singleton
+            | [lhs, rhs] ->
+                let newBody = QsScope.New (s.Body.Statements.InsertRange (0, newStatements), s.Body.KnownSymbols)
+                QsQubitScope.New s.Kind ((lhs, rhs), newBody) |> QsQubitScope |> Seq.singleton
+            | _ ->
+                let lhs = List.map fst myList |> ImmutableArray.CreateRange |> VariableNameTuple
+                let rhs = List.map snd myList |> ImmutableArray.CreateRange |> QubitTupleAllocation |> ResolvedInitializer.New
+                let newBody = QsScope.New (s.Body.Statements.InsertRange (0, newStatements), s.Body.KnownSymbols)
```
bettinaheim: Are the known symbols listed still accurate? I believe you may need to adapt that, same above.

src/QsCompiler/Tests.Compiler/ExecutionTests.fs
rsoiffer: I separated them specifically to make them different test cases, as I was trying to figure out exactly which part of the test case was failing. What's the benefit of combining them into one large test case?

src/QsCompiler/Tests.Compiler/TestCases/Optimizer/Miscellaneous_output.txt
rsoiffer: The pure circuit optimizer got smarter - it now parses Unwrap expressions as UnknownValues instead of just giving up, so it can cancel the adjacent adjoint gates

src/QsCompiler/CompilerOptimizations/Evaluation.fs
rsoiffer: Would "TooManyStatements" be a better name?

src/QsCompiler/CompilerOptimizations/Evaluation.fs
rsoiffer: I suppose the second case shouldn't be reached anymore, as the modifications to evaluateExpression mean that only literal values will be passed to castToBool. Having it yield a CouldNotEvaluate interrupt would be safer than raising an exception in case it's passed an incorrect value for some reason. Should we have the code be cautious or should it raise an exception on input that should be impossible?

src/QsCompiler/CompilerOptimizations/Evaluation.fs
rsoiffer: This is a workaround for limiting the number of statements evaluated in recursive functions. Currently, evaluting a recursive call means going through the ExpressionEvaluator, which doesn't use the Imperative monad, so after trying to evaluate a recursive call we have no information about how many statements were executed during this call. This leads to edge cases where any given stack includes a small number of statement executions, but the total number of evaluations is exponentially large. Dividing by 2 blocks almost all of these edge cases, so we don't use up a ton of time trying to evaluate code. The "good" solution would require tracking the number of statements evaluated in the ExpressionEvaluator, which would require completely rewriting the ExpressionEvaluator to use the Imperative monad.

src/QsCompiler/CompilerOptimizations/Evaluation.fs
rsoiffer: Since we're inside of a function, expression statements will never have side effects (as they can't mutate state), so we can skip evaluating them. The previous code was effectively doing the same - see the call to ignore.

src/QsCompiler/CompilerOptimizations/Evaluation.fs
rsoiffer: By let!, do you mean the entire Imperative monad? It is used for counting the number of statements evaluated, but it also tracks the state of the currently defined variables, also allows for interrupts at any time, and also gives Break statements to exit loops early.

src/QsCompiler/CompilerOptimizations/PureCircuitAPI.fs
```diff
@@ -4,197 +4,175 @@
...
+        recurse cc x Some Tuple
+    | ExprKind.ValueArray x when typeIsArray TypeKind.Pauli ->
+        recurse cc x (function Literal (PauliLiteral p) -> Some p | _ -> None) (PauliArray >> Literal)
+    | ExprKind.ValueArray x when typeIsArray TypeKind.Qubit ->
+        recurse cc x (function Qubit i -> Some i | _ -> None) QubitArray
+    | ExprKind.Identifier _ when expr.ResolvedType.Resolution = TypeKind.Qubit ->
+        let newQubits, i = ensureMatchingIndex cc.qubits
+        Some ({cc with qubits = newQubits}, Qubit i)
+    | _ ->
+        // TODO - add checks to ensure that this doesn't contain a qubit
```
bettinaheim: How about addressing that?

bettinaheim:
@rsoiffer Did you actually add the check to make sure that all qubits for the pure circuits you are handing off are distinct? I don't think I've seen it.

src/QsCompiler/CompilerOptimizations/PureCircuitFinding.fs
bettinaheim: Sorry, I was referring to the fact that it does not check whether the given expression is a partial application.

src/QsCompiler/CompilerOptimizations/SideEffectChecking.fs
bettinaheim: Hmm, it would be good to have a handle to enable or disable this special accomodation for Message. I don't remember - is there such a handle?

src/QsCompiler/CompilerOptimizations/StatementRemoving.fs
bettinaheim: Ok, I agree that you probably want to strip range info, but it would make sense to do that separately form this transformation step e.g. before handing it off to the pure circuit API.

src/QsCompiler/Tests.Compiler/ExecutionTests.fs
bettinaheim: I see. I bundle tests according to cover a particular functionality. Usually one unit tests contains a couple of dozed individual test cases and the output tells which ones failed how. I suppose I did not write a proper output that says which part failed here. If you want to keep it I can do that later though.

src/QsCompiler/Tests.Compiler/TestCases/Optimizer/Miscellaneous_output.txt
bettinaheim: Which reminds me: Did you ever get around to using the metadata attached to operations to cancel self-adjoint operations?

src/QsCompiler/CompilerOptimizations/Evaluation.fs
bettinaheim: The risk with "cautious" code is that is can get a lot harder to maintain and debug, because over time, a lot of the assumptions that were originally made are no longer applicable. Writing permissive like that means that these chances can go undetected for a long time. There is definitively a benefit to writing permissive code if you have a lot of external dependencies or the code is user facing. However, given that this is an internal function in a module that is maintained by us and has no external dependencies whatsoever, I'd rather we know about and fix "minor" bugs instead of ignoring them silently.

src/QsCompiler/CompilerOptimizations/Evaluation.fs
bettinaheim: Ok, seems fine to leave it like that. Thanks for the explanation!

src/QsCompiler/CompilerOptimizations/CallableInlining.fs
rsoiffer: Fixed in latest commit

src/QsCompiler/CompilerOptimizations/CallableInlining.fs
rsoiffer: Fixed in latest commit

src/QsCompiler/CompilerOptimizations/Evaluation.fs
bettinaheim: Right - a comment here would be helpful.

src/QsCompiler/CompilerOptimizations/PureCircuitAPI.fs
rsoiffer: If the index didn't fit into a 32-bit integer, that would mean that the code has at least 2^31 distinct qubit allocations or unknown values. I have to imagine that if the code really did have more than 2 billion distinct expressions, many other parts of the compiler would break first.

src/QsCompiler/CompilerOptimizations/PureCircuitAPI.fs
rsoiffer: This should be an int list instead of an Expression list, as the only valid control is a qubit array. Fixed in latest commit.

src/QsCompiler/CompilerOptimizations/PureCircuitAPI.fs
rsoiffer: Fixed in latest commit

src/QsCompiler/DataStructures/SyntaxExtensions.fs
rsoiffer: Added documentation in latest commit

src/QsCompiler/CompilerOptimizations/PureCircuitFinding.fs
rsoiffer: Fixed in latest commit

src/QsCompiler/CompilerOptimizations/SideEffectChecking.fs
rsoiffer: No, I don't currently have a handle like that. Where would be the best place to put one?

src/QsCompiler/CompilerOptimizations/StatementRemoving.fs
rsoiffer: Fixed in latest commit

src/QsCompiler/CompilerOptimizations/Evaluation.fs
rsoiffer: Fixed in latest commit

src/QsCompiler/CompilerOptimizations/Evaluation.fs
rsoiffer: Fixed in latest commit

src/QsCompiler/CompilerOptimizations/Evaluation.fs
rsoiffer: Fixed in latest commit

src/QsCompiler/CompilerOptimizations/PureCircuitAPI.fs
bettinaheim: Probably a fair point. :)

src/QsCompiler/CompilerOptimizations/SideEffectChecking.fs
bettinaheim: You would need to have the handle on the transformation that eliminates the messages, and propagate it up to the front handle that defines which of transformations to run in which order. The behavior of all other transformations should be independent on that configuration. Is that the case?

src/QsCompiler/CompilerOptimizations/PureCircuitAPI.fs
bettinaheim: Actually, please handle this properly. I did ping the libraries team, because it seemed like a possibility that we might be working with that many qubit allocations on the toffoli simulator, and we are. Thanks!

src/QsCompiler/CompilerOptimizations/SideEffectChecking.fs
rsoiffer: I added the handle in Optimizations.fs, the file that defines which transformations to run in which order.

src/QsCompiler/CompilerOptimizations/PureCircuitAPI.fs
rsoiffer: It's not even possible to create lists that long, and a lot of the library functions I'm using only work with integers. Handling this case would be somewhat difficult, and given that this is an extreme edge case that we won't run into for years, I'm not sure it's worth it. What if the compiler throws an error if there would be an overflow?

src/QsCompiler/CompilerOptimizations/StatementRemoving.fs
rsoiffer: There are a lot of places in the code that can change the symbols visible at the top of a scope. For instance, inlining a callable will add all the symbols of the parent function to that callable, or removing an unused variable will remove symbols from all following scopes. Instead of trying to update this metadata in each transformation, which seems difficult and prone to error, could we entirely regenerate it after running all the optimization steps?

src/QsCompiler/CompilerOptimizations/SideEffectChecking.fs
rsoiffer: I added the handle in Optimizations.fs, the file that defines which transformations to run in which order.

src/QsCompiler/Tests.Compiler/TestCases/Optimizer/Miscellaneous_output.txt
rsoiffer: Fixed in latest commit

src/QsCompiler/CompilerOptimizations/Printer.fs
rsoiffer: Will fix in a future PR

src/QsCompiler/CompilerOptimizations/PureCircuitAPI.fs
rsoiffer: Fixed in latest commit

src/QsCompiler/CompilerOptimizations/PureCircuitAPI.fs
rsoiffer: Fixed in latest commit

src/QsCompiler/CompilerOptimizations/MinorTransformations.fs
```diff
@@ -0,0 +1,230 @@
...
+            scope.Statements
+            |> Seq.map this.onStatement
+            |> Seq.map (fun x -> x.Statement)
+            |> Seq.collect this.TransformStatement
+            |> Seq.map wrapStmt
+        QsScope.New (statements, parentSymbols)
+
+
+/// A SyntaxTreeTransformation that removes all range information from anywhere in the AST
+type internal StripAllRangeInformation() =
```
bettinaheim: With the latest update to master, you can just replace this with a call to StripPositionInfo.Apply (defined in the QsCore project).

src/QsCompiler/CompilerOptimizations/OptimizingTransformations/CallableInlining.fs
```diff
@@ -122,42 +129,87 @@ type internal CallableInliner(callables) =
...
 
-            do! check (countReturnStatements ii.body = 0)
-            let! current = currentCallable
-            do! check (cannotReachCallable callables ii.body current.FullName)
+            let! currentCallable = currentCallable
+            let! renamer = renamer
+            renamer.clearStack()
+            do! check (cannotReachCallable callables ii.body currentCallable.FullName)
             do! check (ii.functors.controlled < 2)
             // TODO - support multiple Controlled functors
```
bettinaheim: What's the reason for this limitation again?

src/QsCompiler/CompilerOptimizations/OptimizingTransformations/CallableInlining.fs
```diff
@@ -122,42 +129,87 @@ type internal CallableInliner(callables) =
 
     // The current callable we're in the process of transforming
     let mutable currentCallable: QsCallable option = None
+    let mutable renamer: VariableRenamer option = None
 
-    /// Inline an expression as a ScopeStatement, with many checks to ensure correctness.
-    /// Returns None if the expression cannot be safely inlined.
-    let safeInline expr =
+    /// The random generator to give unique variable names
+    let random = System.Random()
```
bettinaheim:
```suggestion
```
Looks like this on is not actually used.

src/QsCompiler/CompilerOptimizations/OptimizingTransformations/CallableInlining.fs
```diff
@@ -122,42 +129,87 @@ type internal CallableInliner(callables) =
...
+                |> ImmutableArray.CreateRange
+            return ii, newStatements
         }
 
+    /// Inline an expression representing a callable with no return statements.
+    /// Returns None if the expression cannot be safely inlined.
+    let safeInline expr =
+        maybe {
+            let! ii, newStatements = tryInline expr
+            do! check (countReturnStatements ii.body = 0)
```
bettinaheim: I'd suggest to move getting the inlining info out of tryInline, and swap the order of these two instructions to save some processing.

src/QsCompiler/CompilerOptimizations/OptimizingTransformations/CallableInlining.fs
```diff
@@ -122,42 +129,87 @@ type internal CallableInliner(callables) =
...
+        maybe {
+            let! ii, newStatements = tryInline expr
 
-        override scope.StatementKind = { new StatementKindTransformation() with
-            override stmtKind.ExpressionTransformation x = scope.Expression.Transform x
-            override stmtKind.LocationTransformation x = scope.onLocation x
-            override stmtKind.ScopeTransformation x = scope.Transform x
-            override stmtKind.TypeTransformation x = scope.Expression.Type.Transform x
+            do! check (countReturnStatements ii.body = 1)
+            let lastStatement = newStatements.[newStatements.Length-1]
```
bettinaheim: It would be good to document that the assumption is that dead code elimination is done beforehand.

src/QsCompiler/CompilerOptimizations/Printer.fs
```diff
@@ -1,177 +0,0 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
bettinaheim: Yay!

src/QsCompiler/CompilerOptimizations/Utils.fs
```diff
@@ -12,29 +12,87 @@ open Microsoft.Quantum.QsCompiler.SyntaxExtensions
...
+
+
+/// Returns whether a given expression is a literal (and thus a constant)
+let rec internal isLiteral (callables: Callables) (expr: TypedExpression): bool =
+    expr |> TypedExpression.MapFold (fun ex -> ex.Expression) (fun sub ex ->
+        match ex.Expression with
+        | IntLiteral _ | BigIntLiteral _ | DoubleLiteral _ | BoolLiteral _ | ResultLiteral _ | PauliLiteral _ | StringLiteral _
+        | UnitValue | MissingExpr | Identifier (GlobalCallable _, _)
+        | ValueTuple _ | ValueArray _ | RangeLiteral _ | NewArray _ -> true
+        | Identifier _ when ex.ResolvedType.Resolution = Qubit -> true
```
bettinaheim: How come identifiers of type Qubit specifically/only are treated as literals?

src/QsCompiler/CompilerOptimizations/VariableRenaming.fs
```diff
@@ -12,119 +13,142 @@ open Microsoft.Quantum.QsCompiler.SyntaxTree
...
+            override __.ScopeTransformation x = this.Transform x
+            override __.TypeTransformation x = this.Expression.Type.Transform x
+
+            override this.onSymbolTuple syms =
+                match syms with
+                | VariableName item -> VariableName (NonNullable<_>.New (generateUniqueName item.Value))
+                | VariableNameTuple items -> Seq.map this.onSymbolTuple items |> ImmutableArray.CreateRange |> VariableNameTuple
+                | InvalidItem | DiscardedItem -> syms
+
+            override this.onVariableDeclaration stm =
```
bettinaheim: Please take a look at https://github.com/microsoft/qsharp-compiler/pull/175. 
Merging that in, you should be able to drop the override for onVariableDeclaration, onForStatement, and onQubitScope here.

src/QsCompiler/CompilerOptimizations/StatementRemoving.fs
bettinaheim: Yes, I think that would be fine. I suggest to explicitly strip that information everywhere entirely then. Could you make sure that info is stripped in general? And in particular it needs to be documented. I would say stripping that info cleanly and documenting the behavior is the important part, even if you don't get around to writing a transformation that reconstructs that info.

PR #166, opened 2019-09-07
Disable code completion in Visual Studio 16.2 or older; use a workaround in 16.3

marshallsa:
Due to a bug in Visual Studio 16.2, code completions in Q# aren't dismissed after typing whitespace, and accepting a completion after typing multiple symbols separated by whitespace replaces the whole line instead of just the last symbol. For now, we are disabling completion in the Visual Studio extension until the bug is resolved. (Code completion in the Visual Studio Code extension isn't affected.)

Part of issue #44.

### DIFF
--- a/src/VisualStudioExtension/QsharpVSIX/QsLanguageClient.cs
+++ b/src/VisualStudioExtension/QsharpVSIX/QsLanguageClient.cs
@@ -1,6 +1,11 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using Microsoft.VisualStudio.LanguageServer.Client;
+using Microsoft.VisualStudio.Threading;
+using Microsoft.VisualStudio.Utilities;
+using Newtonsoft.Json.Linq;
+using StreamJsonRpc;
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


marshallsa:
Added a workaround for Visual Studio 16.3 (include " " and "(" as trigger characters).

src/VisualStudioExtension/QsharpVSIX/QsLanguageClient.cs
```diff
@@ -134,5 +140,24 @@ public void OnTelemetry(JToken args)
                 { Debug.Assert(false, $"error sending telemetry: \n{ex}"); }
             }
         }
+
+        private static string GetVisualStudioVersion()
```
bettinaheim: Seems a bit hacky. I think it should be possible to go via DTE.Version without load issues.

marshallsa: Definitely agree that it's hacky, but when I tried `DTE.Version` it only gave me "16.0" (no minor version). :(

bettinaheim: Arg, is that annoying. I didn't find a better solution either. Ok, hacky it is then.

PR #169, opened 2019-09-10
Support for declaration attributes

bettinaheim:
This is the final merge of the feature branch back to master, adding support for declaration attributes. 
- any user defined type can be marked as attribute
- marking it as attribute does not prevent using it just like any other user defined type
- it is marked by attaching the "Attribute" attribute defined in the core namespace to it
- an attribute annotation then consists of a compile time constant of such an attribute
- such annotations for now may only be attached to declarations (types, functions, operations)
- for now they are disabled on specializations


bettinaheim:
I'll merge this since the bulk of the work has been reviewed on the feature branch.

PR #171, opened 2019-09-10
Check that namespace exists before resolving documentation

marshallsa:
This fixes an exception that was thrown if a completion resolve request was received with a qualified name that doesn't exist. The behavior now is to return the completion item unchanged if its qualified name is invalid.

Part of issue #44.

### DIFF
--- a/src/QsCompiler/LanguageServer/EditorState.cs
+++ b/src/QsCompiler/LanguageServer/EditorState.cs
@@ -369,10 +369,11 @@ namespace Microsoft.Quantum.QsLanguageServer
             : null;
 
         /// <summary>
-        /// Resolves additional information for the given completion item.
+        /// Resolves additional information for the given completion item. Returns the original completion item if no
+        /// additional information is available.
         /// <para/>
         /// Returns null if any parameter is null or the file given in the original completion request is invalid or
-        /// ignored, or if the completion to resolve is outdated. 
+        /// ignored.
         /// </summary>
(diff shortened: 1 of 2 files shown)


src/QsCompiler/LanguageServer/EditorState.cs
```diff
@@ -369,10 +369,11 @@ internal Task CloseFileAsync(TextDocumentIdentifier textDocument, Action<string,
             : null;
 
         /// <summary>
-        /// Resolves additional information for the given completion item.
+        /// Resolves additional information for the given completion item. Returns the original completion item if no
+        /// additional information is available.
         /// <para/>
         /// Returns null if any parameter is null or the file given in the original completion request is invalid or
-        /// ignored, or if the completion to resolve is outdated.
```
bettinaheim: So, if the information is outdated, then the original item is returned?

marshallsa: By outdated, do you mean invalid (e.g., the qualified name in the completion item data does not/no longer exists)?

src/QsCompiler/LanguageServer/EditorState.cs
```diff
@@ -369,10 +369,11 @@ internal Task CloseFileAsync(TextDocumentIdentifier textDocument, Action<string,
             : null;
 
         /// <summary>
-        /// Resolves additional information for the given completion item.
+        /// Resolves additional information for the given completion item. Returns the original completion item if no
+        /// additional information is available.
```
bettinaheim:
```suggestion
        /// additional information is available, or if the completion item is no longer valid or accurate.
```

PR #185, opened 2019-09-26
Publish language server as self-contained deployment.

cgranade:
This draft PR modifies the VS Code extension to depend on the language server as a self-contained ZIP hosted on Azure Blob Storage.


cgranade:
As per offline discussion with @bettinaheim, marking as ready for review and targeting a feature branch.

bettinaheim:
Looks like there is still some todo's for Mac.

src/VSCodeExtension/src/languageServer.ts
```diff
@@ -0,0 +1,518 @@
...
+                            vscode.env.openExternal(vscode.Uri.parse(
+                                "https://github.com/microsoft/qsharp-compiler/issues/new?assignees=&labels=bug,Area-IDE&template=bug_report.md&title="
+                            ));
+                            return;
+                            break;
+                    }
+                }
+            }
+            if (!this.findExecutable()) {
+                // TODO: handle this error more gracefully.
```
bettinaheim: As I understand that case should never happen. A more informative comment would be helpful.

bettinaheim:
There are still four things to do, I think. 
One is to adapt the VS extension to actually package the self-contained language server built for windows instead of the one freshly built. 
Another is we need to double check if we've got everything needed for signing. 
Then I think you wanted to leverage the 3.0 features to download the exe instead of the zip, 
and finally the support for Mac mentioned above.

bettinaheim:
@granade Before I merge this, could you please add the third party notice again? 

Copy-Item -Force -Path (Join-Path $RepoRoot "NOTICE.txt") -Destination $PSScriptRoot -Verbose
used to be in BuildDependencies.ps1. 
I think it would simplify things to go with downloading the exe instead of the zip, but I am happy to merge it as is onto the feature branch.

cgranade:
Thanks for catching that, @bettinaheim! Should be fixed as of 430f8a0.

bettinaheim:
> Thanks for catching that, @bettinaheim! Should be fixed as of 430f8a0.

Thanks! And thanks for updating the readme as well.

PR #188, opened 2019-09-28
Compiler support for Q# dlls

bettinaheim:
There are some basic tests related to being able to process both old and new dlls (tested via Project reference). But all in all there isn't that much to test for the compiler alone - more tests are needed for the QDK setup. 
For now I am generating a warning if old dlls are processed with the new compiler version (may lead to runtime errors), but eventually I think we should make it a breaking change.


src/QsCompiler/CommandLineTool/Commands/Build.cs
```diff
@@ -92,11 +92,12 @@ public static int Run(BuildOptions options, ConsoleLogger logger)
             var specifiesTargets = options.Targets != null && options.Targets.Any();
             var loadOptions = new CompilationLoader.Configuration
             {
-                ProjectFile = options.ProjectName == null ? null : new Uri(Path.GetFullPath(options.ProjectName)),
+                ProjectName = options.ProjectName,
                 GenerateFunctorSupport = true,
                 SkipSyntaxTreeTrimming = false,
                 DocumentationOutputFolder = options.DocFolder,
                 BuildOutputFolder = options.OutputFolder ?? (specifiesTargets ? "." : null),
+                DllOutputPath = " ", // generating the dll in the same location as the .bson file
```
anpaz-msft: this is a blank space (vs an empty string). is that on purpose?

bettinaheim: Both works; a dll will be generated as long as the DllOutputPath is not null, and a default path will be chosen if it is empty or whitespace.

src/QsCompiler/CommandLineTool/Program.cs
```diff
@@ -37,17 +37,21 @@ public static class ReturnCode
         /// </summary>
         public const int BINARY_GENERATION_ERRORS = -5;
         /// <summary>
+        /// Return code indicating that generating a dll containing the compiled binary failed.
+        /// </summary>
+        public const int DLL_GENERATION_ERRORS = -6;
```
anpaz-msft: Is there a reason to change this constants? I think right now is fine, but more generally speaking these are breaking changes so the right thing would be to add the new error code at the end.

bettinaheim: Yes, agreed. Since we don't currently package or ship the command line compiler, the only way right now to use it is to build it from source. Given that, I think it's fine for now to change the numbering, but I agree that once we ship it as tool, we should stick with the numbers.

src/QsCompiler/CompilationManager/CompilationUnit.cs
```diff
@@ -24,117 +25,114 @@ public class References
...
+        private static Func<(string, string), string> IsDeclaration(string declarationType) => (attribute) =>
+        {
+            var (typeName, serialization) = attribute;
+            if (!typeName.Equals(declarationType, StringComparison.InvariantCultureIgnoreCase)) return null;
+            return serialization;
+        };
+
+        private static IEnumerable<CallableDeclarationHeader> CallableHeaders(IEnumerable<(string, string)> attributes) =>
+            attributes.Select(IsDeclaration("CallableDeclarationAttribute")).Where(v => v != null)
+                .Select(CallableDeclarationHeader.FromJson).Select(built => built.Item2);
```
alan-geller: One note that might be irrelevant but isn't really well documented: most of the LINQ extension methods have the feature that they evaluate lazily. In many cases this is a significant benefit, but it can be a real problem if there's any possibility of the underlying collection getting modified before you actually complete the enumeration (by turning it into an ImmutableArray, for instance). So for these three static methods, for instance, you'd have to be careful to complete the enumeration before you change the argument. If these are only used in the constructor above, you're fine because the root constructor always completes the enumeration, but I'm not sure where else you use these.

src/QsCompiler/CompilationManager/AssemblyLoader.cs
```diff
@@ -0,0 +1,192 @@
...
+        /// possibly because the referenced assembly has been compiled with an older compiler version. 
+        /// Throws an ArgumentNullException if the given uri is null. 
+        /// Throws a FileNotFoundException if no file with the given name exists. 
+        /// Throws the corresponding exceptions if the information cannot be extracted.
+        /// </summary>
+        public static bool LoadReferencedAssembly(Uri asm, out References.Headers headers)
+        {
+            if (asm == null) throw new ArgumentNullException(nameof(asm));
+            if (!CompilationUnitManager.TryGetFileId(asm, out var id) || !File.Exists(asm.LocalPath))
+            { throw new FileNotFoundException($"the uti '{asm}' given to the assembly loader is invalid or the file does not exist"); }
```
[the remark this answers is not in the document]
bettinaheim: fixed

src/QsCompiler/CompilationManager/CompilationUnit.cs
bettinaheim: Yes, I know. They are indeed private and only used in a save context, so I am omitting the unnecessary array creation.

src/QsCompiler/CompilationManager/AssemblyLoader.cs
```diff
@@ -0,0 +1,192 @@
...
+
+            using (var stream = File.OpenRead(asm.LocalPath))
+            using (var assemblyFile = new PEReader(stream))
+            {
+                var loadedFromResource = FromResource(assemblyFile, out var syntaxTree);
+                var attributes = loadedFromResource ? Enumerable.Empty<(string, string)>() : LoadHeaderAttributes(assemblyFile);
+                headers = loadedFromResource
+                    ? new References.Headers(id, syntaxTree)
+                    : new References.Headers(id, attributes);
+                return loadedFromResource || !attributes.Any();
```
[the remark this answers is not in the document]
bettinaheim: changed to 
"
                if (!FromResource(assemblyFile, out var syntaxTree))
                {
                    var attributes = LoadHeaderAttributes(assemblyFile);
                    headers = new References.Headers(id, attributes);
                    return !attributes.Any(); // just means we have no references
                }
                headers = new References.Headers(id, syntaxTree);
                return true;
"

PR #190, opened 2019-10-02
Sccarda/generic resolution

ScottCarda-MS:
Rewriting calls to global generics to have type-specific identifiers. Robust naming collision handling is not yet addressed.

### DIFF
--- a/src/QsCompiler/Compiler/CompilationLoader.cs
+++ b/src/QsCompiler/Compiler/CompilationLoader.cs
@@ -218,14 +218,14 @@ namespace Microsoft.Quantum.QsCompiler
 
             // executing the specified rewrite steps 
 
+            this.GeneratedSyntaxTree = Transformations.Monomorphization.ResolveGenericsSyntax.Apply(this.GeneratedSyntaxTree);
+
             if (this.Config.GenerateFunctorSupport)
             {
                 this.CompilationStatus.FunctorSupport = 0;
                 var functorSpecGenerated = this.GeneratedSyntaxTree != null && FunctorGeneration.GenerateFunctorSpecializations(this.GeneratedSyntaxTree, out this.GeneratedSyntaxTree);
                 if (!functorSpecGenerated) this.LogAndUpdate(ref this.CompilationStatus.FunctorSupport, ErrorCode.FunctorGenerationFailed, Enumerable.Empty<string>());
             }
-
-            this.GeneratedSyntaxTree = Transformations.Monomorphization.ResolveGenericsSyntax.Apply(this.GeneratedSyntaxTree);
 
             if (!this.Config.SkipSyntaxTreeTrimming)
             {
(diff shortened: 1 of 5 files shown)


src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -0,0 +1,110 @@
...
+            if (types.Any() && ex.Expression is QsExpressionKind<TypedExpression, Identifier, ResolvedType>.CallLikeExpression call)
+            {
+                string prefix = CreateTypePrefix(types);
+
+                // For now, only resolve identifiers of global callables
+                if (call.Item1.Expression is QsExpressionKind<TypedExpression, Identifier, ResolvedType>.Identifier id &&
+                    id.Item1 is Identifier.GlobalCallable globalCallable)
+                {
+                    // Rebuild the method with the updated identifier
+                    ex = new TypedExpression(
```
ScottCarda-MS: Is there a framework I could leverage to create new objects from existing objects with only one field changed? This block of code is rather unpleasant.

bettinaheim: Yes, there are some helper functions around. Let's get together and go over this together.

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -0,0 +1,110 @@
...
+                        QsExpressionKind<TypedExpression, Identifier, ResolvedType>.NewCallLikeExpression(
+                            new TypedExpression(
```
bettinaheim: For the sake of public documentation of what we discussed: 
We want to make this recursive since technically the left hand side of the call might be an partial application that resolves some of the type parameters etc. Hence you really wan to call this.Transform on the subexpression, and you want to pass the information on the type parameter resolution down. 
You should also think about how you want to handle the case where once you get to the identifier not all type parameters are resolved.

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -218,6 +218,8 @@ public CompilationLoader(SourceLoader loadSources, ReferenceLoader loadReference
 
             // executing the specified rewrite steps 
 
+            this.GeneratedSyntaxTree = Transformations.Monomorphization.ResolveGenericsSyntax.Apply(this.GeneratedSyntaxTree);
```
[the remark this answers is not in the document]
ScottCarda-MS: This has been addressed.

src/QsCompiler/Transformations/Monomorphization.cs
ScottCarda-MS: This will need to be addressed later once there is a data structure for keeping tabs on information regarding generics.

src/QsCompiler/Transformations/Monomorphization.cs
ScottCarda-MS: This will need to be addressed later once there is a data structure for keeping tabs on information regarding generics.

PR #192, opened 2019-10-09
Insert doc comments before declaration attributes

bettinaheim:
A minor thing I came across when working: The code action for adding doc comments did not get updated when we added support for attributes, such that the doc comments were inserted immediately before the declaration instead of before all attributes attached to the declaration.

### DIFF
--- a/src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
+++ b/src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
@@ -77,7 +77,7 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
             if (file == null || range?.Start == null || range.End == null) return Enumerable.Empty<CodeFragment>();
             var (start, end) = (range.Start.Line, range.End.Line);
 
-            var fragAtStart = file.TryGetFragmentAt(range?.Start, out var _, includeEnd: true);
+            var fragAtStart = file.TryGetFragmentAt(range.Start, out var _, includeEnd: true);
             var inRange = file.GetTokenizedLine(start).Select(t => t.WithUpdatedLineNumber(start)).Where(ContextBuilder.TokensAfter(range.Start)); // does not include fragAtStart
             inRange = start == end
                 ? inRange.Where(ContextBuilder.TokensStartingBefore(range.End))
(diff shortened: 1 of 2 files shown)


src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
```diff
@@ -408,17 +408,27 @@ internal static IEnumerable<(string, WorkspaceEdit)> DocCommentSuggestions(this
...
+
+            // set declStart to the position of the first attribute attached to the declaration
+            bool EmptyOrFirstAttribute(IEnumerable<CodeFragment> line, out CodeFragment att)
+            {
+                att = line.Reverse().TakeWhile(t => t.Kind is QsFragmentKind.DeclarationAttribute).LastOrDefault();
+                return att != null || !line.Any();
+            }
+            var preceding = file.GetTokenizedLine(declStart.Line).TakeWhile(ContextBuilder.TokensUpTo(new Position(0, declStart.Character)));
+            for (var lineNr = declStart.Line; lineNr-- > 0 && EmptyOrFirstAttribute(preceding, out var precedingAttribute); preceding = file.GetTokenizedLine(lineNr))
+            { if (precedingAttribute != null) declStart = precedingAttribute.GetRange().Start.WithUpdatedLineNumber(lineNr+1); }
```
ScottCarda-MS: This for loop is too dense. For readability, it should be reworked. The post- decrement operator should not be used except as a statement by itself. If possible, stick to the more standard use of the for loop:
for( init loop index; condition; loop index increment [or decrement] ) { ... }

ScottCarda-MS: The braces should be on their own lines.

ScottCarda-MS: This for loop is too dense. For readability, it should be reworked.
The post-decrement operator should not be used except as a statement by itself.
If possible, stick to the more standard use of the for loop: for( init loop index; condition; loop index increment [or decrement] ) { ... }
The braces should be on their own lines.

bettinaheim: Alright, added whitespace, and more code, which may or may not make the logic more transparent. 
Arguable the sole purpose of pre- and post- increment is to not have to have a dedicated statement, so I am keeping that. ;)

PR #197, opened 2019-10-11
Sccarda/generic to concrete definitions

ScottCarda-MS:
Added functionality to replace the type-parameterized definitions of callables with concrete definitions.


src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -6,47 +6,90 @@
...
+            var filter = new ResolveGenericsSyntax(new ScopeTransformation<ResolveGenericsExpression>(new ResolveGenericsExpression(generics)), generics);
             return namespaces.Select(ns => filter.Transform(ns));
         }
 
         public static IEnumerable<QsNamespace> Apply(params QsNamespace[] namespaces)
         {
             return Apply((IEnumerable<QsNamespace>)namespaces);
         }
+        
+        private ResolveGenericsSyntax(ScopeTransformation<ResolveGenericsExpression> scope, GenericsCrate generics) : base(scope) { this.Generics = generics; }
```
bettinaheim: Should generics here be allowed to be null? Unless there is a specific reason to allow it, I'd suggest a null check here and throwing and ArgumentNullException if it is null.

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -6,47 +6,90 @@
...
-            List<QsQualifiedName> generics = new List<QsQualifiedName>();
-            var Callables = namespaces.GlobalCallableResolutions();
-            foreach (var callable in Callables)
-            {
-                if (callable.Value.Signature.TypeParameters.Any())
-                {
-                    generics.Add(callable.Key);
-                }
-            }
+            // TODO: this is specific to Provided implementations, but should be general to all implementation kinds
```
bettinaheim: What is the problem with just removing the "call.Value.Specializations.Any(x => x.Implementation.IsProvided)" check?

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -6,47 +6,90 @@
...
+            GenericsCrate generics = namespaces.GlobalCallableResolutions()
+                .Where(call => call.Value.Signature.TypeParameters.Any() && call.Value.Specializations.Any(x => x.Implementation.IsProvided))
+                .ToDictionary(call => call.Key, call => (call.Value, new Dictionary<Concretion, QsCallable>()));
 
-            var filter = new ResolveGenericsSyntax(new ScopeTransformation<ResolveGenericsExpression>(new ResolveGenericsExpression()));
+            var filter = new ResolveGenericsSyntax(new ScopeTransformation<ResolveGenericsExpression>(new ResolveGenericsExpression(generics)), generics);
```
bettinaheim: Instead of handing down the GenericsCrate instance you might consider handing it a function that given the full name of a callable returns the dictionary of concretions.

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -87,13 +128,97 @@ public override TypedExpression Transform(TypedExpression ex)
...
+        {
+            Concretion target = CreateConcretion(types);
+
+            // TODO: handle if globalCallable is not in Generics
+
+            var (original, concretions) = Generics[globalCallable.Item];
+
+            NonNullable<string> name;
+
+            if (concretions.ContainsKey(target))
```
bettinaheim: Here we may want to actually use the logic in the symbol table - we can chat about that later though (just put it on your todo list).

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -87,13 +128,97 @@ public override TypedExpression Transform(TypedExpression ex)
...
+        }
+    }
+
+    public class ResolveGenericsExpressionKind : ExpressionKindTransformation<ResolveGenericsExpression>
+    {
+        public ResolveGenericsExpressionKind(ResolveGenericsExpression expr) : base(expr) { }
+
+        public override QsExpressionKind<TypedExpression, Identifier, ResolvedType> onIdentifier(Identifier sym, QsNullable<ImmutableArray<ResolvedType>> tArgs)
+        {
+            return base.onIdentifier(sym, tArgs);
```
bettinaheim: I assume that's preparation for filing something in here? :)

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -87,13 +128,97 @@ public override TypedExpression Transform(TypedExpression ex)
...
+    internal class ReplaceTypeParamsSyntax :
+        SyntaxTreeTransformation<MinorTransformations.ReplaceTypeParams>
+    {
+        public static QsCallable Apply(ImmutableDictionary<QsTypeParameter, ResolvedType> typeParams, QsCallable callable)
+        {
+            if (callable == null) throw new ArgumentNullException(nameof(callable));
+
+            var filter = new ReplaceTypeParamsSyntax(new MinorTransformations.ReplaceTypeParams(typeParams));
+
+            // Create a holding namespace to use transform on
```
bettinaheim: Why not directly call the transformation on the callable?

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -87,13 +128,97 @@ public override TypedExpression Transform(TypedExpression ex)
...
+
+        public ReplaceTypeParamsSyntax(MinorTransformations.ReplaceTypeParams scope) : base(scope) { }
+
+        public override ResolvedSignature onSignature(ResolvedSignature s)
         {
-            return "_" + String.Join("", typeDict
-                .ToList()
-                .OrderBy(kvp => kvp.Key.TypeName)
-                .Select(kvp => kvp.Key.TypeName.Value + "_" + kvp.Value.Resolution.ToString() + "_")
+            // Remove the type parameters from the signature
```
bettinaheim: Where is the logic that replace the now concretized type parameters in the argument and return type?

src/QsCompiler/Transformations/Monomorphization.cs
ScottCarda-MS: That would be ideal. I haven't seen how to call a transformation on something other than a QsNamespace. Can you give me a little more feedback on how I would call a transformation on a callable? Being able to transform callables directly would simplify things quite a bit.

src/QsCompiler/Core/SymbolTable.fs
```diff
@@ -1239,7 +1239,7 @@ and NamespaceManager
 
 
     /// Generates a hash for a resolved type. Does not incorporate any positional information.
-    static member internal TypeHash (t : ResolvedType) = t.Resolution |> function
+    static member public TypeHash (t : ResolvedType) = t.Resolution |> function
```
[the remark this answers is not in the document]
ScottCarda-MS: This is probably not necessary anymore as the reason for making this public was to hash Concretions, which is no longer necessary. The change will be reversed.

src/QsCompiler/Optimizations/Utils/HelperTransformations.fs
```diff
@@ -118,7 +118,7 @@ type internal ReferenceCounter() =
 
 
 /// A ScopeTransformation that substitutes type parameters according to the given dictionary
-type internal ReplaceTypeParams(typeParams: ImmutableDictionary<QsTypeParameter, ResolvedType>) =
+type public ReplaceTypeParams(typeParams: ImmutableDictionary<QsTypeParameter, ResolvedType>) =
```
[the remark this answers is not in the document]
ScottCarda-MS: The transformation I wrote on top of this one does just that. Perhaps they should have their definitions merged?

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -6,47 +6,90 @@
...
+    using GenericsCrate = Dictionary<QsQualifiedName, (QsCallable, Dictionary<Concretion, QsCallable>)>;
+
+    #region Monomorphization
+
+    internal class Concretion : HashSet<(NonNullable<string>, LocalTypeKind)>
+    {
+        public override int GetHashCode()
+        {
+            return this.Select(x => (x.Item1.Value.GetHashCode(), NamespaceManager.TypeHash(ResolvedType.New(x.Item2))).ToString().GetHashCode())
+                .Aggregate((sum, i) => unchecked(sum + i)); // TODO: Use better hashing over enumerable
```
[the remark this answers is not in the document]
ScottCarda-MS: No longer necessary as Concretions will not need to be hashed and will now just be a local typedef.

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -6,47 +6,90 @@
...
+        }
+
+        public override bool Equals(object obj)
```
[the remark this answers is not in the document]
ScottCarda-MS: Concretions will be changed to a local typedef instead of a full class.

src/QsCompiler/Transformations/Monomorphization.cs
ScottCarda-MS: The GenericsCreate class and its instances will no longer be used.

src/QsCompiler/Transformations/Monomorphization.cs
ScottCarda-MS: The GenericsCreate class and its instances will no longer be used.

src/QsCompiler/Transformations/Monomorphization.cs
ScottCarda-MS: Oops, this is actually the start of my work towards making the feature work with partial applications. It will be removed from this PR. Thanks!

src/QsCompiler/Optimizations/Utils/HelperTransformations.fs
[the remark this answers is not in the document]
bettinaheim: Yes, I think merging them in this case makes a lot of sense. Thanks!

src/QsCompiler/Transformations/Monomorphization.cs
bettinaheim: Instead of filter.Transform you would just call the corresponding filter.onCallableImplementation. Maybe the naming here is not the best; it takes a QsCallable and returns a QsCallable.

src/QsCompiler/Transformations/Monomorphization.cs
ScottCarda-MS: The inherited ReplaceTypeParams transformation acts on all type resolution nodes, including the nodes found under the signiture's ArgumentType and ReturnType nodes.

src/QsCompiler/Optimizations/Utils/HelperTransformations.fs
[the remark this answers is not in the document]
ScottCarda-MS: Adjusted the existing behavior of the default expression transformation to allow the existing transformations to act on the type parameter resolution dictionaries.

src/QsCompiler/Transformations/Monomorphization.cs
ScottCarda-MS: No longer using this implementation.

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -6,47 +6,187 @@
...
+                {
+                    originalName = call.Key,
+                    typeResolutions = null,
+                    concreteName = call.Key
+                });
+
+            var requests = new Stack<Request>(entryPoints);
+            var responses = new List<Response>();
+
+            while(requests.Count() > 0)
```
bettinaheim:
```suggestion
            while(requests.Any())
```
shouldn't really matter here but still.

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -6,47 +6,187 @@
...
             {
-                if (callable.Value.Signature.TypeParameters.Any())
+                Request currentRequest = requests.Pop();
+
+                // If there is a call to an unknown callable, throw exception
+                if (!globals.ContainsKey(currentRequest.originalName)) throw new ArgumentNullException(); // TODO: need to throw a more valid exception
```
bettinaheim: How about addressing the todo? 
And I'd suggest to go with a globals.TryGetValue to save the duplicate effort to first check if the key exists and then actually get the value further below.

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -6,47 +6,187 @@
...
+using Microsoft.Quantum.QsCompiler.SymbolManagement;
 using Microsoft.Quantum.QsCompiler.SyntaxTokens;
 using Microsoft.Quantum.QsCompiler.SyntaxTree;
 using Microsoft.Quantum.QsCompiler.Transformations.Core;
 
 namespace Microsoft.Quantum.QsCompiler.Transformations.Monomorphization
 {
+    using Concretion = HashSet<(QsTypeParameter, ResolvedType)>;
+
+    internal struct Request
```
bettinaheim: Is there anywhere where you actually set the properties of a Request and a Response below? What's the reason for the struct?
Otherwise I might lean towards having a class with readonly properties.

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -6,47 +6,187 @@
 using System.Collections.Generic;
```
bettinaheim: There seems to be a bunch of usings here that are not really needed.

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -6,47 +6,187 @@
...
+    internal class ReplaceTypeParamImplementationsSyntax :
+        SyntaxTreeTransformation<MinorTransformations.ReplaceTypeParams>
+    {
+        public static Response Apply(Response current)
+        {
+            // Nothing to change if the current callable is already concrete
+            if (current.typeResolutions == null) return current;
+
+            var filter = new ReplaceTypeParamImplementationsSyntax(
+                new MinorTransformations.ReplaceTypeParams(current.typeResolutions.ToImmutableDictionary(param => param.Item1, param => param.Item2)));
```
bettinaheim: I think you mentioned you got rid of Concretions and the HashSet <-> Dictionary conversions in the subsequent PR, correct?

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -6,47 +6,187 @@
...
+
     }
 
-    public class ResolveGenericsExpression :
-        ExpressionTransformation<ExpressionKindTransformation<ResolveGenericsExpression>>
+    internal class ReplaceTypeParamCallsExpression :
+        ExpressionTransformation<ReplaceTypeParamCallsExpressionKind>
     {
-        public ResolveGenericsExpression() : base(ex => new ExpressionKindTransformation<ResolveGenericsExpression>(ex as ResolveGenericsExpression)) { }
+        private Response CurrentResponse;
```
bettinaheim: Looks like all of these (CurrentResponse, Requests, Responses) could be readonly?

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -87,13 +225,91 @@ public override TypedExpression Transform(TypedExpression ex)
...
+                    typeResolutions = target,
+                    concreteName = new QsQualifiedName(globalCallable.Item.Namespace, NonNullable<string>.New(name))
+                });
+            }
+
+            return Identifier.NewGlobalCallable(new QsQualifiedName(globalCallable.Item.Namespace, NonNullable<string>.New(name)));
         }
     }
+
+    internal class ReplaceTypeParamCallsExpressionKind : ExpressionKindTransformation<ReplaceTypeParamCallsExpression>
```
bettinaheim: I am assuming you will add functionality to this class later on?

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -55,8 +195,6 @@ public override TypedExpression Transform(TypedExpression ex)
 
             if (types.Any() && ex.Expression is QsExpressionKind<TypedExpression, Identifier, ResolvedType>.CallLikeExpression call)
```
bettinaheim: What is there are type parameter resolutions but the expression is not a call-like expression?

src/QsCompiler/Transformations/Monomorphization.cs
ScottCarda-MS: The reason for the structs is that structs are generally more light-weight than classes from an efficiency perspective, so I was hoping here to eliminate some overhead with these, which are little more than tuples with named fields. Their members are never assigned to after initialization, so they are treated as immutable, but I can make some changes to reinforce that immutability.

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -55,8 +195,6 @@ public override TypedExpression Transform(TypedExpression ex)
...
             {
-                string prefix = CreateTypePrefix(types);
-
                 // For now, only resolve identifiers of global callables
                 if (call.Item1.Expression is QsExpressionKind<TypedExpression, Identifier, ResolvedType>.Identifier id &&
```
bettinaheim: Seems like there should be a TODO here that I believe is addressed with the next PR?

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -66,7 +204,7 @@ public override TypedExpression Transform(TypedExpression ex)
                         QsExpressionKind<TypedExpression, Identifier, ResolvedType>.NewCallLikeExpression(
                             new TypedExpression(
                                 QsExpressionKind<TypedExpression, Identifier, ResolvedType>.NewIdentifier(
```
bettinaheim: .. and I know that a proper recursion is also part of the next one.

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -87,13 +225,91 @@ public override TypedExpression Transform(TypedExpression ex)
             return base.Transform(ex);
         }
 
-        private string CreateTypePrefix(ImmutableDictionary<QsTypeParameter, ResolvedType> typeDict)
+        private Concretion CreateConcretion(ImmutableDictionary<QsTypeParameter, ResolvedType> typeDict)
```
bettinaheim: That one will be eliminated when you remove the use of Concretions, correct?

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -87,13 +225,91 @@ public override TypedExpression Transform(TypedExpression ex)
...
+            foreach (var kvp in typeDict)
+            {
+                result.Add((kvp.Key, kvp.Value));
+            }
+            return result;
+        }
+
+        private Identifier GetConcreteIdentifier(Identifier.GlobalCallable globalCallable, ImmutableDictionary<QsTypeParameter, ResolvedType> types)
+        {
+            Concretion target = null;
```
bettinaheim: I think SetEquals below might actually throw if target is null.

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -87,13 +225,91 @@ public override TypedExpression Transform(TypedExpression ex)
...
+            if (types != null && !types.IsEmpty)
+            {
+                target = CreateConcretion(types);
+            }
+
+            string name = null;
+
+            // Check for recursive call
```
bettinaheim:
```suggestion
            // Check for self-recursive call
```

src/QsCompiler/Transformations/Monomorphization.cs
ScottCarda-MS: Yes, that's correct.

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -87,13 +225,91 @@ public override TypedExpression Transform(TypedExpression ex)
...
+            {
+                name = CurrentResponse.concreteCallable.FullName.Name.Value;
+            }
+
+            // Search Requests for identifier
+            if (name == null)
+            {
+                name = Requests
+                    .Where(req =>
+                        req.originalName.Namespace.Value == globalCallable.Item.Namespace.Value &&
```
bettinaheim: Since you repeatedly do similar comparisons I suggest to actually define a local function to do so.

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -6,47 +6,187 @@
...
+                currentResponse = ReplaceTypeParamImplementationsSyntax.Apply(currentResponse);
+
+                // Rewrite calls
+                currentResponse = ReplaceTypeParamCallsSyntax.Apply(currentResponse, requests, responses);
+
+                responses.Add(currentResponse);
             }
 
-            var filter = new ResolveGenericsSyntax(new ScopeTransformation<ResolveGenericsExpression>(new ResolveGenericsExpression()));
+            var filter = new ResolveGenericsSyntax(responses
```
bettinaheim: There needs to be a test to make sure only what should be dropped is dropped (doesn't need to be part of this PR, but you should add that to your todo list).

src/QsCompiler/Transformations/Monomorphization.cs
ScottCarda-MS: The need for this is removed in the other PR: #206 .

src/QsCompiler/Transformations/Monomorphization.cs
ScottCarda-MS: Yes the TODO is near the top of the function. It is addressed in #206 .

src/QsCompiler/Transformations/Monomorphization.cs
ScottCarda-MS: I didn't quite understand your question here, but you should know that this section was entirely rewritten in #206 .

src/QsCompiler/Transformations/Monomorphization.cs
ScottCarda-MS: Yes in PR #206 .

src/QsCompiler/Transformations/Monomorphization.cs
ScottCarda-MS: Yes this is removed in PR #206 .

src/QsCompiler/Transformations/Monomorphization.cs
ScottCarda-MS: Yes this class is filled out in PR #206 . Strangely though, when I didn't have this here, even empty as it is, the transform wouldn't recurse properly, and having this here fixed the issue.

src/QsCompiler/Transformations/Monomorphization.cs
ScottCarda-MS: The problem is I don't know what kind of exception to throw here.

src/QsCompiler/Transformations/Monomorphization.cs
ScottCarda-MS: This issue is addressed in PR #206.

src/QsCompiler/Transformations/Monomorphization.cs
bettinaheim: Yes, that might be good. I'm not sure if there are universally applicable rules of thumb for efficiency. I tend to like immutable data structures but that is definitively somewhat rooted in my preference for a functional setup if such a setup makes sense.

PR #206, opened 2019-10-17
Sccarda/generic partial application

ScottCarda-MS:
Added functionality to replace the type-parameterized definitions of callables with concrete definitions.
Split the transformation into three transformations: One that replaces type-parameterized calls, one that replaces type-parameterized definition of callables, and one that coordinates the other two transformations and updates the namespace to have the appropriate callables.
Added functionality to enable use on partial applications of type-parameterized calls.


src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -69,11 +70,16 @@ public static IEnumerable<QsNamespace> Apply(IEnumerable<QsNamespace> namespaces
                     concreteCallable = globals[currentRequest.originalName].WithFullName(name => currentRequest.concreteName)
```
bettinaheim: I suggest to use globals.TryGetValue instead of ContainsKey above to avoid looking up the key twice.

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -166,139 +244,109 @@ public static Response Apply(Response current, Stack<Request> requests, List<Res
...
         public override TypedExpression Transform(TypedExpression ex)
         {
-            // TODO: This should be recursive and use a more robust data structure to handle the type information for generics
-            ImmutableDictionary<QsTypeParameter, ResolvedType> types = ex.TypeParameterResolutions;
+            var range                = this.onRangeInformation(ex.Range);
+            var typeParamResolutions = this.onTypeParamResolutions(ex.TypeParameterResolutions);
+            var exType               = this.Type.Transform(ex.ResolvedType);
+            var inferredInfo         = this.onExpressionInformation(ex.InferredInformation);
+            // Change the order so that Kind is transformed last
+            var kind                 = this.Kind.Transform(ex.Expression);
```
bettinaheim: Might be good to add a comment here saying that the order of recursion matters (onTypeParameterResolutions before the kind transformation). And it would definitively be good to add that on the todo list for test: write a test that fails if the order of these two is switched.

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -166,139 +244,109 @@ public static Response Apply(Response current, Stack<Request> requests, List<Res
...
+            GetConcreteIdentifier = getConcreteIdentifier;
+            CurrentParamTypes = currentParamTypes;
         }
 
-        private Identifier GetConcreteIdentifier(Identifier.GlobalCallable globalCallable, ImmutableDictionary<QsTypeParameter, ResolvedType> types)
+        public override QsExpressionKind<TypedExpression, Identifier, ResolvedType> onIdentifier(Identifier sym, QsNullable<ImmutableArray<ResolvedType>> tArgs)
         {
-            Concretion target = null;
-            if (types != null && !types.IsEmpty)
+            if (sym is Identifier.GlobalCallable global)
```
bettinaheim: It would be good to add an else case that throws an ArgumentException if sym is not a GlobalCallable but there are type argument specified (i.e. tArgs is not Null). If you do that, then that implies that this transformation will only succeed on valid code and fail if there was a compilation error - which is fine - at the benefit of that it is not a silent failure if we start supporting assigning type parameterized objects.

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -166,139 +244,109 @@ public static Response Apply(Response current, Stack<Request> requests, List<Res
...
-            if (name == null)
-            {
-                name = Requests
-                    .Where(req =>
-                        req.originalName.IsEqual(globalCallable.Item) &&
-                        req.typeResolutions.SetEquals(target))
-                    .Select(req => req.concreteName.Name.Value)
-                    .FirstOrDefault();
-            }
+    internal class ReplaceTypeParamCallsExpressionType : ExpressionTypeTransformation<ReplaceTypeParamCallsExpression>
```
bettinaheim: It seems like the a lot cleaner solution for this transformation is to just not use MinorTransformations.ReplaceTypeParams (please make that internal again) and override onTypeParameter here.

src/QsCompiler/Transformations/Monomorphization.cs
ScottCarda-MS: Please check that I have handled this case appropriately.

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -166,139 +270,104 @@ public static Response Apply(Response current, Stack<Request> requests, List<Res
...
+                    foreach (var key in applicableParams.Keys)
+                    {
+                        CurrentParamTypes.Remove(key);
+                    }
+                }
             }
-
-            // Search Requests for identifier
-            if (name == null)
+            else if (sym is Identifier.LocalVariable && !tArgs.IsNull && tArgs.Item.Any())
```
bettinaheim:
```suggestion
            else if (sym is Identifier.LocalVariable && tArgs.IsValue && tArgs.Item.Any())
```

src/QsCompiler/Transformations/Monomorphization.cs
bettinaheim: Looks good.

PR #215, opened 2019-10-23
Added support for the Deprecated attribute in docgen

alan-geller:
This should eventually replace the "Deprecated" section in markdown comments.

The code assumes that the string parameter of the Deprecated attribute, if not empty, contains markdown test that can be pasted in to the item summary and documentation.


src/QsCompiler/Tests.DocGenerator/DocParsingTests.cs
```diff
@@ -352,5 +352,20 @@ ArgDeclType BuildArgument(string name, ResolvedType t)
...
+        {
+            string[] comments = { "# Summary",
+                                  "This is some text",
+                                };
+            string dep = "Don't use this";
+            string warning = "> [!WARNING]\n> Deprecated\n";
+            var dc = new DocComment(comments, true, dep);
+            Assert.Equal(comments[1] + "\r" + warning + "\r" + dep, dc.Summary);
+            Assert.Equal(warning + "\r" + dep, dc.ShortSummary);
+            // The Documentatin string is run through the Markdown generator and so has \r instead of \n
```
bettinaheim:
```suggestion
            // The Documentation string is run through the Markdown generator and so has \r instead of \n
```

src/QsCompiler/Tests.DocGenerator/DocParsingTests.cs
```diff
@@ -352,5 +352,20 @@ ArgDeclType BuildArgument(string name, ResolvedType t)
             var s = stream.ToString();
             Assert.Equal(expected, s);
         }
+
+        [Fact]
+        public void ParseDeprecated()
+        {
+            string[] comments = { "# Summary",
```
bettinaheim: Currently, deprecated warnings for our docs are generated based on a section in the doc comments  
/// # Deprecated
/// Please use @"Microsoft.Quantum.NewName".
It would be good to add a test if the docs generation is well behaved if both are present.

src/QsCompiler/DocumentationParser/DocItem.cs
```diff
@@ -39,13 +45,25 @@ internal abstract class DocItem
...
-        internal DocItem(string nsName, string itemName, string kind, ImmutableArray<string> documentation)
+        internal DocItem(string nsName, string itemName, string kind, ImmutableArray<string> documentation,
+            IEnumerable<QsDeclarationAttribute> attributes)
         {
             namespaceName = nsName;
             name = itemName;
             uid = (namespaceName + "." + name).ToLowerInvariant();
             itemType = kind;
-            comments = new DocComment(documentation);
+            var depAttrs = attributes.Where(Microsoft.Quantum.QsCompiler.BuiltIn.MarksDeprecation).ToList();
```
bettinaheim: You could instead go with SymbolResolution.TryFindRedirect - that would take care of both determining whether the type or callable is deprecated and finding the redirect if one has been defined. It seems good to not duplicate that logic.

src/QsCompiler/DocumentationParser/DocComment.cs
```diff
@@ -317,8 +319,36 @@ void ParseMapSection(IEnumerable<Block> blocks, Dictionary<string, string> accum
                     }
                 }
 
+                if (deprecated && !deprecationProcessed)
```
bettinaheim: So this means that right now what is written in the Deprecated section of a doc comment takes precedence over what is in the deprecated attribute? Given that the compiler behavior only looks at the latter, and I also intend to add potentially some verification of the redirection and a code action if appropriate, I suggest to make it the other way round.

src/QsCompiler/DocumentationParser/DocItem.cs
alan-geller: Are we expecting the argument to @Deprecated to be the name of the replacement item, or a message? I've been assuming that it's a message, since there might not always be a simple replacement item (either there might not be any or there might be more than one).

bettinaheim: No, it's supposed to be the replacement item. Throughout our libraries that works well. If the string is empty, then there isn't a replacement item, if it's not, then it's assumed to be the full name of the replacement.

src/QsCompiler/DocumentationParser/DocComment.cs
```diff
@@ -317,8 +321,36 @@ void ParseMapSection(IEnumerable<Block> blocks, Dictionary<string, string> accum
...
+                    summarySection.AddRange(DeprecatedSection);
+                    if (!string.IsNullOrEmpty(deprecation))
+                    {
+                        summarySection.AddRange(Markdown.Parse(deprecation));
+                    }
+                }
+
                 this.Documentation = ToMarkdown(summarySection.Concat(descriptionSection));
             }
+            else
```
bettinaheim: Am I missing something or is this else case generating a deprecated warning if there are no doc comments on a declaration even if that declaration has not been marked as deprecated?

src/QsCompiler/DocumentationParser/DocItem.cs
alan-geller: OK, I think it's correct.

src/QsCompiler/DocumentationParser/DocComment.cs
alan-geller: They now get merged.

src/QsCompiler/DocumentationParser/DocComment.cs
alan-geller: No, only if there are no doc comments and there was a Deprecated attribute. This has changed a bit, though.

PR #217, opened 2019-10-25
Top level node representing a Q# compilation

bettinaheim:
We will want this handle going forward, so hence it would be good to actually make that change before we ship the compiled Q# for the first time...


bettinaheim:
I need to merge this to update the runtime so I can make the PR to modify the dll generation.

PR #219, opened 2019-10-28
Support for including rewrite steps defined in .Net dlls into the compilation process

bettinaheim:


src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -127,8 +142,8 @@ private class ExecutionStatus
                 WasSuccessful(options.SerializeSyntaxTree, this.Serialization) &&
                 WasSuccessful(options.BuildOutputFolder != null, this.BinaryFormat) &&
                 WasSuccessful(options.DllOutputPath != null, this.DllGeneration) &&
-                !this.BuildTargets.Values.Any(status => !WasSuccessful(true, status))
-                ? 0 : 1;
+                !this.LoadedRewriteSteps.Values.Any(status => !WasSuccessful(true, status))
+                ? true : false;
```
ScottCarda-MS: Shouldn't the result already be bool? If so, this this line might not be necessary.

src/QsCompiler/TestTargets/Simulation/Target/Simulation.csproj
```diff
@@ -3,22 +3,15 @@
 
   <PropertyGroup>
     <AssemblyName>Simulation</AssemblyName>
-    <TargetFramework>netcoreapp3.0</TargetFramework>
-    <OutputType>Exe</OutputType>
+    <TargetFramework>netstandard2.1</TargetFramework>
```
ScottCarda-MS: I'm confused why this is being changed from 3.0 back to 2.1. Is this not getting .Net 3.0 anymore?

src/QsCompiler/Compiler/CompilationLoader.cs
bettinaheim: Oh boy, yes you are right. :)

src/QsCompiler/TestTargets/Simulation/Target/Simulation.csproj
bettinaheim: The numbering is just a bit confusing: netcoreapp3.0 for an application corresponds to netstandard2.1 for a library (there is no netstandard3.0 - these two numbers are not kept in sync...)

src/QsCompiler/Compiler/ExternalRewriteSteps.cs
```diff
@@ -17,50 +17,88 @@ namespace Microsoft.Quantum.QsCompiler
...
+                this.Priority = step?.Priority ?? throw new ArgumentNullException(nameof(step));
+                this.Name = step.Name;
+                this._GetOutputFolder = step.OutputFolder;
+                this._SetOutputFolder = value => step.OutputFolder = value;
+
+                this.ImplementsPreconditionVerification = step.ImplementsPreconditionVerification;
+                this.ImplementsPostconditionVerification = step.ImplementsPostconditionVerification;
+                this.ImplementsTransformation = step.ImplementsTransformation;
+                this._PreconditionVerification = step.PreconditionVerification;
+                this._PostconditionVerification = step.PostconditionVerification;
```
ScottCarda-MS: This class will have both this.PreconditionVerification *and* this._PreconditionVerification. I'm not sure of the reason behind that.

src/QsCompiler/Compiler/ExternalRewriteSteps.cs
```diff
@@ -17,50 +17,88 @@ namespace Microsoft.Quantum.QsCompiler
...
 
             internal LoadedStep(IRewriteStep step, Uri origin)
             {
-                this.Implementation = step ?? throw new ArgumentNullException(nameof(step));
-                this.Origin = origin ?? throw new ArgumentNullException(nameof(origin)); 
+                this.Origin = origin ?? throw new ArgumentNullException(nameof(origin));
+                this.Priority = step?.Priority ?? throw new ArgumentNullException(nameof(step));
+                this.Name = step.Name;
+                this._GetOutputFolder = step.OutputFolder;
+                this._SetOutputFolder = value => step.OutputFolder = value;
```
ScottCarda-MS: Why is the OutputFolder treated differently then the other properties? Why can't you do this.OutputFolder = step.OutputFolder?

src/QsCompiler/Compiler/ExternalRewriteSteps.cs
```diff
@@ -17,50 +17,88 @@ namespace Microsoft.Quantum.QsCompiler
 {
     internal static class RewriteSteps
     {
-        /// <summary>
-        /// Concrete implementation of the rewrite steps options such that the configured options may be null. 
-        /// </summary>
-        internal class RewriteStepOptions : IRewriteStepOptions
```
ScottCarda-MS: Did you get rid of IRewriteStepOptions to simplify the rewrite interface? Are there no more options to consider other than output folder?

src/QsCompiler/Compiler/ExternalRewriteSteps.cs
```diff
@@ -17,50 +17,88 @@ namespace Microsoft.Quantum.QsCompiler
...
-        {
-            public readonly string OutputFolder;
-            internal RewriteStepOptions(CompilationLoader.Configuration config) =>
-                this.OutputFolder = config.BuildOutputFolder;
-        }
-
         /// <summary>
         /// Concrete implementation of a rewrite steps with an additional property specifying the dll it was loaded from. 
         /// </summary>
         internal class LoadedStep : IRewriteStep
```
ScottCarda-MS: Have you considered using class template to get the IRewriteStep methods and fields?

src/QsCompiler/Compiler/ExternalRewriteSteps.cs
```diff
@@ -94,11 +132,18 @@ Uri WithFullPath(string file)
...
                 var relevantTypes = new List<Type>();
                 Diagnostic LoadError(ErrorCode code, params string[] args) => Errors.LoadError(code, args, ProjectManager.MessageSource(target));
-                try { relevantTypes.AddRange(Assembly.LoadFrom(target.LocalPath).GetTypes().Where(t => t.IsAssignableFrom(typeof(IRewriteStep)))); }
+                Diagnostic LoadWarning(WarningCode code, params string[] args) => Warnings.LoadWarning(code, args, ProjectManager.MessageSource(target));
+                try
+                {
+                    var asmType = Assembly.LoadFrom(target.LocalPath).GetTypes();
+                    var exactInterfaceMatch = asmType.Where(t => typeof(IRewriteStep).IsAssignableFrom(t)); // inherited interface is defined in this exact dll
+                    var compatibleInterface = asmType.Where(t => t.GetInterfaces().Any(t => t.FullName == typeof(IRewriteStep).FullName)); // inherited interface may be defined in older compiler version
+                    relevantTypes.AddRange(exactInterfaceMatch.Concat(compatibleInterface));
```
ScottCarda-MS: This double Where + Concat combo could cause duplicate entries in the resulting list. Do you want entries that satisfy both conditions to appear twice? Even if the conditions are mutually exclusive, if you don't want the duplicate behavior it is best to avoid this pattern either by having one Where with both conditions, or two consecutive Where's on the same enumerable.

bettinaheim:
Thanks @ScottCarda-MS, for taking a look!

PR #224, opened 2019-10-29
Sccarda/generic resolution validation

ScottCarda-MS:
Adds a validation transformation for the monomorphization transformation. Adds some tests for the monomorphization feature.


src/QsCompiler/Tests.Compiler/LinkingTests.fs
```diff
@@ -11,93 +11,191 @@ open Microsoft.Quantum.QsCompiler.DataTypes
...
+        monomorphicCompilation.ValidateMonomorphization () |> Assert.True
+        monomorphicCompilation
+
+
+    [<Fact>]
+    member this.``Monomorphization`` () =
+
+        let makeSig input =
+            let ns, name, args, rtrn = input
+            { Namespace = NonNullable<string>.New ns; Name = NonNullable<string>.New name },
```
bettinaheim: nit: Hmm, in this case actually assigning the name, argument and return type to variables, and then returning them as `(fullName, argType, returnType)` would make things more readable to me.

src/QsCompiler/Transformations/CodeTransformations.cs
```diff
@@ -108,6 +106,27 @@ public static bool Monomorphisize(this QsCompilation compilation, out QsCompilat
             }
             return true;
         }
+
+        /// <summary>
+        /// Validates that the monomorphization step cleared the syntax tree of all references to, and instances of type-parameterized callables.
+        /// Any thrown exception is logged using the given onException action and are silently ignored if onException is not specified or null.
```
bettinaheim:
```suggestion
        /// Any thrown exception is logged using the given onException action and silently ignored if onException is not specified or null.
```

src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/Generics.qs
```diff
@@ -0,0 +1,98 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+/// This namespace contains test cases for generic operations
+namespace Microsoft.Quantum.Testing.Generics {
+
+	operation Test1Main() : Unit {
+		body (...) {
```
bettinaheim: nit: you could just drop the explicit specialization and go with
```suggestion
```
and remove the corresponding closing brace below.

src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/Generics.qs
```diff
@@ -0,0 +1,98 @@
...
+
+			(BasicGeneric<_,_>((BasicGeneric<_,String>("Yes",_))("No"), _))((BasicGeneric<Double, _>(_, 2))(1.0));
+
+			BasicGeneric((),());
+			BasicGeneric("","");
+			BasicGeneric(1.0,2);
+
+			let temp3 = NoArgsGeneric<Double>();
+			let bar = (ReturnGeneric<String, _, _>("Yes", (ReturnGeneric<Double, _, _>(12.0, "No", _))(4), _))("Maybe");
+		}
```
bettinaheim:
```suggestion
```

src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/Generics.qs
```diff
@@ -0,0 +1,98 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
bettinaheim: nit: Do you mind running the replacement "\t" to "    " (Ctrl + h and then click on the little star to enable regex)? Thanks!

src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/Generics.qs
```diff
@@ -0,0 +1,98 @@
...
+	operation Test2Main () : Unit is Adj+Ctl {
+        body (...) {
+			using (q = Qubit()) {
+				GenericCallsGeneric(q, 12);
+				let temp = ArrayGeneric(q, ArrayGeneric(q, "twelve"));
+				let temp2 = ArrayGeneric(q, 12);
+			}
+        }
+        
+        adjoint (...) {
```
bettinaheim: I've got an interesting case to check & add a test for: Does the transformation work as expected if a particular type specialization is only invoked as part of the adjoint specialization (or controlled, or controlled adjoint) of a used operation? In particular with additional control qubits, this seems like a valid use case. Since you may not know whether or not a particular functor specialization is executed at compile time, it seems ok to assume that all of the defined functor specializations for a particular set of type arguments will be executed. 
I am happy for this to be a separate PR. ;)

src/QsCompiler/CompilationManager/CompilationUnitManager.cs
```diff
@@ -757,12 +757,20 @@ public class Compilation
...
-                    SemanticDiagnostics[file]) :
-                Enumerable.Empty<Diagnostic>();
+                    ScopeDiagnostics[file]
+                        .Concat(SyntaxDiagnostics[file])
+                        .Concat(ContextDiagnostics[file])
+                        .Concat(HeaderDiagnostics[file])
+                        .Concat(SemanticDiagnostics[file]) :
+                    Enumerable.Empty<Diagnostic>();
+
+            public IEnumerable<Diagnostic> Diagnostics() =>
```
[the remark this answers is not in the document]
ScottCarda-MS: Oh I forgot to do this when I copied it over. Thanks for catching this.

src/QsCompiler/Tests.Compiler/LinkingTests.fs
```diff
@@ -11,93 +11,193 @@ open Microsoft.Quantum.QsCompiler.DataTypes
 open Microsoft.Quantum.QsCompiler.Diagnostics
 open Microsoft.Quantum.QsCompiler.SyntaxExtensions
 open Microsoft.Quantum.QsCompiler.SyntaxTree
+open Microsoft.Quantum.QsCompiler
 open Xunit
 open Xunit.Abstractions
+open Microsoft.Quantum.QsCompiler.SyntaxTokens
+open System.Collections.Immutable
```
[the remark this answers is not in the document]
ScottCarda-MS: You are right these should be sorted.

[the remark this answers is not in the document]
ScottCarda-MS: Hmm, I don't seem to have something that automatically sorts open namespaces in FS. I will sort it manually.

src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/Generics.qs
ScottCarda-MS: Will do! I also have noticed that in some fs files there are lots of spaces at the ends of lines. I have been removing those with: <space>+(?=[\r\n])

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -283,10 +294,18 @@ public CompilationLoader(SourceLoader loadSources, ReferenceLoader loadReference
             {
                 this.CompilationStatus.Monomorphization = 0;
                 void onException(Exception ex) => this.LogAndUpdate(ref this.CompilationStatus.Monomorphization, ex); 
-                var succeeded = this.CompilationOutput != null && this.CompilationOutput.Monomorphisize(out this.CompilationOutput, onException);
+                bool succeeded = this.CompilationOutput != null && this.CompilationOutput.Monomorphisize(out this.CompilationOutput, onException);
```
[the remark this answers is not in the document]
ScottCarda-MS: I specifically changed this from a var to a bool. I am not fond of overuse of the var keyword in C# because it contributes to a style of coding that is hard to read. This particular case wasn't too bad to understand, but to the same effect it could never _not_ be a bool, so choosing to change the data structure somewhere would not change the type here.

src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/Generics.qs
ScottCarda-MS: The changes to SyntaxTree.fs are all of that nature.

src/QsCompiler/Tests.Compiler/LinkingTests.fs
```diff
@@ -11,93 +11,191 @@ open Microsoft.Quantum.QsCompiler.DataTypes
...
+            |> Seq.choose (function | QsCallable c -> Some c | _ -> None)
+            |> Seq.map (fun call -> (call.FullName, call.Signature.ArgumentType, call.Signature.ReturnType))
+
+        let doesCallMatchSig call signature =
+            let (call_fullName : QsQualifiedName), call_argType, call_rtrnType = call
+            let (sig_fullName : QsQualifiedName), sig_argType, sig_rtrnType = signature
+
+            call_fullName.Namespace.Value = sig_fullName.Namespace.Value &&
+            call_fullName.Name.Value.EndsWith sig_fullName.Name.Value &&
+            call_argType = sig_argType &&
```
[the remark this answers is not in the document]
ScottCarda-MS: Yes, I have renamed some of the variables to make it clearer that I am not comparing QsCallable objects, but just their signatures.

src/QsCompiler/Transformations/MonomorphizationValidation.cs
```diff
@@ -0,0 +1,69 @@
...
+
+                return base.onSignature(s);
+            }
+        }
+
+        internal class MonomorphizationValidationExpression :
+            ExpressionTransformation<Core.ExpressionKindTransformation, MonomorphizationValidationExpressionType>
+        {
+            public MonomorphizationValidationExpression() :
+                base(null, expr => new MonomorphizationValidationExpressionType(expr as MonomorphizationValidationExpression))
```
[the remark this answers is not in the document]
ScottCarda-MS: The kind recursion it would prevent is probably not needed for this transform, but this should be filled out anyway as putting null here causes the transformation framework to work in an unintended way. It's my thinking now that this should never be null to avoid strange behavior that is hard to debug. I will fill this argument out.

PR #228, opened 2019-10-31
VS Code extension adaptions

bettinaheim:


src/VSCodeExtension/src/languageServer.ts
```diff
@@ -170,6 +170,7 @@ export class LanguageServer {
         // Before anything else, look at the user's configuration to see
         // if they set a path manually for the language server.
         let packageInfo = getPackageInfo(this.context);
```
cgranade: Honestly, at this point, we might as well error out if packageInfo is undefined.

PR #230, opened 2019-11-01
Lowercase UIDs on deprecation attributes.

cgranade:
This PR causes API documentation generated for newtypes and callables annotated with the `@Deprecated` attribute to use lower-cased links (`@""` in DocFX-Flavored Markdown), consistent with our convention for generating UIDs in API documentation.


src/QsCompiler/DocumentationParser/DocComment.cs
```diff
@@ -258,7 +258,7 @@ void ParseMapSection(IEnumerable<Block> blocks, Dictionary<string, string> accum
 
             var deprecationSummary = String.IsNullOrWhiteSpace(replacement) 
                                     ? DiagnosticItem.Message(WarningCode.DeprecationWithoutRedirect, new string[] { name })
-                                    : DiagnosticItem.Message(WarningCode.DeprecationWithRedirect, new string[] { name, "@\"" + replacement + "\"" });
+                                    : DiagnosticItem.Message(WarningCode.DeprecationWithRedirect, new string[] { name, "@\"" + replacement.ToLowerInvariant() + "\"" });
```
bettinaheim: I believe you will need to adapt the test accordingly.

PR #231, opened 2019-11-06
Sccarda/master to monomorph

ScottCarda-MS:
Merging in master
Added rewrite step for monomorphization


src/QsCompiler/Tests.Compiler/LinkingTests.fs
```diff
@@ -110,11 +112,10 @@ type LinkingTests (output:ITestOutputHelper) =
         compilationDataStructures.Diagnostics() |> Seq.exists (fun d -> d.IsError()) |> Assert.False
 
         Assert.NotNull compilationDataStructures.BuiltCompilation
-        let mutable monomorphicCompilation = compilationDataStructures.BuiltCompilation
-        compilationDataStructures.BuiltCompilation.Monomorphisize &monomorphicCompilation |> Assert.True
+        let monomorphicCompilation = ResolveGenericsSyntax.Apply compilationDataStructures.BuiltCompilation
```
bettinaheim: Could we rename ResolveGenericsSyntax to something like Monomorphize (I am open to other suggestions)?

src/QsCompiler/Compiler/RewriteSteps/MonomorphizationRewriteStep.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        public string OutputFolder { get; set; }
+
+        public bool ImplementsTransformation { get; }
+
+        public bool ImplementsPreconditionVerification { get; }
+
+        public bool ImplementsPostconditionVerification { get; }
+
+        public MonomorphizationRewriteStep(bool doPostConditionValidation = false)
```
bettinaheim: It seems like you added basically an override to disable the post condition verification in all cases here? What is the reason for that?

src/QsCompiler/Tests.Compiler/LinkingTests.fs
ScottCarda-MS: I added a wrapper class around the transformations called MonomorphizationTransfromation.

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -352,15 +387,50 @@ public CompilationLoader(SourceLoader loadSources, ReferenceLoader loadReference
...
+            if (this.CompilationOutput != null)
             {
-                this.CompilationStatus.BuildTargets[buildTarget.Key] = 0;
-                var succeeded = this.PathToCompiledBinary != null && buildTarget.Value != null &&
-                    buildTarget.Value(this.PathToCompiledBinary, ex => this.LogAndUpdate(buildTarget.Key, ex));
-                if (!succeeded) this.LogAndUpdate(buildTarget.Key, ErrorCode.TargetExecutionFailed, new[] { buildTarget.Key });
-            } 
+                for (int i = 0; i < this.ExternalRewriteSteps.Length; i++)
+                {
+                    this.CompilationStatus.LoadedRewriteSteps[i] = this.ExecuteRewriteStep(this.ExternalRewriteSteps[i], this.CompilationOutput, out this.CompilationOutput);
```
bettinaheim: The way it is set up now the CompilationOutput may actually be modified even though the rewrite step failed. Could we please keep it so that the rewrite steps either succeeds, or does not modify the CompilationOutput.

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -418,12 +418,14 @@ private int ExecuteRewriteStep(RewriteSteps.LoadedStep rewriteStep, QsCompilatio
...
-                bool transformationPassed = !executeTransformation || rewriteStep.Transformation(compilation, out transformed);
+                bool transformationPassed = !executeTransformation || rewriteStep.Transformation(compilation, out compilation);
                 if (!transformationPassed) this.LogAndUpdate(ref status, ErrorCode.RewriteStepExecutionFailed, new[] { rewriteStep.Name, messageSource });
 
                 bool executePostconditionVerification = this.Config.EnableAdditionalChecks && transformationPassed && rewriteStep.ImplementsPostconditionVerification;
-                bool postconditionPassed = !executePostconditionVerification || rewriteStep.PostconditionVerification(transformed);
+                bool postconditionPassed = !executePostconditionVerification || rewriteStep.PostconditionVerification(compilation);
                 if (!postconditionPassed) this.LogAndUpdate(ref status, ErrorCode.PostconditionVerificationFailed, new[] { rewriteStep.Name, messageSource });
+
+                transformed = compilation;
```
bettinaheim: You should only set this if the status indicates success.

PR #232, opened 2019-11-06
Features/monomorphization

ScottCarda-MS:
Adding the Monomorphization Transformation to remove type-parameterized information from the syntax tree as part of the compilation.


bettinaheim:
I'll merge this as soon as the build has finished. 
Well done, @ScottCarda-MS!

PR #233, opened 2019-11-06
Syntax Tree Walkers

alan-geller:
Syntax tree walkers parallel syntax tree transformations, but effectively do `iter` instead of `fold`. In some cases they can be a bit more performant.


src/QsCompiler/QsTargeting/Target.fs
```diff
@@ -0,0 +1,24 @@
...
+
+type Target(builtIns : NamespaceManager, capabilityLevel : CapabilityLevel) =
+    member this.BuiltIns with get () = builtIns
+
+    member this.CapabilityLevel with get () = capabilityLevel
+
+    member this.ValidateSpecialization (spec : QsSpecialization) : List<DiagnosticItem> = 
+        let diagnostics = new List<DiagnosticItem>()
+        if spec.RequiredCapability > this.CapabilityLevel
+        then diagnostics.Add(Error ErrorCode.UnexpectedCompilerException)
```
ScottCarda-MS: Are these capability levels something that is documented for the public to see? If so, a more specific error message here could be useful.

src/QsCompiler/QsTargeting/Target.fs
```diff
@@ -0,0 +1,24 @@
...
+        diagnostics
+
+    member this.ValidateSyntaxTree (ns : QsNamespace, cancellationToken : CancellationToken) : List<DiagnosticItem> =
```
ScottCarda-MS: None of the parameters here seem to be used. Was there meant to be more logic here at a later time? If so, please add a ToDo as a reminder.

alan-geller: I'm going to remove this file from the PR, and add it later.

src/QsCompiler/QsTargeting/Target.fs
alan-geller: I'm going to remove this file from the PR, and add it later.
Yes, they will be public.

src/QsCompiler/Core/ExpressionWalker.fs
```diff
@@ -0,0 +1,454 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
ScottCarda-MS: I'm confused about the purpose of these walkers. Does the ExpressionTransformation.fs, StatementTransformation.fs, and TreeTransformation.fs do all of this already?

alan-geller: The walker classes iterate through the syntax tree, without producing any new objects, while the transformation classes fold through the syntax tree, producing a new version of the tree (potentially sharing some structure with the source tree). For situations where you don’t need or want the new tree, the walker classes are slightly simpler and should run somewhat faster while producing less garbage in memory.

src/QsCompiler/QsTargeting/CallGraphWalker.fs
```diff
@@ -0,0 +1,239 @@
...
+    override this.StatementKind = upcast kindXformer
+
+    member this.Holder with get() = graph
+
+/// This syntax tree transformer fills in the CapabilityLevel fields in specializations,
+/// based on information gathered by the associated scope and other transformations.
+type TreeGraphBuilder() =
+    inherit SyntaxTreeWalker()
+
+    let graph = new CallGraph()
```
ScottCarda-MS: I am not sure how you get the graph back to the calling context. I make a new TreeGraphBuilder obj, and use its Walk method to process my syntax tree, but where do I get back the graph?

src/QsCompiler/Core/ExpressionWalker.fs
ScottCarda-MS: Oh, yes I can see the utility in that. It is unfortunate that there must be this code duplication, but I don't see any reasonable way around that. At a future point (not needed for this PR) we may want to consider have a script automatically generate these walker classes from their associated transformation classes, but that would only really be useful if we expect them to change much.

src/QsCompiler/QsTargeting/CallGraphWalker.fs
alan-geller: Argh... I messed that up while refactoring (the call graph walker was originally embedded into the capability leveler). I added the appropriate property, and also cleaned up some old/dead code.

src/QsCompiler/Core/ExpressionWalker.fs
alan-geller: I spent some time trying to come up with a clever way to have one set of classes provide both functions, perhaps with some clever use of generics, but I wasn't clever enough to do so. :-(

src/QsCompiler/QsTargeting/Targeting.fsproj
```diff
@@ -0,0 +1,23 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="CapabilityLeveler.fs" />
+    <Compile Include="CallGraphWalker.fs" />
+    <Compile Include="Target.fs" />
```
ScottCarda-MS: Seems like having this here is causing the build issue. This file was removed, right?

alan-geller: And of course it built fine on my machine because I didn't actually delete the file, I just removed it from git... Fixed now, I think.
Thanks for the catch!

QsCompiler.sln
```diff
@@ -270,6 +296,7 @@ Global
 		{D50583DF-FBEF-45EF-B523-70B2CDBE1DD1} = {76BA96DA-DC1E-4315-A3ED-5F0700A79812}
 		{2E331781-F7ED-4EF1-8451-896636C6D93A} = {76BA96DA-DC1E-4315-A3ED-5F0700A79812}
 		{D2E36476-A65F-4310-9C4C-B721BCC47B00} = {6077A717-50BF-4F87-B439-CA549AF6A4AE}
+		{A83A9BA2-A774-4F07-B801-5A4A73598F62} = {D2E36476-A65F-4310-9C4C-B721BCC47B00}
```
ScottCarda-MS: This line is indicating that there should be another project besides Library1.csproj under src\QsCompiler\TestTargets\Libraries, but no such project is listed there, earlier in this solution file. If there is supposed to be a project, you could add it to the solution and to the PR, otherwise I think all references to {A83A9BA2-A774-4F07-B801-5A4A73598F62} have to be removed for it to build correctly.

alan-geller: Duly removed. Odd that this is ignored when I build locally.

ScottCarda-MS: It build locally for me as well, which is strange. Maybe VSO has a more restrictive process.

src/QsCompiler/QsTargeting/CapabilityLeveler.fs
```diff
@@ -0,0 +1,243 @@
...
+        currentOperationLevel <- c.Attributes |> checkForLevelAttributes
+        let result = base.beforeCallable(c)
+        result
+
+    override this.onSpecializationImplementation(s) =
+        let codeLevel = match s.Implementation with
+                        | Intrinsic -> CapabilityLevel.Minimal
+                        | External -> CapabilityLevel.Unset
+                        | Generated _ -> CapabilityLevel.Unset
+    // TODO: For generated specializations, we need to find the appropriate "body" declaration
```
bettinaheim: You know, this specifically makes me wonder if we might instead want to track it as inferred information. Let's maybe quickly discuss that option.

src/QsCompiler/Core/Dependencies.fs
```diff
@@ -67,7 +67,6 @@ type BuiltIn = {
         TypeParameters = ImmutableArray.Empty
     }
 
-
```
bettinaheim: I you undo that whitespace change, then you don't have conflicts. :)

src/QsCompiler/Core/TreeWalker.fs
```diff
@@ -0,0 +1,196 @@
...
+
+namespace Microsoft.Quantum.QsCompiler.Transformations.Core
+
+open System.Collections.Immutable
+open System.Linq
+open Microsoft.Quantum.QsCompiler.DataTypes
+open Microsoft.Quantum.QsCompiler.SyntaxTokens
+open Microsoft.Quantum.QsCompiler.SyntaxTree
+
+// This are inherited here from TreeTransformation.fs; kept here as comments for documentation.
```
bettinaheim:
```suggestion
```

src/QsCompiler/Core/TreeWalker.fs
```diff
@@ -0,0 +1,196 @@
...
+//type QsArgumentTuple = QsTuple<LocalVariableDeclaration<QsLocalSymbol>>
```
bettinaheim:
```suggestion
```

PR #241, opened 2019-11-09
Verification for unit test attributes

bettinaheim:


src/QsCompiler/Core/SymbolTable.fs
```diff
@@ -823,7 +829,31 @@ and NamespaceManager
...
+                            | _ -> true // invalid type
+                        match signature.Argument.Items |> Seq.toList with 
+                        | [] -> signature.ReturnType |> isUnitType
+                        | [(_, argType)] -> argType |> isUnitType && signature.ReturnType |> isUnitType
+                        | _ -> false
+                    match box decl.Defined with 
+                    | :? CallableSignature as signature when signature |> isUnitToUnit && not (signature.TypeParameters.Any()) -> 
+                        let arg = att.Argument |> AttributeAnnotation.NonInterpolatedStringArgument (fun ex -> ex.Expression)
+                        match (if arg = null then null else arg.ToLowerInvariant()) with 
+                        | "quantumsimulator" | "tracesimulator" | "toffolisimulator" -> // error message below needs to be adapted when this changes!
```
ScottCarda-MS: Seems like the comment here is a "ToDo". If it is, please add the "ToDo" to it to make it easier to find and track.

Issue #40, opened 2019-07-12
Docs gen fails if previously generated docs file exist in the same folder


bettinaheim:
**Describe the bug**

Running documentation generation twice in a row causes the second execution to fail due to a "key already exists" exception.

**To Reproduce**

1. Call the command line compiler specifying --doc "someFolder" (should build fine)
2. Do the same things again without deleting "someFolder" (fails)

**Expected behavior**

I think the best thing would be to just silently replace any content that already exists on disk, and maybe log a warning instead. @alan-geller and @anpaz-msft  for advice. 

**System information**

irrelevant 

**Additional context**

Mostly, we need to clarify what the correct behavior should be. 
The reason docgen reads the files in the folder is to handle the situation where you have a library namespace that spans multiple projects.

alan-geller:
This might need some diagnosis as well as a (probably simple) code fix. Doc gen output files are always opened with mode Create, which should truncate and overwrite if the file exists and create a new file if the file doesn't exist.
One diagnostic tip is to provide a log file name to DoIgnoringExceptions in the Utils.cs file; this will write file-related exceptions (which can include file not found, among others that aren't actually errors) to the log file.

PR #242, opened 2019-11-09
Fix for doc generation failing  on repeated compilations

codingupastorm:
For #40.

This means docgen will succeed when the files already exist. It will overwrite, always using the newest version of the docs.


bettinaheim:
@codingupastorm Awesome, thanks for putting this up! 
I'll take a look shortly to double check everything works fine with our larger build setup, but everyone is going to be really happy to have this fixed.

PR #246, opened 2019-11-12
Environment Intrinsic Mapping Transformation

ScottCarda-MS:
This transformation maps environment definitions for operations and types into the source code-generated syntax tree.


src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/IntrinsicMapping.qs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
bettinaheim: Looking at this, it might make sense to take it as a precondition that the environment tree does not contain any type parametrized callables. What do you think?

src/QsCompiler/Transformations/IntrinsicMappingTransformation.cs
```diff
@@ -0,0 +1,37 @@
...
+                .ToImmutableArray(), target.EntryPoints);
+        }
+
+        private static QsNamespace MergeNamespaces(QsNamespace overriding, QsNamespace accepting)
+        {
+            var overridingNames = overriding.Elements.Select(x => x.FullName).ToImmutableHashSet();
+
+            return accepting.WithElements(_ => accepting.Elements
+                .Where(elem => !overridingNames.Contains(elem.FullName))
+                .Concat(overriding.Elements)
```
bettinaheim: I like the simplicity of it all. I think we agreed to keep around the overriding elements that are not used since the overhead of computing the call graph seems unnecessary in this case. It may be good to document that somewhere, though (I suggest as doc comment on MergeNamespaces, and more importantly even on the IntrinsicMappingTransformation).

src/QsCompiler/Transformations/IntrinsicMappingTransformation.cs
```diff
@@ -0,0 +1,37 @@
...
+// Licensed under the MIT License.
+
+using Microsoft.Quantum.QsCompiler.SyntaxTree;
+using System.Collections.Immutable;
+using System.Linq;
+
+
+namespace Microsoft.Quantum.QsCompiler.Transformations.IntrinsicMappingTransformation
+{
+    public class IntrinsicMappingTransformation
```
bettinaheim: How about renaming it to something like InstructionSetReplacement (mapping to me sounds like it's tied to a geometry rather than an instruction set)?

src/QsCompiler/Compiler/RewriteSteps/IntrinsicMapping.cs
```diff
@@ -0,0 +1,50 @@
...
+{
+    internal class IntrinsicMapping : IRewriteStep
+    {
+        public string Name { get; }
+        public int Priority { get; }
+        public string OutputFolder { get; set; }
+
+        public bool ImplementsTransformation { get; }
+        public bool ImplementsPreconditionVerification { get; }
+        public bool ImplementsPostconditionVerification { get; }
```
bettinaheim: Given that the mapping itself blindly replaces declarations with the same full name, a postcondition verification seems useful (possibly in a separate PR). And I am tempted to suggest to keep an internal state with the original signatures that got replaced in here, such that the postcondition verification doesn't need to walk the whole tree but can just compare those with the replaced ones.

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -71,6 +71,11 @@ public struct Configuration
             /// </summary>
             public bool SkipMonomorphization;
             /// <summary>
+            /// If set to true, the compiler will include all environment-defined callables and types
+            /// in the resulting syntax tree.
+            /// </summary>
+            public bool DoIntrinsicMapping;
```
bettinaheim: I would actually suggest to hold off with making this configurable in the same way that other things are (meaning I would actually not yet write the handle in the compilation loader - it's quick to add after all). It is not yet clear to me whether that makes sense or not.

src/QsCompiler/Compiler/RewriteSteps/IntrinsicMapping.cs
```diff
@@ -0,0 +1,50 @@
...
+        private QsCompilation Envrionment { get; }
+
+        public IntrinsicMapping(QsCompilation environment)
+        {
+            Name = "IntrinsicMapping";
+            Priority = 10; // Not used for hard-coded transformations like this
+            OutputFolder = null;
+            ImplementsTransformation = true;
+            ImplementsPreconditionVerification = false;
+            ImplementsPostconditionVerification = false;
```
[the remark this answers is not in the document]
ScottCarda-MS: I'm going to hold off on that for now.

src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/IntrinsicMapping.qs
ScottCarda-MS: Validating the state of the incoming environment tree could get pretty involved. I think that should be done separately from this transformation, in whatever process we have for generating those environment trees.

src/QsCompiler/Tests.Compiler/LinkingTests.fs
```diff
@@ -14,9 +13,9 @@ open Microsoft.Quantum.QsCompiler.Diagnostics
 open Microsoft.Quantum.QsCompiler.SyntaxExtensions
 open Microsoft.Quantum.QsCompiler.SyntaxTokens
 open Microsoft.Quantum.QsCompiler.SyntaxTree
+open Microsoft.Quantum.QsCompiler.Transformations.IntrinsicMappingTransformation
```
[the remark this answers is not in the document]
ScottCarda-MS: Now that we are using rewrite steps, there are two classes (and two files) that will be similarly named for each step: the step class that implements the IRewriteStep, and the transformation class. To more clearly distinguish between the two, I will name the transformation class with a 'Transformation' at the end.

I will change "Intrinsic Mapping" to "Intrinsic Resolution" throughout the solution.

src/QsCompiler/Transformations/IntrinsicMappingTransformation.cs
ScottCarda-MS: See other conversation regarding naming.

src/QsCompiler/Tests.Compiler/Signatures.fs
```diff
@@ -0,0 +1,147 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
ScottCarda-MS: I've moved the mentioned files to a new folder.

src/QsCompiler/Compiler/RewriteSteps/IntrinsicMapping.cs
ScottCarda-MS: I'm not clear what you are suggesting we check for in post-verification. Could you please clarify?

bettinaheim: Say there was by accident a callable with the same full name in the environment and in the compilation but they had different signatures, then the tree after pulling in the environment specific declarations is broken. It would be good to have the option to verify that. 
Concretely, that means that you could have an internal state in the rewrite step that keeps the signatures of all replaced callables (e.g. a dictionary of QsQualifiedName to ResolvedSignature) that is filled during the Transformation, and then a postcondition verification that given the built compilation, checks that all callables in that declaration are either not present in the dictionary or have the same signature as those in the dictionary.

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -329,14 +329,15 @@ public CompilationLoader(SourceLoader loadSources, ReferenceLoader loadReference
             foreach (var diag in this.VerifiedCompilation?.Diagnostics() ?? Enumerable.Empty<Diagnostic>())
```
bettinaheim: nit:  I think the changes here are no longer needed, and I tend to try and limit the scope of defined things if possible.

src/QsCompiler/Transformations/MonomorphizationValidation.cs
```diff
@@ -1,12 +1,12 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-using Microsoft.Quantum.QsCompiler.DataTypes;
-using Microsoft.Quantum.QsCompiler.SyntaxTokens;
-using Microsoft.Quantum.QsCompiler.SyntaxTree;
 using System;
 using System.Collections.Immutable;
 using System.Linq;
+using Microsoft.Quantum.QsCompiler.DataTypes;
```
bettinaheim: Thanks for moving them down. :)

src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/IntrinsicResolution.qs
```diff
@@ -0,0 +1,178 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
bettinaheim: The one thing that is not currently giving any kind of error is if there is an intrinsic callable in the compilation that is not contained in teh environment tree, correct? 
I would be good if we could also verify this, but it may make sense to add this in a separate PR when we've had a conversation on how to integrate the existing targeting.

src/QsCompiler/Transformations/IntrinsicResolutionTransformation.cs
```diff
@@ -0,0 +1,81 @@
...
+                MergeElements(overriding.Elements.Where(x => x.IsQsCallable), accepting.Elements.Where(x => x.IsQsCallable), true)
+                .Concat(MergeElements(overriding.Elements.Where(x => x.IsQsCustomType), accepting.Elements.Where(x => x.IsQsCustomType), false))
+                .ToImmutableArray());
+        }
+
+        private static IEnumerable<QsNamespaceElement> MergeElements(IEnumerable<QsNamespaceElement> overriding, IEnumerable<QsNamespaceElement> accepting, bool checkSignature = false)
+        {
+            var overridingNames = overriding.ToImmutableDictionary(x => x.GetFullName());
+
+            if (checkSignature) // Check that overriding elems and accepting elems with the same names have the same signatures if they are callables
```
bettinaheim: It seems to me that to have a correct behavior this parameter would currently have to be false for user defined types. It also looks like there isn't a corresponding check for user defined types. I would suggest to eliminate the checkSignature parameter and instead go with if (overrideElem is QsNamespaceElement.QsCallable overrideCall) { ... }. 
I would also add the corresponding check for user defined types. The comparison here needs to include the named items (i.e. you need to check the TypeItems).

src/QsCompiler/Transformations/IntrinsicResolutionTransformation.cs
```diff
@@ -0,0 +1,81 @@
...
+                .Where(elem => !overridingNames.ContainsKey(elem.GetFullName()))
+                .Concat(overriding);
+        }
+
+        private static bool CompareSigs(ResolvedSignature first, ResolvedSignature second)
+        {
+            return
+                StripPositionInfo.Apply(first.ArgumentType).Equals(StripPositionInfo.Apply(second.ArgumentType)) &&
+                StripPositionInfo.Apply(first.ReturnType).Equals(StripPositionInfo.Apply(second.ReturnType)) &&
+                !first.TypeParameters.Select((val, i) => val.Equals(second.TypeParameters[i])).Any(x => !x) && // ToDo: this should be a more robust check
```
bettinaheim:
```suggestion
                first.TypeParameters.Select((val, i) => val.Equals(second.TypeParameters[i])).All() && // ToDo: this should be a more robust check
```

src/QsCompiler/Tests.Compiler/LinkingTests.fs
```diff
@@ -102,96 +53,112 @@ type LinkingTests (output:ITestOutputHelper) =
...
-                    genericsNs, "ArrayGeneric", [|"Qubit"; "Double"|], "Int";
-                    genericsNs, "ArrayGeneric", [|"Qubit"; "Unit"|], "Int";
-                |]
-            |] |> Seq.map (fun case -> Seq.map (fun _sig -> makeSig _sig) case)
-
-        for testCase in LinkingTests.GetEntryPoints "Monomorphization.qs" |> Seq.zip targetSignatures do
-            this.CompileAndTestMonomorphization (snd testCase) |>
-            LinkingTests.SignatureCheck (fst testCase)
+    [<Fact>]
+    member this.``Intrinsic Resolution Param UDT`` () =
```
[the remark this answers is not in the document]
ScottCarda-MS: I'm a little confused by this. Test 4 tests that mapping works correctly with UDT params. Test 5 is the positive test for mapping with specialization (test that it works when it should work). Test 6 is the negative test for mapping with specializations (tests that it errors when it should fail). Maybe I misunderstood which three tests you are referring to?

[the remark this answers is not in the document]
bettinaheim: No you didn't. My comment was suggesting to possibly aggregate this under on [<Fact>], that's all: 
```suggestion
    [<Fact>]
    member this.``Intrinsic Resolution`` () =
        this.RunIntrinsicResolutionTest 1
        this.RunIntrinsicResolutionTest 2
        Assert.Throws<Exception> (fun _ -> this.RunIntrinsicResolutionTest 3) |> ignore
```
My suggestion is based on that three tests test different aspects of the same functionality, but there are certainly arguments against this aggregation into one test containing several test cases.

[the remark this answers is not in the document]
ScottCarda-MS: Added in Category trait for the tests related to Intrinsic Resolution.

PR #248, opened 2019-11-13
Make PreEvaluation more customizable

msoeken:
This PR adds a function called `PreEvaluation.Script (script : Callables -> OptimizingTransformation list) (arg : QsCompilation)` which takes as input a sequence of transformation passes. It changes `PreEvaluation.All` to be a specialization of the `Script` function that takes as input the script which is currently specified in the code base.

This change requires to make the OptimizingTransformation type public to write custom optimizing passes outside of the code base. The namespace `MinorTransformations` is changed to `Tools` and more documentation is added.


src/QsCompiler/Optimizations/Utils/HelperTransformations.fs
```diff
@@ -75,8 +75,11 @@ type internal MutationChecker() =
     }
 
 
-/// Represents a transformation meant to optimize a syntax tree
-type internal OptimizingTransformation() =
+/// A transformation base meant to optimize a syntax tree
```
bettinaheim:
```suggestion
/// A transformation base for optimizing syntax tree transformations.
```

src/QsCompiler/Optimizations/Utils/Utils.fs
```diff
@@ -23,7 +23,7 @@ type internal InitKind = QsInitializerKind<ResolvedInitializer, TypedExpression>
 
 
 /// Represents the dictionary of all callables in the program
-type internal Callables (m: ImmutableDictionary<QsQualifiedName, QsCallable>) =
+type Callables (m: ImmutableDictionary<QsQualifiedName, QsCallable>) =
```
bettinaheim: This specific type has been eliminated with the latest updates I think.

src/QsCompiler/Optimizations/PreEvaluation.fs
```diff
@@ -21,12 +21,13 @@ open Microsoft.Quantum.QsCompiler.SyntaxTree
...
-        // TODO: these should be configurable
-        let removeFunctions = false
-        let maxSize = 40
+    /// Attempts to pre-evaluate the given sequence of namespaces
+    /// as much as possible with a script of optimizing transformations
+    ///
+    /// Some of the optimizing transformations need a dictionary of all
+    /// callables by name.  Consequently, the script is generated by a
+    /// function that takes as input such a dictionary of callables.
+    static member Script (script : Callables -> OptimizingTransformation list) (arg : QsCompilation) =
```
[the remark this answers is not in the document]
msoeken: Done.

src/QsCompiler/Optimizations/Utils/Utils.fs
msoeken: No, it's been used only within the Optimization project.

bettinaheim: Hmm, I would rather not introduce another public type Callables. While such an additional buffer makes sense for the optimization project alone, it's current functionality is rather limited and it is mostly there in anticipation of what might make sense in the future. I'd rather introduce an additional public class when it is actually needed. 
Would you hence mind just eliminating this class instead? Passing through ImmutableDictionary<QsQualifiedName, QsCallable> for now seems fine.

msoeken: Addressed in d964c6e

PR #250, opened 2019-11-14
Capability Leveler

ScottCarda-MS:
Adds a transformer/walker combo that computes the required classical capability level for each specialization.

Depends on PR: #233


samarsha:
Closing this in favor of #600, which implements the same functionality.

PR #255, opened 2019-11-20
Fixes QsCompiler build scripts in macos, linux

eginez:
Hi all

A couple of minor changes I had to add to make the project `QsCompiler.sln` build on mac os x.


bettinaheim:
@eginez Sorry for the Mac compatibility issue & thanks for addressing it!
It actually looks like windows is fine with teh '/' slashes in the project files, so I'd suggest to not make a destinction by platform and just go with '/' instead of '\' everywhere in those two project files.

eginez:
Sure thing I'll collapse it all  to forward slash

PR #258, opened 2019-11-26
Support arbitrary execution targets for unit tests

bettinaheim:
This PR adds support to specify arbitrary execution targets for unit tests via a fully qualified name. 
It also adapts the expected name of the attribute to be Test instead of TestOperation, and adapts the templates accordingly.

### DIFF
--- a/src/QsCompiler/Tests.Compiler/TestCases/LocalVerification.qs
+++ b/src/QsCompiler/Tests.Compiler/TestCases/LocalVerification.qs
@@ -878,127 +878,154 @@ namespace Microsoft.Quantum.Testing.LocalVerification {
 
     // unit tests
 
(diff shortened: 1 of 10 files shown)


src/QsCompiler/Tests.Compiler/TestCases/LocalVerification.qs
```diff
@@ -878,127 +878,154 @@ namespace Microsoft.Quantum.Testing.LocalVerification {
 
     // unit tests
 
-    @ TestOperation("QuantumSimulator")
+    @ Test("QuantumSimulator")
```
cgranade:
```suggestion
    @Test("QuantumSimulator")
```

PR #264, opened 2019-12-07
Language server command line and logging

bettinaheim:
Creating more logging information upon server launch. Also, different command line package.


src/VisualStudioExtension/.gitignore
```diff
@@ -1,4 +1,3 @@
 
 # Files generated using .v.template
-Common/AssemblyCommon.cs
```
ScottCarda-MS: Why was this being ignored before? And do we need to replace this with an updated path?

bettinaheim: It is still being ignored now (just by the global .gitignore). It is ignored because it is generated from an AssemblyCommon.cs.v.template file, which is under version control.

Issue #205, opened 2019-10-17
Code actions for wrong capitalization


bettinaheim:
I would be nice if code actions were provided in the case where a keyword or known identifier is misspelled e.g. lower case instead of upper case etc.

bettinaheim:
Ping me if someone is interested in picking this up - I am happy to help.

samarsha:
It would also be cool, in addition to the code action, if the error message showed a list of suggestions for unknown identifiers.

cgranade:
@samarsha: Agreed entirely. Unknown identifier is by far the most common error in IQ#, for instance, but there's not much to help users out at that point, unfortunately.

PR #266, opened 2019-12-16
Add suggestions for alternative capitalization that would fix errors.

jwittner:
Sorry for not reaching out on the issue beforehand as called for in the contribution guide, my original intention was to just understand the workflow of the repo and how to enhance the VS extension. If someone else is already working on this, no big deal - I learned a lot about the code base.

I'm sure there are details and cases I'm missing for this contribution (my first to the project), but I believe this is a step towards resolving #205. Please let me know if you'd like any changes or if you uncover any flaws. =)

I was testing with return types and constructors:

image

image


bettinaheim:
@jwittner At a first glace: this looks awesome!
Nobody is on it, and I'll assign the issue to you shortly. I should get around to reviewing the PR a bit later today or tomorrow. Thanks so much for adding this!

src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
```diff
@@ -189,6 +230,50 @@ private static IEnumerable<TextEdit> OpenDirectiveSuggestions(this FileContentMa
...
+            (this FileContentManager file, CompilationUnit compilation, int lineNr, IEnumerable<Diagnostic> diagnostics)
+        {
+            var suggestions = NamespaceSuggestionsForUnknownIdentifiers(file, compilation, lineNr, diagnostics);
+            if (!suggestions.Any())
+            {
+                suggestions = CapitalizationSuggestionsForUnknownIdentifiers(file, compilation, diagnostics);
+            }
+
+            return suggestions;
+        }
```
samarsha, line 275:
```suggestion
                (this FileContentManager file, CompilationUnit compilation, int lineNr, IEnumerable<Diagnostic> diagnostics) =>
            NamespaceSuggestionsForUnknownIdentifiers(file, compilation, lineNr, diagnostics)
                .Concat(CapitalizationSuggestionsForUnknownIdentifiers(file, compilation, diagnostics));
```

src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
```diff
@@ -189,6 +230,50 @@ private static IEnumerable<TextEdit> OpenDirectiveSuggestions(this FileContentMa
...
+        /// The given line number is used to determine the containing namespace. 
+        /// Returns an empty enumerable if any of the given arguments is null. 
+        /// </summary>
+        internal static IEnumerable<(string, WorkspaceEdit)> SuggestionsForUnknownIdentifiers
+            (this FileContentManager file, CompilationUnit compilation, int lineNr, IEnumerable<Diagnostic> diagnostics)
+        {
+            var suggestions = NamespaceSuggestionsForUnknownIdentifiers(file, compilation, lineNr, diagnostics);
+            if (!suggestions.Any())
+            {
+                suggestions = CapitalizationSuggestionsForUnknownIdentifiers(file, compilation, diagnostics);
```
samarsha, line 271: Since `lineNr` is not used here, it looks like the capitalization suggestions are for the entire file? When I tested it, VS Code only showed suggestions for the current line, so it might be doing some filtering, but it would probably be better to only send the relevant suggestions to the client.

samarsha (dismissed):
Thank you for the PR! Sorry for the long delay in reviewing it.

It looks good overall. I have a few minor suggestions for changes. Also, just to clarify, it looks like it only provides suggestions for UDT names. I think we can extend this to UDT names and callables relatively easily (one of my comments is about this). For built-in types, other keywords, and local variable names, that can probably be a separate change, so we may want to keep #205 open or make a separate issue for the remaining kinds of suggestions.

Since it has been a while, no worries if you are no longer able to work on this PR. Let us know if you're still interested, or I can make the changes instead. :)

src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
```diff
@@ -189,6 +230,50 @@ private static IEnumerable<TextEdit> OpenDirectiveSuggestions(this FileContentMa
                 .Select(edit => (edit.NewText.Trim().Trim(';'), file.GetWorkspaceEdit(edit)));
         }
 
+        /// <summary>
+        /// Returns a sequence of replacement Type suggestions for how errors for unknown types and callable in the given diagnostics can be fixed,
+        /// given the file for which those diagnostics were generated and the corresponding compilation.
+        /// Returns an empty enumerable if any of the given arguments is null.
+        /// </summary>
+        internal static IEnumerable<(string, WorkspaceEdit)> CapitalizationSuggestionsForUnknownIdentifiers
```
samarsha, line 238: For me, it wasn't suggesting anything when the type was part of an opened namespace (rather than defined in the same file). So for example, this gave me no suggestions for `Complex`:

```
namespace Project {
    open Microsoft.Quantum.Math;

    @EntryPoint()
    operation HelloQ() : Unit {
        let x = complex(2.0, 5.0);
    }
}
```

I'm not sure what is causing that, though.

samarsha:
Hi @jwittner, since I haven't heard back from you, I'm planning to update this PR based on the feedback so that we can merge it soon. Thanks!

src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
samarsha: Nevermind, it looks like the list of diagnostics is provided by the code action context, so it shouldn't be for the whole file.

src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
```diff
@@ -76,6 +78,42 @@ private static WorkspaceEdit GetWorkspaceEdit(this FileContentManager file, para
                 : ImmutableArray<NonNullable<string>>.Empty;
         }
 
+        /// <summary>
+        /// Returns all types that match an alternative capitalization of this type.
+        /// </summary>
+        private static IEnumerable<string> TypesWithDifferentCapitalization(CompilationUnit compilation, string type) =>
+            compilation.GetTypes().Keys
+                .Select(t => t.Name.Value)
+                .Where(t => t != type && t.Equals(type, StringComparison.OrdinalIgnoreCase));
```
samarsha, line 87: I think there are a couple issues with this:

1. It uses `compilation.GetTypes()` which only includes types defined in the current compilation unit, not types in references.
2. It looks for any matching type names based on the second part of the qualified name, but it doesn't check if the namespace of that type is opened.

src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
samarsha: Probably caused by point 1 here: https://github.com/microsoft/qsharp-compiler/pull/266#discussion_r513816570

src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
samarsha: > It uses `compilation.GetTypes()` which only includes types defined in the current compilation unit, not types in references.

On second thought, this might be fine, since there is another code action that offers to open namespaces for you (but does a case-sensitive search). So in two steps you can go from an un-opened, mis-cased name to an open and correctly cased name.

src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
```diff
@@ -182,57 +219,84 @@ private static IEnumerable<TextEdit> OpenDirectiveSuggestions(this FileContentMa
...
+            return suggestionsForIds.Concat(suggestionsForTypes);
         }
 
+        /// <summary>
+        /// Returns a sequence of suggestions on how errors for unknown types and callable in the given diagnostics can be fixed,
+        /// given the file for which those diagnostics were generated and the corresponding compilation.
+        /// The given line number is used to determine the containing namespace.
+        /// Returns an empty enumerable if any of the given arguments is null.
+        /// </summary>
+        internal static IEnumerable<(string, WorkspaceEdit)> UnknownIdSuggestions(
```
bettinaheim, line 284: I know these methods are internal, but it would still be good to try and follow the same naming pattern for all methods: all other methods are named SuggestionsFor* opposed to *Suggestions. Could we stick with one and have all methods follow the same pattern? To me the former is more clear in some cases. It also looks like DocCommentSuggestions is the odd one out that would be good to adapt as well.

samarsha: What do you think about the private methods that follow the *Suggestions convention, including `OpenDirectiveSuggestions` which was here before?

bettinaheim: All in all I think it is a good effort to have consistent naming wherever possible. I am ok with both versions, but I think we should make them all the same.

bettinaheim (approved):

src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
samarsha: I agree that consistency is important - the reason why I started using *Suggestions for some of them is because I think I would prefer *Suggestions over SuggestionsFor* in all cases.

samarsha: I renamed all of them to have *Suggestions at the end.

PR #273, opened 2019-12-27
Sort documentation items when merging TOCs and namespaces.

cgranade:


src/QsCompiler/DocumentationParser/DocNamespace.cs
```diff
@@ -29,7 +31,7 @@ internal class DocNamespace
         /// </summary>
         /// <param name="ns">The namespace to be represented</param>
         /// <param name="sourceFiles">If specified, only the items in the specified source files are included.</param>
-        internal DocNamespace(QsNamespace ns, IEnumerable<string> sourceFiles = null)
+        internal DocNamespace(QsNamespace ns, IEnumerable<string>? sourceFiles = null)
```
bettinaheim: IEnumerables are nullable anyway - why the additional nullable?

cgranade: I added `#nullable enable` at the top since YamlDotNet now uses that new C# 8 feature. That makes references non-nullable by default (though violations are warnings rather than errors)

cgranade:
Thanks for looking over, @bettinaheim!

PR #275, opened 2020-01-01
Build tweaks.

anpaz-msft:
A couple of changes to the build scripts to align them with the scripts from the rest of the qsharp repositories.


anpaz-msft:
I'll target feature/qdk-pipeline, then mere from there.

PR #285, opened 2020-01-15
Adding a handle for diagnostics to the IRewriteStep interface

bettinaheim:
This PR adds an additional handle to the IRewriteStep interface that allows compiler plugins to generate diagnostics that will be displayed to the user.


src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -436,22 +436,44 @@ private Status ExecuteRewriteStep(RewriteSteps.LoadedStep rewriteStep, QsCompila
...
+                    this.LogAndUpdate(ref status, Warning(WarningCode.PreconditionVerificationFailed, new[] { rewriteStep.Name, messageSource }));
+                    return status;
+                }
 
-                var executeTransformation = preconditionPassed && rewriteStep.ImplementsTransformation;
-                var transformationPassed = !executeTransformation || rewriteStep.Transformation(compilation, out transformed);
-                if (!transformationPassed) this.LogAndUpdate(ref status, ErrorCode.RewriteStepExecutionFailed, new[] { rewriteStep.Name, messageSource });
+                var transformationFailed = rewriteStep.ImplementsTransformation && !rewriteStep.Transformation(compilation, out transformed);
+                var postconditionFailed = this.Config.EnableAdditionalChecks && rewriteStep.ImplementsPostconditionVerification && !rewriteStep.PostconditionVerification(transformed);
+                LogDiagnostics();
```
ScottCarda-MS: Do you want to log diagnostics regardless of failure here?

src/QsCompiler/Compiler/PluginInterface.cs
```diff
@@ -1,73 +1,127 @@
...
+        public bool ImplementsPostconditionVerification { get; }
+
         /// <summary>
         /// Verifies whether a given compilation satisfies the precondition for executing this rewrite step. 
         /// <see cref="ImplementsPreconditionVerification"/> indicates whether or not this method is implemented. 
+        /// If the precondition verfication succeeds, then the invocation of an implemented transformation (if any) 
+        /// with the given compilation should complete without throwing an exception. 
+        /// The precondition verification should never throw an exception, 
+        /// but instead indicate if the precondition is satisfied via the returned value. 
+        /// More detailed information can be provided via loggging.
```
ScottCarda-MS: spelling : logging

src/QsCompiler/Compiler/CompilationLoader.cs
bettinaheim: Yes, I believe we will want to do that.

src/QsCompiler/Compiler/PluginInterface.cs
bettinaheim:
```suggestion
        /// More detailed information can be provided via logging. 
```

bettinaheim: Thanks!

PR #286, opened 2020-01-16
Exposing optimizing transformations under Experimental namespace

bettinaheim:
This PR just does some renaming of namespaces and exposes some of the optimizing transformations under the namespace Microsoft.Quantum.QsCompiler.Exerimental.


bettinaheim:
Given that this is a book-keeping exercise and about what interface we expose publicly, I'll merge this so we can build a release candidate any time soon-ish.

bettinaheim:
Thanks, @msoeken, for taking a quick look!

PR #287, opened 2020-01-18
Classically Controlled Unit Tests

ScottCarda-MS:
Unit tests for the feature


src/QsCompiler/Tests.Compiler/ClassicalControlTests.fs
```diff
@@ -132,13 +133,13 @@ type ClassicalControlTests () =
             <| Seq.take i lst
             <| Seq.skip (i+1) lst
 
-        for content in contents do
-            calls
+        for content in calls do
```
marshallsa:
```suggestion
        for call in calls do
```

src/QsCompiler/Tests.Compiler/ClassicalControlTests.fs
```diff
@@ -110,20 +109,22 @@ type ClassicalControlTests () =
...
-        let mutable calls = generatedCalls |> Seq.map (fun x -> x, x |> (GetBodyFromCallable >> GetLinesFromSpecialization))
+    let IdentifyGeneratedByCalls generatedCallables calls =
+        let mutable callables = generatedCallables |> Seq.map (fun x -> x, x |> (GetBodyFromCallable >> GetLinesFromSpecialization))
         let hasContent call (content : seq<int * string * string>) =
             let (_, lines : string[]) = call
-            Seq.forall (fun (i, ns, name) -> CheckIfLineIsCall ns name lines.[i] |> (fun (x,_,_) -> x)) content
+            Seq.forall (fun (i, ns, name) -> CheckIfLineIsCall ns name lines.[i] |> (fun (x, _, _) -> x)) content
 
-        Assert.True(Seq.length calls = Seq.length contents) // This should be true if this method is call correctly
+        Assert.True(Seq.length callables = Seq.length calls) // This should be true if this method is call correctly
```
marshallsa:
```suggestion
        Assert.True(Seq.length callables = Seq.length calls) // This should be true if this method is called correctly
```

PR #288, opened 2020-01-21
Fix Unit Test - Apply If One Else Zero

ScottCarda-MS:
This PR relies on PR #287.

Fixes an issue brought to light by the Classical Control unit test <i>#</i>9, where the conversion from an if/else statement to an ApplyIfElse statement assumed the condition compared against a result literal Zero.


ScottCarda-MS:
This PR is being abandoned in favor of PR #289.

PR #290, opened 2020-01-23
Updates from master

bettinaheim:
This will also make the transformation for classical control execute if the execution target is a quantum processor. Please don't squash this PR - a proper merge commit will make it a lot easier to set up the PR back to master.


src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -373,13 +373,15 @@ public CompilationLoader(SourceLoader loadSources, ReferenceLoader loadReference
             if (this.Config.ConvertClassicalControl)
             {
                 var rewriteStep = new RewriteSteps.LoadedStep(new ClassicallyControlled(), typeof(IRewriteStep), thisDllUri);
-                this.CompilationStatus.ConvertClassicalControl = this.ExecuteRewriteStep(rewriteStep, this.CompilationOutput, out this.CompilationOutput);
+                this.CompilationStatus.ConvertClassicalControl = this.ExecuteRewriteStep(rewriteStep, this.CompilationOutput, out var transformed);
+                if (this.CompilationStatus.ConvertClassicalControl == Status.Succeeded) this.CompilationOutput = transformed;
```
ScottCarda-MS: Doesn't this.ExecuteRewriteStep already perform this logic? If there is something wrong with that logic, it should be fixed in the ExecuteRewriteStep method. We should not have to do this here.

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -373,13 +373,15 @@ public CompilationLoader(SourceLoader loadSources, ReferenceLoader loadReference
...
             }
 
             if (!this.Config.SkipMonomorphization && this.CompilationOutput?.EntryPoints.Length != 0)
             {
                 var rewriteStep = new RewriteSteps.LoadedStep(new Monomorphization(), typeof(IRewriteStep), thisDllUri);
-                this.CompilationStatus.Monomorphization = this.ExecuteRewriteStep(rewriteStep, this.CompilationOutput, out this.CompilationOutput);
+                this.CompilationStatus.Monomorphization = this.ExecuteRewriteStep(rewriteStep, this.CompilationOutput, out var transformed);
+                if (this.CompilationStatus.Monomorphization == Status.Succeeded) this.CompilationOutput = transformed;
```
ScottCarda-MS: Same as above

src/QsCompiler/Compiler/CompilationLoader.cs
bettinaheim: If you look through the code then you will see that ExecuteRewriteStep does not perform this logic, and I would consider that the constructor is ultimately the piece that is responsible for the flow throughout the compilation stages (when to continue the compilation, when to abort, which steps to invoke for which configurations etc.). The placement of the logic here hence seems appropriate, though with now more than one piece implementing this logic a local helper function would indeed be a good idea. I added that.

bettinaheim:
@ScottCarda-MS I pulled in the latest changes, and made the necessary adaptions. I also reenabled execution tests. After merging this everything is ready for a PR to master. Do you want to set that up?

Issue #259, opened 2019-12-04
Access modifiers for callables and types


cgranade:
**Is your feature request related to a problem? Please describe.**
When developing Q# libraries, it is common to introduce small functions and operations that are not intended for direct use, but which are helpful with partial application. At the moment, these functions and operations are given names starting with `_` or ending with `Impl`, but that makes it easy to accidentally take a dependency on these functions and operations from other namespaces or compilation units.

**Describe the solution you'd like**
I would suggest the inclusion of an accessibility modifier attribute (or possibly multiple, depending on the level of granularity needed), paralleling the use of special-purpose keywords in other languages. In particular,
```Q#
namespace Microsoft.Quantum.Core {
 
    @Attribute()
    newtype Private = Unit;

}
```

In code that calls a `@Private()` function or operation from another namespace or compilation unit, a warning could be raised indicating that about a possible accidental dependency on internal implementation details.

The new accessibility modifier attributes could also be used to simplify documentation generation logic, such that no documentation is generated for `@Private()` callables rather than depending on name-matching (paralleling the new `@Test("…")` attribute).

**Describe alternatives you've considered**
- The current approach could continue to be used, in which the leading underscore convention is given special meaning by the compiler and documentation generation tool. This suffers the same disadvantages as relying on `*Test` names for unit testing.
- New keywords (e.g.: `private`) could be introduced instead of new attributes.

bettinaheim:
Thanks, @cgranade for filing this feature request! Tagging this for language impact.
As discussed, I'd love to have something like this, but it is not yet clear when we will be able to resource it. 
In the meantime, it would be great to accumulate thoughts and feedback - and upvotes - from the community regarding the need(s) for this feature. 

My thought so far were to make the distinction between private, internal and public, with private meaning restricting visibility to file + namespace, internal meaning the restriction to namespace + compilation unit, and public meaning the current state without any restrictions.

bettinaheim:
Here is what we have converged to from the language design side. 
Please let me know if you feel this won't meet your needs and if you have any questions or feedback.  
	

- By default, everything will be visible everywhere, and there is no keyword planned for this case. 
- We introduce a keyword "internal" that may directly precede (i.e. after any attached attributes) any operation, function or type declaration. Types and callables decorated with this keyword will be visible anywhere within the compilation unit in which they are declared. 
[...]

PR #292, opened 2020-01-24
Add access modifiers to compiler data structures

marshallsa:
Adds fields to the records for callables and custom types to store their access modifiers (and potentially other kinds of modifiers in the future). There's no logic for parsing access modifiers from source, or enforcing access modifiers when resolving symbols yet. See #259.


src/QsCompiler/Tests.Compiler/ExecutionTests.fs
```diff
@@ -40,7 +40,7 @@ type ExecutionTests (output:ITestOutputHelper) =
         AssertEqual expectedOutput out
 
 
-    [<Fact>]
+    // TODO: [<Fact>]
```
ScottCarda-MS: This should be [<Fact(Skip="Reason")>]. This will make the skipped test more visible and less likely to be missed in the future.

src/QsCompiler/Tests.Compiler/ExecutionTests.fs
```diff
@@ -139,7 +139,7 @@ type ExecutionTests (output:ITestOutputHelper) =
             "
 
 
-    [<Fact>]
+    // TODO: [<Fact>]
```
ScottCarda-MS: This should be [<Fact(Skip="Reason")>]. This will make the skipped test more visible and less likely to be missed in the future.

src/QsCompiler/Tests.Compiler/ExecutionTests.fs
marshallsa: I'm not 100% clear on if the disabled tests should be part of the final PR or if they can be re-enabled now that there's a build of this branch available. @bettinaheim

bettinaheim: @marshallsa The runtime packages you can use to reenable the tests are 0.10.2001.2501-alpha.

PR #293, opened 2020-01-25
Disabling incremental build support and design time builds

bettinaheim:
The incremental build doesn't seem to behave as it should, causing some occasional issues when debugging (the debugger complaining fails changed, because the design time build re-generated C# files). 
Also: adding a AdditionalQscArguments property, and switching to using the Sdk for the Example project.


bettinaheim:
Merging this since the changes have been discussed, and I don't like to wait too long with building the release candidate.

PR #296, opened 2020-01-30
Transformation Restructure

ScottCarda-MS:
Removed hierarchical structure in transformations by putting all the handlers into BaseTransformation.

Added a class (that should be renamed) to better allow transformation to share code.

Includes an example transformation in the new format: MonomorphizationValidation.


src/QsCompiler/Core/BaseTransformation.fs
```diff
@@ -0,0 +1,683 @@
...
+    member val private Handlers : MyHandler<'T> list = [] with get, set
+    member val private CombinedHandleValid = true with get, set
+    member val private CombinedHandle = defaultHandler with get, set
+
+    member private this.UpdateCombinedHandle () =
+        this.CombinedHandle <- match this.Handlers with
+                               | [f] -> f.Call
+                               | f1 :: fn -> List.fold (>>) f1.Call <| List.map (fun (f : MyHandler<'T>) -> f.Call) fn
+                               | [] -> defaultHandler
+        this.CombinedHandleValid <- true
```
marshallsa: Is the added complexity of caching a combined handler worth it? It seems that you can avoid traversing the list each time, but you still have to call all of the functions in sequence in both cases. Is there a performance difference?

ScottCarda-MS:
We went with a different setup.

PR #297, opened 2020-01-31
Add a performance measurement option to the command line tool

cesarzc:
This change adds a performance measurement option (-p --perf) to the command line tool.

The overview of the change is that the compiler raises events when a task or sub-task starts or ends. Those events are tracked by a sub-component of the command line tool (CompilationTracker.cs) when the perf option is present. After the compilation process is done, the command line tool invokes the compilation tracker to publish the results of the tracking.

In a bit more detail, the compilation tracker builds a hierarchy of tasks and sub-tasks, and then writes that to a json file that contains the start time, end time and duration of each task and sub-task.

Here's an example of what the json looks like:
`[
  {
    "Task": {
      "ParentName": null,
      "Name": "OverallCompilation",
      "UtcStart": "2020-01-31T00:41:20.8402258Z",
      "UtcEnd": "2020-01-31T00:41:32.0618253Z",
      "DurationInMs": 11221
    },
    "Children": {
      "SourcesLoading": {
        "Task": {
          "ParentName": "OverallCompilation",
          "Name": "SourcesLoading",
          "UtcStart": "2020-01-31T00:41:20.8481749Z",
          "UtcEnd": "2020-01-31T00:41:21.4133199Z",
          "DurationInMs": 565
        },
        "Children": {}
      },
      "ReferenceLoading": {
        "Task": {
          "ParentName": "OverallCompilation",
          "Name": "ReferenceLoading",
          "UtcStart": "2020-01-31T00:41:21.4133437Z",
          "UtcEnd": "2020-01-31T00:41:23.1862459Z",
          "DurationInMs": 1772
        },
        "Children": {}
      },
      "Build": {
        "Task": {
          "ParentName": "OverallCompilation",
          "Name": "Build",
          "UtcStart": "2020-01-31T00:41:23.186258Z",
          "UtcEnd": "2020-01-31T00:41:30.1440662Z",
          "DurationInMs": 6957
        },
        "Children": {}
      },
      "OutputGeneration": {
        "Task": {
          "ParentName": "OverallCompilation",
          "Name": "OutputGeneration",
          "UtcStart": "2020-01-31T00:41:30.1440802Z",
          "UtcEnd": "2020-01-31T00:41:32.0618242Z",
          "DurationInMs": 1917
        },
        "Children": {
          "SyntaxTreeSerialization": {
            "Task": {
              "ParentName": "OutputGeneration",
              "Name": "SyntaxTreeSerialization",
              "UtcStart": "2020-01-31T00:41:30.1440903Z",
              "UtcEnd": "2020-01-31T00:41:32.0235554Z",
              "DurationInMs": 1879
            },
            "Children": {}
          },
          "BinaryGeneration": {
            "Task": {
              "ParentName": "OutputGeneration",
              "Name": "BinaryGeneration",
              "UtcStart": "2020-01-31T00:41:32.0235699Z",
              "UtcEnd": "2020-01-31T00:41:32.0618029Z",
              "DurationInMs": 38
            },
            "Children": {}
          }
        }
      }
    }
  }
]`

### DIFF
--- a/src/QsCompiler/CommandLineTool/Commands/Build.cs
+++ b/src/QsCompiler/CommandLineTool/Commands/Build.cs
@@ -141,7 +141,10 @@ namespace Microsoft.Quantum.QsCompiler.CommandLineCompiler
                 EnableAdditionalChecks = false // todo: enable debug mode?
             }; 
 
+            // ToDo: check options before using the OnCompilationEvent handler.
             var loaded = new CompilationLoader(options.LoadSourcesOrSnippet(logger), options.References, loadOptions, logger, CompilationTracker.OnCompilationEvent);
+            // ToDo: check options before publishing results.
+            CompilationTracker.PublishResults();
             return ReturnCode.Status(loaded);
         }
     }
--- a/src/QsCompiler/CommandLineTool/Commands/Build.cs
+++ b/src/QsCompiler/CommandLineTool/Commands/Build.cs
@@ -135,7 +135,7 @@ namespace Microsoft.Quantum.QsCompiler.CommandLineCompiler
                 DocumentationOutputFolder = options.DocFolder,
                 BuildOutputFolder = options.OutputFolder ?? (usesPlugins ? "." : null),
                 DllOutputPath = options.EmitDll ? " " : null, // set to e.g. an empty space to generate the dll in the same location as the .bson file
-                RewriteSteps = options.Plugins?.Select(step => (step, (string)null)) ?? ImmutableArray<(string, string)>.Empty, 
+                RewriteSteps = options.Plugins?.Select(step => (step, (string)null)) ?? ImmutableArray<(string, string)>.Empty,
                 EnableAdditionalChecks = false // todo: enable debug mode?
             }; 
 
--- a/src/QsCompiler/CommandLineTool/Commands/Build.cs
+++ b/src/QsCompiler/CommandLineTool/Commands/Build.cs
@@ -108,8 +108,6 @@ namespace Microsoft.Quantum.QsCompiler.CommandLineCompiler
             );
         }
 
-        
-
 
         // publicly accessible routines 
 
--- a/src/QsCompiler/CommandLineTool/Commands/Build.cs
+++ b/src/QsCompiler/CommandLineTool/Commands/Build.cs
@@ -108,6 +108,8 @@ namespace Microsoft.Quantum.QsCompiler.CommandLineCompiler
             );
         }
 
+        
+
 
         // publicly accessible routines 
 
@@ -139,7 +141,7 @@ namespace Microsoft.Quantum.QsCompiler.CommandLineCompiler
                 EnableAdditionalChecks = false // todo: enable debug mode?
             }; 
 
-            var loaded = new CompilationLoader(options.LoadSourcesOrSnippet(logger), options.References, loadOptions, logger);
+            var loaded = new CompilationLoader(options.LoadSourcesOrSnippet(logger), options.References, loadOptions, logger, CompilationTracker.OnCompilationEvent);
             return ReturnCode.Status(loaded);
         }
     }
--- a/src/QsCompiler/CommandLineTool/Commands/Build.cs
+++ b/src/QsCompiler/CommandLineTool/Commands/Build.cs
@@ -139,12 +139,16 @@ namespace Microsoft.Quantum.QsCompiler.CommandLineCompiler
                 DllOutputPath = options.EmitDll ? " " : null, // set to e.g. an empty space to generate the dll in the same location as the .bson file
                 RewriteSteps = options.Plugins?.Select(step => (step, (string)null)) ?? ImmutableArray<(string, string)>.Empty,
                 EnableAdditionalChecks = false // todo: enable debug mode?
-            }; 
+            };
 
-            // ToDo: check options before using the OnCompilationEvent handler.
+            // Note that performance is evaluated only if the performance folder option is present.
+            CompilationLoader.CompilationProcessEventHandler onCompilationProcessEvent = options.PerfFolder != null ? CompilationTracker.OnCompilationEvent : (CompilationLoader.CompilationProcessEventHandler)null;
             var loaded = new CompilationLoader(options.LoadSourcesOrSnippet(logger), options.References, loadOptions, logger, CompilationTracker.OnCompilationEvent);
-            // ToDo: check options before publishing results.
-            CompilationTracker.PublishResults();
+            if (options.PerfFolder != null)
+            {
+                CompilationTracker.PublishResults(options.PerfFolder);
+            }
+
             return ReturnCode.Status(loaded);
         }
     }
--- a/src/QsCompiler/CommandLineTool/Commands/Build.cs
+++ b/src/QsCompiler/CommandLineTool/Commands/Build.cs
@@ -139,11 +139,23 @@ namespace Microsoft.Quantum.QsCompiler.CommandLineCompiler
                 EnableAdditionalChecks = false // todo: enable debug mode?
             }; 
 
-            var onCompilationProcessEvent = options.PerfFolder != null ? CompilationTracker.OnCompilationTaskEvent : (CompilationLoader.CompilationTaskEventHandler)null;
-            var loaded = new CompilationLoader(options.LoadSourcesOrSnippet(logger), options.References, loadOptions, logger, CompilationTracker.OnCompilationTaskEvent);
             if (options.PerfFolder != null)
             {
-                CompilationTracker.PublishResults(options.PerfFolder);
+                CompilationLoader.CompilationTaskEvent += CompilationTracker.OnCompilationTaskEvent;
+            }
+
+            var loaded = new CompilationLoader(options.LoadSourcesOrSnippet(logger), options.References, loadOptions, logger);
+            if (options.PerfFolder != null)
+            {
+                try
+                {
+                    CompilationTracker.PublishResults(options.PerfFolder);
+                }
+                catch (Exception ex)
+                {
+                    logger.Log(ErrorCode.PublishingPerfResultsFailed, new string[]{options.PerfFolder});
+                    logger.Log(ex);
+                }
             }
 
             return ReturnCode.Status(loaded);
--- a/src/QsCompiler/CommandLineTool/Commands/Build.cs
+++ b/src/QsCompiler/CommandLineTool/Commands/Build.cs
@@ -141,9 +141,8 @@ namespace Microsoft.Quantum.QsCompiler.CommandLineCompiler
                 EnableAdditionalChecks = false // todo: enable debug mode?
             };
 
-            // Note that performance is evaluated only if the performance folder option is present.
-            CompilationLoader.CompilationProcessEventHandler onCompilationProcessEvent = options.PerfFolder != null ? CompilationTracker.OnCompilationEvent : (CompilationLoader.CompilationProcessEventHandler)null;
-            var loaded = new CompilationLoader(options.LoadSourcesOrSnippet(logger), options.References, loadOptions, logger, CompilationTracker.OnCompilationEvent);
+            CompilationLoader.CompilationTaskEventHandler onCompilationProcessEvent = options.PerfFolder != null ? CompilationTracker.OnCompilationTaskEvent : (CompilationLoader.CompilationTaskEventHandler)null;
+            var loaded = new CompilationLoader(options.LoadSourcesOrSnippet(logger), options.References, loadOptions, logger, CompilationTracker.OnCompilationTaskEvent);
             if (options.PerfFolder != null)
             {
                 CompilationTracker.PublishResults(options.PerfFolder);
--- a/src/QsCompiler/CommandLineTool/Commands/Build.cs
+++ b/src/QsCompiler/CommandLineTool/Commands/Build.cs
@@ -61,6 +61,10 @@ namespace Microsoft.Quantum.QsCompiler.CommandLineCompiler
             [Option("emit-dll", Required = false, Default = false, SetName = CODE_MODE,
             HelpText = "Specifies whether the compiler should emit a .NET Core dll containing the compiled Q# code.")]
             public bool EmitDll { get; set; }
+
+            [Option('p', "perf", Required = false, SetName = CODE_MODE,
+            HelpText = "Destination folder where the output of the performance assessment will be generated.")]
+            public string PerfFolder { get; set; }
         }
 
         /// <summary>
--- a/src/QsCompiler/CommandLineTool/Commands/Build.cs
+++ b/src/QsCompiler/CommandLineTool/Commands/Build.cs
@@ -137,7 +137,7 @@ namespace Microsoft.Quantum.QsCompiler.CommandLineCompiler
                 DllOutputPath = options.EmitDll ? " " : null, // set to e.g. an empty space to generate the dll in the same location as the .bson file
                 RewriteSteps = options.Plugins?.Select(step => (step, (string)null)) ?? ImmutableArray<(string, string)>.Empty,
                 EnableAdditionalChecks = false // todo: enable debug mode?
-            };
+            }; 
 
             CompilationLoader.CompilationTaskEventHandler onCompilationProcessEvent = options.PerfFolder != null ? CompilationTracker.OnCompilationTaskEvent : (CompilationLoader.CompilationTaskEventHandler)null;
             var loaded = new CompilationLoader(options.LoadSourcesOrSnippet(logger), options.References, loadOptions, logger, CompilationTracker.OnCompilationTaskEvent);
--- a/src/QsCompiler/CommandLineTool/Commands/Build.cs
+++ b/src/QsCompiler/CommandLineTool/Commands/Build.cs
@@ -135,7 +135,7 @@ namespace Microsoft.Quantum.QsCompiler.CommandLineCompiler
                 DocumentationOutputFolder = options.DocFolder,
                 BuildOutputFolder = options.OutputFolder ?? (usesPlugins ? "." : null),
                 DllOutputPath = options.EmitDll ? " " : null, // set to e.g. an empty space to generate the dll in the same location as the .bson file
-                RewriteSteps = options.Plugins?.Select(step => (step, (string)null)) ?? ImmutableArray<(string, string)>.Empty,
+                RewriteSteps = options.Plugins?.Select(step => (step, (string)null)) ?? ImmutableArray<(string, string)>.Empty, 
                 EnableAdditionalChecks = false // todo: enable debug mode?
             }; 
 
--- a/src/QsCompiler/CommandLineTool/Commands/Build.cs
+++ b/src/QsCompiler/CommandLineTool/Commands/Build.cs
@@ -139,7 +139,7 @@ namespace Microsoft.Quantum.QsCompiler.CommandLineCompiler
                 EnableAdditionalChecks = false // todo: enable debug mode?
             }; 
 
-            CompilationLoader.CompilationTaskEventHandler onCompilationProcessEvent = options.PerfFolder != null ? CompilationTracker.OnCompilationTaskEvent : (CompilationLoader.CompilationTaskEventHandler)null;
+            var onCompilationProcessEvent = options.PerfFolder != null ? CompilationTracker.OnCompilationTaskEvent : (CompilationLoader.CompilationTaskEventHandler)null;
             var loaded = new CompilationLoader(options.LoadSourcesOrSnippet(logger), options.References, loadOptions, logger, CompilationTracker.OnCompilationTaskEvent);
             if (options.PerfFolder != null)
             {
--- a/src/QsCompiler/CommandLineTool/CompilationTracker.cs
+++ b/src/QsCompiler/CommandLineTool/CompilationTracker.cs
@@ -27,19 +27,19 @@ namespace Microsoft.Quantum.QsCompiler.CommandLineCompiler
 
             public CompilationProcess(string parentName, string name)
             {
-                this.ParentName = parentName;
-                this.Name = name;
-                this.UtcStart = DateTime.UtcNow;
-                this.UtcEnd = null;
-                this.DurationInMs = null;
-                this.Watch = Stopwatch.StartNew();
+                ParentName = parentName;
+                Name = name;
+                UtcStart = DateTime.UtcNow;
+                UtcEnd = null;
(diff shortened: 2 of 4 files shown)


src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -0,0 +1,281 @@
...
+        /// Creates a hierarchical structure that contains the compilation tasks.
+        /// </summary>
+        private static IList<CompilationTaskNode> BuildCompilationTasksHierarchy()
+        {
+            IList<CompilationTaskNode> compilationTasksForest = new List<CompilationTaskNode>();
+            Queue<CompilationTaskNode> toFindChildrenNodes = new Queue<CompilationTaskNode>();
+
+            // First add the roots (top-level tasks) of all trees to the forest.
+
+            foreach (KeyValuePair<string, CompilationTask> entry in CompilationTasks)
```
ScottCarda-MS: The 'var' keyword is typically used for the iteration variable of a foreach loop.
```suggestion
            foreach (var entry in CompilationTasks)
```

src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -0,0 +1,281 @@
...
+                    compilationTasksForest.Add(node);
+                    toFindChildrenNodes.Enqueue(node);
+                }
+            }
+
+            // Iterate through the tasks until all of them have been added to the hierarchy.
+
+            while (toFindChildrenNodes.Count > 0)
+            {
+                CompilationTaskNode parentNode = toFindChildrenNodes.Dequeue();
```
ScottCarda-MS:
```suggestion
                var parentNode = toFindChildrenNodes.Dequeue();
```

src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -0,0 +1,281 @@
...
+            IList<CompilationTaskNode> compilationTasksForest = new List<CompilationTaskNode>();
+            Queue<CompilationTaskNode> toFindChildrenNodes = new Queue<CompilationTaskNode>();
+
+            // First add the roots (top-level tasks) of all trees to the forest.
+
+            foreach (KeyValuePair<string, CompilationTask> entry in CompilationTasks)
+            {
+                if (entry.Value.ParentName == null)
+                {
+                    CompilationTaskNode node = new CompilationTaskNode(entry.Value);
```
ScottCarda-MS:
```suggestion
                    var node = new CompilationTaskNode(entry.Value);
```

src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -0,0 +1,281 @@
...
+        private static readonly IList<Warning> Warnings = new List<Warning>();
+
+        // Private methods.
+
+        /// <summary>
+        /// Creates a hierarchical structure that contains the compilation tasks.
+        /// </summary>
+        private static IList<CompilationTaskNode> BuildCompilationTasksHierarchy()
+        {
+            IList<CompilationTaskNode> compilationTasksForest = new List<CompilationTaskNode>();
```
ScottCarda-MS: Unless necessary, the standard of this code base is to use the 'var' keyword for local declarations.
```suggestion
            var compilationTasksForest = new List<CompilationTaskNode>();
```

src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -0,0 +1,281 @@
...
+            Queue<CompilationTaskNode> toFindChildrenNodes = new Queue<CompilationTaskNode>();
```
ScottCarda-MS:
```suggestion
            var toFindChildrenNodes = new Queue<CompilationTaskNode>();
```

src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -0,0 +1,281 @@
...
+                    toFindChildrenNodes.Enqueue(node);
+                }
+            }
+
+            // Iterate through the tasks until all of them have been added to the hierarchy.
+
+            while (toFindChildrenNodes.Count > 0)
+            {
+                CompilationTaskNode parentNode = toFindChildrenNodes.Dequeue();
+                foreach (KeyValuePair<string, CompilationTask> entry in CompilationTasks)
```
ScottCarda-MS:
```suggestion
                foreach (var entry in CompilationTasks)
```

src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -0,0 +1,281 @@
...
+                {
+                    if (parentNode.Task.Name.Equals(entry.Value.ParentName))
+                    {
+                        CompilationTaskNode childNode = new CompilationTaskNode(entry.Value);
```
ScottCarda-MS:
```suggestion
                        var childNode = new CompilationTaskNode(entry.Value);
```

src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -0,0 +1,281 @@
...
+
+            return compilationTasksForest;
+        }
+
+        /// <summary>
+        /// Handles a compilation task start event.
+        /// </summary>
+        private static void CompilationEventStartHandler(CompilationLoader.CompilationTaskEventArgs eventArgs)
+        {
+            string key = CompilationTask.GenerateKey(eventArgs.ParentTaskName, eventArgs.TaskName);
```
ScottCarda-MS:
```suggestion
            var key = CompilationTask.GenerateKey(eventArgs.ParentTaskName, eventArgs.TaskName);
```

src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -0,0 +1,281 @@
...
+
+            CompilationTasks.Add(key, new CompilationTask(eventArgs.ParentTaskName, eventArgs.TaskName));
+        }
+
+        /// <summary>
+        /// Handles a compilation task end event.
+        /// </summary>
+        private static void CompilationEventEndHandler(CompilationLoader.CompilationTaskEventArgs eventArgs)
+        {
+            string key = CompilationTask.GenerateKey(eventArgs.ParentTaskName, eventArgs.TaskName);
```
ScottCarda-MS:
```suggestion
            var key = CompilationTask.GenerateKey(eventArgs.ParentTaskName, eventArgs.TaskName);
```

src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -0,0 +1,281 @@
...
+            if (!CompilationTasks.TryGetValue(key, out CompilationTask process))
```
ScottCarda-MS: Consider using the 'var' keyword for 'out' arguments.
```suggestion
            if (!CompilationTasks.TryGetValue(key, out var process))
```

src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -0,0 +1,281 @@
...
+        {
+            CompilationEventTypeHandlers[args.Type](args);
+        }
+
+        /// <summary>
+        /// Publishes the results to text files in the specified folder.
+        /// </summary>
+        public static void PublishResults(string outputFolder)
+        {
+            IList<CompilationTaskNode> compilationProcessesForest = BuildCompilationTasksHierarchy();
```
ScottCarda-MS:
```suggestion
            var compilationProcessesForest = BuildCompilationTasksHierarchy();
```

src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -0,0 +1,281 @@
...
+            string outputPath = Path.GetFullPath(outputFolder);
```
ScottCarda-MS:
```suggestion
            var outputPath = Path.GetFullPath(outputFolder);
```

src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -0,0 +1,281 @@
...
+            DirectoryInfo outputDirectoryInfo = new DirectoryInfo(outputPath);
```
ScottCarda-MS:
```suggestion
            var outputDirectoryInfo = new DirectoryInfo(outputPath);
```

src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -0,0 +1,281 @@
...
+            if (!outputDirectoryInfo.Exists)
+            {
+                outputDirectoryInfo.Create();
+            }
+
+            using (StreamWriter file = File.CreateText(Path.Combine(outputPath, "compilationPerf.json")))
```
ScottCarda-MS: Consider using the 'var' keyword for 'using' block variables.
```suggestion
            using (var file = File.CreateText(Path.Combine(outputPath, "compilationPerf.json")))
```

src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -0,0 +1,281 @@
...
+            {
+                JsonSerializer serializer = new JsonSerializer
```
ScottCarda-MS:
```suggestion
                var serializer = new JsonSerializer
```

src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -0,0 +1,281 @@
...
+                {
+                    Formatting = Formatting.Indented
+                };
+
+                serializer.Serialize(file, compilationProcessesForest);
+            }
+
+            if (Warnings.Count > 0) {
+                using (StreamWriter file = File.CreateText(Path.Combine(outputPath, "compilationPerfWarnings.json")))
```
ScottCarda-MS:
```suggestion
                using (var file = File.CreateText(Path.Combine(outputPath, "compilationPerfWarnings.json")))
```

src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -0,0 +1,281 @@
...
+                {
+                    JsonSerializer serializer = new JsonSerializer
```
ScottCarda-MS:
```suggestion
                    var serializer = new JsonSerializer
```

cesarzc:
> There are some places where the use of the 'var' keyword is more standard in this code base than explicit type declarations. Please consider the suggestions changes I made.

Thanks for the feedback Scott. I made the changes that you suggested.

src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -0,0 +1,281 @@
...
+                {
+                    var node = new CompilationTaskNode(entry.Value);
+                    compilationTasksForest.Add(node);
+                    toFindChildrenNodes.Enqueue(node);
+                }
+            }
+
+            // Iterate through the tasks until all of them have been added to the hierarchy.
+
+            while (toFindChildrenNodes.Count > 0)
```
bettinaheim: FYI, there is also TryDequeue, though it seems like Count is also O(1) plus it really doesn't seem relevant which one you use in practice in this case.

src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -0,0 +1,281 @@
...
+        }
+
+        // Public methods.
+
+        /// <summary>
+        /// Handles a compilation task event.
+        /// </summary>
+        public static void OnCompilationTaskEvent(object sender, CompilationLoader.CompilationTaskEventArgs args)
+        {
+            CompilationEventTypeHandlers[args.Type](args);
```
bettinaheim: I would go with something like
```suggestion
            if (CompilationEventTypeHandlers.TryGetValue(args.Type, out var handler)) handler(args);
            else Warnings.Add(new Warning(WarningType.UnknownCompilationTaskEventType, args.Type.ToString())); 
```

src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -0,0 +1,281 @@
...
+            /// Contains the duration of the task in milliseconds.
+            /// </summary>
+            public long? DurationInMs;
+            /// <summary>
+            /// Stopwatch used to measure the duration of the task.
+            /// </summary>
+            private readonly Stopwatch Watch;
+
+            /// <summary>
+            /// Generates a key that uniquely identifies a task in the compilation process based on the task's name and its parent's name.
```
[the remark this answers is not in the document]
cesarzc: > If the goal is for this to be unique, that would mean that e.g. for parsing the task name would also contain the name of the source file to parse?

Yes. In that scenario I see the parsing of all source files as one task, and each individual file parsing as a sub-task of it.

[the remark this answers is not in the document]
cesarzc: > Also, what is the reason for wanting this to be public? I realize the class is private, but suppose at a later time it is made public, I would think that this particular method should probably remain limited to this compilation unit (e.g. because I think it's signature might change over time). Come to think of it, it seems like this method might make sense to be part of the CompilationTaskNode class instead, since that one seems to be the one that would handle the logic on how to define the tree structure for compilation tasks.

In this case it needs to be public since this method is used by the event handlers to get the key to both look for a task in the dictionary.

Even though the CompilationTaskNode is the one used to build the tree structure for compilation tasks, the key is a convenient way to store compilation tasks in a flat structure until we need to build the tree. Having said that, I think it makes sense to keep the GenerateKey method as part of the CompilationTask class.

src/QsCompiler/CommandLineTool/Commands/Build.cs
```diff
@@ -135,7 +139,13 @@ public static int Run(BuildOptions options, ConsoleLogger logger)
                 EnableAdditionalChecks = false // todo: enable debug mode?
             }; 
 
-            var loaded = new CompilationLoader(options.LoadSourcesOrSnippet(logger), options.References, loadOptions, logger);
+            CompilationLoader.CompilationTaskEventHandler onCompilationProcessEvent = options.PerfFolder != null ? CompilationTracker.OnCompilationTaskEvent : (CompilationLoader.CompilationTaskEventHandler)null;
+            var loaded = new CompilationLoader(options.LoadSourcesOrSnippet(logger), options.References, loadOptions, logger, CompilationTracker.OnCompilationTaskEvent);
```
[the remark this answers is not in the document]
cesarzc: Correct, thank you for catching that!

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -182,6 +182,37 @@ private class ExecutionStatus
                 this.LoadedRewriteSteps.All(status => WasSuccessful(true, status));
         }
 
+        /// <summary>
```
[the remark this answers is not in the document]
cesarzc: Moved it, the change will be part of a commit that addresses all comments.

src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -0,0 +1,281 @@
...
+        /// <summary>
+        /// Handles a compilation task event.
+        /// </summary>
+        public static void OnCompilationTaskEvent(object sender, CompilationLoader.CompilationTaskEventArgs args)
+        {
+            CompilationEventTypeHandlers[args.Type](args);
+        }
+
+        /// <summary>
+        /// Publishes the results to text files in the specified folder.
```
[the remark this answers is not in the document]
cesarzc: Documented the exceptions that might be thrown. The change will be part of a commit that addresses all comments.

src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -0,0 +1,281 @@
...
+        /// </summary>
+        public static void PublishResults(string outputFolder)
+        {
+            var compilationProcessesForest = BuildCompilationTasksHierarchy();
+            var outputPath = Path.GetFullPath(outputFolder);
+            var outputDirectoryInfo = new DirectoryInfo(outputPath);
```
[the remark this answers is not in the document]
cesarzc: Thanks for the suggestion, this makes the code more compact :) 

The change will be part of a commit that addresses all comments.

src/QsCompiler/CommandLineTool/Commands/Build.cs
```diff
@@ -135,7 +139,13 @@ public static int Run(BuildOptions options, ConsoleLogger logger)
...
+            if (options.PerfFolder != null)
+            {
+                CompilationTracker.PublishResults(options.PerfFolder);
```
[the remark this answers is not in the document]
cesarzc: Added the try catch.

The change will be part of a commit that addresses all comments.

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -324,8 +359,17 @@ public Status LoadedRewriteStep(string name, string source = null)
...
         /// Throws an ArgumentNullException if either one of the given loaders is null or returns null.
         /// </summary>
-        public CompilationLoader(SourceLoader loadSources, ReferenceLoader loadReferences, Configuration? options = null, ILogger logger = null)
+        public CompilationLoader(SourceLoader loadSources, ReferenceLoader loadReferences, Configuration? options = null, ILogger logger = null, CompilationTaskEventHandler onCompilationTaskEvent = null)
         {
+
+            // adding the event handlers.
+            if (onCompilationTaskEvent != null)
+            {
+                CompilationTaskEvent += onCompilationTaskEvent;
```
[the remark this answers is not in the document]
cesarzc: I think it is a good idea. The code will look cleaner since the event handler wouldn't have to be passed to the constructor. I will make the changes to make this member static.

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -393,28 +442,45 @@ public CompilationLoader(SourceLoader loadSources, ReferenceLoader loadReference
...
                 var docsGenerated = this.VerifiedCompilation != null && DocBuilder.Run(docsFolder, this.VerifiedCompilation.SyntaxTree.Values, this.VerifiedCompilation.SourceFiles, onException: onDocException);
                 if (!docsGenerated) this.LogAndUpdate(ref this.CompilationStatus.Documentation, ErrorCode.DocGenerationFailed, Enumerable.Empty<string>());
+                RaiseCompilationTaskEnd("OutputGeneration", "DocumentationGeneration");
             }
 
+            RaiseCompilationTaskEnd("OverallCompilation", "OutputGeneration");
+
             // invoking rewrite steps in external dlls
 
             for (int i = 0; i < this.ExternalRewriteSteps.Length; i++)
```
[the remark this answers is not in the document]
cesarzc: Added rewrite steps events.

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -827,5 +895,21 @@ public static string GeneratedFile(NonNullable<string> fileId, string outputFold
             File.WriteAllText(targetFile, content);
             return targetFile;
         }
+
+        /// <summary>
+        /// Raises a compilation task start event.
+        /// </summary>
+        private void RaiseCompilationTaskStart (string parentTaskName, string taskName)
```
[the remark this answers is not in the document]
cesarzc: I like that syntax too, but I am not used to it yet. I will change it.

src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -0,0 +1,281 @@
...
+        private static readonly IDictionary<CompilationLoader.CompilationTaskEventType, CompilationTaskEventTypeHandler> CompilationEventTypeHandlers = new Dictionary<CompilationLoader.CompilationTaskEventType, CompilationTaskEventTypeHandler>
+        {
+            { CompilationLoader.CompilationTaskEventType.Start, CompilationEventStartHandler },
+            { CompilationLoader.CompilationTaskEventType.End, CompilationEventEndHandler }
+        };
+
+        /// <summary>
+        /// Contains the compilation tasks tracked through the handled events.
+        /// </summary>
+        private static readonly IDictionary<string, CompilationTask> CompilationTasks = new Dictionary<string, CompilationTask>();
```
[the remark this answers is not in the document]
cesarzc: I added a global lock to make CompilationTracker thread safe. I think it makes sense to have a global lock instead of concurrent data structures since we need a lock to protect the operation of building the tree anyway.

src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -0,0 +1,281 @@
...
+        /// <summary>
+        /// Contains the warnings generated while handling the compiler tasks events.
+        /// </summary>
+        private static readonly IList<Warning> Warnings = new List<Warning>();
```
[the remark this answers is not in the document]
cesarzc: I added a global lock to make access to this member thread-safe.

src/QsCompiler/CommandLineTool/CompilationTracker.cs
cesarzc: C# Queue class does not seem to have a TryDequeue method, only the ConcurrentQueue which there is no need for in this case. I will keep using Count in this case :)

src/QsCompiler/CommandLineTool/CompilationTracker.cs
cesarzc: Thanks for the suggestion, this is much better than the original one :)

src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -0,0 +1,281 @@
...
+
+        // Private members.
+
+        /// <summary>
+        /// Defines a handler for a type of compilation task event.
+        /// </summary>
+        private delegate void CompilationTaskEventTypeHandler(CompilationLoader.CompilationTaskEventArgs eventArgs);
+
+        /// <summary>
+        /// Contains a handler that takes care of each type of task event.
```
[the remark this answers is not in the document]
cesarzc: I added the assumption to the doc comment and changed the code per your suggestion below.

src/QsCompiler/CommandLineTool/CompilationTracker.cs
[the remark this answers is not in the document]
cesarzc: I changed it to internal per your suggestion.

PR #300, opened 2020-02-01
Updating Sdk readme and adding an example for defining custom compiler extensions

bettinaheim:
Mathias ran into an issue where a rewrite step defined in a project reference failed with a FileNotFoundException despite that that dll was present in the project output directory. Updated to section on how to troubleshoot extensions to suggest how that case may be fixed. 

I also decided that it would be helpful to have an example for how to define custom compilation steps, and cleaned up and added the demo. 
Also, updating the used packages to the latest release.

### DIFF
--- a/src/QuantumSdk/README.md
+++ b/src/QuantumSdk/README.md
@@ -93,6 +93,9 @@ The NuGet version of the Sdk package.
 
 The following properties can be configured to customize the build: 
 
+- `AdditionalQscArguments`:    
+May contain additional arguments to pass to the Q# command line compiler. Valid additional arguments are `--emit-dll`, or `--no-warn` followed by any number of integers specifying the warnings to ignore.   
+
 - `CsharpGeneration`:    
 Specifies whether to generate C# code as part of the compilation process. Setting this property to false may prevent certain interoperability features or integration with other pieces of the Quantum Development Kit. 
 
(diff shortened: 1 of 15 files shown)


src/QuantumSdk/README.md
```diff
@@ -79,7 +79,14 @@ When a rewrite steps fails to execute, setting the `QscVerbosity` to "Detailed"
 ```
   <QscVerbosity>Detailed</QscVerbosity>
 ```
-By default, the compiler will search the project output directory for a suitable assembly in case a dependency cannot be found. In case of a `FileNotFoundException`, and option may be to add the corresponding reference to the project, or to copy the missing dll to the output directory. 
+A `FileNotFoundException` will be raised if a compiler extension attempts to load a reference that either could not be found, or could not be loaded for other reasons. 
+By default, the compiler will search the project output directory for a suitable assembly in case a dependency cannot be found.
+If such an exception occurs during a compilation step loaded from a package reference, adding the package containing the dll that could not be found to the project or to copying the missing dll to the output directory may resolve the issue.
```
marshallsa:
```suggestion
If such an exception occurs during a compilation step loaded from a package reference, adding the package containing the dll that could not be found to the project or copying the missing dll to the output directory may resolve the issue. 
```

ScottCarda-MS: After the comma, this line is confusing. What are you supposed to do if the exception occurs? Consider starting after the comma with 'the issue may be resolved by...' rather than putting that piece at the end.

ScottCarda-MS: Consider replacing 'dll that could not be found' with 'missing dll'.

bettinaheim: Yes, that makes it better. Thanks!

bettinaheim:
@ScottCarda-MS, @marshallsa Thanks for taking a quick look!
After writing the blog post I decided that it would be good to have an example for defining custom compilation steps after all, so I cleaned up and added the demo project. My apologies for adding a junk of code after you two reviewed!

PR #302, opened 2020-02-03
Updating name dependencies for classically controlled operations

bettinaheim:
I believe there is one typo, and the namespace where classically controlled operations live has changed since this work started.


bettinaheim:
> The code that defines the ApplyIf... operations lists the name for the built in operation as 'ApplyIfElseCA', without the 'R'. The namespace change can be addressed in the other PR.
> This PR could probably be closed.

You are right - that looks like a typo on the runtime repo; let's fix it there, and then I'll close this PR.

PR #312, opened 2020-02-12
Transformation setup clean-up - Part I

bettinaheim:
I added a revised setup for the transformations provided in the Core project. The idea is to build all transformations on top of this setup, which in particular eliminates the C# wrapper in the transformation project. 
A syntax tree transformation is structured into the same subtransformations as now, but the shared internal state is now handled much cleaner, eliminating the need for the rather cumbersome type parametrizations on construction. Right now the actual transformation on individual nodes is still the same, with the idea to revise that slightly in a subsequent step. 
What is also still missing is a neat way to disable certain subtransformations.
So far I've migrated the transformations in SearchAndReplace, and I'll see that I can migrate some more.

### DIFF
--- a/src/QsCompiler/Core/TransformationDefinition.fs
+++ b/src/QsCompiler/Core/TransformationDefinition.fs
@@ -2,18 +2,6 @@
 // Licensed under the MIT License.
 
 namespace Microsoft.Quantum.QsCompiler.Transformations.Core
-
-open System.Collections.Immutable
-open System.Numerics
-open System.Runtime.CompilerServices
-open Microsoft.Quantum.QsCompiler.DataTypes
-open Microsoft.Quantum.QsCompiler.SyntaxExtensions
-open Microsoft.Quantum.QsCompiler.SyntaxTokens
-open Microsoft.Quantum.QsCompiler.SyntaxTree
-open System
-
-type private ExpressionKind = QsExpressionKind<TypedExpression,Identifier,ResolvedType>
-type private ExpressionType = QsTypeKind<ResolvedType, UserDefinedType, QsTypeParameter, CallableInformation>
     
 
 type QsSyntaxTreeTransformation<'T> (state : 'T) as this =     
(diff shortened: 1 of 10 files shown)


src/QsCompiler/Core/TransformationDefinition.fs
```diff
@@ -0,0 +1,188 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.QsCompiler.Transformations.Core
+    
+
+type QsSyntaxTreeTransformation<'T> private (state : 'T, dummy) =
```
ScottCarda-MS: Is dummy needed here? It seems like it is unused.

src/QsCompiler/Tests.Compiler/TestCases/ExecutionTests/LoggingBasedTests.qs
```diff
@@ -33,10 +33,12 @@ namespace Microsoft.Quantum.Testing.ExecutionTests {
             ULog("V1");
 
             within {
+                let dummy = 0;
```
ScottCarda-MS: I'm not sure what this dummy variable is used to test. Can you please explain?

src/QsCompiler/Transformations/BasicTransformations.cs
```diff
@@ -7,15 +7,33 @@
...
+        public class TransformationState
+        {
+            internal readonly HashSet<NonNullable<string>> SourceFiles = 
+                new HashSet<NonNullable<string>>();
+        }
+
+
+        private GetSourceFiles() :
+            base(new TransformationState())
+        { }
```
ScottCarda-MS: nit: It might look cleaner to have this as a single line. Is there a style convention that we have that would mandate it look like this?

src/QsCompiler/Transformations/BasicTransformations.cs
```diff
@@ -7,15 +7,33 @@
...
+
+        public override NamespaceTransformation<TransformationState> NewNamespaceTransformation() =>
+            new NamespaceTransformation(this);
```
ScottCarda-MS: It isn't clear to me what this accomplishes. Doesn't the base NewNamespaceTransformation do the same thing?

cgranade:
> Please remove trailing whitespace in the files added/modified. A regular expression search/replace can make it quick and painless.

I'm not sure about VS2019, but there's a "Trim Trailing Whitespace" command in the VS Code palette that's pretty handy for that.

src/QsCompiler/Tests.Compiler/TestCases/ExecutionTests/LoggingBasedTests.qs
bettinaheim: It's a good idea to make sure the transformation that handles eliminating the apply-within statements deals appropriately with variable names. It is valid to defined a variable with the same name in both blocks, and correspondingly variables need to be renamed upon inclining the blocks.

src/QsCompiler/Transformations/BasicTransformations.cs
bettinaheim: I meant to go over the project and either put all colon on the old line or on the new line. Right now I think it's a wild mixture. The line can get pretty long if the call to the base constructor is on the same line, hence why I generally put it on a new line, but either one works for me.

bettinaheim:
> I agree, this is much cleaner. The one comment I have is about naming -- having QsSyntaxTreeTransformation, but just NamespaceTransformation, StatementTransformation, etc., is a bit hard to remember. I assume this is because you have the existing SyntaxTreeTransformation that is really a NamespaceTransformation; while it will be a breaking change, I suggest renaming the existing SyntaxTreeTransformation class as part of this PR.
> It also might be nice to have QsSyntaxTreeTransformation have a Transform method that takes a SyntaxTree as input and returns a new SyntaxTree as output...

Yes, I agree on all of that, I just haven't gotten around to cleaning it all up - hence the PR against a feature branch opposed to into master. Right now this also mostly strictly adds code, but the ultimate goal is to replace the old setup, and remove a good junk of code that was needed to deal with internal state in the old setup. This refactoring should thus result in an overall negative code contribution (reducing the amount of source code). Once we've migrated all the transformation I can go ahead and remove the code that is no longer needed, clean up the subtransformations (there is still no separation between the recursion/walker logic and the actual action/transformation on the node itself). Going over everything and renaming as appropriate seems like a good think to do as a final step before merging to master.

src/QsCompiler/Core/TransformationDefinition.fs
bettinaheim: I haven't found a more elegant way of dealing with this yet - hopefully I'll have a good inspiration at some point. Basically F# is not happy to have two constructors with the same signature (for obvious reasons), but I need a) a base constructor that allows me to build a QsSyntaxTreeTransformation in the first place such that I can get a handle to it to construct the subtransformations, and b) an additional constructor that actually initializes the subtransformations with the ones that have the correct handle back to the parent. 
So yeah, not exactly the prettiest way to achieve this, but also the only one that came to mind.

bettinaheim:
> Please remove trailing whitespace in the files added/modified. A regular expression search/replace can make it quick and painless.
> 
> I'm not sure about VS2019, but there's a "Trim Trailing Whitespace" command in the VS Code palette that's pretty handy for that.

Thanks, @cgranade!

PR #313, opened 2020-02-13
Enforce access modifiers in the symbol table

marshallsa:
This PR adds checks in the symbol table so that inaccessible types/callables can't be resolved. There are also new error messages for when you try to use an inaccessible type/callable.

See #259.

### DIFF
--- a/src/QsCompiler/Core/SymbolTable.fs
+++ b/src/QsCompiler/Core/SymbolTable.fs
@@ -756,6 +756,14 @@ and NamespaceManager
                 | Value (declSource, deprecation) -> buildAndReturn (ns.Name, declSource, deprecation, [||]) 
                 | Null -> None, [| symRange |> orDefault |> QsCompilerDiagnostic.Error (ErrorCode.UnknownTypeInNamespace, [symName.Value; qualifier.Value]) |]
 
+    /// Returns whether a declaration is visible from the calling location, given whether the calling location is in the
+    /// same assembly and/or namespace as the declaration, and the declaration's modifiers.
+    let IsDeclarationVisible isSameAssembly isSameNs modifiers =
+        match modifiers.Access with
+        | DefaultAccess -> true
+        | Internal -> isSameAssembly
+        | Private -> isSameAssembly && isSameNs
+
     /// Given the name of the namespace as well as the source file in which the attribute occurs, resolves the given attribute.
     /// Generates suitable diagnostics if a suitable attribute cannot be determined, 
     /// or if the attribute argument contains expressions that are not supported, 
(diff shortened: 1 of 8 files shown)


src/QsCompiler/Core/SymbolTable.fs
```diff
@@ -18,6 +18,19 @@ open Microsoft.Quantum.QsCompiler.SyntaxTree
 open Newtonsoft.Json
 
 
+/// Represents the outcome of resolving a symbol in the symbol table.
+type ResolutionResult<'T> =
+    /// The symbol resolved successfully.
+    | Found of 'T
+    /// The symbol is ambiguous, and more than one resolution is possible. Contains the list of possible namespaces in
+    /// which the symbol could be resolved to.
```
bettinaheim: I know what you mean, but the sentence is a bit strange to me.

src/QsCompiler/Core/SymbolTable.fs
```diff
@@ -414,43 +427,57 @@ and Namespace private
...
-        match this.ContainsCallable cName with
+        match this.TryFindCallable cName with
         | Value _ -> (Parts.Values.SelectMany (fun partial -> 
             partial.GetSpecializations cName |> Seq.map (fun (kind, decl) -> kind, (partial.Source, decl)))).ToImmutableArray()        
         | Null -> ArgumentException "no callable with the given name exist within the namespace" |> raise
 
-    /// If this namespace contains a declaration for the given type name, 
-    /// returns a Value with the name of the source file or the name of the file within a referenced assembly
-    /// in which it is declared as well as a string option indicating the redirection for the type if it has been deprecated. 
+    /// If this namespace contains a declaration for the given type name, returns a Value with:
```
bettinaheim: I like the doc comment!

src/QsCompiler/Core/SymbolTable.fs
```diff
@@ -1342,16 +1427,16 @@ and NamespaceManager
     member this.NamespacesContainingCallable cName = 
         // FIXME: we need to handle the case where a callable/type with the same qualified name is declared in several references!
         syncRoot.EnterReadLock()
-        try let containsCallable (ns : Namespace) = ns.ContainsCallable cName |> QsNullable<_>.Map (fun _ -> ns.Name)
-            (Namespaces.Values |> QsNullable<_>.Choose containsCallable).ToImmutableArray()
+        try let tryFindCallable (ns : Namespace) = ns.TryFindCallable cName |> QsNullable<_>.Map (fun _ -> ns.Name)
+            (Namespaces.Values |> QsNullable<_>.Choose tryFindCallable).ToImmutableArray()
         finally syncRoot.ExitReadLock()
 
     /// Returns the names of all namespaces in which a type with the given name is declared.
```
bettinaheim: It would be good to mention whether or not private/internal things are included.
Edit: As discussed, the behavior of these should probably be changed (I believe they are only used for a code action, but you need to double check that). If you want to do that in a separate PR, then please add a comment on NamespacesContainingType and NamespacesContainingCallable.

src/QsCompiler/Core/SymbolTable.fs
```diff
@@ -1342,16 +1427,16 @@ and NamespaceManager
     member this.NamespacesContainingCallable cName =
```
bettinaheim: It would be good to mention in the doc comment whether private/internal things are included.

PR #318, opened 2020-02-14
Prevent use of less accessible types in more accessible declarations

marshallsa:
This prevents private or internal types from being exposed in places they shouldn't be; for example, a public operation returning an internal type, or an internal UDT using a private UDT as part of its underlying type.

See #259.

### DIFF
--- a/src/QsCompiler/Core/SymbolTable.fs
+++ b/src/QsCompiler/Core/SymbolTable.fs
@@ -770,40 +770,92 @@ and NamespaceManager
     let TryResolveTypeName (parentNS, source) ((nsName, symName), symRange) = 
         let orDefault (range : QsNullable<_>) = range.ValueOr QsCompilerDiagnostic.DefaultRange
         let checkQualificationForDeprecation qual = BuiltIn.Deprecated |> PossibleQualifications (parentNS, source) |> Seq.contains qual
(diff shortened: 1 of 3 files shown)


src/QsCompiler/Core/SymbolTable.fs
```diff
@@ -909,33 +961,37 @@ and NamespaceManager
...
-        let processUDT = TryResolveTypeName (parent.Namespace, source) >> function 
-            | Some (udt, _), errs -> UserDefinedType udt, errs
-            | None, errs -> InvalidType, errs 
-        let processTP (symName, symRange) = 
-            if tpNames |> Seq.contains symName then TypeParameter {Origin = parent; TypeName = symName; Range = symRange}, [||]
-            else InvalidType, [| symRange.ValueOr QsCompilerDiagnostic.DefaultRange |> QsCompilerDiagnostic.Error (ErrorCode.UnknownTypeParameterName, [symName.Value]) |]
-        syncRoot.EnterReadLock()
-        try SymbolResolution.ResolveType (processUDT, processTP) qsType
-        finally syncRoot.ExitReadLock()
+    /// Throws a NonSupportedException if the QsType to resolve contains a MissingType.
```
bettinaheim:
```suggestion
    /// Throws a NotSupportedException if the QsType to resolve contains a MissingType.
```

PR #319, opened 2020-02-16
Transformation API clean-up - part I

bettinaheim:
This PR cleans up the naming and the handles of the 6 subtransformations. 
I migrated all F# transformations, and in particular all transformations in the Optimization project. 
Also, I realized that F# seems to fail when accessing overridden handles as part of a constructor (after the initial construction), even though the same works fine when used from within C#. I hence revised that setup.

What I haven't don't yet is making a distinction between the action on a node and the recursion, or and attempt to unify the transformations and walkers (and correspondingly I haven't touched the walkers so far).


src/QsCompiler/Optimizations/OptimizingTransformations/CallableInlining.fs
```diff
@@ -10,7 +10,7 @@ open Microsoft.Quantum.QsCompiler.Experimental.Utils
 open Microsoft.Quantum.QsCompiler.SyntaxExtensions
 open Microsoft.Quantum.QsCompiler.SyntaxTokens
 open Microsoft.Quantum.QsCompiler.SyntaxTree
-open Microsoft.Quantum.QsCompiler.Transformations.Core
+//open Microsoft.Quantum.QsCompiler.Transformations.Core
```
ScottCarda-MS: Did you forget to remove this commented-out open directive?

bettinaheim:
```suggestion
```

PR #321, opened 2020-02-20
Unifying transformation and walker setup

bettinaheim:


src/QsCompiler/Core/TransformationOptions.fs
```diff
@@ -0,0 +1,51 @@
...
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.QsCompiler.Transformations.Core
+
+
+/// Used to configure the behavior of the default implementations for transformations.
+type TransformationOptions = {
+    /// Disables the transformation at the transformation root, 
+    /// meaning the transformation won't recur into leaf nodes or subnodes. 
+    Disable : bool
```
ScottCarda-MS: I am not sure why this option is needed. Can you explain in what scenarios it would be useful for?

bettinaheim: Each subtransformation can be configured upon construction. You would set these to disabled e.g. if you don't want to do expression transformations. This is replacing the bool that the old setup took as constructor argument. Does this clarify things?

src/QsCompiler/Core/TransformationOptions.fs
```diff
@@ -0,0 +1,51 @@
...
+    static member NoRebuild = {
+        Disable = false
+        DisableRebuild = true
+    }
+
+
+/// Tools for adapting the default implementations for transformations 
+/// based on the specified options. 
+module internal Utils = 
+    type internal INode =
```
alan-geller: Could you get the same effect with Fold and Walk just being functions, and having Node in the various transformation classes being a function variable that you assign Fold or Walk to, just as you do now? You'd then call Node instead of Node.BuildOr -- which probably means it should be called BuildOr if you do it that way.

It doesn't make much of a difference, since you only ever have one instance of Fold and Walk anyway, so there's no memory or GC issue. It just seems a little more "F#-y".

src/QsCompiler/Core/StatementTransformation.fs
```diff
@@ -25,63 +28,67 @@ type StatementKindTransformationBase internal (unsafe) =
...
+            | QubitRegisterAllocation ex as orig -> QubitRegisterAllocation |> Node.BuildOr orig (this.Expressions.Transform ex)
+            | QubitTupleAllocation is as orig    -> QubitTupleAllocation |> Node.BuildOr orig (is |> Seq.map this.onQubitInitializer |> ImmutableArray.CreateRange)
+            | InvalidInitializer                 -> InvalidInitializer
+        ResolvedInitializer.New |> Node.BuildOr init transformed
 
-    abstract member onExpressionStatement : TypedExpression -> QsStatementKind
-    default this.onExpressionStatement ex = this.Expressions.Transform ex |> QsExpressionStatement
+    abstract member onPositionedBlock : QsNullable<TypedExpression> * QsPositionedBlock -> QsNullable<TypedExpression> * QsPositionedBlock
+    default this.onPositionedBlock (intro : QsNullable<TypedExpression>, block : QsPositionedBlock) = 
+        let location = this.Statements.onLocation block.Location
```
alan-geller: Shouldn't this check `options.Disable` before recursing? And similarly in other methods in this class.

src/QsCompiler/Core/NamespaceTransformation.fs
```diff
@@ -183,7 +206,7 @@ type NamespaceTransformationBase internal (unsafe) =
         let specializations = c.Specializations |> Seq.map this.dispatchSpecialization
```
alan-geller: It would be convenient if it were possible to guarantee the order of specializations, at least that the body is always seen first. I realize this might be awkward, and it can be worked around, but being able to look at the body before the other specializations is sometimes very convenient in transformations.

src/QsCompiler/Core/SyntaxTreeTransformation.fs
```diff
@@ -4,14 +4,16 @@
 namespace Microsoft.Quantum.QsCompiler.Transformations.Core
 
 
+// setup for syntax tree transformations with internal state
+
 type QsSyntaxTreeTransformation<'T> private (state : 'T, unsafe : string) =
```
alan-geller: Can this be renamed `SyntaxTreeTransformation` since none of the other classes have the Qs prefix?

src/QsCompiler/Core/SyntaxTreeTransformation.fs
```diff
@@ -40,36 +42,44 @@ type QsSyntaxTreeTransformation<'T> private (state : 'T, unsafe : string) =
...
+    new (state : 'T) = new QsSyntaxTreeTransformation<'T>(state, TransformationOptions.Default)
 
 
-and TypeTransformation<'T> internal (unsafe) =
-    inherit TypeTransformationBase()
+and TypeTransformation<'T> internal (options, unsafe) =
+    inherit TypeTransformationBase(options)
     let mutable _Transformation : QsSyntaxTreeTransformation<'T> option = None
 
     member this.Transformation
```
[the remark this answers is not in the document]
marshallsa: I wonder if it would also be possible to have the property be called `State`, so that transformations can access their state using `this.State.Foo`? Since the state seems like it will be referenced a lot by transformations (if it's a stateful transformation), I think it would be good for it to be short, unless there's some reason to be explicit about internal state vs. other state.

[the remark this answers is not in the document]
bettinaheim: Adding the shortcut makes sense. I might have gone with something like SharedState then (which is a bit longer) to distinguish it form any state within the subtransformation only. Thoughts?

src/QsCompiler/Core/SyntaxTreeTransformation.fs
```diff
@@ -40,36 +42,44 @@ type QsSyntaxTreeTransformation<'T> private (state : 'T, unsafe : string) =
...
+    new (parentTransformation : QsSyntaxTreeTransformation<'T>) = 
+        new TypeTransformation<'T>(parentTransformation, TransformationOptions.Default)
             
+    new (sharedInternalState : 'T) = 
+        new TypeTransformation<'T>(sharedInternalState, TransformationOptions.Default)
 
-and ExpressionKindTransformation<'T> internal (unsafe) =
-    inherit ExpressionKindTransformationBase(unsafe)
+
+and ExpressionKindTransformation<'T> internal (options, unsafe) =
```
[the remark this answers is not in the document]
bettinaheim: sure.

src/QsCompiler/Core/TransformationOptions.fs
bettinaheim: No, unfortunately not. F# is a bit particular with what can be generic. A member can be generic, or a let in a module, but not a let in a class. The interface safes me the evaluation of an additional if on each invokation of BuildOr.

src/QsCompiler/Core/SyntaxTreeTransformation.fs
bettinaheim: Yes, I forgot to rename it. Thanks!

src/QsCompiler/Core/SyntaxTreeTransformation.fs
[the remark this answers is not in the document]
marshallsa: I think `SharedState` makes sense.

marshallsa:
I think this was already the case with the old transformations, but is there a reason why all of the `onFoo` methods start with a lowercase "o" instead of an uppercase "O"?

src/QsCompiler/Core/NamespaceTransformation.fs
bettinaheim: done

src/QsCompiler/Tests.Compiler/ExecutionTests.fs
```diff
@@ -40,7 +40,7 @@ type ExecutionTests (output:ITestOutputHelper) =
         AssertEqual expectedOutput out
 
 
-    [<Fact>]
+    //[<Fact>]
```
ScottCarda-MS: For visibility, please mark the test as 'Skipped' and provide a short reason instead of commenting it out.

src/QsCompiler/Core/SyntaxTreeTransformation.fs
```diff
@@ -170,58 +178,62 @@ and StatementTransformation<'T> internal (options, unsafe) =
...
-    new (sharedInternalState : 'T) = 
-        new NamespaceTransformation<'T>(sharedInternalState, TransformationOptions.Default)
+    new (sharedState : 'T) = 
+        new NamespaceTransformation<'T>(sharedState, TransformationOptions.Default)
 
 
 // setup for syntax tree transformations without internal state
 
-type QsSyntaxTreeTransformation private (unsafe : string) =
+type SyntaxTreeTransformation private (_internal_ : string) =
```
ScottCarda-MS: It looks like the non-generic classes did not get the shortcut for SharedState.

ScottCarda-MS: In other words, they are missing:
`member this.SharedState = this.Transformation.SharedState`

src/QsCompiler/Tests.Compiler/ExecutionTests.fs
bettinaheim: This is temporary and will be reverted shortly.

src/QsCompiler/Core/SyntaxTreeTransformation.fs
bettinaheim: The non-generic versions do not have a shared state.

src/QsCompiler/Core/SyntaxTreeTransformation.fs
```diff
@@ -40,36 +42,44 @@ type QsSyntaxTreeTransformation<'T> private (state : 'T, unsafe : string) =
...
 
 
-and TypeTransformation<'T> internal (unsafe) =
-    inherit TypeTransformationBase()
+and TypeTransformation<'T> internal (options, unsafe) =
+    inherit TypeTransformationBase(options)
     let mutable _Transformation : QsSyntaxTreeTransformation<'T> option = None
 
     member this.Transformation
         with get () = _Transformation.Value
```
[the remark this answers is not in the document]
bettinaheim: I added doc comments.

PR #322, opened 2020-02-20
Classical Control Migrate

ScottCarda-MS:
Migrated the Classical Control transformations.


src/QsCompiler/Transformations/ClassicallyControlledTransformation.cs
```diff
@@ -4,615 +4,472 @@
...
+                                Utilities.CreateValueTupleExpression(condId, condArgs));
+
+                            targetArgs = ImmutableArray.Create(condArgs.ResolvedType);
                         }
                         else
                         {
-                            controlOpInfo = (result == QsResult.Zero)
-                            ? BuiltIn.ApplyIfZero
-                            : BuiltIn.ApplyIfOne;
+                            return null; // ToDo: Diagnostic message - default body exists, but is not valid
```
bettinaheim: What does it mean for the body not to be valid? I don't understand what you mean with this comment.

src/QsCompiler/Transformations/ClassicallyControlledTransformation.cs
```diff
@@ -4,615 +4,472 @@
...
-                        controlArgs = CreateValueTupleExpression(
-                            conditionExpression,
-                            CreateValueTupleExpression(condId, condArgs));
-
-                        targetArgs = ImmutableArray.Create(condArgs.ResolvedType);
                     }
                     else
                     {
-                        return null; // ToDo: Diagnostic message - default body exists, but is not valid
+                        return null; // ToDo: Diagnostic message - cond body not valid
```
bettinaheim: What does that mean? Also, I assume this is talking about the conditional block?

src/QsCompiler/Transformations/ClassicallyControlledTransformation.cs
bettinaheim: Do you mean the default block?

src/QsCompiler/Transformations/ClassicallyControlledTransformation.cs
```diff
@@ -4,615 +4,472 @@
...
+                        if (statement.Statement is QsStatementKind.QsConditionalStatement)
+                        {
+                            var stm = ReshapeConditional(statement);
+                            stm = this.onStatement(stm);
+
+                            var (isCondition, cond, conditionScope, defaultScope) = IsConditionWtihSingleBlock(stm);
+
+                            if (isCondition)
+                            {
+                                /*ToDo: this could be a separate function.*/
```
bettinaheim: Ok, so why isn't it a separate function? I.e. why is it left as a todo?

ScottCarda-MS: This will be addressed in the Classical Control Rewrite feature.

src/QsCompiler/Transformations/ClassicallyControlledTransformation.cs
ScottCarda-MS: Right, the body of the default block is not valid, but it exists. In this case we would not be able to do the conversion on either because we went with an all-or-nothing approach to converting an if-structure. If both the conditional body and the default body can't be converted, then we don't convert either of them. Here we need to raise a diagnostics message informing the user why this if-structure couldn't be converted.

src/QsCompiler/Transformations/ClassicallyControlledTransformation.cs
ScottCarda-MS: This is referring to the body of the conditional block. Or in other words, the body that is conditional.

PR #326, opened 2020-02-22
Hide private and internal declarations when generating documentation

marshallsa:
Part of issue #259.

### DIFF
--- a/src/QsCompiler/DocumentationParser/DocNamespace.cs
+++ b/src/QsCompiler/DocumentationParser/DocNamespace.cs
@@ -185,10 +185,11 @@ namespace Microsoft.Quantum.QsCompiler.Documentation
         }
 
         /// <summary>
-        /// Writes the YAML file for this namespace.
+        /// Writes the YAML file for this namespace to a stream.
         /// </summary>
-        /// <param name="directoryPath">The directory to write the file to</param>
-        internal void WriteToFile(string directoryPath)
+        /// <param name="stream">The stream to write to.</param>
+        /// <param name="rootNode">The preexisting YAML mapping node.</param>
+        internal void WriteToStream(Stream stream, YamlMappingNode? rootNode = null)
         {
             string ToSequenceKey(string itemTypeName)
             {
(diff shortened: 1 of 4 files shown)


src/QsCompiler/DocumentationParser/DocNamespace.cs
```diff
@@ -185,10 +185,11 @@ void WriteItem(DocItem i)
         }
 
         /// <summary>
-        /// Writes the YAML file for this namespace.
+        /// Writes the YAML file for this namespace to a stream.
         /// </summary>
-        /// <param name="directoryPath">The directory to write the file to</param>
-        internal void WriteToFile(string directoryPath)
+        /// <param name="stream">The stream to write to.</param>
+        /// <param name="rootNode">The preexisting YAML mapping node.</param>
```
cgranade: It might be good to clarify that the value of `rootNode` represents the preexisting contents of a namespace YAML file.

src/QsCompiler/DocumentationParser/DocNamespace.cs
```diff
@@ -211,8 +212,7 @@ string ToSequenceKey(string itemTypeName)
                 return "";
             }
 
-            var rootNode = Utils.ReadYamlFile(directoryPath, name) as YamlMappingNode ?? new YamlMappingNode();
-
+            rootNode ??= new YamlMappingNode();
```
cgranade: Ah, I hadn't realized that `??` could be used as an assignment operator, neat!

src/QsCompiler/DocumentationParser/DocNamespace.cs
```diff
@@ -280,14 +280,22 @@ string ToSequenceKey(string itemTypeName)
...
+            yamlStream.Save(output, false);
+        }
+
+        /// <summary>
+        /// Writes the YAML file for this namespace.
+        /// </summary>
+        /// <param name="directoryPath">The directory to write the file to</param>
+        internal void WriteToFile(string directoryPath)
+        {
+            var rootNode = Utils.ReadYamlFile(directoryPath, name) as YamlMappingNode;
```
cgranade: This does read much more clearly now, thanks for separating it out!

PR #330, opened 2020-02-23
Apply Conditionally Logic for Classical Control Transformation

ScottCarda-MS:
Added logic to handle case where two non-literal results are being compared, converting them to use the ApplyConditionally interface operations.


src/QsCompiler/Transformations/Utilities.cs
```diff
@@ -55,6 +55,30 @@ public static class Utilities
...
             );
 
+        // All of the resolved types of the given expression must match.
+        public static TypedExpression CreateValueArray(params TypedExpression[] expressions)
+        {
+            ResolvedType type = null;
+            if (expressions.Any())
+            {
+                type = expressions.First().ResolvedType;
+                QsCompilerError.Verify(expressions.All(expr => expr.ResolvedType.Equals(type)));
```
bettinaheim: This will fail in general.

src/QsCompiler/Transformations/Utilities.cs
```diff
@@ -55,6 +55,30 @@ public static class Utilities
                 QsNullable<Tuple<QsPositionInfo, QsPositionInfo>>.Null
             );
 
+        // All of the resolved types of the given expression must match.
+        public static TypedExpression CreateValueArray(params TypedExpression[] expressions)
```
bettinaheim: All of this methods are really not general, and should not be public.

src/QsCompiler/Transformations/Utilities.cs
```diff
@@ -55,6 +55,30 @@ public static class Utilities
...
+            {
+                type = ResolvedType.New(ResolvedTypeKind.UnitType);
+            }
+
+            return new TypedExpression
+            (
+                ExpressionKind.NewValueArray(expressions.ToImmutableArray()),
+                TypeArgsResolution.Empty,
+                ResolvedType.New(ResolvedTypeKind.NewArrayType(type)),
+                new InferredExpressionInformation(false, false),
```
bettinaheim: E.g. this here is not accurate in general.

src/QsCompiler/Transformations/Utilities.cs
```diff
@@ -84,6 +108,27 @@ public static TypeArgsResolution GetCombinedType(TypeArgsResolution outer, TypeA
             }).Concat(outerDict.Select(x => Tuple.Create(x.Key.Item1, x.Key.Item2, x.Value))).ToImmutableArray();
         }
 
+        public static (TypedExpression, TypedExpression) GetNoOp()
```
bettinaheim: There is no reason for this to be public.

src/QsCompiler/Transformations/ClassicallyControlledTransformation.cs
```diff
@@ -226,41 +315,15 @@ private TypedExpression CreateApplyIfExpression(QsResult result, TypedExpression
                         return null; // ToDo: Diagnostic message - cond body not valid
```
bettinaheim:
```suggestion
                        return null; // ToDo: Diagnostic message - conditional block not valid
```
To be consistent with the naming in the SyntaxTree.

src/QsCompiler/Transformations/ClassicallyControlledTransformation.cs
```diff
@@ -272,6 +335,81 @@ private QsStatement CreateApplyIfStatement(QsStatement statement, QsResult resul
...
+                        // ToDo: Diagnostic message
+                        return statement; // The reshaping of the conditional did not succeed.
+                    }
+                }
+
+                #endregion
+
+                #region Condition Checking Logic
+
+                private (bool, TypedExpression, QsScope, QsScope) IsConditionWtihSingleBlock(QsStatement statement)
```
bettinaheim: Typo
```suggestion
                private (bool, TypedExpression, QsScope, QsScope) IsConditionWithSingleBlock(QsStatement statement)
```

src/QsCompiler/Transformations/Utilities.cs
ScottCarda-MS: As a built in operation with no arguments, there should be a built in way of generating a call to NoOp. Because this call is not context sensitive, a utility method for getting this call should be generally available to the transformations.

src/QsCompiler/Transformations/Utilities.cs
ScottCarda-MS: Isn't ResolvedType compared with structural comparison? Doesn't that mean that it is generally safe to compare two ResolvedType objects with the .Equals method?

src/QsCompiler/Transformations/Utilities.cs
ScottCarda-MS: Could you elaborate on this, please?

src/QsCompiler/Transformations/Utilities.cs
ScottCarda-MS: Let's discuss offline.

src/QsCompiler/Transformations/Utilities.cs
bettinaheim: To initialize and array literal it is sufficient for items to have a common base type. Since our arrays are immutable they can be covariant in principle.

src/QsCompiler/Transformations/Utilities.cs
bettinaheim: HasLocalQuantumDependency is a sticky bit - it is pessimistic in that if it is set to true, the expression may have dependencies, but if it is false, it is guaranteed not to. 
If any of the items in the array literal you are constructing here have quantum dependencies then the entire array expression is considered to (potentially) have a local quantum dependency.

src/QsCompiler/Transformations/Utilities.cs
bettinaheim: I don't understand the argument here: There are a lot of library operations with no arguments. I also can't think of any transformation that would have a need to generate an identifier for NoOp. When you write code, it is always useful to keep in mind how code may evolve in the future, and make sure it can easily be extended to cover future needs. However, please don't add code to cover hypothetical scenarios that may or may not be needed in the future - we will adapt and extend the code as these needs come up.

src/QsCompiler/Transformations/Utilities.cs
ScottCarda-MS: This is being track in a separate task.

src/QsCompiler/Transformations/Utilities.cs
ScottCarda-MS: This is being track in a separate task.

src/QsCompiler/Transformations/Utilities.cs
ScottCarda-MS: This is being track in a separate task.

PR #331, opened 2020-02-24
Disable unnecessary transformations and cleaning up API

bettinaheim:
Since we do a pretty heavy breaking change for our transformation setup, I went ahead and also made a pass to clean up the API. 
This PR also disables subtransformations that are not needed.


src/QsCompiler/Tests.Compiler/ClassicalControlTests.fs
```diff
@@ -9,7 +9,7 @@ open Microsoft.Quantum.QsCompiler
 open Microsoft.Quantum.QsCompiler.CompilationBuilder
 open Microsoft.Quantum.QsCompiler.DataTypes
 open Microsoft.Quantum.QsCompiler.SyntaxTree
-open Microsoft.Quantum.QsCompiler.Transformations.ClassicallyControlledTransformation
```
ScottCarda-MS: Please revert the renaming for the Classically Controlled, Monomorphization, Monomorphization Validation, and Intrinsic Resolution transformation, as well as the changes to the Utility class so that the changes can be placed in a separate PR.

src/QsCompiler/DocumentationParser/DocItem.cs
```diff
@@ -26,7 +26,7 @@ internal abstract class DocItem
         protected readonly string replacement;
 
         /// <summary>
-        /// The item's kind, as a string (Utilities.OperationKind, .FunctionKind, or .UdtKind)
+        /// The item's kind, as a string (Utils.OperationKind, .FunctionKind, or .UdtKind)
```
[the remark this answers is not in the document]
bettinaheim: Just to be consistent with the rest of the code base. 
Everything so far is named Utils rather than Utilities.

bettinaheim:
> Please revert the renaming for the Classically Controlled, Monomorphization, Monomorphization Validation, and Intrinsic Resolution transformations, as well as the changes to the Utility class, so that the changes can be placed in a separate PR.

What is the reason to want it a separate PR? There is no benefit to that in terms of resolving merge conflicts, and logically the API clean up belongs here.

src/QsCompiler/Transformations/ClassicallyControlledUtils.cs
```diff
@@ -31,11 +35,11 @@ public static class Utilities
                     : QsNullable<ImmutableArray<ResolvedType>>.Null),
                 typeArgsMapping,
                 resolvedType,
-                new InferredExpressionInformation(false, false),
+                new InferredExpressionInformation(false, false),
```
marshallsa: Trailing space here.

src/QsCompiler/Transformations/IntrinsicResolutionTransformation.cs
```diff
@@ -9,9 +9,9 @@
 using Microsoft.Quantum.QsCompiler.SyntaxTree;
 
 
-namespace Microsoft.Quantum.QsCompiler.Transformations.IntrinsicResolutionTransformation
+namespace Microsoft.Quantum.QsCompiler.Transformations.IntrinsicResolution
```
marshallsa: Should the file be renamed to IntrinsicResolution.cs as well? The naming convention in this folder seems split between files that end in Transformation and those that don't. Since the namespaces are being shortened it could make sense to shorten the filenames too.

src/QsCompiler/Transformations/MonomorphizationTransformation.cs
```diff
@@ -10,15 +10,16 @@
 using Microsoft.Quantum.QsCompiler.SyntaxTokens;
 using Microsoft.Quantum.QsCompiler.SyntaxTree;
 using Microsoft.Quantum.QsCompiler.Transformations.Core;
+using Microsoft.Quantum.QsCompiler.Transformations.SearchAndReplace;
 
 
-namespace Microsoft.Quantum.QsCompiler.Transformations.MonomorphizationTransformation
+namespace Microsoft.Quantum.QsCompiler.Transformations.Monomorphization
```
marshallsa: Same suggestion about renaming file to Monomorphization.cs.

src/QsCompiler/Transformations/MonomorphizationValidationTransformation.cs
```diff
@@ -10,23 +10,23 @@
 using Microsoft.Quantum.QsCompiler.Transformations.Core;
 
 
-namespace Microsoft.Quantum.QsCompiler.Transformations.MonomorphizationValidationTransformation
+namespace Microsoft.Quantum.QsCompiler.Transformations.Monomorphization.Validation
```
marshallsa: Same suggestion about renaming file to MonomorphizationValidation.cs.

src/QsCompiler/Transformations/IntrinsicResolutionTransformation.cs
```diff
@@ -9,9 +9,9 @@
...
 {
-    public class IntrinsicResolutionTransformation
+    public class ReplaceWithTargetIntrinsics
```
marshallsa: It seems like this class is not actually a transformation. I'm guessing it's a utility class? Could it be made static?

src/QsCompiler/Transformations/ClassicallyControlledTransformation.cs
```diff
@@ -9,21 +9,24 @@
 using Microsoft.Quantum.QsCompiler.SyntaxTokens;
 using Microsoft.Quantum.QsCompiler.SyntaxTree;
 using Microsoft.Quantum.QsCompiler.Transformations.Core;
+using Microsoft.Quantum.QsCompiler.Transformations.SearchAndReplace;
 
 
-namespace Microsoft.Quantum.QsCompiler.Transformations.ClassicallyControlledTransformation
+namespace Microsoft.Quantum.QsCompiler.Transformations.ClassicallyControlled
```
marshallsa: I sort of would like to rename this file to ClassicallyControlled.cs as well, although in this case the Transformation part provides contrast with ClassicallyControlledUtils.cs.

src/QsCompiler/Transformations/IntrinsicResolutionTransformation.cs
bettinaheim: My preference would be to rename the files, too, dropping the *Transformation ending. It seems silly to have all files within a folder named Transformations ending in Transformations as well. The reason they were named this way I believe is that there is a RewriteStep folder in the Compiler project that contains files with the same name which implement the IRewriteStep interface invoking these transformations. Personally, I think having both files have the same name is fine.

src/QsCompiler/Transformations/MonomorphizationTransformation.cs
bettinaheim: I'll resolve the rest of the comments, leaving one open until we decide.

src/QsCompiler/Transformations/IntrinsicResolutionTransformation.cs
marshallsa: Oh, I see. Yes, I think it's fine to have the same filenames since it should be clear from context which one is the transformation. If we don't do that then for consistency I'd argue that Conjugations.cs, FunctorGeneration.cs, etc. should be renamed to ConjugationsTransformation.cs, FunctorGenerationTransformation.cs, etc. I prefer the shorter names. :)

Then we also have files like BasicTransformation.cs and CodeTransformations.cs, but having files named Basic.cs and Code.cs would probably be confusing.

PR #336, opened 2020-02-25
updated the app template to use async Main as entry point

filipw:
This doesn't really solve any problem, but - at least to me - avoiding the explicit blocking is a slightly more elegant version for the intro template.
There is no corresponding open issue for that, and given the small change I felt like there was no need to open one.


vadym-kl:
A related enchancement request https://github.com/microsoft/qsharp-runtime/issues/141.
Will the change you suggest change the way error messages are shown ?

PR #337, opened 2020-02-25
Update namespace suggestions code action for access modifiers

marshallsa:
This updates the code action that adds missing open directives, so that it doesn't suggest namespaces which have an inaccessible declaration with the given name.

Part of #259.

### DIFF
--- a/src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
+++ b/src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
@@ -306,16 +306,18 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
 
             // Ensure that the IndexRange library function exists in this compilation unit.
             var nsName = file.TryGetNamespaceAt(range.Start);
+            if (nsName == null)
+            {
+                return Enumerable.Empty<(string, WorkspaceEdit)>();
+            }
             var indexRange = compilation.GlobalSymbols.TryGetCallable(
                 new QsQualifiedName(BuiltIn.IndexRange.Namespace, BuiltIn.IndexRange.Name),
                 NonNullable<string>.New(nsName),
                 file.FileName);
-            if (nsName == null || !indexRange.IsFound)
+            if (!indexRange.IsFound)
             {
                 return Enumerable.Empty<(string, WorkspaceEdit)>();
             }
-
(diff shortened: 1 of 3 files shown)


src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
```diff
@@ -39,22 +40,24 @@ private static WorkspaceEdit GetWorkspaceEdit(this FileContentManager file, para
 
         /// <summary>
         /// Returns all namespaces in which a callable with the name of the symbol at the given position in the given file belongs to.
-        /// Returns an empty collection if any of the arguments is null or if no unqualified symbol exists at that location. 
+        /// Returns an empty collection if any of the arguments is null or if no valid unqualified symbol exists at that location.
```
bettinaheim: What does valid here mean? (same below)

src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
```diff
@@ -294,9 +299,22 @@ string CharacteristicsAnnotation(Characteristics c)
...
+            if (file == null || compilation == null || range?.Start == null)
+            {
+                return Enumerable.Empty<(string, WorkspaceEdit)>();
+            }
+
+            // Ensure that the IndexRange library function exists in this compilation unit.
+            var nsName = file.TryGetNamespaceAt(range.Start);
+            var indexRange = compilation.GlobalSymbols.TryGetCallable(
+                new QsQualifiedName(BuiltIn.IndexRange.Namespace, BuiltIn.IndexRange.Name),
+                NonNullable<string>.New(nsName),
```
bettinaheim: You would need a null check for the nsName before that.

src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
marshallsa: Since `NamespaceSuggestionsForIdAtPosition` now returns an empty enumerable if `TryGetNamespace` returns null, I was trying to document that change in the doc comment.

Now that I think about it I'm not sure if this actually changed the behavior of the function. If an unqualified symbol appears outside a namespace, does `TryGetQsSymbolInfo` recognize it?

marshallsa: Seems like it may be possible for `TryGetQsSymbolInfo` to return info outside of a namespace (maybe?), so I updated the doc comments to be more clear about that case.

PR #340, opened 2020-02-26
Revised transformation setup 

bettinaheim:
This PR contains a major refactoring of our infrastructure for syntax tree transformations.
See the individual commits for details.
The content of the branch to merge matches the one on the features/transformationSetup branch.


src/QsCompiler/Core/TransformationOptions.fs
```diff
@@ -0,0 +1,51 @@
...
+
+/// Used to configure the behavior of the default implementations for transformations.
+type TransformationOptions = {
+    /// Disables the transformation at the transformation root, 
+    /// meaning the transformation won't recur into leaf nodes or subnodes. 
+    Disable : bool
+    /// Indicates that the transformation is used to walk the syntax tree, but does not modify any of the nodes. 
+    /// If set to true, the nodes will not be rebuilt during the transformation. 
+    /// Setting this to true constitutes a promise that the return value of all methods will be ignored. 
+    DisableRebuild : bool
```
marshallsa: I wonder if something like `ReadOnly` would be a better name for this option.

src/QsCompiler/Optimizations/Utils/HelperFunctions.fs
```diff
@@ -50,19 +51,22 @@ let rec internal isLiteral (callables: ImmutableDictionary<QsQualifiedName, QsCa
 /// Otherwise, returns constants without any changes.
 /// If the given variable is already defined, its name is shadowed in the current scope.
 /// Throws an InvalidOperationException if there aren't any scopes on the stack.
-let internal defineVar check constants (name, value) =
-    if not (check value) then constants else constants |> Map.add name value
+let internal defineVar check (constants : IDictionary<_,_>) (name, value) =
+    //if not (check value) then constants else constants |> Map.add name value
```
marshallsa: Should this commented out code still be here?

src/QsCompiler/Optimizations/Utils/HelperFunctions.fs
```diff
@@ -50,19 +51,22 @@ let rec internal isLiteral (callables: ImmutableDictionary<QsQualifiedName, QsCa
...
     | VariableName name, _ ->
         op constants (name.Value, values)
     | VariableNameTuple namesTuple, Tuple valuesTuple ->
         if namesTuple.Length <> valuesTuple.Length then
             ArgumentException "names and values have different lengths" |> raise
-        Seq.zip namesTuple valuesTuple |> Seq.fold (onTuple op) constants
-    | _ -> constants
+        for sym, value in Seq.zip namesTuple valuesTuple do
+            onTuple op constants (sym, value)
+        //Seq.zip namesTuple valuesTuple |> Seq.fold (onTuple op) constants
```
marshallsa: Also here.

src/QsCompiler/Core/TransformationOptions.fs
```diff
@@ -0,0 +1,51 @@
...
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.QsCompiler.Transformations.Core
+
+
+/// Used to configure the behavior of the default implementations for transformations.
+type TransformationOptions = {
+    /// Disables the transformation at the transformation root, 
+    /// meaning the transformation won't recur into leaf nodes or subnodes. 
+    Disable : bool
```
marshallsa: Also, I have a slight preference for booleans named positively, so this could possibly be renamed to Enabled and invert all the logic everywhere it's used.

src/QsCompiler/Core/TransformationOptions.fs
bettinaheim: I think it might be more confusing, since all data structures themselves are immutable?

src/QsCompiler/Optimizations/Utils/HelperFunctions.fs
bettinaheim: No, thanks!

src/QsCompiler/Core/TransformationOptions.fs
bettinaheim: I was keeping it with this one because the default should be that it is enabled, and the default boolean is false.

marshallsa: Does it need to be initialized by a default boolean value somewhere? I think the default could be made true by changing `TransformationOptions.Default` to have `Enabled = true`. It's a really minor issue so feel free to ignore this though. :)

src/QsCompiler/Core/TransformationOptions.fs
marshallsa: That makes sense. My second choice would be to change the default value to true and rename this option to `RebuildTree`, e.g., to avoid double-negatives like "disable rebuild is false => rebuild is enabled," but it's not a big deal either way.

bettinaheim: changed to Rebuild

src/QsCompiler/Core/TransformationOptions.fs
bettinaheim: changed to Enable

PR #345, opened 2020-02-26
Add the internal access modifier for types, operations, and functions

### LINKED ISSUE #259: Access modifiers for callables and types

marshallsa:
Resolves #259.

TODO:
- [x] Name mangling for imported internal declarations (#370)
- [x] Support for re-using internal names in more than one reference (#376)


src/QsCompiler/CompilationManager/CompilationUnit.cs
```diff
@@ -143,6 +153,8 @@ public References(ImmutableDictionary<NonNullable<string>, Headers> refs, Action
     /// </summary>
     public class CompilationUnit : IReaderWriterLock, IDisposable
     {
+        internal static readonly NameDecorator ReferenceDecorator = new NameDecorator("QsReference");
```
bettinaheim: Looks like this one can actually be private.

src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
```diff
@@ -38,24 +39,33 @@ private static WorkspaceEdit GetWorkspaceEdit(this FileContentManager file, para
...
+        /// 
+        /// Returns the name of the identifier as an out parameter if an unqualified symbol exists at that location.
         /// </summary>
         private static IEnumerable<NonNullable<string>> NamespaceSuggestionsForIdAtPosition
             (this FileContentManager file, Position pos, CompilationUnit compilation, out string idName)
         {
             var variables = file?.TryGetQsSymbolInfo(pos, true, out CodeFragment _)?.UsedVariables;
             idName = variables != null && variables.Any() ? variables.Single().Symbol.AsDeclarationName(null) : null;
-            return idName != null && compilation != null
+            var nsName = file.TryGetNamespaceAt(pos);
```
bettinaheim: This additional query for the namespace is no longer needed (it was needed with private I believe). I'd suggest to remove it again for performance reasons.

src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
```diff
@@ -64,7 +74,8 @@ private static IEnumerable<NonNullable<string>> NamespaceSuggestionsForTypeAtPos
             typeName = types != null && types.Any() &&
                 types.Single().Type is QsTypeKind<QsType, QsSymbol, QsSymbol, Characteristics>.UserDefinedType udt
                 ? udt.Item.Symbol.AsDeclarationName(null) : null;
-            return typeName != null && compilation != null
+            var nsName = file.TryGetNamespaceAt(pos);
```
bettinaheim: same here.

bettinaheim:
Two minor comments, but other than that looks good! Well done, @marshallsa!

src/QsCompiler/CompilationManager/CompilationUnit.cs
marshallsa: It's used by the linking tests.

bettinaheim: Hu, I could have sworn I built it. But I'll take your word for it.

PR #348, opened 2020-02-27
updated the compilation options for the generated DLL to use release mode

filipw:
I think it makes sense to always emit it as Release.
it could also be exposed as a compiler flag but that seems unnecessary.


bettinaheim:
@filipw Yes, that makes sense. 
I was wondering if you are using that capability (the capability to generate a dll with only the Q# resources)? If so, I would be curious to learn more.

filipw:
at the moment mostly playing around, trying to figure out scenarios myself. will let you know when something materielizes 😀

bettinaheim:
> at the moment mostly playing around, trying to figure out scenarios myself. will let you know when something materielizes 😀

Cool, please do! That particular setup is similarly somewhat experimental. It is not currently actively used as part of our packages and may be subject to change in the future. Independent on this particular handle, I am always eager to hear feedback from explorations. :)

PR #355, opened 2020-02-28
Classical Control Inequality Condition Support

ScottCarda-MS:
Added support for inequality comparison between Results to be converted to classical control API operation calls.

### DIFF
--- a/src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/ClassicalControl.qs
+++ b/src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/ClassicalControl.qs
@@ -910,3 +910,94 @@ namespace Microsoft.Quantum.Testing.ClassicalControl {
         }
     }
 }
+
+// =================================
+
+// Inequality with ApplyConditionally
+namespace Microsoft.Quantum.Testing.ClassicalControl {
+    open SubOps;
+
+    operation Bar(r : Result) : Unit { }
+
+    operation Foo() : Unit {
+        let r1 = Zero;
+        let r2 = Zero;
+
+        if (r1 != r2) {
+            Bar(r1);
+        }
+        else {
+            SubOp1();
+        }
+    }
+}
+
+// =================================
+
+// Inequality with Apply If One Else Zero
+namespace Microsoft.Quantum.Testing.ClassicalControl {
+    open SubOps;
+
+    operation Bar(r : Result) : Unit { }
+
+    operation Foo() : Unit {
+        let r = Zero;
+
+        if (r != Zero) {
+            Bar(r);
+        }
+        else {
+            SubOp1();
+        }
+    }
+}
+
+// =================================
+
+// Inequality with Apply If Zero Else One
+namespace Microsoft.Quantum.Testing.ClassicalControl {
+    open SubOps;
+
+    operation Bar(r : Result) : Unit { }
+
+    operation Foo() : Unit {
+        let r = Zero;
+
+        if (r != One) {
+            Bar(r);
+        }
+        else {
+            SubOp1();
+        }
+    }
+}
+
+// =================================
+
+// Inequality with ApplyIfOne
+namespace Microsoft.Quantum.Testing.ClassicalControl {
+    open SubOps;
+
+    operation Foo() : Unit {
+        let r = Zero;
+
+        if (r != Zero) {
+            SubOp1();
+        }
+    }
+}
+
+// =================================
+
+// Inequality with ApplyIfZero
+namespace Microsoft.Quantum.Testing.ClassicalControl {
+    open SubOps;
+
+    operation Foo() : Unit {
+        let r = Zero;
+
+        if (r != One) {
+            SubOp1();
+        }
+    }
+}
--- a/src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/ClassicalControl.qs
+++ b/src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/ClassicalControl.qs
@@ -1001,3 +1001,18 @@ namespace Microsoft.Quantum.Testing.ClassicalControl {
         }
     }
 }
+
+// =================================
+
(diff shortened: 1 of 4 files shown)


src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/ClassicalControl.qs
```diff
@@ -909,4 +909,95 @@ namespace Microsoft.Quantum.Testing.ClassicalControl {
...
+namespace Microsoft.Quantum.Testing.ClassicalControl {
+    open SubOps;
+
+    operation Bar(r : Result) : Unit { }
+
+    operation Foo() : Unit {
+        let r1 = Zero;
+        let r2 = Zero;
+
+        if (r1 != r2) {
```
bettinaheim: Am I correct in that this basically gets translated into "if (r1 == r2) { SubOp1(); } else { Bar(r1); }", and the replaced accordingly?

src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/ClassicalControl.qs
```diff
@@ -909,4 +909,95 @@ namespace Microsoft.Quantum.Testing.ClassicalControl {
...
+// Inequality with Apply If One Else Zero
+namespace Microsoft.Quantum.Testing.ClassicalControl {
+    open SubOps;
+
+    operation Bar(r : Result) : Unit { }
+
+    operation Foo() : Unit {
+        let r = Zero;
+
+        if (r != Zero) {
```
bettinaheim: In this case (and the other Apply If cases) you flip the constant, correct? (such that this becomes "if (r == One) { Bar(r); } else { SubOp1(); }")

src/QsCompiler/Transformations/ClassicallyControlled.cs
```diff
@@ -171,32 +171,41 @@ private TypedExpression CreateControlCall(BuiltIn opInfo, IEnumerable<OpProperty
...
                 /// Creates an operation call from the conditional control API for non-literal Result comparisons.
                 /// </summary>
-                private TypedExpression CreateApplyConditionallyExpression(TypedExpression conditionExpr1, TypedExpression conditionExpr2, QsScope conditionScope, QsScope defaultScope)
+                private TypedExpression CreateApplyConditionallyExpression(TypedExpression conditionExpr1, TypedExpression conditionExpr2, QsScope equalityScope, QsScope inequalityScope)
                 {
-                    var (isConditionValid, conditionId, conditionArgs) = IsValidScope(conditionScope);
-                    var (isDefaultValid, defaultId, defaultArgs) = IsValidScope(defaultScope);
+                    if (equalityScope == null && inequalityScope == null)
+                    {
+                        return null; // Not sure how this would happen
```
bettinaheim: What happens if you return null? Is the if-statement just left unchanged? 
Since this shouldn't happen, a QsCompilerDiagnostic would probably make sense here.

src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/ClassicalControl.qs
ScottCarda-MS: Yes, that's right. If there was no 'else' clause to flip with the conditional body, the missing 'else' is treated like a NoOp.

src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/ClassicalControl.qs
ScottCarda-MS: Yes, that is correct.

src/QsCompiler/Transformations/ClassicallyControlled.cs
ScottCarda-MS: When it goes to create the surrounding statement, in CreateControlStatement, it checks if the created expression is null. If so, it just returns the original statement.

ScottCarda-MS: The comment was in reference to the condition. I am not sure how the condition would be met. I don't know how we could have an if-statement whose bodies are all null.

src/QsCompiler/Transformations/ClassicallyControlled.cs
```diff
@@ -413,41 +420,92 @@ private QsStatement ConvertConditionalToControlCall(QsStatement statement)
...
-                /// Checks if the expression is an equality comparison where one side is a Result literal.
-                /// If it is, returns true along with the Result literal and the other expression in the
-                /// equality, otherwise returns false with nulls.
+                /// Checks if the expression is an equality or inequality comparison where one side is a
+                /// Result literal. If it is, returns true along with the Result literal and the other
+                /// expression in the (in)equality, otherwise returns false with nulls. If it is an
+                /// inequality, the returned result value will be the opposite of the result literal found.
                 /// </summary>
-                private (bool, QsResult, TypedExpression) IsConditionedOnResultLiteralExpression(TypedExpression expression)
+                private bool IsConditionedOnResultLiteralExpression(TypedExpression expression, out QsResult literal, out TypedExpression nonLiteral)
```
marshallsa: What happens if both expressions are literals? The name `nonLiteral` wouldn't be correct in that case.

src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/ClassicalControl.qs
```diff
@@ -909,4 +909,95 @@ namespace Microsoft.Quantum.Testing.ClassicalControl {
...
+namespace Microsoft.Quantum.Testing.ClassicalControl {
+    open SubOps;
+
+    operation Foo() : Unit {
+        let r = Zero;
+
+        if (r != One) {
+            SubOp1();
+        }
+    }
```
marshallsa: Could you add a test where the literal is on the left side?

PR #356, opened 2020-03-03
use local path for UNC files in language server

filipw:
At the moment most of the language server features do not work when opening the project on the UNC path  (shared network, a Mac folder on a Parallels Windows VM etc). See this GIF:

4fDZHrWr6G

This is caused by using `Uri.AbsolutePath` instead of `LocalPath`. 

There is really no way to have a test for this, but I tested using  `"quantumDevKit.languageServerPath"` and it seems to solve the issue, see below (notice the UNC path of this project):

Screenshot 2020-03-03 20 50 46


filipw:
actually, I posted screenshots from VS Code only, but it's the same for VS too

bettinaheim:
Hi @filipw, 
Sorry for the delay with processing this. I meant to double check whether everything works still fine in our generated code, and didn't get around to it yet. Hopefully it shouldn't be too long.

filipw:
thanks a lot 👍

ricardo-espinoza:
Thanks a lot @filipw for reporting this issue and contributing with the fix. We’ve reviewed and tested it and it looks good.
 
Marking as Approved.

PR #361, opened 2020-03-09
Remove the private access modifier

marshallsa:
See the new comment on #259 for why it's being removed.

### DIFF
--- a/src/QsCompiler/Tests.Compiler/AccessModifierTests.fs
+++ b/src/QsCompiler/Tests.Compiler/AccessModifierTests.fs
@@ -19,7 +19,7 @@ type AccessModifierTests (output) =
                                                  [File.ReadAllLines("ReferenceTargets.txt").ElementAt(1)],
                            output)
 
-    member private this.Expect name (diagnostics : IEnumerable<DiagnosticItem>) = 
+    member private this.Expect name (diagnostics : IEnumerable<DiagnosticItem>) =
         let ns = "Microsoft.Quantum.Testing.AccessModifiers" |> NonNullable<_>.New
         let name = name |> NonNullable<_>.New
         this.Verify (QsQualifiedName.New (ns, name), diagnostics)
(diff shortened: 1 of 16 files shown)


src/VSCodeExtension/syntaxes/qsharp.tmLanguage.json.v.template
```diff
@@ -50,7 +50,7 @@
         },
         {
           "name": "keyword.other.qsharp",
-          "match": "\\b(new|not|and|or|w\/|private|internal)\\b"
+          "match": "\\b(internal|new|not|and|or|w\/)\\b"
```
bettinaheim: I wonder if it might make more sense to have it listed where operation is listed as well (meaning they would get the same coloring - not sure how it usually is).

src/QsCompiler/Tests.Compiler/AccessModifierTests.fs
```diff
@@ -19,55 +19,38 @@ type AccessModifierTests (output) =
...
                            output)
 
-    member private this.Expect name (diagnostics : IEnumerable<DiagnosticItem>) = 
+    member private this.Expect name (diagnostics : IEnumerable<DiagnosticItem>) =
         let ns = "Microsoft.Quantum.Testing.AccessModifiers" |> NonNullable<_>.New
         let name = name |> NonNullable<_>.New
         this.Verify (QsQualifiedName.New (ns, name), diagnostics)
 
     [<Fact>]
     member this.``Callables`` () =
```
bettinaheim: It might be nice to add a suitable reference manually to the compilation unit (you would have to adapt CompilerTests.Compile to take references as argument), such that you can have a test or two where something should fail.

marshallsa: Sorry, I'm not sure I understand - the tests where an inaccessible callable or type is used are in the "References" test. (Maybe I should get rid of the separate test and move them to the callables and types tests, though.)

marshallsa: The tests that make sure internal callables and types can't be used in references have now been moved up to the callables and types tests, respectively.

PR #362, opened 2020-03-10
General Hoist Transformation

ScottCarda-MS:
Separates out the hoisting (lifting) logic of the classical control transformation into its own transformation that can be used by future features.

Work that still needs to be done:
- Add IsSelfAdjointable to the BuiltIn Class so it can be used in the new transformation.
- Separate out unit tests.


ScottCarda-MS:
Closed in favor of #364.

PR #366, opened 2020-03-10
Access modifiers for callable lifting PR

bettinaheim:


src/QsCompiler/Transformations/ContentLifting.cs
```diff
@@ -64,17 +63,22 @@ public CallableDetails(QsCallable callable)
...
-            public bool InAdjoint = false;
-            public bool InControlled = false;
-            public bool InWithinBlock = false;
+            internal CallableDetails CurrentCallable = null;
+
+            protected internal bool InBody = false;
+            protected internal bool InAdjoint = false;
+            protected internal bool InControlled = false;
+            protected internal bool InControlledAdjoint = false;
+            protected internal bool InWithinBlock = false;
```
ScottCarda-MS: Since combining access modifiers in C# can result in some confusing behavior (confusing to me at least 😁), I just want to double check this does what you intend. My understanding is that this will make these fields accessible to anything in the assembly, _and_ _also_ accessible to any derived class outside of the assembly. Is that what you are going for here?

src/QsCompiler/Transformations/ContentLifting.cs
```diff
@@ -64,17 +63,22 @@ public CallableDetails(QsCallable callable)
...
+
+            public List<QsCallable> GeneratedOperations = null;
```
ScottCarda-MS: Should this also be internal?

src/QsCompiler/Transformations/ContentLifting.cs
bettinaheim: It was indeed what I was going for. I did debate internally what they should be, and I could go either way, but then I ultimately didn't want to make them fully public (I can be convinced though).

src/QsCompiler/Transformations/ContentLifting.cs
```diff
@@ -347,6 +351,14 @@ public override QsSpecialization OnControlledSpecialization(QsSpecialization spe
                 return rtrn;
             }
 
+            public override QsSpecialization OnControlledAdjointSpecialization(QsSpecialization spec)
```
ScottCarda-MS: The logic for the lifting, at least as of now, doesn't actually need to know when it is in this specialization. But I guess it could be nice to have more complete information. I'll leave it up to you if you want to include this and the associated flag.

src/QsCompiler/Transformations/ContentLifting.cs
bettinaheim: Unfortunatly, I think you basically want access to it if you derive from the LiftContent. I don't entirely like that anyone can just add and remove things, but the alternative is to make is protected internal and force people to add a handle to access it when they inherit from it. I am fine either way.

src/QsCompiler/Transformations/ContentLifting.cs
```diff
@@ -64,17 +63,22 @@ public CallableDetails(QsCallable callable)
 
         public class TransformationState
         {
-            public bool IsValidScope = true;
-            public List<QsCallable> GeneratedOperations = null;
-            public ImmutableArray<LocalVariableDeclaration<NonNullable<string>>> GeneratedOpParams =
+            // ToDo: It should be possible to make these three properties private, 
+            // if we absorb the corresponding logic into LiftBody. 
+            internal bool IsValidScope = true;
```
ScottCarda-MS: Currently, this field needs to have the same access set as the whole LiftContent transformation, as the transformation could be inherited outside the assembly. In that case, that class would need access to this field to validate before calling LiftBody.

src/QsCompiler/Transformations/ContentLifting.cs
bettinaheim: Yes, I realized that looking at what queries these handles. It just felt like such an implementation detail that you want to know what specialization you are in except if it's a controlled adjoint speciliazation.

src/QsCompiler/Transformations/ContentLifting.cs
bettinaheim: Right. This is the one piece where I would suggest to look into whether there is a way to not have to expose that handle. Reading your replies on your PR possibly clarifies it a bit whether that makes sense. For this PR I'll make it public again, but it would be good if we could find a way around that.

src/QsCompiler/Transformations/ContentLifting.cs
bettinaheim: going with protected internal

PR #367, opened 2020-03-11
Support for target packages

### LINKED ISSUE #307: Command line parameter --format MsBuild has no effect in response file
**Describe the bug**

When "--format MsBuild" is specified in response file rather than on command line the error reporting is still in "Human Readable" format. Response files are widely used in MsBuild and Visual Studio fails to parse and report these errors correctly.

**To Reproduce**

1. Create response file
qsc.rsp:
--input a.qs
--format MsBuild

2. Create Q# file a.qs with errors

3. Run "dotnet qsc.dll build --response-files qsc.rsp"

4. See that errors are in "Human Readable" format:

Warning QS3101:
C:/Users/dmitryv/Documents/Q/TryQSC/a.qs
Position: [ln 2, cn 92]
Deprecated syntax. Use "is Ctl+Adj" to denote the operation characteristics instead.

5. Move "--format MsBuild" to command line

6. Observe errors in MS Build format:

C:/Users/dmitryv/Documents/Q/TryQSC/a.qs(2,92): warning QS3101: Deprecated syntax. Use "is Ctl+Adj" to denote the operation characteristics instead.

**Expected behavior**

I expect --format switch to have the same effect on command file and in response file.

**Screenshots**

WrongFormat

**System information**

.nuget\packages\microsoft.quantum.sdk\0.10.2001.1505-alpha

bettinaheim:
This PR adds support for pulling in packages that contain target specific implementations. 
This also fixes #307 (happy to make a separate PR).


src/QsCompiler/CommandLineTool/Options.cs
```diff
@@ -17,6 +17,81 @@
...
+            {
+                var pieces = keyValue?.Split("=");
+                var valid = pieces != null && pieces.Length == 2;
+                success = valid && parsed.TryAdd(pieces[0].Trim().Trim('"'), pieces[1].Trim().Trim('"')) && success;
+            }
+            return success;
+        }
+
+        /// <summary>
+        /// Returns null if TargetPackage is not null or empty, and
```
swernli: I think this should read "Returns null if TargetPackage is null or empty" instead of "not null or empty"

src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -273,7 +273,11 @@ type ErrorCode =
     | UnknownCompilerPlugin = 7015
     | CouldNotLoadCompilerPlugin = 7016
     | CouldNotInstantiateRewriteStep = 7017
-    | UnexpectedCompilerException = 7018
+    | CouldNotFineTargetPackage = 7018
```
swernli: Should be `CouldNotFindTargetPackage`

src/QsCompiler/CommandLineTool/Options.cs
bettinaheim: Yes, thank you!

src/QsCompiler/DataStructures/Diagnostics.fs
bettinaheim: ups, thanks!

PR #368, opened 2020-03-11
Changes to BuiltIn Records

ScottCarda-MS:
IsSelfAdjoint is a field of the InferredCallableInformation object, which we need to fill out when we generate calls to these built-in objects. The IsSelfAdjoint can't be determined from the signature, so we need to provide it explicitly, in the case of operations.

I need a review on the structural changes to the BuiltIn record, before I go and change the references.


src/QsCompiler/Core/Dependencies.fs
```diff
@@ -35,184 +46,161 @@ type BuiltIn = {
...
     }
 
     // hard dependencies in Microsoft.Quantum.Canon
 
     static member NoOp = {
-        Name = "NoOp" |> NonNullable<string>.New
-        Namespace = BuiltIn.CanonNamespace
-        TypeParameters = ImmutableArray.Create("T" |> NonNullable<string>.New)
+        FullName = {Name = "NoOp" |> NonNullable<string>.New; Namespace = BuiltIn.CanonNamespace}
+        Kind = Operation (TypeParameters = ImmutableArray.Create("T" |> NonNullable<string>.New), IsSelfAdjoint = false)
```
alan-geller: No op isn't self adjoint?

ScottCarda-MS: From looking at https://github.com/microsoft/QuantumLibraries/blob/master/Standard/src/Canon/NoOp.qs
it looks like it isn't declared as such.

bettinaheim: It should be...

PR #370, opened 2020-03-13
Name mangling for imported internal declarations (after symbol resolution)

marshallsa:
This is an alternative to PR #360, part of issue #259. Instead of name mangling as soon as references are loaded (before symbol resolution), name mangling is run just before the syntax tree is built in `QsCompilation.Build`. The benefit is that it is easier to show specific error messages for inaccessible symbols, instead of an "unknown symbol" message.

Because this happens after symbol resolution, the symbol table needs to be updated to accommodate multiple declarations for a single callable or type if some of them belong to a project/package reference. Thus, the logic for resolving symbols in the symbol table has been refactored so that it is more flexible.

TODO in this PR:

- [X] Allow shadowing imported internal callables
- [x] Allow shadowing imported internal types
- [x] Run name mangling in `QsCompilation.Build`
- [x] Add tests for name mangling
- [x] Add tests for redeclarations in more than one reference
- [x] Decide what to do about some assertions that were disabled (marked with TODO comments)

TODO in the next PR:

- Allow redeclarations of internal items in more than one reference
- Group internal specializations with the same name in different references correctly


src/QsCompiler/Core/SymbolTable.fs
```diff
@@ -13,23 +13,102 @@ open Microsoft.Quantum.QsCompiler.Diagnostics
...
+    /// referencing it.
     | Inaccessible
-    /// No declaration with that name was found.
+    /// A result indicating that no declaration with that name was found.
     | NotFound
-    
+
+
+/// Tools for processing resolution results.
+module private ResolutionResult =
```
bettinaheim: It looks like almost all of these are really members to ResultionResult<'T>, and one or two would be static members. What was the reason to split them into a separate module, opposed to having them be members?
Also, a better place to put these would probably be SymbolResolutions.fs before the SymbolResolution module.

src/QsCompiler/Core/SymbolTable.fs
```diff
@@ -261,21 +343,25 @@ and Namespace private
     let mutable TypesDefinedInAllSourcesCache = null
     let mutable CallablesDefinedInAllSourcesCache = null
 
-    /// Given a non-nullable string, returns true if either a callable or a type with that name already exists
-    /// either in a source file, or in a referenced assembly.
-    let IsDefined arg = 
-        CallablesInReferences.ContainsKey arg || TypesInReferences.ContainsKey arg || 
-        Parts.Values.Any (fun partial -> partial.ContainsCallable arg || partial.ContainsType arg)
+    /// Returns true if the name is available for use in a new declaration.
+    let isNameAvailable name =
```
bettinaheim: Could we keep with upper-casing? I know the lower case first letter is more aligned with F#, but we use upper case letters in the places where the expectation is that the property/method will be used from within C#. Just having everything upper case makes it more consistent I think.

src/QsCompiler/Core/SymbolTable.fs
```diff
@@ -13,23 +13,102 @@ open Microsoft.Quantum.QsCompiler.Diagnostics
...
+        | _ -> None
+
+    /// Converts the resolution result to a 0- or 1-element list containing the Found value if the result is a Found.
+    let private toList = toOption >> Option.toList
+
+    /// Sorts the sequence of results in order of Found, Ambiguous, Inaccessible, and NotFound.
+    ///
+    /// If the sequence contains a Found result, the rest of the sequence after it is not automatically evaluated.
+    /// Otherwise, the whole sequence may be evaluated by calling sort.
+    let private sort results =
```
bettinaheim: Hehe, I see someone is having fun with F# :)

src/QsCompiler/CompilationManager/CompilationUnit.cs
```diff
@@ -686,5 +701,71 @@ internal LocalDeclarations TryGetLocalDeclarations(FileContentManager file, Posi
...
+            Regex.Replace(Path.GetFileNameWithoutExtension(source), "[^A-Za-z0-9_]", "");
+
+        /// <summary>
+        /// Embeds the tag name into the qualified name.
+        /// </summary>
+        /// <param name="tag">The tag name to embed.</param>
+        /// <param name="name">The qualified name in which to embed the tag name.</param>
+        /// <returns>A new qualified name with the tag embedded into it.</returns>
+        internal static QsQualifiedName EmbedTag(string tag, QsQualifiedName name) =>
+            // TODO: We might need to change the format to make it easier to reverse later.
```
bettinaheim: Regarding easily reversible, e.g. the pattern used in UniqueVariableNames has a tested regex pattern.

src/QsCompiler/CompilationManager/CompilationUnit.cs
```diff
@@ -686,5 +701,71 @@ internal LocalDeclarations TryGetLocalDeclarations(FileContentManager file, Posi
...
+                .GroupBy(GetTagForSource)
+                .SelectMany(group =>
+                    group.Count() == 1
+                    ? new[] { (key: group.Single(), value: group.Key) }
+                    : group.Select((source, index) => (key: source, value: group.Key + index)))
+                .ToImmutableDictionary(item => item.key, item => item.value);
+
+            ImmutableDictionary<QsQualifiedName, QsQualifiedName> GetMappingForSourceGroup(
+                IGrouping<string, (QsQualifiedName name, string source, AccessModifier access)> group) =>
+                // TODO: Is there another way besides file extension to check if the source file is a reference?
```
bettinaheim: Not currently. What I meant to do, however, is to generated a Q# attribute for declarations that come from references to tag them. I'll make a task under clean-up to do that.

src/QsCompiler/CompilationManager/CompilationUnit.cs
```diff
@@ -686,5 +701,71 @@ internal LocalDeclarations TryGetLocalDeclarations(FileContentManager file, Posi
...
             return this.PositionedDeclarations(parentCallable, callablePos, specPos, declarations); 
         }
+
+        /// <summary>
+        /// Returns a short tag name for the given source file. The tag is not unique. If multiple source files are
+        /// being tagged, make sure there are no name conflicts.
+        /// </summary>
+        /// <param name="source">The source file to get the tag for.</param>
+        /// <returns>A short tag name for the given source file.</returns>
+        internal static string GetTagForSource(string source) =>
```
bettinaheim: I'd make this a static method in RenameReferences, and the same for EmbedTag.

src/QsCompiler/Core/SymbolTable.fs
marshallsa: I based it loosely off of the F# Seq module, which is organized similarly. I don't think the `seq` type has any members (static or instance), either. In this situation I prefer the function style because it's easier to use with the `|>` operator.

src/QsCompiler/Core/SymbolTable.fs
marshallsa: [F# naming conventions](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting#naming-conventions) say to use camelCase everywhere except members and type names.

In terms of consistency, we already use lower case names in the Q# compiler for local let-bindings inside of functions, and let-bindings at the type level are basically local bindings inside the type constructor.

src/QsCompiler/Core/SymbolTable.fs
bettinaheim: Ok, sounds good. I would still make sense to move it to SymbolResolution.fs.

src/QsCompiler/Core/SymbolTable.fs
bettinaheim: As long as it's not visible, I am fine either way. But for let-bindings inside modules I'd like to go with upper case, since these may be exposed to C#, and it should not be visible for anyone using the compiler packages whether the code was writing in F# or C#.

src/QsCompiler/CompilationManager/CompilationUnit.cs
marshallsa: RenameReferences is meant to be a general transformation that doesn't care what kind of renaming it is doing, so I put the name-mangling logic in the Compilation class, since that is where we do care that the mangling is done in a certain way (based on source file).

src/QsCompiler/Core/SymbolTable.fs
marshallsa: Yeah, surprisingly that page recommends even public module-level let-bindings be lowercased, but [on this page](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/component-design-guidelines#guidelines-for-f-facing-libraries) they recommend uppercase for identifiers that can be used from other .NET languages.

Should I rename all of my ResolutionResult module functions to start with an uppercase letter?

src/QsCompiler/CompilationManager/CompilationUnit.cs
marshallsa: Actually, I should probably unify this with GenerateUniqueName and StripUniqueName in the UniqueVariableNames transformation.

src/QsCompiler/Core/SymbolTable.fs
bettinaheim: Yes, I think it would be good to do that for the functions in the ResolutionResult. Thanks!

src/QsCompiler/CompilationManager/CompilationUnit.cs
marshallsa: It looks like I can use `Externals.Declarations.ContainsKey(source)` instead.

src/QsCompiler/CompilationManager/CompilationUnit.cs
marshallsa: I updated it to use the same name format as UniqueVariableNames, could you take another look?

src/QsCompiler/Tests.Compiler/Tests.Compiler.fsproj
```diff
@@ -175,8 +178,12 @@
     <PropertyGroup>
       <ExecutionTarget>$(MSBuildThisFileDirectory)..\TestTargets\Simulation\Example\bin\$(Configuration)\netcoreapp3.0\Example.dll</ExecutionTarget>
       <LibraryTarget>$(MSBuildThisFileDirectory)..\TestTargets\Libraries\Library1\bin\$(Configuration)\netcoreapp3.0\Library1.dll</LibraryTarget>
+      <LibraryTargetCopy>$(MSBuildThisFileDirectory)..\TestTargets\Libraries\Library1\bin\$(Configuration)\netcoreapp3.0\Library2.dll</LibraryTargetCopy>
```
bettinaheim: The folder here should be Library2 as well, and we should set the target framework to netcoreapp3.1 as well. Also, it would be nice to rename the property names to LibraryTarget1 and LibraryTarget2 or similar.

src/QsCompiler/Tests.Compiler/Tests.Compiler.fsproj
```diff
@@ -175,8 +178,12 @@
...
     </PropertyGroup>
-    <WriteLinesToFile File="$(OutputPath)ReferenceTargets.txt" Lines="$(ExecutionTarget); $(LibraryTarget)" Overwrite="true" />
+    <Copy SourceFiles="$(LibraryTarget)" DestinationFiles="$(LibraryTargetCopy)"/>
```
bettinaheim: Ah, I see. I was going to suggest to actually manually make literally another library project and add it to the solution instead of copying it here.

src/QsCompiler/Transformations/SearchAndReplace.cs
```diff
@@ -363,6 +363,59 @@ public override QsStatementKind OnValueUpdate(QsValueUpdate stm)
 
     // routines for replacing symbols/identifiers
 
+    /// <summary>
+    /// Provides simple name decoration (or name mangling) by prefixing names with a label and number.
+    /// </summary>
+    public class NameDecorator
```
bettinaheim: Thanks for splitting this out!

src/QsCompiler/CompilationManager/CompilationUnit.cs
bettinaheim: Done.

src/QsCompiler/Tests.Compiler/Tests.Compiler.fsproj
marshallsa: I was trying to avoid going to all the trouble when I only need a copy. :) Do you think copying the file isn't a good idea?

bettinaheim: Having a second library just seems cleaner to me, and not all that much trouble. I meant to make one for testing purposes, but so far there just wasn't a reason to.

marshallsa: Oh, actually, that is a good point... I forgot Library1 contains public methods besides the internal ones in the access modifier tests. That would probably make the test fail even if internal redeclarations were supported correctly. I'll make a separate library.

marshallsa: I decided to write a completely different test that more precisely checks that specializations with the same name are grouped correctly based on their assembly. (The old test was only checking that the name still existed in the syntax tree.) I added it to the linking tests. The extra library file is no longer needed. Please take another look. :)

src/QsCompiler/Tests.Compiler/Tests.Compiler.fsproj
marshallsa: With the new test, the extra library target is no longer needed, so I removed this line. I think Library1 will get updated to .NET Core 3.1 when master is merged into the feature branch again.

src/QsCompiler/Tests.Compiler/Tests.Compiler.fsproj
bettinaheim: Looks good to me.

bettinaheim:
@marshallsa Found the issue: The changes here seem to have gotten lost. I'll make a PR to this branch shortly.

PR #371, opened 2020-03-14
Updating gitignore to skip .vscode & .ionide

swernli:


.gitignore
```diff
@@ -33,6 +33,11 @@ drops/
 # Visual Studio 2017 auto generated files
 Generated\ Files/
 
+# Visual Studio Code options directory
+.vscode/
```
cgranade: I apologize that I didn't see this PR before completing, but I find it a really helpful best practice to include `.vscode/extensions.json` to recommend what extensions users should have to work with the repo.

swernli: Interesting... I didn't know that was a feature vscode supported! I do agree that suggesting extensions is good, but I also find it tricky when folks provide existing vscode settings that might not fit my environment (like build and test tasks). Would the tactic there be to gitignore the specific json files that relate to workspace settings instead of a blanket ignore of the whole folder?
Also, while the Ionide extension is helpful for processing F# projects and the Omnisharp extension is recommended for C# projects, they have trouble playing well together. It took me a while to figure out how to get Ionide not to show errors for the C# projects that it fails to load and on the flip side how to avoid seeing thousands of errors registered by Omnisharp because it didn't understand namespaces that were coming from F# projects. I think

swernli: Also, while the Ionide extension is helpful for processing F# projects and the Omnisharp extension is recommended for C# projects, they have trouble playing well together. It took me a while to figure out how to get Ionide not to show errors for the C# projects that it fails to load and on the flip side how to avoid seeing thousands of errors registered by Omnisharp because it didn't understand namespaces that were coming from F# projects. I think we should add some additional guidance to the readme.md for vscode users to help with this set up.

cgranade: > Interesting... I didn't know that was a feature vscode supported! I do agree that suggesting extensions is good, but I also find it tricky when folks provide existing vscode settings that might not fit my environment (like build and test tasks). Would the tactic there be to gitignore the specific json files that relate to workspace settings instead of a blanket ignore of the whole folder?

That could make sense, yeah. The other option might be to gitignore `.vscode`, but to override `.vscode/extensions.json`. What makes that particularly nice is that anything in `extensions.json` is only a recommendation made by the current workspace, and is overriden by `settings.json` at either the workspace or user level.

PR #375, opened 2020-03-16
Pulling in changes from PR 358

bettinaheim:


src/QsCompiler/Transformations/SearchAndReplace.cs
```diff
@@ -542,62 +530,145 @@ internal UserDefinedType RenameUdt(UserDefinedType udt)
...
+                qualifiedName: State.GetNewName(type.QualifiedName),
+                attributes: type.Attributes.Select(Namespaces.OnAttribute).ToImmutableArray(),
+                modifiers: type.Modifiers,
+                sourceFile: type.SourceFile,
+                position: type.Position,
+                symbolRange: type.SymbolRange,
+                type: Types.OnType(type.Type),
+                typeItems: Namespaces.OnTypeItems(type.TypeItems),
+                documentation: Namespaces.OnDocumentation(type.Documentation));
+        }
```
marshallsa: I removed these methods, since PR #370 moves the name mangling step late enough that the syntax tree already exists, and we can run the transformation on that instead of the declaration headers.

bettinaheim: Yes, I know, but I need them for another feature later down the line. They are not needed here, but it was just easier to keep them to limit my confusion regarding what got merged where.

marshallsa: Ah, OK. I would still prefer to remove them from the access modifiers PR, so that that PR is not bigger than it needs to be, but the PR that needs them can add them back. Is that OK?

bettinaheim: removed

PR #376, opened 2020-03-16
Support more than one reference with the same internal name

marshallsa:
Enables the skipped test added by #370. Part of issue #259.

### DIFF
--- a/src/QsCompiler/CompilationManager/CompilationUnit.cs
+++ b/src/QsCompiler/CompilationManager/CompilationUnit.cs
@@ -493,7 +493,11 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
             var specializations =
                 GlobalSymbols
                 .ImportedSpecializations(header.QualifiedName)
-                .Where(specialization => specialization.Item1.SourceFile.Equals(header.SourceFile))
+                .Where(specialization =>
+                    // Either the callable is externally accessible, or all of its specializations must be defined in
+                    // the same reference as the callable.
+                    Namespace.IsDeclarationAccessible(false, header.Modifiers.Access) ||
+                    specialization.Item1.SourceFile.Equals(header.SourceFile))
                 .Select(specialization =>
                 {
                     var (specHeader, implementation) = specialization;
--- a/src/QsCompiler/CompilationManager/CompilationUnit.cs
+++ b/src/QsCompiler/CompilationManager/CompilationUnit.cs
@@ -483,25 +483,29 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
         {
             // TODO: this needs to be adapted if we want to support external specializations
             if (header == null) throw new ArgumentNullException(nameof(header));
-            var importedSpecs = this.GlobalSymbols.ImportedSpecializations(header.QualifiedName);
             if (Namespace.IsDeclarationAccessible(false, header.Modifiers.Access))
             {
-                var definedSpecs = this.GlobalSymbols.DefinedSpecializations(header.QualifiedName);
+                var definedSpecs = GlobalSymbols.DefinedSpecializations(header.QualifiedName);
                 QsCompilerError.Verify(definedSpecs.Length == 0,
                                        "external specializations are currently not supported");
             }
 
-            var specializations = importedSpecs.Select(imported =>
-            {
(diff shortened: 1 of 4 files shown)


src/QsCompiler/Core/SymbolTable.fs
```diff
@@ -240,10 +240,12 @@ type private PartialNamespace private
...
 and Namespace private
-    (name, parts : IEnumerable<KeyValuePair<NonNullable<string>,PartialNamespace>>,
-     CallablesInReferences : ImmutableDictionary<NonNullable<string>, CallableDeclarationHeader>,
-     SpecializationsInReferences : ILookup<NonNullable<string>, SpecializationDeclarationHeader * SpecializationImplementation>,
-     TypesInReferences : ImmutableDictionary<NonNullable<string>, TypeDeclarationHeader>) =
+    (name,
+     parts : IEnumerable<KeyValuePair<NonNullable<string>,PartialNamespace>>,
+     CallablesInReferences : ILookup<NonNullable<string>, CallableDeclarationHeader>,
+     SpecializationsInReferences : ILookup<NonNullable<string>,
+                                           SpecializationDeclarationHeader * SpecializationImplementation>,
```
bettinaheim: So this is now just a bag that contains all specializations for all callables with that (potentially internal) name?

src/QsCompiler/CompilationManager/CompilationUnit.cs
```diff
@@ -483,25 +483,29 @@ private QsCallable GetImportedCallable(CallableDeclarationHeader header)
         {
             // TODO: this needs to be adapted if we want to support external specializations
             if (header == null) throw new ArgumentNullException(nameof(header));
-            var importedSpecs = this.GlobalSymbols.ImportedSpecializations(header.QualifiedName);
             if (Namespace.IsDeclarationAccessible(false, header.Modifiers.Access))
             {
-                var definedSpecs = this.GlobalSymbols.DefinedSpecializations(header.QualifiedName);
+                var definedSpecs = GlobalSymbols.DefinedSpecializations(header.QualifiedName);
```
bettinaheim: nit/just something to think about: I kind of like the explicit "this" since it makes it transparent what are instance methods and which ones are static.

src/QsCompiler/CompilationManager/CompilationUnit.cs
```diff
@@ -483,25 +483,29 @@ private QsCallable GetImportedCallable(CallableDeclarationHeader header)
...
-                    : header.Signature;
-                return new QsSpecialization(specHeader.Kind, header.QualifiedName, specHeader.Attributes,
-                    specHeader.SourceFile, specHeader.Location, specHeader.TypeArguments, specSignature,
-                    implementation, specHeader.Documentation, QsComments.Empty);
-            })
-            .ToImmutableArray();
+            var specializations =
+                GlobalSymbols
+                .ImportedSpecializations(header.QualifiedName)
+                .Where(specialization => specialization.Item1.SourceFile.Equals(header.SourceFile))
```
bettinaheim: Why are you filtering by source file? I realize that we currently only support that scenario, but I'd rather not.

src/QsCompiler/Core/SymbolTable.fs
marshallsa: Yes.

src/QsCompiler/CompilationManager/CompilationUnit.cs
marshallsa: I updated the condition to filter by source only if the callable is internal. Even if we support external specializations, specializations for an internal callable can only be defined within the same assembly, because it's not visible from anywhere else. So if I don't filter by source, a callable could get specializations that don't belong to it - does that make sense?

src/QsCompiler/CompilationManager/CompilationUnit.cs
marshallsa: I added it back.

src/QsCompiler/CompilationManager/CompilationUnit.cs
bettinaheim: Ah, I see. Yes, that makes sense, thanks!

marshallsa:
Yep! (Hopefully!)

PR #377, opened 2020-03-17
Migrate Functor Generation

ScottCarda-MS:


src/QsCompiler/Compiler/RewriteSteps/FunctorGeneration.cs
```diff
@@ -0,0 +1,62 @@
...
+            }
+
+            var providedOperations = new QsNamespace[] { requiredNamespace }.Callables().Select(c => c.FullName);
+            var requiredBuiltIns = new List<QsQualifiedName>()
+            {
+                BuiltIn.Length.FullName,
+                BuiltIn.RangeReverse.FullName
+            };
+
+            return requiredBuiltIns.All(builtIn => providedOperations.Any(provided => provided.Equals(builtIn)));
```
bettinaheim: It's two entries and not a whole lot of namespaces, so it doesn't really matter, but as a general setup, I'd expect that making providedOperations something like an ImmutableSet and leveraging set operations should be more efficient.

ScottCarda-MS: Hmm, I am skeptical that using set operations would be more efficient than using the Any method in general for these kinds of checks. Do you have any resources that you could post?

PR #380, opened 2020-03-17
Fix build failure with .NET Core 3.1.200

marshallsa:
This fixes the following build error with the recent .NET Core 3.1.200 update:

> C:\Users\samarsha\Desktop\qsharp-compiler\src\QsCompiler\Optimizations\Utils\Evaluation.fs(65,26): error FS0410: The type 'FunctionInterrupt' is less accessible than the value, member or type 'member internal FunctionEvaluator.EvaluateExpression : expr:TypedExpression -> Imp<TypedExpression>' it is used in. [C:\Users\samarsha\Desktop\qsharp-compiler\src\QsCompiler\Optimizations\Optimizations.fsproj]

The problem is that FunctionEvaluator.EvaluateExpression is internal, but its return type is type parameterized on the private type FunctionInterrupt. Making FunctionInterrupt internal fixes the problem. I'm guessing that an F# update fixed a bug that allowed the original code to compile in the first place.


marshallsa:
Actually, EvaluateExpression is only used within the class, so it seems like a better option is to make both FunctionInterrupt and EvaluateExpression private.

bettinaheim:
> Actually, EvaluateExpression is only used within the class, so it seems like a better option is to make both FunctionInterrupt and EvaluateExpression private. @bettinaheim

Sure, sounds good.

PR #386, opened 2020-03-22
Call Graph Walker

ScottCarda-MS:
Adding the basics of the call graph walker.

### DIFF
--- a/src/QsCompiler/DataStructures/SyntaxTree.fs
+++ b/src/QsCompiler/DataStructures/SyntaxTree.fs
@@ -67,8 +67,8 @@ type QsQualifiedName = {
     /// the declared name of the namespace element
     Name : NonNullable<string>
 }
-    with 
-    override this.ToString () = 
+    with
+    override this.ToString () =
         sprintf "%s.%s" this.Namespace.Value this.Name.Value
 
 
@@ -115,6 +115,7 @@ type QsLocation = {
 
 
 /// used to represent the use of a type parameter within a fully resolved Q# type
+[<CustomComparison; CustomEquality>]
 type QsTypeParameter = {
 // TODO: origin needs adapting if we ever allow to declare type parameters on specializations
 
@@ -126,9 +127,29 @@ type QsTypeParameter = {
     /// -> is Null for auto-generated type information, i.e. in particular for inferred type information
     Range : QsRangeInfo
 }
+    with
+    interface IComparable<QsTypeParameter> with
+        member this.CompareTo other =
+            compare (this.Origin, this.TypeName) (other.Origin, other.TypeName)
+    interface IComparable with
+        member this.CompareTo obj =
+            match obj with
+            | null                        -> 1
+            | :? QsTypeParameter as other -> (this :> IComparable<_>).CompareTo other
+            | _                           -> invalidArg "obj" "not a QsTypeParameter"
+    interface IEquatable<QsTypeParameter> with
+        member this.Equals other =
+            (this.Origin, this.TypeName) = (other.Origin, other.TypeName)
+    override this.Equals obj =
+        match obj with
+        | :? QsTypeParameter as other -> (this :> IEquatable<_>).Equals other
+        | _                           -> false
+    override this.GetHashCode() =
+        hash (this.Origin, this.TypeName)
 
 
 /// used to represent the use of a user defined type within a fully resolved Q# type
+[<CustomComparison; CustomEquality>]
 type UserDefinedType = {
     /// the name of the namespace in which the type is declared
     Namespace : NonNullable<string>
@@ -138,6 +159,25 @@ type UserDefinedType = {
     /// -> is Null for auto-generated type information, i.e. in particular for inferred type information
     Range : QsRangeInfo
 }
+    with
+    interface IComparable<UserDefinedType> with
+        member this.CompareTo other =
+            compare (this.Namespace, this.Name) (other.Namespace, other.Name)
+    interface IComparable with
+        member this.CompareTo obj =
+            match obj with
+            | null                        -> 1
+            | :? UserDefinedType as other -> (this :> IComparable<_>).CompareTo other
+            | _                           -> invalidArg "obj" "not a UserDefinedType"
+    interface IEquatable<UserDefinedType> with
+        member this.Equals other =
+            (this.Namespace, this.Name) = (other.Namespace, other.Name)
+    override this.Equals obj =
+        match obj with
+        | :? UserDefinedType as other -> (this :> IEquatable<_>).Equals other
+        | _                           -> false
+    override this.GetHashCode() =
+        hash (this.Namespace, this.Name)
 
 
 /// Fully resolved operation characteristics used to describe the properties of a Q# callable.
@@ -344,8 +384,8 @@ type InferredExpressionInformation = {
 type TypedExpression = {
     /// the content (kind) of the expression
     Expression : QsExpressionKind<TypedExpression, Identifier, ResolvedType>
-    /// contains all type arguments implicitly or explicitly determined by the expression, 
-    /// i.e. the origin, name and concrete type of all type parameters whose type can either be inferred based on the expression, 
+    /// contains all type arguments implicitly or explicitly determined by the expression,
+    /// i.e. the origin, name and concrete type of all type parameters whose type can either be inferred based on the expression,
     /// or who have explicitly been resolved by provided type arguments
     TypeArguments : ImmutableArray<QsQualifiedName * NonNullable<string> * ResolvedType>
     /// the type of the expression after applying the type arguments
@@ -361,13 +401,13 @@ type TypedExpression = {
 
     /// Contains a dictionary mapping the origin and name of all type parameters whose type can either be inferred based on the expression,
     /// or who have explicitly been resolved by provided type arguments to their concrete type within this expression
-    member this.TypeParameterResolutions = 
+    member this.TypeParameterResolutions =
         this.TypeArguments.ToImmutableDictionary((fun (origin, name, _) -> origin, name), (fun (_,_,t) -> t))
 
     /// Returns true if the expression is a call-like expression, and the arguments contain a missing expression.
     /// Returns false otherwise.
     static member public IsPartialApplication kind =
-        let rec containsMissing ex = 
+        let rec containsMissing ex =
             match ex.Expression with
             | MissingExpr -> true
             | ValueTuple items -> items |> Seq.exists containsMissing
@@ -564,7 +604,7 @@ and QsStatementKind =
 | QsRepeatStatement      of QsRepeatStatement
 | QsConjugation          of QsConjugation
 | QsQubitScope           of QsQubitScope // includes both using and borrowing scopes
-| EmptyStatement 
+| EmptyStatement
 
 
 and QsStatement = {
@@ -645,7 +685,7 @@ type QsSpecialization = {
     /// Contains the location information for the declared specialization.
     /// The position offset represents the position in the source file where the specialization is declared,
     /// and the range contains the range of the corresponding specialization header.
-    /// For auto-generated specializations, the location is set to the location of the parent callable declaration. 
+    /// For auto-generated specializations, the location is set to the location of the parent callable declaration.
     Location : QsNullable<QsLocation>
     /// contains the type arguments for which the implementation is specialized
     TypeArguments : QsNullable<ImmutableArray<ResolvedType>>
--- a/src/QsCompiler/DataStructures/SyntaxTree.fs
+++ b/src/QsCompiler/DataStructures/SyntaxTree.fs
@@ -302,18 +302,6 @@ type ResolvedType = private {
     /// By construction never contains any arity-0 or arity-1 tuple types.
     member this.Resolution = this._TypeKind
 
-    /// Recursively removes all position information for the resolved type and any sub-types,
-    /// setting the Range value to Null.
-    member this.RemovePositionInfo () =
-        match this.Resolution with
-        | QsTypeKind.TypeParameter tp -> { tp with Range = QsNullable<_>.Null } |> QsTypeKind.TypeParameter |> ResolvedType.New
-        | QsTypeKind.UserDefinedType udt -> { udt with Range = QsNullable<_>.Null } |> QsTypeKind.UserDefinedType |> ResolvedType.New
-        | QsTypeKind.TupleType ts -> ts |> Seq.map (fun x -> x.RemovePositionInfo()) |> fun x -> x.ToImmutableArray() |> TupleType |> ResolvedType.New
-        | QsTypeKind.ArrayType b -> b.RemovePositionInfo() |> ArrayType |> ResolvedType.New
-        | QsTypeKind.Function (it, ot) -> (it.RemovePositionInfo(), ot.RemovePositionInfo()) |> QsTypeKind.Function |> ResolvedType.New
-        | QsTypeKind.Operation ((it, ot), fList) -> ((it.RemovePositionInfo(), ot.RemovePositionInfo()), fList) |> QsTypeKind.Operation |> ResolvedType.New
-        | _ -> this
-
     /// Builds a ResolvedType based on a compatible Q# type kind, and replaces the (inaccessible) record constructor.
     /// Replaces an arity-1 tuple by its item type.
     /// Throws an ArgumentException if the given type kind is an empty tuple.
--- a/src/QsCompiler/DataStructures/SyntaxTree.fs
+++ b/src/QsCompiler/DataStructures/SyntaxTree.fs
@@ -115,7 +115,6 @@ type QsLocation = {
 
 
 /// used to represent the use of a type parameter within a fully resolved Q# type
-[<CustomComparison; CustomEquality>]
 type QsTypeParameter = {
 // TODO: origin needs adapting if we ever allow to declare type parameters on specializations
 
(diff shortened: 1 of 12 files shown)


src/QsCompiler/DataStructures/SyntaxTree.fs
```diff
@@ -302,6 +302,18 @@ type ResolvedType = private {
     /// By construction never contains any arity-0 or arity-1 tuple types.
     member this.Resolution = this._TypeKind
 
+    /// Recursively removes all position information for the resolved type and any sub-types,
+    /// setting the Range value to Null.
+    member this.RemovePositionInfo () =
```
bettinaheim: This should not be here. Please use StripPositionInfo.Apply instead.

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -37,6 +38,10 @@ private struct Response
 
         public static QsCompilation Apply(QsCompilation compilation)
         {
+            CallGraph callGraph = BuildCallGraph.Apply(compilation);
+
+            var cycles = callGraph.GetCallCycles();
```
bettinaheim: What's the purpose of having this here? It isn't used by the monomorphization so far.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -0,0 +1,319 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
bettinaheim: Looks like there are still a bunch of todo's left throughout this file. What should I look at exactly, and what isn't ready for review yet?

src/QsCompiler/Transformations/Monomorphization.cs
ScottCarda-MS: removed

src/QsCompiler/Transformations/CallGraphWalker.cs
ScottCarda-MS: I rewrote some of the ToDo's to make them clearer. If there is one or two in particular that you need more explanation, on feel to post a comment 😊.

avasch01:
> // Licensed under the MIT License.

Should you remove three unchanged files for this PR?

---
Refers to: src/QsCompiler/Transformations/Monomorphization.cs:2 in 22c03cd. [](commit_id = 22c03cd7b410a3b6800b45c5981c50fd308f1630, deletion_comment = False)

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -0,0 +1,344 @@
...
+                                callStack.Remove(node);
+                            }
+                        }
+                    }
+                }
+
+                finished.Add(node);
+            }
+
+            foreach (var key in _Dependencies.Keys)
```
avasch01: >var key [](start = 21, length = 7)

A comment here explaining that we are just iterating over all nodes in the graph would be useful for understanding. An even better comment would be to note that in reality we just need to iterate over entry points. Also if "var key" was "CallGraphNode node" I would read it much smoother. Why is it called a key? What is important about it, that it is a key in some dictionary or that it is a node in the graph?

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -0,0 +1,344 @@
...
+        }
+
+        public struct CallGraphNode
+        {
+            public QsQualifiedName CallableName;
+            public QsSpecializationKind Kind;
+            public QsNullable<ImmutableArray<ResolvedType>> TypeArgs;
+        }
+
+        private Dictionary<CallGraphNode, Dictionary<CallGraphNode, ImmutableArray<CallGraphEdge>>> _Dependencies =
```
avasch01: >Dictionary<CallGraphNode, Dictionary<CallGraphNode, ImmutableArray<CallGraphEdge>>> [](start = 16, length = 83)

Do I understand it correctly, that this is a dictionary mapping source node to pairs of target node and information about the edge between source node and target node? Would be great to add a comment stating it. This is a non-trivial place for understanding.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -0,0 +1,344 @@
...
+        private class BuildGraph : SyntaxTreeTransformation<BuildGraph.TransformationState>
+        {
+            public class TransformationState
+            {
+                internal QsSpecialization spec;
+
+                internal bool inCall = false;
+                internal bool adjoint = false;
+                internal bool controlled = false;
+
```
avasch01: Any good reason not to use QsSpecializationKind here?

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -0,0 +1,344 @@
...
+                }
+            }
+
+            private class ExpressionKindTransformation : ExpressionKindTransformation<TransformationState>
+            {
+                public ExpressionKindTransformation(SyntaxTreeTransformation<TransformationState> parent) : base(parent, TransformationOptions.NoRebuild) { }
+
+                private ExpressionKind HandleCall(TypedExpression method, TypedExpression arg)
+                {
+                    var contextInCall = SharedState.inCall;
```
avasch01: >var [](start = 20, length = 3)

Is it bool? Just one more character and readability is greatly improved.

ScottCarda-MS:
> > // Licensed under the MIT License.
> 
> Should you remove three unchanged files for this PR?
> 
> Refers to: src/QsCompiler/Transformations/Monomorphization.cs:2 in [22c03cd](https://github.com/microsoft/qsharp-compiler/commit/22c03cd7b410a3b6800b45c5981c50fd308f1630). [](commit_id = [22c03cd](https://github.com/microsoft/qsharp-compiler/commit/22c03cd7b410a3b6800b45c5981c50fd308f1630), deletion_comment = False)

I had, at various different points, made logic changes to these files, but they were resolved over the coarse of development. When I change a file, I like to make sure there are no trailing whitespaces left in the file. Those are the remaining changes in these files, and I would like to keep them, as they should not have trailing whitespace.

src/QsCompiler/Transformations/CallGraphWalker.cs
ScottCarda-MS: I would have preferred to use explicit types everywhere throughout the code, but the style for this code base is to use var when possible, despite the hit to readability.

src/QsCompiler/Transformations/CallGraphWalker.cs
ScottCarda-MS: With the adjoint, it is easier to flip back and forth a Boolean than to change back and forth a QsSpecializationKind.QsBody to QsSpecializationKind.QsAdjoint or QsSpecializationKind.QsControlled to QsSpecializationKind.QsControlledAdjoint as we encounter adjoint functor applications. It is easier to have all that logic done at the end, when we are not going to encounter any more functor applications. I will rename the fields to make it clearer that these are in reference to functor applications and not what specialization of a callable we are in.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -0,0 +1,349 @@
...
+
+        public struct CallGraphNode
+        {
+            public QsQualifiedName CallableName;
+            public QsSpecializationKind Kind;
+            public QsNullable<ImmutableArray<ResolvedType>> TypeArgs;
+        }
+
+        /// <summary>
+        /// This is a dictionary mapping source nods to information about target nodes. This information is represented
```
avasch01: >nods [](start = 48, length = 4)

nodes

ScottCarda-MS: Thanks!

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -0,0 +1,302 @@
...
+            //                var noTypeParams = new CallGraphNode { CallableName = k.Item1.CallableName, Kind = k.Item1.Kind, TypeArgs = QsNullable<ImmutableArray<ResolvedType>>.Null };
+            //                WalkDependencyTree(noTypeParams, accum, maxDepType);
+            //            }
+            //        }
+            //    }
+            //
+            //    return accum;
+            //}
+            //
+            //return WalkDependencyTree(callerSpec, new HashSet<(CallGraphNode, DependencyType)>(), DependencyType.NoTypeParameters).ToImmutableArray();
```
samarsha, line 145: Does all of the code need to be commented out, or just the part that isn't implemented yet (resolving type parameters)?

ScottCarda-MS: I'd rather keep this around for now as it might help with implementing the method in the future, once we have the proper tools for doing this. It won't go into the master branch like this.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -0,0 +1,302 @@
...
+        }
+    }
+
+    /// <summary>
+    /// This transformation walks through the compilation without changing it, building up a call graph as it does.
+    /// This call graph is then returned to the user.
+    /// </summary>
+    public static class BuildCallGraph
+    {
+        public static CallGraph Apply(QsCompilation compilation)
```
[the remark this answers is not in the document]
ScottCarda-MS: Thanks! Finding a way to work with the access modifiers in this setup can be a bit tricky sometimes. 😊

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -0,0 +1,302 @@
...
+                else
+                {
+                    deps[calledKey] = ImmutableArray.Create(edge);
+                }
+            }
+            else
+            {
+                var newDeps = new Dictionary<CallGraphNode, ImmutableArray<CallGraphEdge>>();
+                newDeps[calledKey] = ImmutableArray.Create(edge);
+                _Dependencies[callerKey] = newDeps;
```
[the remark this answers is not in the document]
ScottCarda-MS: No thanks. 😊 I think keeping this as multiple lines really helps with readability on this one.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -0,0 +1,302 @@
...
+            {
+                walker.Namespaces.OnNamespace(ns);
+            }
+
+            return walker.SharedState.graph;
+        }
+
+        private class BuildGraph : SyntaxTreeTransformation<BuildGraph.TransformationState>
+        {
+            public class TransformationState
```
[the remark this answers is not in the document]
ScottCarda-MS: I can apply this change for now, but I suspect the structure is going to get changed a bit later on.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -0,0 +1,302 @@
...
+        {
+            public ImmutableDictionary<Tuple<QsQualifiedName, NonNullable<string>>, ResolvedType> ParamResolutions;
+        }
+
+        public struct CallGraphNode
+        {
+            public QsQualifiedName CallableName;
+            public QsSpecializationKind Kind;
+            public QsNullable<ImmutableArray<ResolvedType>> TypeArgs;
+        }
```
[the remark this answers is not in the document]
ScottCarda-MS: 🤔 Hmm, not sure about this one. I like these structs being tied to the CallGraph class in this way, and having outside code needing to use the qualified names as a reminder of what these are used for. In the future, these may even be private to the class. The structure of this whole thing will probably get changed at a later point.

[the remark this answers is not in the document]
samarsha: The [coding guidelines for .NET](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/nested-types) recommend avoiding nested types unless they are needed for access control (i.e. making a type visible only to a single class). Public nested types are discouraged (there's a gray squiggly in Visual Studio when you use them) because namespaces should be used for grouping instead of classes.

I think that requiring users of the class to write `CallGraph.CallGraphEdge` and `CallGraph.CallGraphNode` (or a static import) is unnecessary; even with `CallGraph.Edge` and `CallGraph.Node`, it's still unusual to need to use a dot in the type name (or a static import).

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -0,0 +1,291 @@
...
+
+namespace Microsoft.Quantum.QsCompiler.Transformations.CallGraphWalker
+{
+    using ExpressionKind = QsExpressionKind<TypedExpression, Identifier, ResolvedType>;
+    using ResolvedTypeKind = QsTypeKind<ResolvedType, UserDefinedType, QsTypeParameter, CallableInformation>;
+
+    /// <summary>
+    /// Struct containing information that exists on edges in a call graph.
+    /// </summary>
+    public struct CallGraphEdge
```
samarsha, line 22: I took a look at how the call graph is being used in #395. It looks like there are unused methods (`GetDirectDependencies` and `GetAllDependencies`). The rest of the structs, classes and methods look like they can be internal instead of public, since they are only used by the tests, and possibly might only be used by other transformations in this project (e.g. monomorphization) in the future?

ScottCarda-MS: Getting dependencies is a required feature of the call graph and will be needed to interact with the call graph. It is too early to say definitively what the access modifiers should be, but I think all of the ones listed as public are supposed to be public, as in we want to make them available to be used by 3rd party developers.

samarsha: I think that, depending on how the graph is used, getting specific dependencies/edges in the graph is not necessarily a requirement. For example, there are higher-level methods that can be used instead, like `GetCallCycles`. Do you have an example of when the public methods need to be used by a third party, or by other projects in the compiler/QDK?

Since there is uncertainty in what the access modifiers should be, I recommend making them internal until there is a need for them to be public.

ScottCarda-MS: Leaving the access modifiers as public for now.

samarsha: If they stay public, I would recommend adding TODO comments as a reminder to check the access modifiers later and make them more restrictive if possible. (It would be fine with me to have either a TODO comment before each method or one TODO at the top of the file.) Without changing the access modifiers or adding some reminder to the code, I'm not sure the issue is resolved from my point of view.

ScottCarda-MS: Added a ToDo for the file for access modifiers.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -0,0 +1,394 @@
...
+
+        /// <summary>
+        /// Adds a dependency to the call graph using the relevant information from the caller's specialization and the called specialization.
+        /// </summary>
+        public void AddDependency(
+            QsQualifiedName callerName, QsSpecializationKind callerKind, QsNullable<ImmutableArray<ResolvedType>> callerTypeArgs,
+            QsQualifiedName calledName, QsSpecializationKind calledKind, QsNullable<ImmutableArray<ResolvedType>> calledTypeArgs,
+            CallGraphEdge edge)
+        {
+            // ToDo: Setting TypeArgs to Null because the type specialization is not implemented yet
```
bettinaheim, line 188: Would you mind using the specific phrasing below sucht that it's easy to find these places in the code?
```suggestion
            // Setting TypeArgs to Null because the type specialization is not implemented yet
            // ToDo: this needs adaption if we want to support type specializations
```

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -0,0 +1,394 @@
...
+
+        private class BuildGraph : SyntaxTreeTransformation<TransformationState>
+        {
+            public BuildGraph() : base(new TransformationState())
+            {
+                this.Namespaces = new NamespaceTransformation(this);
+                this.Statements = new StatementTransformation<TransformationState>(this, TransformationOptions.NoRebuild);
+                this.StatementKinds = new StatementKindTransformation<TransformationState>(this, TransformationOptions.NoRebuild);
+                this.Expressions = new ExpressionTransformation<TransformationState>(this, TransformationOptions.NoRebuild);
+                this.ExpressionKinds = new ExpressionKindTransformation(this);
```
bettinaheim, line 287:
```suggestion
                this.ExpressionKinds = new ExpressionKindTransformation(this, TransformationOptions.NoRebuild);
```

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -0,0 +1,394 @@
...
+                walker.Namespaces.OnNamespace(ns);
+            }
+            return walker.SharedState.graph;
+        }
+
+        private class BuildGraph : SyntaxTreeTransformation<TransformationState>
+        {
+            public BuildGraph() : base(new TransformationState())
+            {
+                this.Namespaces = new NamespaceTransformation(this);
```
bettinaheim, line 283:
```suggestion
                this.Namespaces = new NamespaceTransformation(this, TransformationOptions.NoRebuild);
```

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -0,0 +1,394 @@
...
+            {
+                var contextInCall = SharedState.inCall;
+                SharedState.inCall = true;
+                this.Expressions.OnTypedExpression(method);
+                SharedState.inCall = contextInCall;
+                this.Expressions.OnTypedExpression(arg);
+                return ExpressionKind.InvalidExpr;
+            }
+
+            public override ExpressionKind OnOperationCall(TypedExpression method, TypedExpression arg) => HandleCall(method, arg);
```
bettinaheim, line 328: You would also need the same thing for partial applications. It hence makes more sense to just override OnCallLikeExpression instead of OnOperationCall and OnFunctionCall.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -0,0 +1,394 @@
...
+                var result = base.OnControlledApplication(ex);
+                SharedState.hasControlledDependency = contextControlled;
+                return result;
+            }
+
+            public override ExpressionKind OnIdentifier(Identifier sym, QsNullable<ImmutableArray<ResolvedType>> tArgs)
+            {
+                if (sym is Identifier.GlobalCallable global)
+                {
+                    // ToDo: Type arguments need to be resolved for the whole expression to be accurate, though this will not be needed until type specialization is implemented
```
bettinaheim, line 353:
```suggestion
                    // Type arguments need to be resolved for the whole expression to be accurate
                    // ToDo: this needs adaption if we want to support type specializations
```

src/QsCompiler/Transformations/CallGraphWalker.cs
ScottCarda-MS: Expression Kind and Namespace sub-transformations don't take another parameter. This 'No Rebuild' behavior is hard-coded into them because they are specific to this walker.

src/QsCompiler/Transformations/CallGraphWalker.cs
ScottCarda-MS: Expression Kind and Namespace sub-transformations don't take another parameter. This 'No Rebuild' behavior is hard-coded into them because they are specific to this walker.

PR #395, opened 2020-03-31
Cycle Tests

ScottCarda-MS:
Tests an algorithm for finding cycles in call graph. The current algorithm in this PR does not meet the requirements, and a more complete algorithm will need to replace it in the future.


src/QsCompiler/Tests.Compiler/TestUtils/Signatures.fs
```diff
@@ -114,6 +114,7 @@ let public GenericsNs = "Microsoft.Quantum.Testing.Generics"
 let public IntrinsicResolutionNs = "Microsoft.Quantum.Testing.IntrinsicResolution"
 let public ClassicalControlNs = "Microsoft.Quantum.Testing.ClassicalControl"
 let public InternalRenamingNs = "Microsoft.Quantum.Testing.InternalRenaming"
+let public CycleDetection = "Microsoft.Quantum.Testing.CycleDetection"
```
samarsha, line 117: Just for consistency:
```suggestion
let public CycleDetectionNs = "Microsoft.Quantum.Testing.CycleDetection"
```

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -255,8 +256,57 @@ private void RecordDependency(CallGraphNode callerKey, CallGraphNode calledKey,
...
         {
-            // ToDo: need to implement a robust algorithm for finding cycles
-            throw new NotImplementedException();
+            var callStack = new Dictionary<CallGraphNode, CallGraphNode>();
+            var finished = new HashSet<CallGraphNode>();
+            var cycles = new List<ImmutableArray<CallGraphNode>>();
+
+            void processDependencies(CallGraphNode node)
+            {
+                if (_Dependencies.TryGetValue(node, out var dependencies))
```
samarsha, line 265: If the node is not in the graph, does this indicate an error, or is it that library functions aren't tracked in the graph, so it's valid for nodes to call other nodes that don't exist?

samarsha, line 265: To decrease nesting, it might be better to do something like
```
if (!_Dependencies.TryGetValue(node, out var dependencies))
{
    finished.Add(node);
    return;
}

// ...
finished.Add(node);

File: src/QsCompiler/Transformations/CallGraphWalker.cs [hunk 3]
```diff
@@ -255,8 +256,57 @@ private void RecordDependency(CallGraphNode callerKey, CallGraphNode calledKey,
...
+                {
+                    foreach (var temp in dependencies)
+                    {
+                        var (curr, _) = temp;
```
samarsha, line 269:
```suggestion
                    foreach (var curr in dependencies.Keys)
                    {
```
but this means `curr` cannot be reassigned, so the "Cycle detected" section would need to be changed also.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -255,8 +256,57 @@ private void RecordDependency(CallGraphNode callerKey, CallGraphNode calledKey,
...
+                            {
+                                // Cycle detected
+
+                                var cycle = new List<CallGraphNode>() { curr };
+                                while (callStack.TryGetValue(curr, out var next))
+                                {
+                                    if (curr.Equals(next)) break;
+                                    cycle.Add(next);
+                                    curr = next;
+                                }
```
samarsha, line 283: Consider splitting this out into a separate (probably local) function like `List<CallGraphNode> getCallCycle(CallGraphNode node)`.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -255,8 +256,57 @@ private void RecordDependency(CallGraphNode callerKey, CallGraphNode calledKey,
...
+
+            void processDependencies(CallGraphNode node)
+            {
+                if (_Dependencies.TryGetValue(node, out var dependencies))
+                {
+                    foreach (var temp in dependencies)
+                    {
+                        var (curr, _) = temp;
+
+                        if (!finished.Contains(curr))
```
samarsha, line 271: Consider decreasing nesting:
```suggestion
                        if (finished.Contains(curr))
                        {
                            continue;
                        }
                        
                        // ...
```

src/QsCompiler/Transformations/CallGraphWalker.cs
ScottCarda-MS: Yah I initially had the tuple deconstruction in the foreach, but it also did not let me update those later. So I just went with a variable that I immediately deconstruct into the value that I actually want.

ScottCarda-MS: Kinda dumb that .NET does not let me do this without a dumb looking extra line like this.

src/QsCompiler/Transformations/CallGraphWalker.cs
ScottCarda-MS: All the nodes have to be in the graph, or this will error. I added logic to make sure that all the nodes are in the graph.

src/QsCompiler/Transformations/CallGraphWalker.cs
samarsha: If you factor out the code that is reassigning `curr` (which I think would be good - I made a comment about that), you can use `Keys`.

src/QsCompiler/Transformations/CallGraphWalker.cs
ScottCarda-MS: The nesting isn't too bad here, and I prefer to have the logic that makes use of the out variable be in the conditional block. It makes the logic cleaner and more encapsulated.

src/QsCompiler/Transformations/CallGraphWalker.cs
ScottCarda-MS: It's not much code, and local functions can be confusing and clutter up the code.

samarsha: It doesn't have to be a local function. It can be a private method. :) IMO, having several small, well-named functions that are composed together enhances readability compared to one large function, especially since this section of code reassigns `curr`; the mutation of `curr` can be isolated to a function that is a few lines long, which makes it easier to reason about.

src/QsCompiler/Transformations/CallGraphWalker.cs
ScottCarda-MS: I renamed temp to make it feel less hacky. I think the one line solution here is simpler than adding a local function.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -255,8 +256,56 @@ private void RecordDependency(CallGraphNode callerKey, CallGraphNode calledKey,
...
+            var cycles = new List<ImmutableArray<CallGraphNode>>();
+
+            void processDependencies(CallGraphNode node)
+            {
+                if (_Dependencies.TryGetValue(node, out var dependencies))
+                {
+                    foreach (var dependency in dependencies)
+                    {
+                        var (curr, _) = dependency;
+
```
cesarzc, line 270: Style: In general, I have seen that lines are only left blank before a comment or after a block of code enclosed in {} (eg. if statements, cycles, methods). In order to have a consistent and easier to read style across the code base, please consider removing blank lines that do not fit the description that I mentioned.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -255,8 +256,56 @@ private void RecordDependency(CallGraphNode callerKey, CallGraphNode calledKey,
...
+            var finished = new HashSet<CallGraphNode>();
+            var cycles = new List<ImmutableArray<CallGraphNode>>();
+
+            void processDependencies(CallGraphNode node)
+            {
+                if (_Dependencies.TryGetValue(node, out var dependencies))
+                {
+                    foreach (var dependency in dependencies)
+                    {
+                        var (curr, _) = dependency;
```
cesarzc, line 269: A small suggestion regarding the variable name "curr". It is not vary clear what that variable represents, so maybe renaming it to "dependencyNode" might better convey what the variable represents.

Since the "dependency" word is used a lot in this code fragment, maybe the "dependency" variable could be named "entry"?

src/QsCompiler/Transformations/CallGraphWalker.cs
samarsha: Sorry, I don't see the added logic?

src/QsCompiler/Transformations/CallGraphWalker.cs
samarsha: I don't think there is any rule against blank lines; it is used commonly to separate sections of code, see [here](https://github.com/microsoft/qsharp-compiler/blob/master/src/QsCompiler/CompilationManager/CompilationUnit.cs#L715) and [here](https://github.com/microsoft/qsharp-compiler/blob/master/src/QsCompiler/Core/SymbolTable.fs#L1224) for example.

src/QsCompiler/Transformations/CallGraphWalker.cs
samarsha: In this case, we know that `curr` is related to a dependency because it was extracted from a variable called `dependency`, so I think that `dependencyNode` might be too long for a local variable name. Maybe something like `target` would work, since it is the target of a dependency edge in the graph.

ScottCarda-MS: `curr`, as well as `prev` and `next` are commonly used identifiers in algorithms that walk through nodes of a data structure. I was going to rename it to `target` as per Sarah's suggestion, but after thinking about it, that seems less clear to me given this precedent.

src/QsCompiler/Transformations/CallGraphWalker.cs
ScottCarda-MS: Sorry, looks like that didn't come into play until later: https://github.com/microsoft/qsharp-compiler/blob/sccarda/JohnsonsAlgorithm/src/QsCompiler/Transformations/CallGraphWalker.cs#L405

ScottCarda-MS:
I am going to short-circuit the discussions surrounding the `GetCallCycles` implementation as it is being replaced with Johnson's Algorithm in a PR that will be merged in shortly.

src/QsCompiler/Transformations/CallGraphWalker.cs
cesarzc: I think I wrote my comment in a way that sounded like a rule, sorry about that. Regarding blank lines, my recommendation in general is to be consistent with its use (e.g. do not add a blank space after a comment if in general blank lines are not added in that case), and try to not over use blank lines because I think it makes the code a bit less readable if there are a lot of them (however this can be subjective so maybe some people find the code more readable).

I don't think there is a need for a rule on blank lines but my main point is to be consistent with their use :)

PR #396, opened 2020-03-31
Implement Johnson's Algorithm for Finding All Simple Cycles

ScottCarda-MS:
Johnson's algorithm found here:
https://www.cs.tufts.edu/comp/150GA/homeworks/hw1/Johnson%2075.PDF

Tarjan's algorithm found here:
https://en.wikipedia.org/wiki/Tarjan%27s_strongly_connected_components_algorithm


src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -17,6 +16,258 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.CallGraphWalker
...
+        /// This implementation was based on the pseudo-code found here:
+        /// https://en.wikipedia.org/wiki/Tarjan%27s_strongly_connected_components_algorithm
+        ///
+        /// This returns a list of SCCs represented by sets of nodes, each paired with the
+        /// value of their lowest valued-node. The list is sorted such that each SCC comes
+        /// before any of its children (reverse topological ordering).
+        /// </summary>
+        private List<(HashSet<int> SCC, int MinNode)> TarjanSCC(Dictionary<int, List<int>> inputGraph)
+        {
+            var index = 0; // This algorithm needs its own separate indexing of nodes
```
irinayat, line 112: This comment isn't very useful. I'd say something along the lines, that the index represents the order in which nodes are discovered by the algorithm.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -17,6 +16,258 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.CallGraphWalker
...
+            foreach (var scc in sccs)
+            {
+                SccStack.Push(scc);
+            }
+        }
+
+        private Dictionary<int, List<int>> GetSubGraph(Dictionary<int, List<int>> inputGraph, HashSet<int> subGraphNodes) =>
+            inputGraph
+                .Where(kvp => subGraphNodes.Contains(kvp.Key))
+                .ToDictionary(kvp => kvp.Key, kvp => kvp.Value.Where(dep => subGraphNodes.Contains(dep)).ToList());
```
irinayat, line 96: Is it helpful to define an SCC as a hashset? If it was a list, we could traverse it linearly and have the same O(1) get from the inputGraph to create a new dictionary. Having it as a hashset is likely wasteful (most SCCs will contain a single element, anyway).

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -17,6 +16,258 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.CallGraphWalker
...
+        /// Johnson's algorithm for finding all cycles in a graph.
+        /// This uses Tarjan's algorithm for finding strongly-connected components, or SCCs,
+        /// of the graph and processes those SCCs individually for finding cycles.
+        ///
+        /// This returns a list of cycles, each represented as a list of nodes. Nodes
+        /// for this algorithm are integers. The cycles are guaranteed to not contain
+        /// any duplicates and the last node in the list is assumed to be connected
+        /// to the first.
+        /// </summary>
+        public List<List<int>> GetAllCycles(Dictionary<int, List<int>> graph)
```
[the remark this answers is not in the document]
ScottCarda-MS: The purpose is to find an return all the cycles in the call graph. Cycles are generally okay, as this would indicate recursion, which is allowed. However we need to analyze the cycles for certain qualities involving the type arguments used to in the call cycle. This code for this analysis is being developed separately by @bettinaheim .

src/QsCompiler/Transformations/CallGraphWalker.cs
ScottCarda-MS: Maybe I am misunderstanding your comment. I am currently under the impression that a HashSet and a List have similar size complexities, and that checking for an item in a List was O(n). If this is not the case, do you have any resources I could look at to better understand this?

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -17,6 +16,258 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.CallGraphWalker
     using ExpressionKind = QsExpressionKind<TypedExpression, Identifier, ResolvedType>;
     using ResolvedTypeKind = QsTypeKind<ResolvedType, UserDefinedType, QsTypeParameter, CallableInformation>;
 
+    /// <summary>
+    /// Class for using Johnson's algorithm for finding all simple cycles in a graph.
```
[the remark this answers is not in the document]
ScottCarda-MS: Removed the stuff about Tarjan's algorithm from `GetAllCycles`.

src/QsCompiler/Transformations/CallGraphWalker.cs
ScottCarda-MS: When writing this, I though I could be clever about the indexing and use the node's ids, but I discovered that will not work. I wanted to put something here to that effect, in case someone else had the same idea.

src/QsCompiler/Transformations/CallGraphWalker.cs
irinayat-MS: Dictionaries, sets, etc. usually have worse locality compared to array-based structures. I'd also expect them to have more memory overhead per structure. Basically, yes, it's O(1) for iterating over both but consts might be considerably different.

src/QsCompiler/Transformations/CallGraphWalker.cs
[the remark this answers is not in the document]
irinayat-MS: I'm just curious, could you give an example of a problematic cycle? Maybe, link to the issue# that has lead this work?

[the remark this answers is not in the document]
irinayat-MS: Also, if most cycles aren't going to be problematic, enumerating all of them would be wasteful.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -17,6 +16,256 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.CallGraphWalker
     using ExpressionKind = QsExpressionKind<TypedExpression, Identifier, ResolvedType>;
     using ResolvedTypeKind = QsTypeKind<ResolvedType, UserDefinedType, QsTypeParameter, CallableInformation>;
 
+    /// <summary>
+    /// Implementation of Johnson's algorithm for finding all simple cycles in a graph.
```
irinayat-MS, line 20: I'd also add a comment about the expected complexity of the algorithm.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -17,6 +16,256 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.CallGraphWalker
...
+            }
+
+            return SCCs;
+        }
+
+        /// <summary>
+        /// Johnson's algorithm for finding all cycles in an SCC, or strongly-connected component.
+        ///
+        /// It will process an individual SCC for cycles by looking at each of their nodes
+        /// starting with the minimum-valued node of the SCC as the current node. Children of
```
irinayat-MS, line 192: "minimum-valued" isn't very clear here. As far as I understand the SCC algorithm, it's the first node we discover in a component _given_ the specific traversal order of nodes in the input graph. Why starting from it is important? This SCC subgraph is a new dictionary and the traversal order of nodes in it doesn't necessarily match the traversal order in the overall input graph.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -17,6 +16,256 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.CallGraphWalker
...
+                var foundCycle = false;
+                nodeStack.Push(currNode);
+                blockedSet.Add(currNode);
+
+                foreach (var child in intputSCC[currNode])
+                {
+                    if (child == startNode)
+                    {
+                        foundCycle = true;
+                        cycles.Add(nodeStack.Reverse().ToList());
```
irinayat-MS, line 229: Why is it necessary to reverse the stack?

src/QsCompiler/Tests.Compiler/TestCases/CycleDetection.qs
```diff
@@ -112,4 +112,75 @@ namespace Microsoft.Quantum.Testing.CycleDetection {
         Bar();
         Baz();
     }
+}
+
+// =================================
+
+// Johnson's Graph Cycles
+namespace Microsoft.Quantum.Testing.CycleDetection {
```
irinayat-MS, line 120: With a single complicated test like this is hard to analyse the coverage and corner cases of the algorithm. While it's worth keeping it, I'd add a few smaller test cases that would also illustrate to the reader some of the concerns the algorithm has to address. For example:
1. no cycles
2. two independent elementary cycles in different SCCs
3. function that calls itself
4. two cycles that have exactly one vertex in common
5. two cycles that have two vertices in common (thus producing 4 different elementary cycles)
6. "sausage link" graph {a->b, b->a; b->c, c->b; c->d, d->c} -- in this example only three elementary cycles

ScottCarda-MS: There are other tests in the file that meet many of your examples. This one is called the Johnson's Graph because it is an example that came from the paper. I like the idea of the "sausage link" test and will add that. I will also add something for your fifth case.

src/QsCompiler/Transformations/CallGraphWalker.cs
ScottCarda-MS: The cycles will be backwards if I don't. I'm not sure if there was something else to be done here.

src/QsCompiler/Transformations/CallGraphWalker.cs
ScottCarda-MS: It seemed to me, from the paper it is just a place to start. I'm not sure if there is any importance beyond that.

src/QsCompiler/Transformations/CallGraphWalker.cs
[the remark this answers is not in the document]
ScottCarda-MS: I'm not sure the issue is written down anywhere in any detail. If you want to know more, the best way would probably be to have a quick call with me or Bettina or both.

src/QsCompiler/Transformations/CallGraphWalker.cs
ScottCarda-MS: I am not following. Let's discuss your suggestion over a call.

src/QsCompiler/Tests.Compiler/TestCases/CycleDetection.qs
irinayat-MS: I'd still suggest to add a dedicated test for the case of two separate SCCs with a cycle in each (it's probably covered in the mega-test but I find it hard to analyze that one)

src/QsCompiler/Transformations/CallGraphWalker.cs
irinayat-MS: That was my impression from the paper as well. Maybe, remove 'minimum-valued' from the comment to decrease confusion?

src/QsCompiler/Transformations/CallGraphWalker.cs
ScottCarda-MS: Resolved offline: Because we need to do filtering on both the keys of the input graph, and the adjacency list values, it is easier to have the subgraph nodes structure be a hash set for doing `.Contains` operations.

src/QsCompiler/Transformations/CallGraphWalker.cs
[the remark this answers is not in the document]
ScottCarda-MS: Resolved offline: we want to go for a general solution that can be used at a later time as a template for a more specific solution that may be able to take advantage of scenario-specific optimizations.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -17,6 +16,259 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.CallGraphWalker
...
+        private void PushSCCs(Dictionary<int, List<int>> graph)
+        {
+            var sccs = TarjanSCC(graph).OrderByDescending(x => x.MinNode);
+            foreach (var scc in sccs)
+            {
+                SccStack.Push(scc);
+            }
+        }
+
+        private Dictionary<int, List<int>> GetSubGraph(Dictionary<int, List<int>> inputGraph, HashSet<int> subGraphNodes) =>
```
avasch01, line 93: >GetSubGraph [](start = 43, length = 11)

A comment here would really help me understand what is happening. What kind of subgraph are we trying to get? It would also help to have a more descriptive name for this function. I don't care how long it will become - it is only used once in the code, but it would help understanding.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -17,6 +16,259 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.CallGraphWalker
...
+    /// until there are no more SCCs (and no more nodes) left in the graph.
+    ///
+    /// The algorithm has time complexity of O((n + e)(c + 1)) with n nodes, e edges,
+    /// and c elementary cycles.
+    ///
+    /// The starting point for this algorithm in this class is the GetAllCycles method.
+    /// </summary>
+    internal class JohnsonCycleFind
+    {
+        private Stack<(HashSet<int> SCC, int MinNode)> SccStack = new Stack<(HashSet<int> SCC, int MinNode)>();
```
avasch01, line 41: >SccStack [](start = 55, length = 8)

Why is this a stack? Are you really using the properties of a stack here, or it could be a simple array? This is a question, not a request for a change yet.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -17,6 +16,259 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.CallGraphWalker
...
+        {
+            // Possible Optimization: Each SCC could be processed in parallel.
+
+            // Possible Optimization: Getting the subgraph for an SCC only needs
+            // to consider the parent graph, not the full graph. A recursive strategy
+            // may be better than using a stack object.
+
+            var cycles = new List<List<int>>();
+
+            PushSCCs(graph);
```
avasch01, line 60: >graph [](start = 21, length = 5)

Is this graph a strongly connected component?

src/QsCompiler/Transformations/CallGraphWalker.cs
ScottCarda-MS: We are always only pushing and popping off of this data structure, in a manor indicative of a stack. Using an array would be more messy here, and not communicate as clearly the use of the data structure.

src/QsCompiler/Transformations/CallGraphWalker.cs
ScottCarda-MS: Not necessarily. The method gets all the SCCs of the given graph and pushes those to the stack. I renamed the method to try to make that clearer.

src/QsCompiler/Transformations/CallGraphWalker.cs
ScottCarda-MS: I added a doc comment, renamed the method, and added some comments to the code, but I'm not sure it is any clearer.

avasch01: The comment definitely helps. Thanks. A better name (as I understand now) would be GetSubGraphLimitedToNodes.

---
In reply to: [405006096](https://github.com/microsoft/qsharp-compiler/pull/396#discussion_r405006096) [](ancestors = 405006096)

src/QsCompiler/Transformations/CallGraphWalker.cs
avasch01: Johnson algorithm requires processing components in the order of the smallest-numbered vertex. Stack structure (as used in this implementation) does not guarantee that.

---
In reply to: [404521809](https://github.com/microsoft/qsharp-compiler/pull/396#discussion_r404521809) [](ancestors = 404521809)

ScottCarda-MS: Since there cannot be any cycles between SCCs, there should not be any reason that we could not process all of one SCC before moving on to the next SCC. Nodes within an SCC are still processed by their smallest numbered node first.

ScottCarda-MS: Oh but you were referring to the order of the SCCs, not the nodes. In the Johnson's algorithm the ordering is important because it keeps track of which node to process with a global index. When they finish with a node, it is removed from the graph and the index in incremented. This implementation does not use a global index to keep track of the what node to process next. That information is provided with each SCC from Tarjan's algorithm. It can return the minimum node in each SCC with no extra time cost because it keeps track of that while processing each node.

I am taking advantage of the fact that this implementation of Tarjan's is used specifically for Johnson's, so I can tailor the output of Tarjan's to help with Johnson's. There is a minimal size increase in carrying the integer minimum for each SCC, but this gives us the possible optimization of processing the SCCs in parallel.

avasch01: Agree. returning minimums from Tarjan is a good idea. That one number is not a huge overhead, and is worth it.

---
In reply to: [405306351](https://github.com/microsoft/qsharp-compiler/pull/396#discussion_r405306351) [](ancestors = 405306351)

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -19,6 +18,243 @@ namespace Microsoft.Quantum.QsCompiler.Transformations
...
+        /// This returns a list of cycles, each represented as a list of nodes. Nodes
+        /// for this algorithm are integers. The cycles are guaranteed to not contain
+        /// any duplicates and the last node in the list is assumed to be connected
+        /// to the first.
+        /// </summary>
+        
+        public List<List<int>> GetAllCycles(Dictionary<int, List<int>> graph) =>
+            TarjanSCC(graph)
+                .OrderByDescending(x => x.MinNode)
+                .SelectMany(item =>
```
avasch01, line 56: >SelectMany [](start = 17, length = 10)

As per our email discussion, I would favor the old non-recursive approach.

ScottCarda-MS: Moved to using the 'filter' implementation we discussed.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -19,6 +18,254 @@ namespace Microsoft.Quantum.QsCompiler.Transformations
...
+        }
+
+        /// <summary>
+        /// Uses Tarjan's algorithm for finding strongly-connected components, or SCCs of a
+        /// graph to get all SCC, orders them by their node with the smallest id, and pushes
+        /// them to the SCC stack.
+        /// </summary>
+        private void PushSCCsFromGraph(Dictionary<int, List<int>> graph, HashSet<int> filter)
+        {
+            var sccs = TarjanSCC(graph, filter).OrderByDescending(x => x.MinNode);
```
avasch01, line 79: >var sccs [](start = 12, length = 8)

Please consider using explicit type here as per our recently adopted style guide: "Do not use var when the type is not apparent from the right side of the assignment." See [https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions)

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -19,6 +18,254 @@ namespace Microsoft.Quantum.QsCompiler.Transformations
...
+            foreach (var scc in sccs)
```
avasch01, line 80: >var scc [](start = 21, length = 7)

Please consider using explicit type here as per our recently adopted style guide: "Do not use implicit typing to determine the type of the loop variable in foreach loops."

Once you replace this with explicit type, you will notice a peculiar thing: the scc variable in this function and scc variable in the previous function mean two different things (different types). This is hard to see when everything is vars rather than explicit types. This is one reason vars can lead to bugs or misunderstandings.

I recommend following the stylequide for explicit types throughout this new file.

ScottCarda-MS: I will need to talk with @bettinaheim on how our team's style will be changed by these new style guidelines. I would also like to make this change across the whole feature branch, rather than just on this subset, so I will make a separate task for that and defer doing this to a later PR.

src/QsCompiler/Transformations/CallGraphWalker.cs
ScottCarda-MS: Addressed in comment below.

PR #397, opened 2020-04-03
General tool for combining and verifying type resolution dictionaries

bettinaheim:
Currently, the implementation requires that for a cycle, the verification is invoked for each node. I'll see if I can find some time to be more clever about it. As a first thing to work with, this should do however, and also gives a good base line to test.

### DIFF
--- a/src/QsCompiler/Tests.Compiler/CallGraphTests.fs
+++ b/src/QsCompiler/Tests.Compiler/CallGraphTests.fs
@@ -51,11 +51,11 @@ type CallGraphTests (output:ITestOutputHelper) =
 
     static member AssertResolution (parent, expected, [<ParamArray>] resolutions) = 
         let success = CallGraphTests.CheckResolution (parent, expected, resolutions)
-        Assert.True(success, "overall status indicated as not successful")
+        Assert.True(success, "Combining type resolutions was not successful.")
 
     static member AssertResolutionFailure (parent, expected, [<ParamArray>] resolutions) = 
         let success = CallGraphTests.CheckResolution (parent, expected, resolutions)
-        Assert.False(success, "overall status indicated as success")
+        Assert.False(success, "Combining type resolutions should have failed.")
 
 
     [<Fact>]
--- a/src/QsCompiler/Tests.Compiler/CallGraphTests.fs
+++ b/src/QsCompiler/Tests.Compiler/CallGraphTests.fs
@@ -101,6 +101,21 @@ type CallGraphTests (output:ITestOutputHelper) =
         CallGraphTests.AssertResolution(Foo, expected, res1, res2)
 
         let res1 = resolution [
+            (FooA, BarA |> TypeParameter)
+        ]
+        let res2 = resolution [
+            (FooB, BarA |> TypeParameter)
+        ]
+        let res3 = resolution [
+            (BarA, Int)
+        ]
+        let expected = resolution [
+            (FooA, Int)
+            (FooB, Int)
+        ]
+        CallGraphTests.AssertResolution(Foo, expected, res1, res2, res3)
+
+        let res1 = resolution [
             (FooA, FooA |> TypeParameter)
         ]
         let expected = resolution [
@@ -150,3 +165,56 @@ type CallGraphTests (output:ITestOutputHelper) =
             (FooA, Double)
         ]
         CallGraphTests.AssertResolutionFailure(Foo, expected, res1, res2)
+
+        let res1 = resolution [
+            (FooA, BarA |> TypeParameter)
+        ]
+        let res2 = resolution [
+            (BarA, FooB |> TypeParameter)
+        ]
+        let expected = resolution [
+            (FooA, FooB |> TypeParameter)
+        ]
+        CallGraphTests.AssertResolutionFailure(Foo, expected, res1, res2)
+
+        let res1 = resolution [
+            (FooA, BarA |> TypeParameter)
+        ]
+        let res2 = resolution [
+            (BarA, BazA |> TypeParameter)
+        ]
+        let res3 = resolution [
+            (BazA, FooB |> TypeParameter)
+        ]
+        let expected = resolution [
+            (FooA, FooB |> TypeParameter)
+        ]
+        CallGraphTests.AssertResolutionFailure(Foo, expected, res1, res2, res3)
+
+        let res1 = resolution [
+            (FooA, BarA |> TypeParameter)
+        ]
+        let res2 = resolution [
+            (BarA, BazA |> TypeParameter)
+        ]
+        let res3 = resolution [
+            (BazA, FooA |> TypeParameter)
+        ]
+        let expected = resolution [
+            (FooA, FooA |> TypeParameter)
+        ]
+        CallGraphTests.AssertResolution(Foo, expected, res1, res2, res3)
+
+        let res1 = resolution [
+            (FooA, BarA |> TypeParameter)
+        ]
+        let res2 = resolution [
+            (BarA, BazA |> TypeParameter)
+        ]
+        let res3 = resolution [
+            (BazA, BarC |> TypeParameter) // FIXME: THIS SHOULD NOT BE OK
+        ]
+        let expected = resolution [
+            (FooA, BarC |> TypeParameter)
+        ]
+        CallGraphTests.AssertResolution(Foo, expected, res1, res2, res3)
--- a/src/QsCompiler/Tests.Compiler/CallGraphTests.fs
+++ b/src/QsCompiler/Tests.Compiler/CallGraphTests.fs
@@ -1,0 +1,65 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.QsCompiler.Testing
+
+open System.Collections.Generic
+open System.Collections.Immutable
+open Microsoft.Quantum.QsCompiler.DataTypes
+open Microsoft.Quantum.QsCompiler.SyntaxExtensions
+open Microsoft.Quantum.QsCompiler.SyntaxTokens
+open Microsoft.Quantum.QsCompiler.SyntaxTree
+open Microsoft.Quantum.QsCompiler.Transformations
+open Xunit
+open Xunit.Abstractions
+
+
+type CallGraphTests (output:ITestOutputHelper) =
+
+    let qualifiedName name = 
+        ("NS" |> NonNullable<string>.New, name |> NonNullable<string>.New) |> QsQualifiedName.New
+
+    let typeParameter (id : string) = 
+        let pieces = id.Split(".")
+        Assert.True(pieces.Length = 2)
+        let parent = qualifiedName pieces.[0]
+        let name = pieces.[1] |> NonNullable<string>.New
+        QsTypeParameter.New (parent, name, Null)
+    
+    let resolution (res : (QsTypeParameter * QsTypeKind<_,_,_,_>) list) =
+        res.ToImmutableDictionary((fun (tp,_) -> tp.Origin, tp.TypeName), snd >> ResolvedType.New)
+
+    let compareDict (d1 : ImmutableDictionary<_,_>) (d2 : ImmutableDictionary<_,_>) = 
+        let keysMismatch = new HashSet<_>(d1.Keys)
+        keysMismatch.SymmetricExceptWith d2.Keys 
+        keysMismatch.Count = 0 && d1 |> Seq.exists (fun kv -> d2.[kv.Key] <> kv.Value) |> not
+
+    let Foo  = qualifiedName "Foo"
+    let Bar  = qualifiedName "Bar"
+
+    let FooA = typeParameter "Foo.A"
+    let FooB = typeParameter "Foo.B"
+    let BarA = typeParameter "Bar.A"
+    let BarC = typeParameter "Bar.C"
+    let BazA = typeParameter "Baz.A"
+
+
+    [<Fact>]
+    member this.``Type resolution`` () = 
+
+        let res1 = resolution [
+            (FooA, BarA |> TypeParameter)
+            (FooB, Int)
+        ]
+        let res2 = resolution [
+            (BarA, Int)
+        ]
+        let expected = resolution [
+            (FooA, Int)
+            (FooB, Int)
+        ]
+
+        let mutable combined = ImmutableDictionary.Empty
+        let success = CallGraph.TryCombineTypeResolutions(Foo, &combined, res1, res2)
+        Assert.True(success)
+        Assert.True(compareDict expected combined)
--- a/src/QsCompiler/Tests.Compiler/CallGraphTests.fs
+++ b/src/QsCompiler/Tests.Compiler/CallGraphTests.fs
@@ -59,7 +59,8 @@ type CallGraphTests (output:ITestOutputHelper) =
 
 
     [<Fact>]
-    member this.``Type resolution`` () = 
+    [<Trait("Category","Type resolution")>]
+    member this.``Type resolution: resolution to concrete`` () = 
 
         let res1 = resolution [
             (FooA, BarA |> TypeParameter)
@@ -74,6 +75,40 @@ type CallGraphTests (output:ITestOutputHelper) =
         ]
         CallGraphTests.AssertResolution(Foo, expected, res1, res2)
 
+    [<Fact>]
+    [<Trait("Category","Type resolution")>]
+    member this.``Type resolution: resolution to type parameter`` () = 
+
+        let res1 = resolution [
+            (FooA, BarA |> TypeParameter)
+        ]
+        let res2 = resolution [
+            (BarA, BazA |> TypeParameter)
+        ]
+        let expected = resolution [
+            (FooA, BazA |> TypeParameter)
+        ]
+        CallGraphTests.AssertResolution(Foo, expected, res1, res2)
+
+    [<Fact>]
+    [<Trait("Category","Type resolution")>]
+    member this.``Type resolution: resolution via identity mapping`` () = 
+
+        let res1 = resolution [
+            (FooA, FooA |> TypeParameter)
+        ]
+        let res2 = resolution [
+            (FooA, Int)
+        ]
+        let expected = resolution [
+            (FooA, Int)
+        ]
+        CallGraphTests.AssertResolution(Foo, expected, res1, res2)
+
+    [<Fact>]
+    [<Trait("Category","Type resolution")>]
+    member this.``Type resolution: multi-stage resolution`` () = 
+
         let res1 = resolution [
             (FooA, BarA |> TypeParameter)
         ]
@@ -86,6 +121,10 @@ type CallGraphTests (output:ITestOutputHelper) =
             (FooB, Int)
         ]
         CallGraphTests.AssertResolution(Foo, expected, res1, res2)
+
+    [<Fact>]
+    [<Trait("Category","Type resolution")>]
+    member this.``Type resolution: multiple resolutions to concrete`` () = 
 
         let res1 = resolution [
             (FooA, BarA |> TypeParameter)
@@ -99,6 +138,30 @@ type CallGraphTests (output:ITestOutputHelper) =
             (FooB, Int)
         ]
         CallGraphTests.AssertResolution(Foo, expected, res1, res2)
+
+    [<Fact>]
+    [<Trait("Category","Type resolution")>]
+    member this.``Type resolution: multiple resolutions to type parameter`` () = 
+
+        let res1 = resolution [
+            (FooA, BarA |> TypeParameter)
+            (FooB, BarA |> TypeParameter)
+        ]
+        let res2 = resolution [
+            (BarA, BazA |> TypeParameter)
+        ]
+        let res3 = resolution [
+            (BazA, Double)
+        ]
+        let expected = resolution [
+            (FooA, Double)
+            (FooB, Double)
+        ]
+        CallGraphTests.AssertResolution(Foo, expected, res1, res2, res3)
+
+    [<Fact>]
+    [<Trait("Category","Type resolution")>]
+    member this.``Type resolution: multi-stage resolution of multiple resolutions to concrete`` () = 
 
         let res1 = resolution [
             (FooA, BarA |> TypeParameter)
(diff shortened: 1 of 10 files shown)


src/QsCompiler/Transformations/ClassicallyControlled.cs
```diff
@@ -112,25 +89,27 @@ private TypeArgsResolution GetCombinedTypeResolution(TypeArgsResolution outer, T
...
-                            && id.Item1 is Identifier.GlobalCallable global)
+                        if (
+                            newCallIdentifier.Expression is ExpressionKind.Identifier id
+                            && id.Item1 is Identifier.GlobalCallable global
+                            && (callTypeArguments.Any() || idTypeArguments.Any()))
                         {
+                            // We are dissolving the application of arguments here, so the call's type argument
+                            // resolutions have to be moved to the 'identifier' sub expression.
+                            var combined = CallGraph.TryCombineTypeResolutions(global.Item, out var combinedTypeArguments, idTypeArguments, callTypeArguments);
+                            QsCompilerError.Verify(combined, "failed to combine type parameter resolution");
```
ScottCarda-MS, line 105: I think you had a ToDo for this, but this error should probably be generated by the combining logic, rather than here, in the calling context. That may remove the need to use the "Try-" method patter for the combining method allowing us to move the out-parameter to the proper return value.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -31,11 +31,113 @@ public struct CallGraphNode
...
+        // TODO: 
+        // This is the method that should be invoked to verify cycles of interest,
+        // i.e. where each callable in the cycle is type parametrized.
+        // It should probably generate diagnostics; I'll add the doc comment once its use is fully defined. 
+        internal static bool VerifyCycle(CallGraphNode rootNode, params CallGraphEdge[] edges)
+        {
+            var parent = rootNode.CallableName;
+            bool EqualsParent(QsQualifiedName origin) => origin.Namespace.Value == parent.Namespace.Value && origin.Name.Value == parent.Name.Value;
+            var validResolution = TryCombineTypeResolutions(parent, out var combined, edges.Select(edge => edge.ParamResolutions).ToArray());
+            var resolvedToConcrete = combined.Values.All(res => !(res.Resolution is ResolvedTypeKind.TypeParameter tp) || EqualsParent(tp.Item.Origin));
```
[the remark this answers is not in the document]
bettinaheim: The parent's type parameter is considered to be concrete within a parent. Also, this will return true for resolutions to concrete types as well, so resolvedToParent would not be accurate.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -31,11 +31,113 @@ public struct CallGraphNode
...
+            return validResolution && resolvedToConcrete;
+            //var isClosedCycle = validCycle && combined.Values.Any(res => res.Resolution is ResolvedTypeKind.TypeParameter tp && EqualsParent(tp.Item.Origin));
+            // TODO: check that monomorphization correctly processes closed cycles - meaning add a test...
+        }
+
+        /// <summary>
+        /// Given a sequence of type resolutions specifying e.g. subsequent concretizations as part of a nested expression,
```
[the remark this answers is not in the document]
bettinaheim: I have no idea; what seems more like a valid and accurate English word? I was just assuming the verb concretize exists and the corresponding noun would be concretization.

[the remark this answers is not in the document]
bettinaheim: done

src/QsCompiler/Transformations/ClassicallyControlled.cs
bettinaheim: I think it seems fair to have as a QsCompilerError. If this fails, then that means that the syntax tree passed into this step was incorrect. Each rewrite step should be able to rely on that the given syntax tree is indeed valid, otherwise that's a huge burden and unnecessary overhead to implement full error handling in each step.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -31,11 +31,113 @@ public struct CallGraphNode
...
+        /// Returns the constructed dictionary as out parameter. Returns true if the combination of the given resolutions is valid, 
+        /// i.e. if there are no conflicting resolutions and type parameters of the parent callables are uniquely resolved 
+        /// to either a concrete type, a type parameter of another callable, or themselves.
+        /// Throws an ArgumentNullException if the given parent is null. 
+        /// Throws an ArgumentException if the given resolutions imply that type parameters from multiple callables are 
+        /// simultaneously treated as concrete types. 
+        /// NOTE: This routine prioritizes the verifications to ensure the correctness of the resolution over performance.  
+        /// </summary>
+        public static bool TryCombineTypeResolutions
+            (QsQualifiedName parent, out ImmutableDictionary<Tuple<QsQualifiedName, NonNullable<string>>, ResolvedType> combined,
```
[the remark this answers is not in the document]
bettinaheim: I am not sure what you are suggesting. That would be a major breaking change since it would require changing the SyntaxTree data structures. Is that what you are suggesting? In any case, that would definitively need to be its own PR.

src/QsCompiler/Transformations/ClassicallyControlled.cs
```diff
@@ -112,25 +89,27 @@ private TypeArgsResolution GetCombinedTypeResolution(TypeArgsResolution outer, T
...
 
                         // This relies on anything having type parameters must be a global callable.
-                        var newCallIdentifier = call.Item1;
-                        if (combinedTypeArguments.Any()
-                            && newCallIdentifier.Expression is ExpressionKind.Identifier id
-                            && id.Item1 is Identifier.GlobalCallable global)
+                        if (
+                            newCallIdentifier.Expression is ExpressionKind.Identifier id
+                            && id.Item1 is Identifier.GlobalCallable global
+                            && (callTypeArguments.Any() || idTypeArguments.Any()))
```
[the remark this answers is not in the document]
bettinaheim: done

bettinaheim:
> Getting to be a lot of comments, so I thought I would break it up a bit. I still need to look over the inner details of the TryCombineTypeResolutions method and the unit tests.
> A lot of them are small things, or things that I can do I my end after the PR is in, if you are eager to get the PR. If there is any comment you'd rather I address after the fact, just put a response saying as much.
> Once thing in particular I am hoping we can change is the signature for TryCombineTypeResolutions. I know we tend to use the Try-method pattern a lot in this code base, and sometime it is especially convenient, but using out parameters is generally to be avoided if possible, especially if the method also has a variable number of parameters as the TryCombineTypeResolutions does.
> Here are two resources to consider when considering out-parameters:
> 
> Avoid Out Parameters: https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-2.0/ms229053(v=vs.80)
> Out Parameters Best Practices: https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/parameter-design

Ah, that's a tough one. What is your concrete suggestion? Instead having the signature be
`        public static (bool, ImmutableDictionary<Tuple<QsQualifiedName, NonNullable<string>>, ResolvedType>) TryCombineTypeResolutions
            (QsQualifiedName parent, params ImmutableDictionary<Tuple<QsQualifiedName, NonNullable<string>>, ResolvedType>[] resolutions)
`?
It just seem weird to me since all the C# methods like TryParse, TryGet, Try* follow the pattern of returning a boolean and given the value - if any - as out parameter. I would also argue that this method is not exactly "designing for a general audience", given its purpose.

src/QsCompiler/Transformations/ClassicallyControlled.cs
ScottCarda-MS: I am not suggesting that we get rid of the check, but that it should be moved to the logic in the CallGraph, which has the logic that it is checking. The check doesn't verify any logic specific to the Classically Controlled.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -31,11 +31,110 @@ public struct CallGraphNode
...
+            return validResolution && resolvedToConcrete;
+            //var isClosedCycle = validCycle && combined.Values.Any(res => res.Resolution is ResolvedTypeKind.TypeParameter tp && EqualsParent(tp.Item.Origin));
+            // TODO: check that monomorphization correctly processes closed cycles - meaning add a test...
+        }
+
+        /// <summary>
+        /// Combines subsequent concretions as part of a nested expression, or concretions as part of a cycle in the call graph,
+        /// into a single dictionary containing the resolution for the type parameters of the specified parent callable. 
+        /// The given resolutions are expected to be ordered starting with the dictionary containing the initial mapping for the 
+        /// type parameters of the specified parent callable (the "innermost resolutions"). This mapping may potentially be to
```
ScottCarda-MS, line 52: Based on the code, it seems like it is important to have the resolutions that are guaranteed to have resolutions for native type parameters, which would be the 'outermost resolutions', come first in the ordering. Is that interpretation correct?

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -31,11 +31,110 @@ public struct CallGraphNode
...
+        /// Returns the constructed dictionary as out parameter. Returns true if the combination of the given resolutions is valid, 
+        /// i.e. if there are no conflicting resolutions and type parameters of the parent callables are uniquely resolved 
+        /// to either a concrete type, a type parameter of another callable, or themselves.
+        /// Throws an ArgumentNullException if the given parent is null. 
+        /// Throws an ArgumentException if the given resolutions imply that type parameters from multiple callables are 
+        /// simultaneously treated as concrete types. 
+        /// NOTE: This routine prioritizes the verifications to ensure the correctness of the resolution over performance.  
+        /// </summary>
+        public static bool TryCombineTypeResolutions
+            (QsQualifiedName parent, out ImmutableDictionary<Tuple<QsQualifiedName, NonNullable<string>>, ResolvedType> combined,
```
ScottCarda-MS, line 63: I'm having a difficult time understanding what the role of `parent` is. What is this callable in relation to the given resolutions?

ScottCarda-MS: A code example would go a long way here.

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
```diff
@@ -0,0 +1,220 @@
...
+        let res2 = resolution [
+            (BarA, BazA |> TypeParameter)
+        ]
+        let res3 = resolution [
+            (BazA, BarC |> TypeParameter) // FIXME: THIS SHOULD NOT BE OK
+        ]
+        let expected = resolution [
+            (FooA, BarC |> TypeParameter)
+        ]
+        CallGraphTests.AssertResolution(Foo, expected, res1, res2, res3)
```
ScottCarda-MS, line 220: There should be a scenario where FooA is resolved to BarA. Would that scenario be expected to have a resolution success or failure?

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
```diff
@@ -0,0 +1,220 @@
...
+        let success = CallGraphTests.CheckResolution (parent, expected, resolutions)
+        Assert.True(success, "overall status indicated as not successful")
+
+    static member AssertResolutionFailure (parent, expected, [<ParamArray>] resolutions) = 
+        let success = CallGraphTests.CheckResolution (parent, expected, resolutions)
+        Assert.False(success, "overall status indicated as success")
+
+
+    [<Fact>]
+    member this.``Type resolution`` () =
```
ScottCarda-MS, line 62: Each of these scenarios in this test should be its own test as they are all testing different cases. If one of them fails, it will be much more clear what scenario fails.

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
```diff
@@ -0,0 +1,220 @@
...
+    let Bar  = qualifiedName "Bar"
+
+    let FooA = typeParameter "Foo.A"
+    let FooB = typeParameter "Foo.B"
+    let BarA = typeParameter "Bar.A"
+    let BarC = typeParameter "Bar.C"
+    let BazA = typeParameter "Baz.A"
+
+    static member CheckResolution (parent, expected, [<ParamArray>] resolutions) = 
+        let compareDict (d1 : ImmutableDictionary<_,_>) (d2 : ImmutableDictionary<_,_>) =
```
ScottCarda-MS, line 43: This should probably either be not a local function, or it could omit the first parameter and assume `expected` from the parent context is the first dictionary.

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
```diff
@@ -0,0 +1,220 @@
...
+    let BarA = typeParameter "Bar.A"
+    let BarC = typeParameter "Bar.C"
+    let BazA = typeParameter "Baz.A"
+
+    static member CheckResolution (parent, expected, [<ParamArray>] resolutions) = 
+        let compareDict (d1 : ImmutableDictionary<_,_>) (d2 : ImmutableDictionary<_,_>) = 
+            let keysMismatch = new HashSet<_>(d1.Keys)
+            keysMismatch.SymmetricExceptWith d2.Keys 
+            keysMismatch.Count = 0 && d1 |> Seq.exists (fun kv -> d2.[kv.Key] <> kv.Value) |> not
+        let mutable combined = ImmutableDictionary.Empty
```
ScottCarda-MS, line 47: Mutable Immutable dictionary seems odd, haha.

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
```diff
@@ -0,0 +1,220 @@
...
+        let success = CallGraph.TryCombineTypeResolutions(parent, &combined, resolutions)
+        Assert.True(compareDict expected combined, "combined resolutions did not match the expected ones")
+        success
+
+    static member AssertResolution (parent, expected, [<ParamArray>] resolutions) = 
+        let success = CallGraphTests.CheckResolution (parent, expected, resolutions)
+        Assert.True(success, "overall status indicated as not successful")
```
ScottCarda-MS, line 54:
```suggestion
        Assert.True(success, "Combining type resolutions was not successful.")
```

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
```diff
@@ -0,0 +1,220 @@
...
+
+    static member AssertResolutionFailure (parent, expected, [<ParamArray>] resolutions) = 
+        let success = CallGraphTests.CheckResolution (parent, expected, resolutions)
+        Assert.False(success, "overall status indicated as success")
```
ScottCarda-MS, line 58:
```suggestion
        Assert.False(success, "Combining type resolutions was successful when it was expected to fail.")
```

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -31,11 +31,110 @@ public struct CallGraphNode
...
+                // if a certain type parameter is resolved by the currently processed dictionary. 
+                var mayBeReplaced = combinedBuilder
+                    .Where(kv => kv.Value.Resolution.IsTypeParameter)
+                    .ToLookup(
+                        kv => AsTypeResolutionKey(((ResolvedTypeKind.TypeParameter)kv.Value.Resolution).Item),
+                        entry => entry.Key);
+
+                // We need to ensure that the mappings for extenal type parameters are processed first, 
+                // to cover an edge case that would otherwise be indicated as a conflicting resolution.
+                var entriesToProcess = resolution.OrderByDescending(entry => !entry.Key.Item1.Equals(parent));
```
ScottCarda-MS, line 86: Because the looping logic also discriminates on this same predicate, it might be cleaner to just split this dictionary into two dictionaries based on this predicate and looping over these two separately, splitting out the looping logic appropriately.

src/QsCompiler/Transformations/CallGraphWalker.cs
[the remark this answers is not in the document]
ScottCarda-MS: In my experience, putting a -tion at the end of a verb makes a noun that describes the process of performing that verb, not necessarily the outcome of performing the verb, although, because English is pretty inconsistent, I am certain there are plenty of counter examples.

Concretization seems like it is referring to the process of performing the verb 'concretize' rather than the outcome.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -31,11 +31,110 @@ public struct CallGraphNode
...
+                res.Resolution is ResolvedTypeKind.TypeParameter tp && tp.Item.Origin.Equals(typeParam.Item1) && tp.Item.TypeName.Equals(typeParam.Item2);
+
+            foreach (var resolution in resolutions)
+            {
+                // Contains a lookup of all the keys in the combined resolutions whose value needs to be updated 
+                // if a certain type parameter is resolved by the currently processed dictionary. 
+                var mayBeReplaced = combinedBuilder
+                    .Where(kv => kv.Value.Resolution.IsTypeParameter)
+                    .ToLookup(
+                        kv => AsTypeResolutionKey(((ResolvedTypeKind.TypeParameter)kv.Value.Resolution).Item),
```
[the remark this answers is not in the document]
bettinaheim: There Where above does the filtering such that this case is safe.

src/QsCompiler/Transformations/CallGraphWalker.cs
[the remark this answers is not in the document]
samarsha: Yes, but that refers to the verb *to concrete*, which also seems to be primarily used for physical processes. The abstract use seems uncommon. For example, the [concretion](https://en.wikipedia.org/wiki/Concretion) Wikipedia article is only about physical concretions, while [concretization](https://en.wikipedia.org/w/index.php?title=Concretization&redirect=no) redirects to an article about the concretization of abstract ideas.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -31,11 +31,110 @@ public struct CallGraphNode
...
+                        entry => entry.Key);
+
+                // We need to ensure that the mappings for external type parameters are processed first, 
+                // to cover an edge case that would otherwise be indicated as a conflicting resolution.
+                var entriesToProcess = resolution.OrderBy(entry => entry.Key.Item1.Equals(parent));
```
ScottCarda-MS, line 86: Reposting this comment as the original got marked as 'outdated'.

Because the looping logic also discriminates on this same predicate, it might be cleaner to just split this dictionary into two dictionaries based on this predicate and looping over these two separately, splitting out the looping logic appropriately.

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
bettinaheim: I'm not sure I fully understand what case you had in mind. Would you mind typing out the full test case including what the input dictionaries should contain?

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
bettinaheim: sure

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
bettinaheim: :)

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
ScottCarda-MS:
```
[<Fact>]
member this.``Needs Better Name`` () =
    let res1 = resolution [
        (FooA, BarA |> TypeParameter)
    ]
    let res2 = resolution [
        (BarA, BazA |> TypeParameter)
    ]
    let expected = resolution [
        (FooA, BazA |> TypeParameter)
    ]
    CallGraphTests.AssertResolution(Foo, expected, res1, res2)
```

bettinaheim: added two more test cases

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
bettinaheim: done

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
bettinaheim: done.

src/QsCompiler/Transformations/CallGraphWalker.cs
bettinaheim: No, thinking of call-like expressions, the dictionary to pass first is indeed the one of the innermost expression, and the parent would be the callable invoked in the call, like the doc comment states.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -31,11 +31,110 @@ public struct CallGraphNode
...
+                foreach (var entry in entriesToProcess)
+                {
+                    var resolutionToTypeParam = entry.Value.Resolution as ResolvedTypeKind.TypeParameter;
+                    var isResolutionToNative = resolutionToTypeParam != null && resolutionToTypeParam.Item.Origin.Equals(parent);
+
+                    // resolution of a type parameter that belongs to the parent callable
+                    if (entry.Key.Item1.Equals(parent))
+                    {
+                        // A native type parameter cannot be resolved to another native type parameter, since this would constrain them. 
+                        var nontrivialResolutionToNative = isResolutionToNative && entry.Key.Item2.Value != resolutionToTypeParam.Item.TypeName.Value;
```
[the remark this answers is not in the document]
bettinaheim: done

src/QsCompiler/Transformations/ClassicallyControlled.cs
bettinaheim: The method in the call graph specifically can also be used to validate part of the syntax tree. Hence, it won't raise a QsCompilerError but merely indicate if the validation fails (either via bool or via generated diagnstics). However, this invokation here indeed should never fail, hence raising a QsCompilerError here (serving the same purpose as Debug.Assert) seems appropriate. It is not a matter of QsCompilerError checking any logic of the Classically Controlled; it is a matter of detecting errors as early as we can and not siliently ignoring inconsistent state/input/etc. if there is an easy way to check it.

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
bettinaheim: done

src/QsCompiler/Transformations/CallGraphWalker.cs
bettinaheim: done

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -31,11 +31,112 @@ public struct CallGraphNode
...
+                // Contains a lookup of all the keys in the combined resolutions whose value needs to be updated 
+                // if a certain type parameter is resolved by the currently processed dictionary. 
+                var mayBeReplaced = combinedBuilder
+                    .Where(kv => kv.Value.Resolution.IsTypeParameter)
+                    .ToLookup(
+                        kv => AsTypeResolutionKey(((ResolvedTypeKind.TypeParameter)kv.Value.Resolution).Item),
+                        entry => entry.Key);
+
+                // We need to ensure that the mappings for external type parameters are processed first, 
+                // to cover an edge case that would otherwise be indicated as a conflicting resolution.
```
ScottCarda-MS, line 97: I'm still not getting this. Why do you need to process this for loop first, before the other one? To my thinking you would need to process the other for loop first. Perhaps an explanation of this edge case mentioned here would help clear this up?

src/QsCompiler/Transformations/CallGraphWalker.cs
bettinaheim: If we wanted to see if we can be more clever than having to check the resolution for each node in a cycle, then parent would get eliminated. I'd rather not start to have code example in documenting comments - the tests give these examples.

src/QsCompiler/Transformations/CallGraphWalker.cs
bettinaheim: If you don't sort or change which loop is executed first, you will see which test fails. :)

PR #398, opened 2020-04-09
Renamed 'Concretion' to 'Concretization'

ScottCarda-MS:


bettinaheim:
@ScottCarda-MS Can we close this PR? It would be nice to do this renaming to use consistent terminology, but I think it is easier to do it on the current version rather than resolve the conflicts here. Could you set up a new PR for that?

ScottCarda-MS:
> @ScottCarda-MS Can we close this PR? It would be nice to do this renaming to use consistent terminology, but I think it is easier to do it on the current version rather than resolve the conflicts here. Could you set up a new PR for that?

The intension was to address this after the call graph was implemented into master. You are right in that it will need to be redone with the current version rather than here. I'd like to keep this up as a reminder to myself until #654 is completed, after which I will create a new PR.

ScottCarda-MS:
Closing this in favor of #682

PR #399, opened 2020-04-10
Additional verification for entry points

bettinaheim:
What is still missing the check that only projects of type exe can have entry points (making it a warning otherwise), and the restriction of entry point return values to Result, Result[] and tuples thereof for execution on quantum processors.


src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -171,6 +179,12 @@ public struct Configuration
             internal bool SerializeSyntaxTree =>
                 BuildOutputFolder != null || DllOutputPath != null;
 
+            /// <summary>
+            /// Indicates whether the compiler will remove if-statements and replace them with calls to appropriate intrinsic operations.
+            /// </summary>
+            internal bool ConvertClassicalControl =>
```
avasch01, line 185: >ConvertClassicalControl [](start = 26, length = 23)

Would it be better to call this IntrinsicClassicalControl? ust a little more precise. "Convert" does not give outsider an idea what is converted to what.

src/QsCompiler/Core/SymbolTable.fs
```diff
@@ -980,22 +981,47 @@ and NamespaceManager
...
+
+                        // verify that the entry point has only a default body specialization
+                        let hasCharacteristics = signature.Characteristics.Characteristics |> function | EmptySet | InvalidSetExpr -> false | _ -> true
+                        match Namespaces.TryGetValue parent.Namespace with
+                        | false, _ -> ArgumentException "no namespace with the given name exists" |> raise
+                        | true, ns when not ((ns.SpecializationsDefinedInAllSources parent.Name).Any(fst >> (<>)QsBody) || hasCharacteristics) -> ()
+                        | _ -> errs.Add (decl.Position, signature.Characteristics.Range.ValueOr decl.Range |> QsCompilerDiagnostic.Error (ErrorCode.InvalidEntryPointSpecialization, []))
+
+                        // validate entry point argument and return type
+                        let validateArgAndReturnTypes argumentRestriction (qsType : QsType) =
```
samarsha, line 993: `argumentRestriction` is somewhat misleading; maybe something like `isInArg`.

Alternatively, this could be two functions `validateArgType` and `validateReturnType`, where `validateArgType` calls `validateReturnType` in addition to checking the argument-specific restrictions.

src/QsCompiler/Core/SymbolTable.fs
```diff
@@ -980,22 +981,47 @@ and NamespaceManager
...
                         let signatureErrs = inErrs.Concat outErrs
-                        let hasCharacteristics = signature.Characteristics.Characteristics |> function | EmptySet | InvalidSetExpr -> false | _ -> true
-                        match Namespaces.TryGetValue parent.Namespace with
-                        | false, _ -> ArgumentException "no namespace with the given name exists" |> raise
-                        | true, ns when not ((ns.SpecializationsDefinedInAllSources parent.Name).Any(fst >> (<>)QsBody) || hasCharacteristics) -> ()
-                        | _ -> errs.Add (decl.Position, signature.Characteristics.Range.ValueOr decl.Range |> QsCompilerDiagnostic.Error (ErrorCode.InvalidEntryPointSpecialization, []))
+
+                        // validate entry point argument names
+                        let asCommandLineArg (str : string) = str.ToLowerInvariant() |> Seq.filter((<>)'_') |> String.Concat
+                        let reservedCommandLineArgs = [CommandLineArguments.SimulatorOption] |> List.map asCommandLineArg
```
samarsha, line 1012: This looks like it would be easy to forget to update when new options are added. Could the list of all options be moved to the `CommandLineArguments` module?

src/QsCompiler/Core/SymbolTable.fs
```diff
@@ -980,22 +981,47 @@ and NamespaceManager
...
                         let outErrs = signature.ReturnType |> validateArgAndReturnTypes false 
                         let signatureErrs = inErrs.Concat outErrs
-                        let hasCharacteristics = signature.Characteristics.Characteristics |> function | EmptySet | InvalidSetExpr -> false | _ -> true
-                        match Namespaces.TryGetValue parent.Namespace with
-                        | false, _ -> ArgumentException "no namespace with the given name exists" |> raise
-                        | true, ns when not ((ns.SpecializationsDefinedInAllSources parent.Name).Any(fst >> (<>)QsBody) || hasCharacteristics) -> ()
-                        | _ -> errs.Add (decl.Position, signature.Characteristics.Range.ValueOr decl.Range |> QsCompilerDiagnostic.Error (ErrorCode.InvalidEntryPointSpecialization, []))
+
+                        // validate entry point argument names
+                        let asCommandLineArg (str : string) = str.ToLowerInvariant() |> Seq.filter((<>)'_') |> String.Concat
```
samarsha, line 1011:
```suggestion
                        let asCommandLineArg (str : string) = str.ToLowerInvariant() |> String.filter ((<>) '_')
```

src/QsCompiler/DataStructures/ReservedKeywords.fs
```diff
@@ -231,6 +231,11 @@ module GeneratedAttributes =
     let LoadedViaTestNameInsteadOf = "__LoadedViaTestNameInsteadOf__"
 
 
+/// contains reserved names for command line arguments
```
samarsha, line 234: Might be good to clarify that they are command-line arguments for the standalone executable specifically.

src/QsCompiler/Core/SymbolTable.fs
```diff
@@ -980,22 +981,47 @@ and NamespaceManager
...
+                        let reservedCommandLineArgs = [CommandLineArguments.SimulatorOption] |> List.map asCommandLineArg
+                        let nameAndRange (sym : QsSymbol) = sym.Symbol |> function
+                            | Symbol name -> Some (asCommandLineArg name.Value, sym.Range)
+                            | _ -> None
+                        let simplifiedArgNames = signature.Argument.Items.Select fst |> Seq.choose nameAndRange |> Seq.toList
+                        let verifyArgument i (arg, range : QsNullable<_>) = 
+                            if i > 0 && simplifiedArgNames.[..i-1] |> Seq.map fst |> Seq.contains arg
+                            then errs.Add (decl.Position, range.ValueOr decl.Range |> QsCompilerDiagnostic.Error (ErrorCode.DuplicateEntryPointArgumentName, []))
+                            elif reservedCommandLineArgs.Contains arg
+                            then errs.Add (decl.Position, range.ValueOr decl.Range |> QsCompilerDiagnostic.Error (ErrorCode.ReservedEntryPointArgumentName, []))
```
samarsha, line 1021: This should be a warning instead of an error.

src/QsCompiler/Compiler/CompilationLoader.cs
bettinaheim: This is an internal property and hence will not be visible to an outsider. Additionally, there is a documenting comment that is far more informative than the name itself. :)

avasch01: But would not it be better if even internal variables were named in a more informative way? Outsiders need to debug through this code once in a while. Would not it be better if they understood what is going on without looking through files to find the informative comment? Or let me ask the same question from the other end: is there a good reason to call it Convert rather than Intrinsic?

---
In reply to: [406999603](https://github.com/microsoft/qsharp-compiler/pull/399#discussion_r406999603) [](ancestors = 406999603)

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -581,6 +603,14 @@ private void LogAndUpdate(ref Status current, ErrorCode code, IEnumerable<string
             current = Status.Failed;
         }
 
+        /// <summary>
+        /// Logs a warning with the given warning code and message parameters, and updates the status passed as reference accordingly.
+        /// </summary>
+        private void LogAndUpdate(ref Status current, WarningCode code, IEnumerable<string> args)
```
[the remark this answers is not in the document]
bettinaheim: ups, thank you!

src/QuantumSdk/README.md
```diff
@@ -109,6 +109,9 @@ May contain additional arguments to pass to the Q# command line compiler. Valid
 - `CsharpGeneration`:    
 Specifies whether to generate C# code as part of the compilation process. Setting this property to false may prevent certain interoperability features or integration with other pieces of the Quantum Development Kit. 
 
+- `DefaultSimulator`:
+Specifies the simulator to use by default for execution.
```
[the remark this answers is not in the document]
bettinaheim: I added "Valid values are QuantumSimulator, ToffoliSimulator, ResourcesEstimator, or the fully qualified name of a custom simulator. ". 
It would be good to add a test for a custom simulator in your PR.

src/QsCompiler/Core/SymbolTable.fs
```diff
@@ -980,22 +981,47 @@ and NamespaceManager
                 elif tId |> isBuiltIn BuiltIn.EntryPoint then
```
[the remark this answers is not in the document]
bettinaheim: I split out a validateEntryPoint function.

src/QsCompiler/CommandLineTool/Options.cs
```diff
@@ -50,6 +50,14 @@ public class CompilationOptions : Options
         HelpText = "Additional properties to populate the AssemblyConstants dictionary with. Each item is expected to be of the form \"key:value\".")]
         public IEnumerable<string> AdditionalAssemblyProperties { get; set; }
 
+        [Option("runtime", Required = false, SetName = CODE_MODE, 
+        HelpText = "Specifies the classical capabilites of the runtime. Determines what QIR profile to compile to.")]
+        public AssemblyConstants.RuntimeCapabilities RuntimeCapabilites { get; set; }
+
+        [Option("build-exe", Required = false, Default = false, SetName = CODE_MODE,
+        HelpText = "Specifies whether to build a Q# command line application.")]
+        public bool MakeExecutable { get; set; }
```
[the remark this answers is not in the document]
bettinaheim: I assume you mean non-standalone, but with a custom driver. No, actually this would be true even if you have a custom driver. The goal eventually would be to not have C# and Q# code in the same project, but instead have a separate project when you want to have a custom driver. Things will remain backwards compatible for now, but a warning should be generated if there is a exe project with no Q# entry point (e.g. because there is a C# driver in the same project).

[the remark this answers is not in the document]
bettinaheim: I updated the diagnostic to be a bit more meaningful.

src/QsCompiler/Core/SymbolTable.fs
```diff
@@ -980,22 +981,47 @@ and NamespaceManager
...
                     match box decl.Defined with
                     | :? CallableSignature as signature when not (signature.TypeParameters.Any()) ->
-                        let validateArgAndReturnTypes preventTuples (qsType : QsType) =
+
+                        // verify that the entry point has only a default body specialization
+                        let hasCharacteristics = signature.Characteristics.Characteristics |> function | EmptySet | InvalidSetExpr -> false | _ -> true
+                        match Namespaces.TryGetValue parent.Namespace with
+                        | false, _ -> ArgumentException "no namespace with the given name exists" |> raise
+                        | true, ns when not ((ns.SpecializationsDefinedInAllSources parent.Name).Any(fst >> (<>)QsBody) || hasCharacteristics) -> ()
+                        | _ -> errs.Add (decl.Position, signature.Characteristics.Range.ValueOr decl.Range |> QsCompilerDiagnostic.Error (ErrorCode.InvalidEntryPointSpecialization, []))
```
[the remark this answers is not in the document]
bettinaheim: done.

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -432,6 +446,14 @@ public CompilationLoader(SourceLoader loadSources, ReferenceLoader loadReference
             foreach (var diag in this.VerifiedCompilation?.Diagnostics() ?? Enumerable.Empty<Diagnostic>())
             { this.LogAndUpdate(ref this.CompilationStatus.Validation, diag); }
 
+            if (this.Config.IsExecutable && this.CompilationOutput?.EntryPoints.Length == 0)
+            {
+                this.Logger?.Log(WarningCode.MissingEntryPoint, new string[0]);
+            }
+            // TODO: 
+            // give warnings and ignore entry points in libraries, 
+            // and check additional restriction on the return type for execution on quantum processors.
```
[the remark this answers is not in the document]
bettinaheim: There is an error if there is more than one entry point.

src/QsCompiler/Core/SymbolTable.fs
bettinaheim: renamed

src/QsCompiler/Core/SymbolTable.fs
bettinaheim: Makes sense. I also move the the list of built in simulators to the keywords.

src/QsCompiler/DataStructures/ReservedKeywords.fs
bettinaheim: Opposed to an executable with a driver? Standalone should be the default, so I don't think that should be added (the rest is interop capabilities, and not usually refered to as any kind of executable).

src/QsCompiler/Core/SymbolTable.fs
bettinaheim: Right, for backward compatibility when we add a new one. Done.

src/QsCompiler/DataStructures/ReservedKeywords.fs
```diff
@@ -231,6 +231,11 @@ module GeneratedAttributes =
...
+module CommandLineArguments = 
+    let SimulatorOption = "simulator"
```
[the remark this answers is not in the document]
bettinaheim: done

src/QsCompiler/DataStructures/ReservedKeywords.fs
samarsha: Opposed to command-line arguments to the Q# command-line compiler, for example. It's not clear to me from the context in this file what "command-line arguments" refers to.

src/QsCompiler/DataStructures/ReservedKeywords.fs
```diff
@@ -239,6 +238,28 @@ module AssemblyConstants =
...
+    let ReservedArguments = ImmutableArray.Create SimulatorOption
+    let ReservedArgumentAbbreviations = ImmutableArray.Create SimulatorOption.[0] 
+
+    let BuiltInSimulators =
+        // Note: If this is adapted, then the error message for InvalidExecutionTargetForTest needs to be adapted as well.
+        [AssemblyConstants.QuantumSimulator; AssemblyConstants.ToffoliSimulator; AssemblyConstants.ResourcesEstimator]
+        |> ImmutableHashSet.CreateRange
```
samarsha, line 262: It might be better to have the error message convert `BuiltInSimulators` to a string directly, so the list is not stored in two places.

src/QuantumSdk/README.md
```diff
@@ -109,6 +109,9 @@ May contain additional arguments to pass to the Q# command line compiler. Valid
 - `CsharpGeneration`:    
 Specifies whether to generate C# code as part of the compilation process. Setting this property to false may prevent certain interoperability features or integration with other pieces of the Quantum Development Kit. 
 
+- `DefaultSimulator`:
+Specifies the simulator to use by default for execution. Valid values are QuantumSimulator, ToffoliSimulator, ResourcesEstimator, or the fully qualified name of a custom simulator.
```
samarsha, line 113: Currently, the standalone executable treats the resource estimator differently from the other simulators. Do you think it would be better for all of them to be under the `simulate` command, so to display a resource estimate is `simulate --simulator ResourceEstimator`? (I think I'm OK with this idea, the original reason was because the resource estimator has a custom output compared to the other two, but the check for what output to display can be based on the simulator option rather than the simulate/resources commands.)

For custom simulators, I think we might need to clarify that the simulator must be an `IOperationFactory` and has a zero-argument constructor.

samarsha: We can also say here that the default value of `DefaultSimulator` is `QuantumSimulator`.

src/QsCompiler/DataStructures/ReservedKeywords.fs
bettinaheim: Hmm, thinking about it, I removed the comment, because they might not exactly match and if we were to enable an additional simulator, we should have the proper unit tests to test that functionality in any case.

src/QsCompiler/DataStructures/ReservedKeywords.fs
bettinaheim: edited to "contains reserved names for command line arguments of Q# projects"

src/QuantumSdk/README.md
bettinaheim: As discussed, let's have the resource estimator listed under the simulator option.

src/QsCompiler/DataStructures/ReservedKeywords.fs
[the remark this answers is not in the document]
samarsha: The short name is not accessible separately from the list of all short names. The way that the option is created in the standalone executable is:

```
new[] { "--simulator", "-s" }
```

It would be good to have one property for the long name and one for the short name in `CommandLineArguments`. In general, the short name is not always the first letter of the long name, so I think it would be better not to programmatically create the short names.

src/QsCompiler/DataStructures/ReservedKeywords.fs
samarsha: This change didn't go through. The latest version is still:

```
/// contains reserved names for command line arguments
```

samarsha: Oh, I found the change in #402 instead.

src/QsCompiler/Compiler/CompilationLoader.cs
bettinaheim: @avasch01 Sorry, I somehow did not see this reply when I closed the thread and merged.

PR #402, opened 2020-04-14
Diagnostics based on project information

bettinaheim:
This addresses the remaining todo left in PR #399, and also propagates the processor architecture.


src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -612,6 +614,7 @@ type DiagnosticItem =
             | ErrorCode.MultipleEntryPoints                       -> "Invalid entry point. An entry point {0} already exists in {1}."
             | ErrorCode.InvalidEntryPointSpecialization           -> "Entry points cannot have any other specializations besides the default body."
             | ErrorCode.DuplicateEntryPointArgumentName           -> "Invalid name for entry point argument. A similar argument name is already in use."
+            | ErrorCode.EntryPointInLibrary                       -> "Invalid entry point. Only Q# command line applications can have entry points."
```
samarsha, line 617:
```suggestion
            | ErrorCode.EntryPointInLibrary                       -> "Invalid entry point. Only executable Q# projects can have entry points."
```

Someone could write a custom driver that has a GUI. :)

src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -328,6 +329,7 @@ type WarningCode =
     | MissingEntryPoint = 6202
     | IgnoredEntryPoint = 6203
     | ReservedEntryPointArgumentName = 6204
+    | NonResultTypeReturnedInEntryPoint = 6205
```
samarsha, line 332: I'm curious, what happens if someone ignores this warning?

src/QsCompiler/CompilationManager/CompilationUnit.cs
```diff
@@ -221,11 +221,16 @@ public class CompilationUnit : IReaderWriterLock, IDisposable
...
         internal References Externals { get; private set; }
         internal NamespaceManager GlobalSymbols { get; private set; }
+
         private readonly Dictionary<QsQualifiedName, QsCallable> CompiledCallables;
         private readonly Dictionary<QsQualifiedName, QsCustomType> CompiledTypes;
+
         private readonly ReaderWriterLockSlim SyncRoot;
         private readonly HashSet<ReaderWriterLockSlim> DependentLocks;
 
+        internal readonly AssemblyConstants.RuntimeCapabilities RuntimeCapabilities;
```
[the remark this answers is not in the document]
bettinaheim: done

src/QsCompiler/DataStructures/Diagnostics.fs
bettinaheim: Our end-to-end setup for QE is still a bit in flux, so I am not sure I know at this point in time. The reason I made it a warning is because it seems like a valid use case that people might want to compile for a certain architecture but only get resource estimations rather than running it on quantum hardware.

samarsha: OK. If it's a valid use case, do we want a way to disable the warning, to tell the compiler that they don't intend to run it on hardware?

src/QsCompiler/DataStructures/Diagnostics.fs
bettinaheim: alright, done.

src/QsCompiler/CompilationManager/ProjectManager.cs
```diff
@@ -11,31 +11,49 @@
...
 using Microsoft.Quantum.QsCompiler.Diagnostics;
+using Microsoft.Quantum.QsCompiler.ReservedKeywords;
 using Microsoft.VisualStudio.LanguageServer.Protocol;
 
 
 namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
 {
     public class ProjectInformation
     {
+        internal class Properties
```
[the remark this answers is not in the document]
bettinaheim: done.

Issue #387, opened 2020-03-24
Q# Language Server encounters ArgumentOutOfRangeException while editing a file


rhpreston:
**Describe the bug**

While working on a Q# project, the following banner will pop up, seemingly at random:

```
The Q# Language Server has encountered an error. Diagnostics will be reloaded upon saving the file. Details on the encountered error have been logged to the output window.
```

Subsequently, more identical banners pop up for each character I type, until I save the file and there's a chance it will stop.

Checking the output, the detailed error message is:

```
Error: 
=============================
Microsoft.Quantum.QsCompiler.QsCompilerException: error while running queued task
QsCompilerException: error during scope tracking update
ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter 'startChar')
Stack trace: 
   at Microsoft.Quantum.QsCompiler.CompilationBuilder.Utils.GetChangedText(String lineText, Int32 startChar, Int32 endChar, String insert)
   at Microsoft.Quantum.QsCompiler.CompilationBuilder.FileContentManager.DequeueUnprocessedChanges(Int32& lineNr, String& textToInsert)
   at Microsoft.Quantum.QsCompiler.CompilationBuilder.ScopeTracking.UpdateScopeTacking(FileContentManager file, TextDocumentContentChangeEvent change)
   at Microsoft.Quantum.QsCompiler.QsCompilerError.RaiseOnFailure(Action action, String header)

Stack trace: 
   at Microsoft.Quantum.QsCompiler.QsCompilerError.Raise(String message, Exception innerException)
   at Microsoft.Quantum.QsCompiler.QsCompilerError.RaiseOnFailure(Action action, String header)
   at Microsoft.Quantum.QsCompiler.CompilationBuilder.FileContentManager.Update(TextDocumentContentChangeEvent change)
   at Microsoft.Quantum.QsCompiler.CompilationBuilder.CompilationUnitManager.<>c__DisplayClass19_0.<SourceFileDidChangeAsync>b__0()
   at Microsoft.Quantum.QsCompiler.QsCompilerError.RaiseOnFailure(Action action, String header)

   at Microsoft.Quantum.QsCompiler.QsCompilerError.Raise(String message, Exception innerException)
   at Microsoft.Quantum.QsCompiler.QsCompilerError.RaiseOnFailure(Action action, String header)
[...]

PR #410, opened 2020-04-22
Show no more than one internal error per minute

bettinaheim:
The latest VS version do may an issue with notifications sent out of order. Since this will result in an error in our server, we want to avoid spamming users with pop up. I hence added a timer...


bettinaheim:
This is also to mitigate #387 somewhat.

PR #413, opened 2020-04-23
Minor updates to example

bettinaheim:


bettinaheim:
Need to do this when we have the released packages...

PR #415, opened 2020-04-24
Revert "validation for array item types in entry point arguments (#411)"

samarsha:
The restrictions on array types were lifted by microsoft/qsharp-runtime#175.


samarsha:
#416 does a more precise revert than this one.

PR #419, opened 2020-04-27
Type Resolution in Call Graph

ScottCarda-MS:
Integrated the type parameter resolution logic into different parts of the call graph code. Reorganized the call graph code to make it more manageable. Added tests


src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -0,0 +1,683 @@
...
+
+
+// ToDo: Review access modifiers
+
+namespace Microsoft.Quantum.QsCompiler.Transformations.CallGraphNS
+{
+    using ResolvedTypeKind = QsTypeKind<ResolvedType, UserDefinedType, QsTypeParameter, CallableInformation>;
+    using TypeParameterResolutions = ImmutableDictionary<Tuple<QsQualifiedName, NonNullable<string>>, ResolvedType>;
+
+    public static class TypeParamStuff
```
ScottCarda-MS, line 20: Need to have better name for `TypeParamStuff`.

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -0,0 +1,683 @@
...
+
+        public static bool TryCombineTypeResolutions(QsQualifiedName target, out TypeParameterResolutions combined, params TypeParameterResolutions[] resolutions)
+        {
+            if (target == null) throw new ArgumentNullException(nameof(target));
+            var success = Temp(out combined, resolutions);
+            combined = combined.Where(kvp => kvp.Key.Item1.Equals(target)).ToImmutableDictionary();
+            return success;
+        }
+
+        public static bool Temp(out TypeParameterResolutions combined, params TypeParameterResolutions[] resolutionDictionaries)
```
ScottCarda-MS, line 68: Need to have better name for `Temp`.

ScottCarda-MS: Renamed this to `TryCombineTypeResolutions`, and the original `TryCombineTypeResolutions` is now `TryCombineTypeResolutionsWithTarget`.

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
```diff
@@ -49,6 +49,9 @@ type CallGraphTests (output:ITestOutputHelper) =
     let resolution (res : (QsTypeParameter * QsTypeKind<_,_,_,_>) list) =
         res.ToImmutableDictionary((fun (tp,_) -> tp.Origin, tp.TypeName), snd >> ResolvedType.New)
 
+    let otherResolution (res : (QsQualifiedName * NonNullable<string> * QsTypeKind<_,_,_,_>) list) =
```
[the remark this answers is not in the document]
ScottCarda-MS: Renamed to `ResolutionFromParamName`.

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
```diff
@@ -100,270 +110,536 @@ type CallGraphTests (output:ITestOutputHelper) =
...
+        Assert.True(CheckResolutionMatch expected given, "Given resolutions did not match the expected resolutions.")
+
+    let CheckResolutionListMatch (resList1 : ImmutableDictionary<_,_> list) (resList2 : ImmutableDictionary<_,_> list) =
+        let sameLength = resList1.Length = resList2.Length
+        sameLength && resList1 |> List.forall (fun res1 -> resList2 |> List.exists (fun res2 -> CheckResolutionMatch res1 res2))
+
+    let AssertExpectedResolutionList (expected : ImmutableDictionary<_,_> list) (given : ImmutableDictionary<_,_> list) =
+        Assert.True(CheckResolutionListMatch expected given, "Given resolutions did not match the expected resolutions.")
+
+    let OtherAssertExpectedResolutions name (given : Dictionary<_,ImmutableArray<_>>) expected =
```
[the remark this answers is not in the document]
ScottCarda-MS: Renamed to `AssertExpectedDepencency`.

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: Renamed to `TypeParamUtils`.

src/QsCompiler/Compiler/FunctorGeneration.cs
```diff
@@ -9,7 +9,7 @@
 using Microsoft.Quantum.QsCompiler.ReservedKeywords;
 using Microsoft.Quantum.QsCompiler.SyntaxTokens;
 using Microsoft.Quantum.QsCompiler.SyntaxTree;
-
+using Microsoft.Quantum.QsCompiler.Transformations.CodeTransformationsNS;
```
bettinaheim, line 12: This seems like an odd namespace name? And it doesn't actually seem to be used anywhere?

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -13,515 +13,10 @@
 
 // ToDo: Review access modifiers
 
-namespace Microsoft.Quantum.QsCompiler.Transformations
+namespace Microsoft.Quantum.QsCompiler.Transformations.CallGraphNS
```
bettinaheim, line 16: The name ending in *NS seems a bit odd to me (any namespace could/should then end in NS). Is it intended to stay?

src/QsCompiler/Compiler/FunctorGeneration.cs
ScottCarda-MS: It is used throughout the file. Not sure why it is showing up as unused on your setup. Maybe rebuilding will fix it?

src/QsCompiler/Transformations/CallGraphWalker.cs
ScottCarda-MS: I got the idea that ending a namespace with NS was a MS standard with single class namespaces from this post: https://stackoverflow.com/a/48113597/5399520
After looking into it a bit more, I'm not sure that is the case. I dropped the NS from the namespace and rewrote the class.

ScottCarda-MS: These are some unusual transformations because they are static extension methods. I noticed this transformation class did not have a namespace for itself under `Transformations`, so I added one to fix that. But with them being extension methods, I am not sure. Is this where all extension methods that will be transformation will go?

ScottCarda-MS: I decided to drop the CodeTransformations namespace because that's what made sense to me. I kept the rename for the CodeTransformations class to CodeTransformationExtensions because that is the convention for static classes that just provide extension methods, and it makes it more clear that that is what that class does.

src/QsCompiler/Compiler/FunctorGeneration.cs
ScottCarda-MS: This namespace has been dropped.

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -0,0 +1,569 @@
...
+    /// </summary>
+    public struct CallGraphEdge
+    {
+        public TypeParameterResolutions ParamResolutions;
+    }
+
+    /// <summary>
+    /// Struct containing information that exists on nodes in a call graph.
+    /// </summary>
+    public struct CallGraphNode
```
bettinaheim, line 126: At least for CallGraphNode you have public methods below that expect members to be non-null. For that reason and also to align with how the rest of the code base works I would suggest to make this a class (and also CallGraphEdge) with a constructor that enforces that none of the members is ever null. Alternatively, you could look into whether it make sense to use C#'s nullable feature for that purpose.

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -0,0 +1,569 @@
...
+                res.Resolution is ResolvedTypeKind.TypeParameter tp && tp.Item.Origin.Equals(typeParam.Item1) && tp.Item.TypeName.Equals(typeParam.Item2);
+            static bool IsConstrictiveResolution(Tuple<QsQualifiedName, NonNullable<string>> typeParam, ResolvedType res) =>
+                res.Resolution is ResolvedTypeKind.TypeParameter tp && typeParam.Item1.Equals(tp.Item.Origin) && typeParam.Item2.Value != tp.Item.TypeName.Value;
+
+            foreach (var resolutionDictionary in resolutionDictionaries)
+            {
+                // Contains a lookup of all the keys in the combined resolutions whose value needs to be updated
+                // if a certain type parameter is resolved by the currently processed dictionary.
+                var mayBeReplaced = combinedBuilder
+                    .Where(kv => kv.Value.Resolution.IsTypeParameter)
```
bettinaheim, line 78: Looks like this version still contains the bug I had introduced that type parameters contained in the Resolution are not properly picked up. A ToDo would be good to not forget about it (I already forgot about it until I looking at the code just now).

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -0,0 +1,569 @@
...
+
+                // Do any replacements for type parameters that may be replaced with values in the current dictionary.
+                // This needs to be done first to cover an edge case.
+                foreach (var (typeParam, paramRes) in resolutionDictionary.Where(entry => mayBeReplaced.Contains(entry.Key)))
+                {
+                    // Get all the parameters whose value is dependent on the current resolution's type parameter,
+                    // and update their values with this resolution's value.
+                    foreach (var keyInCombined in mayBeReplaced[typeParam])
+                    {
+                        // Check that we are not doing an excessively constrictive resolution.
```
bettinaheim, line 91:
```suggestion
                        // Check that we are not doing a resolution that constricts what values a type parameter can have.
```

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -0,0 +1,569 @@
...
+    {
+        public static CallGraphEdgeComparer Instance { get; } = new CallGraphEdgeComparer();
+
+        private CallGraphEdgeComparer() { }
+
+        public bool Equals(CallGraphEdge x, CallGraphEdge y) =>
+            x.ParamResolutions.ToImmutableHashSet().SetEquals(y.ParamResolutions.ToImmutableHashSet());
+
+        public int GetHashCode(CallGraphEdge obj) =>
+            obj.ParamResolutions.Aggregate(0, (hash, kvp) => hash ^ kvp.GetHashCode());
```
bettinaheim, line 147: Hmm, the bitwise or here doesn't seem particularly safe. 
Instead of defining and equality comparer, you could go with something like 
`private bool AreSameEdges(CallGraphEdge e1, CallGraphEdge e2) =>
     e1.ParamResolutions.OrderBy(kvp => kvp.Key).SequenceEqual(e2.ParamResolutions.OrderBy(kvp => kvp.Key));
`
and 
` else if (!edges.Any(e => AreSameEdges(e, edge)))`
in RecordDependency instead.

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -0,0 +1,569 @@
...
+        private QsNullable<ImmutableArray<ResolvedType>> RemovePositionFromTypeArgs(QsNullable<ImmutableArray<ResolvedType>> tArgs) =>
+            tArgs.IsValue
+            ? QsNullable<ImmutableArray<ResolvedType>>.NewValue(tArgs.Item.Select(StripPositionInfo.Apply).ToImmutableArray())
+            : tArgs;
+
+        private CallGraphEdge RemovePositionFromEdge(CallGraphEdge edge) =>
+            new CallGraphEdge() { ParamResolutions = edge.ParamResolutions.ToImmutableDictionary(kvp => kvp.Key,
+                kvp => StripPositionInfo.Apply(kvp.Value)) };
+
+        private void RecordDependency(CallGraphNode callerKey, CallGraphNode calledKey, CallGraphEdge edge)
```
bettinaheim, line 417: This method relies on that the edge and all entries in _Dependencies do not contain any positional information. You should add a doc comment to that effect.

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -0,0 +1,569 @@
...
+
+                return cycles;
+            }
+        }
+
+        /// <summary>
+        /// This is a dictionary mapping source nodes to information about target nodes. This information is represented
+        /// by a dictionary mapping target node to the edges pointing from the source node to the target node.
+        /// </summary>
+        private readonly Dictionary<CallGraphNode, Dictionary<CallGraphNode, ImmutableArray<CallGraphEdge>>> _Dependencies =
```
bettinaheim, line 405: It looks like you want to give the promise that all entries in this dictionary have been stripped of any positional information. Is that correct? If so, you should document it.

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -0,0 +1,569 @@
...
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.Quantum.QsCompiler.DataTypes;
+using Microsoft.Quantum.QsCompiler.SyntaxTokens;
+using Microsoft.Quantum.QsCompiler.SyntaxTree;
+
+
+// ToDo: Review access modifiers
+
+namespace Microsoft.Quantum.QsCompiler.Transformations.CallGraphNS
```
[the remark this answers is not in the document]
ScottCarda-MS: Do you have any suggestions? Everything in this namespace is for the sake of the Call Graph class, but we can't call the namespace CallGraph because that is the name of the class. It causes lots of problems if the namespace shares a name with a contained type.

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: Hmm, that wording is confusing to me because all resolutions constrict the type of a type parameter. But I suppose you made the suggestions because the original wording was confusing to you? I had based the original wording off of the what I remembered the error message was for this scenario in other languages, but I can't seem to find this error message now. Could be that I am misremembering.

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: The reason they are structs is so that they work much better as dictionary keys then if they were classes, which is their primary use. I am not seeing where the members of this are expected to be non-null. Can you please point these out to me?

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: The bitwise xor is a particularly well known way of combining hashes, and has the feature of ignoring order, which is desired here. Hashing is much faster than sorting, so it make sense to avoid using OrderBy in favor of this approach.

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: It is really more that CallGraphEdge's ParamResolutions should not have any positional information. I have added that to its summary.

src/QsCompiler/Compiler/FunctorGeneration.cs
```diff
@@ -9,7 +9,7 @@
 using Microsoft.Quantum.QsCompiler.ReservedKeywords;
 using Microsoft.Quantum.QsCompiler.SyntaxTokens;
 using Microsoft.Quantum.QsCompiler.SyntaxTree;
-
+using Microsoft.Quantum.QsCompiler.Transformations;
```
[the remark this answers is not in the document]
ScottCarda-MS: The namespace for the CodeTransformation extension methods was changed to include the `.Transformations` which is accurate as they are transformations and they live in the Transformations project. If a code file want to have access to transformations, they should have to include the Transformations namespace.

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -0,0 +1,569 @@
...
+        /// </summary>
+        public static bool TryCombineTypeResolutionsWithTarget(QsQualifiedName target, out TypeParameterResolutions combined, params TypeParameterResolutions[] resolutions)
+        {
+            if (target == null) throw new ArgumentNullException(nameof(target));
+            var success = TryCombineTypeResolutions(out combined, resolutions);
+            combined = combined.Where(kvp => kvp.Key.Item1.Equals(target)).ToImmutableDictionary();
+            return success;
+        }
+
+        public static bool TryCombineTypeResolutions(out TypeParameterResolutions combined, params TypeParameterResolutions[] resolutionDictionaries)
```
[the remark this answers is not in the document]
ScottCarda-MS: I've added a doc comment for this method that mirrors much of the doc comment for the related method. I makes perfect sense to me that they would both be publicly available, as this this the more generic version and has all of the logic related to performing the task, whereas the other is the more situation-specific method that only has logic related to requiring a target callable.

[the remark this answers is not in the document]
ScottCarda-MS: There will also be an audit of access modifiers closer to the end of the feature development.

src/QsCompiler/Transformations/CallGraph.cs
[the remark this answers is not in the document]
bettinaheim: How about Microsoft.Quantum.QsCompiler.DependencyAnalysis? There are more things down the line that we will need for dependency analysis.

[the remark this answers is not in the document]
ScottCarda-MS: Without the "Transformations"? This namespace also covers the the call graph walker. Should that have a separate namespace from the types in this file?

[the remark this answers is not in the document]
ScottCarda-MS: That might make extra sense if we are going to have the call graph be something that lives beyond a single rewrite step. If that's the case, it should not live in a "Transformations" ns. But that still leaves us with the namespace for the call graph walker.

src/QsCompiler/Transformations/CallGraph.cs
bettinaheim: I see. RemovePositionFromEdge and CombineEdges (via TryCombineTypeResolutions). You don't gain anything by having CallGraphEdge be a struct, since the ParamResolutions don't support support structural equality in any case.

src/QsCompiler/Transformations/CallGraph.cs
bettinaheim: The excessive didn't mean much to me (aside from that I already know what it does). 
maybe something along the lines of "constricts multiple type parameters"?

src/QsCompiler/Transformations/CallGraph.cs
bettinaheim: Please don't claim something is the "well known way of" doing things without at least a reference. 
Well known or not, it is still problematic since a ^ b ^ a = b, a ^ a = 0, etc. 
Please don't close a conversation expressing concern without fully understanding and addressing that concern.

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: Reworded the comment for clarity.

src/QsCompiler/Transformations/CallGraph.cs
bettinaheim: This comment has not been addressed either.

src/QsCompiler/Transformations/CallGraph.cs
bettinaheim: Fully understanding here means to clarify what the reason was for the concern. My initial comment was pretty short and not detailing much of what the concern is, instead giving an alternative suggestion. When you would like to keep something in its current state, then it would be good to ask what exactly the reason was for the concern that wasn't specified in detail before coming to a conclusion.

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: I've converted these to classes, added the constructors that require non-null arguments, and added non-null checks to the public methods.

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: Removing position information has been moved to the `AddDependency` methods. This need to have position information stripped is documented on the CallGraphEdge class.

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: This logic was moved to the CallGraphEdge class. I added some comments there explaining why this method of hashing is well-suited to our needs in this situation.

ScottCarda-MS: Here is where the logic was moved to: https://github.com/microsoft/qsharp-compiler/blob/sccarda/TypeResolution/src/QsCompiler/Transformations/CallGraph.cs#L159.

src/QsCompiler/Transformations/CallGraph.cs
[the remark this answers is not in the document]
bettinaheim: Before making this public, I would want the additional tests that I shared with you a while ago to pass.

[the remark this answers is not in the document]
ScottCarda-MS: I'm not sure which tests you are referring to. Could you please link them here?

src/QsCompiler/Transformations/CallGraph.cs
bettinaheim: I am still somewhat sceptic about the override for Equality and GetHashCode in general, opposed to using something like a helper that I suggested. By far the most likely cases are the case where there is only one or at most two type parameters, so just ordering them instead might be more efficient. I'll leave it up to @irinayat-MS to work with you to assess what makes sense.

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -0,0 +1,678 @@
...
+            AddDependency(
+                callerSpec.Parent, callerSpec.Kind, callerSpec.TypeArguments,
+                calledName, calledKind, calledTypeArgs,
+                typeParamRes);
+        }
+
+        /// <summary>
+        /// Adds a dependency to the call graph using the relevant information from the caller's specialization and the called specialization.
+        /// All parameters are expected to be non-null; the QsNullable parameters may take on their associated null value.
+        /// Throws ArgumentNullException.
```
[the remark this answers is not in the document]
ScottCarda-MS: An ArgumentNullException will be thrown if any of the non-nullable arguments are null.

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -0,0 +1,678 @@
...
+            }
+
+            combined = combinedBuilder.ToImmutable();
+            return success;
+        }
+    }
+
+    /// <summary>
+    /// Contains the information that exists on edges in a call graph.
+    /// The ParamResolutions are expected to be non-null and have all of their position information removed.
```
bettinaheim, line 133: Since the class constructor enforces that, I believe what you wanted to document was that other code can/should indeed rely on the ParamResolutions being non-null, and having position information stripped:
```suggestion
    /// The ParamResolutions are non-null and have all of their position information removed.
```

src/QsCompiler/Transformations/CallGraph.cs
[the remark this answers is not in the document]
bettinaheim: It would be good the edit the doc comment. :)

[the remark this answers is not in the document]
ScottCarda-MS: done

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: done

src/QsCompiler/Transformations/CallGraph.cs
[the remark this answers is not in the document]
bettinaheim: Yes, I mean the tests in this branch and file: https://github.com/microsoft/qsharp-compiler/blob/bd1290a6e19ee9c5ecfe26141dec87768649397d/src/QsCompiler/Tests.Compiler/CallGraphTests.fs#L56 
Specifically, if you have a cycle in the call graph, and are calling a method that is independent on the parent, then it shouldn't matter where you "break" the cycle to get a list. The tests in the file hence check that it indeed doesn't matter. As I mentioned, that is non-trivial to get to work, so my suggestion is to instead keep with a version that requires explicitly specifying the parent.

src/QsCompiler/Transformations/CallGraph.cs
irinayat-MS: As far as I understand, the overrides are necessary to use CallGraphEdge objects as keys in dictionaries. We (Scott and I) reviewed the implementation of both methods and convinced ourselves that they are correct, that is, objects that are considered equal always return equal hashes. As far as quality of the hash function is concerned, I don't have any intuitions. Assuming high quality of C# GetHashCode for tuples and that there are likely to be only a handful of items in ParamResolutions dictionary, XOR should produce a reasonable hash as well...

src/QsCompiler/Transformations/CallGraph.cs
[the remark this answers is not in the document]
ScottCarda-MS: There were two new tests at that location that have been added. Once of the tests highlighted a known issue with the current algorithm, which is that resolution dictionaries need to be resolved internally in addition to being resolved to each other. This issue, along with others should be resolved in their own separate PR to make reviewing easier.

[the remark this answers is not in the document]
ScottCarda-MS: In the mean time, I have made these methods internal.

src/QsCompiler/Transformations/CallGraph.cs
bettinaheim: I believe you forgot to push a change.

src/QsCompiler/Transformations/CallGraph.cs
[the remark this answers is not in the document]
bettinaheim: I believe you forgot to push a change here.

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: Sorry this got resolved on the wrong branch before. It has been cherry-picked into this branch now.

src/QsCompiler/Transformations/CallGraph.cs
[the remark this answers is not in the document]
ScottCarda-MS: Sorry this got resolved on the wrong branch before. It has been cherry-picked into this branch now.

PR #421, opened 2020-04-28
Order built-in rewrite steps according to their priority

bettinaheim:
This change allows compiler extensions to plug in at (almost) any stage during the compilation.

### DIFF
--- a/src/QsCompiler/Compiler/CompilationLoader.cs
+++ b/src/QsCompiler/Compiler/CompilationLoader.cs
@@ -80,7 +80,8 @@ namespace Microsoft.Quantum.QsCompiler
         public static Func<string, Assembly> LoadAssembly { get; set; }
         /// <summary>
         /// Sorts the given list of step according to their relative priority give by getPriority. 
-        /// Throws the corresponding exception if getPriority is null. 
+        /// Throws the corresponding exception if <paramref name="getPriority" /> is <c>null</c>. 
+
         /// </summary>
         internal static void SortRewriteSteps<T>(List<T> steps, Func<T, int> getPriority) =>
             steps?.Sort((fst, snd) => getPriority(snd) - getPriority(fst));
--- a/src/QsCompiler/Compiler/CompilationLoader.cs
+++ b/src/QsCompiler/Compiler/CompilationLoader.cs
@@ -78,7 +78,12 @@ namespace Microsoft.Quantum.QsCompiler
         /// If LoadAssembly is not null, it will be used to load the dlls that are search for classes defining rewrite steps.
         /// </summary>
         public static Func<string, Assembly> LoadAssembly { get; set; }
-
+        /// <summary>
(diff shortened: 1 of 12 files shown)


src/QsCompiler/Compiler/PluginInterface.cs
```diff
@@ -9,6 +9,18 @@
 
 namespace Microsoft.Quantum.QsCompiler
 {
+    /// <summary>
+    /// Lists the priorities for built-in rewrite steps. 
+    /// </summary>
+    public static class RewriteStepPriorities
+    {
+        public const int ControlFlowSubstitutions = 1100;
```
cgranade, line 17: I'd suggest some brief API documentation comments for these.

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -451,49 +456,68 @@ public CompilationLoader(SourceLoader loadSources, ReferenceLoader loadReference
...
 
+            RaiseCompilationTaskEnd("OverallCompilation", "Build");
+
+            // executing the specified rewrite steps
+
+            var steps = new List<(int, Func<QsCompilation>)>();
+
             if (this.Config.ConvertClassicalControl)
             {
                 var rewriteStep = new RewriteSteps.LoadedStep(new ClassicallyControlled(), typeof(IRewriteStep), thisDllUri);
```
cgranade, line 477: Would it be possible to refactor these if blocks into an extension method? They all look very similar.

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -78,7 +78,12 @@ public CompilationTaskEventArgs(CompilationTaskEventType type, string parentTask
         /// If LoadAssembly is not null, it will be used to load the dlls that are search for classes defining rewrite steps.
         /// </summary>
         public static Func<string, Assembly> LoadAssembly { get; set; }
-
+        /// <summary>
+        /// Sorts the given list of step according to their relative priority give by getPriority. 
+        /// Throws the corresponding exception if getPriority is null.
```
cgranade, line 83:
```suggestion
        /// Throws the corresponding exception if <paramref name="getPriority" /> is <c>null</c>. 

```

cgranade: BTW, I've found it useful on IQ# to start using `#nullable enable` to help document what reference types support `null` and which cannot be `null`.

src/QsCompiler/Compiler/PluginInterface.cs
bettinaheim: done

src/QsCompiler/Compiler/CompilationLoader.cs
bettinaheim: Annoyingly not really, since it is not happy to have ref inside a local method. So I don't really get around the second line.

PR #422, opened 2020-04-29
Sccarda/intrinsic fix monomorphization

ScottCarda-MS:
Prevent Intrinsic and certain Built In callables from being monomorphized or removed from the tree.

### DIFF
--- a/src/QsCompiler/Transformations/Monomorphization.cs
+++ b/src/QsCompiler/Transformations/Monomorphization.cs
@@ -41,6 +41,11 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.Monomorphization
 
             var globals = compilation.Namespaces.GlobalCallableResolutions();
 
+            var intrinsicCallableSet = globals
+                .Where(kvp => kvp.Value.Specializations.Any(spec => spec.Implementation.IsIntrinsic))
+                .Select(kvp => kvp.Key)
+                .ToImmutableHashSet();
+
             var entryPoints = compilation.EntryPoints
                 .Select(call => new Request
                 {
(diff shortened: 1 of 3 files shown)


src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -157,31 +162,32 @@ public static QsCompilation Apply(QsCompilation compilation)
...
             }
 
             public class TransformationState
             {
-                public readonly ImmutableDictionary<NonNullable<string>, IEnumerable<QsCallable>> NamespaceCallables;
+                public readonly ILookup<NonNullable<string>, QsCallable> NamespaceCallables;
+                public readonly ImmutableHashSet<QsQualifiedName> ExemptCallableSet;
 
-                public TransformationState(ImmutableDictionary<NonNullable<string>, IEnumerable<QsCallable>> namespaceCallables)
+                public TransformationState(ILookup<NonNullable<string>, QsCallable> namespaceCallables, ImmutableHashSet<QsQualifiedName> exemptCallableSet)
```
bettinaheim, line 177: I like the use of the lookup!

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -359,9 +378,13 @@ private class ExpressionKindTransformation : ExpressionKindTransformation<Transf
                             .Where(kvp => kvp.Key.Item1.Equals(global.Item))
                             .ToImmutableDictionary(kvp => kvp.Key, kvp => kvp.Value);
 
-                        // Create a new identifier
-                        sym = SharedState.GetConcreteIdentifier(global, applicableParams);
-                        tArgs = QsNullable<ImmutableArray<ResolvedType>>.Null;
+                        // We want to skip over intrinsic callables. They will not be monomorphized.
```
bettinaheim, line 381: It's probably worth pointing this out in the doc comment of the transformation.

src/QsCompiler/Transformations/MonomorphizationValidation.cs
```diff
@@ -16,26 +16,52 @@ public class ValidateMonomorphization : SyntaxTreeTransformation<ValidateMonomor
     {
```
bettinaheim, line 16: Looks like the validation is also complete! Great!

src/QsCompiler/Transformations/Monomorphization.cs
ScottCarda-MS: I have added it to the doc comment.

PR #424, opened 2020-04-30
Adding some telemetry info

bettinaheim:


vxfield:
Looking good to me. On Tuesday I'll make sure we are able to fetch the data from our telemetry and hopefully you can add this to the May release.

bettinaheim:
@vxfield I unfortunately didn't get around to doing something fancier, but it would be good to at least take this into the next release. Could you take another look?

PR #426, opened 2020-05-04
Covering some additional cases related to type parameters for intrinic callables

bettinaheim:


src/QsCompiler/Transformations/MonomorphizationValidation.cs
```diff
@@ -53,14 +53,7 @@ private class NamespaceTransformation : NamespaceTransformation<TransformationSt
...
                 // Don't validate intrinsics
-                if (!SharedState.IntrinsicCallableSet.Contains(c.FullName))
-                {
-                    return base.OnCallableDeclaration(c);
-                }
-                else
-                {
-                    return c;
-                }
+                return SharedState.IntrinsicCallableSet.Contains(c.FullName) ? c : base.OnCallableDeclaration(c);
```
ScottCarda-MS, line 56: nit: It would make it a bit easier for me to read if this were left as an if/else.

ScottCarda-MS: Although, getting rid of the negation and swapping the statements works well.

src/QsCompiler/Transformations/MonomorphizationValidation.cs
```diff
@@ -94,10 +87,10 @@ private class TypeTransformation : TypeTransformation<TransformationState>
         {
             public TypeTransformation(SyntaxTreeTransformation<TransformationState> parent) : base(parent, TransformationOptions.NoRebuild) { }
 
-            public override QsTypeKind<ResolvedType, UserDefinedType, QsTypeParameter, CallableInformation> OnTypeParameter(QsTypeParameter tp)
-            {
-                throw new Exception("Type Parameter types must be resolved");
-            }
+            public override QsTypeKind<ResolvedType, UserDefinedType, QsTypeParameter, CallableInformation> OnTypeParameter(QsTypeParameter tp) =>
+                SharedState.IntrinsicCallableSet.Contains(tp.Origin)
```
ScottCarda-MS, line 91: I am not sure why this would be needed. The only time type parameters would come up would be in the implementations of exempt callables, and those are either intrinsic, or not type-parameterized. Could you point out when this logic would be needed?

src/QsCompiler/Core/Dependencies.fs
```diff
@@ -36,7 +36,8 @@ type BuiltIn = {
     /// These should be non-Generic callables only.
     static member RewriteStepDependencies =
         ImmutableHashSet.Create (
-            BuiltIn.RangeReverse.FullName
+            BuiltIn.RangeReverse.FullName,
+            BuiltIn.Length.FullName
```
ScottCarda-MS, line 40: Because `Length` is generic and not intrinsic, it will cause problems later on if it is not monomorphized. Having it here means that the compilation will contain generics all the way through to the end.

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -210,14 +210,9 @@ private class NamespaceTransformation : NamespaceTransformation<TransformationSt
...
                 {
-                    if (elem is QsNamespaceElement.QsCallable call)
-                    {
-                        return SharedState.ExemptCallableSet.Contains(call.Item.FullName);
-                    }
-                    else
-                    {
-                        return true;
-                    }
+                    return elem is QsNamespaceElement.QsCallable call
```
ScottCarda-MS, line 213: nit: It would make it a bit easier for me to read if this were left as an if/else.

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -52,7 +52,7 @@ public static QsCompilation Apply(QsCompilation compilation)
             var globals = compilation.Namespaces.GlobalCallableResolutions();
 
             var exemptCallableSet = globals
```
ScottCarda-MS, line 54: Since this doesn't contain all the exempt callables anymore, it might be more appropriate to rename this, and associated variables/fields as `intrinsicCallableSet` to match the monomorphization validation.

src/QsCompiler/Transformations/MonomorphizationValidation.cs
```diff
@@ -94,10 +87,10 @@ private class TypeTransformation : TypeTransformation<TransformationState>
...
             public TypeTransformation(SyntaxTreeTransformation<TransformationState> parent) : base(parent, TransformationOptions.NoRebuild) { }
 
-            public override QsTypeKind<ResolvedType, UserDefinedType, QsTypeParameter, CallableInformation> OnTypeParameter(QsTypeParameter tp)
-            {
-                throw new Exception("Type Parameter types must be resolved");
-            }
+            public override QsTypeKind<ResolvedType, UserDefinedType, QsTypeParameter, CallableInformation> OnTypeParameter(QsTypeParameter tp) =>
+                SharedState.IntrinsicCallableSet.Contains(tp.Origin) 
+                ? QsTypeKind<ResolvedType, UserDefinedType, QsTypeParameter, CallableInformation>.NewTypeParameter(tp)
+                : throw new Exception("Type Parameter types must be resolved");
```
ScottCarda-MS, line 93: Isn't this being thrown with the generic `Length` still being in the syntax tree after monomorphization?

src/QsCompiler/Core/Dependencies.fs
swernli: I think Length is intrinsic, but intrinsic in the sense of being `body intrinsic` and not in the sense of being part of `Microsoft.Quantum.Intrinsic`. It's defined that way in [Core.qs](https://github.com/microsoft/qsharp-runtime/blob/master/src/Simulation/QsharpCore/Core.qs#L58)

ScottCarda-MS: Ah you are right, it is intrinsic. But that means that it doesn't need to be in this list as it will be picked up with all the other intrinsics at the [beginning of the monomorphization transformation](https://github.com/microsoft/qsharp-compiler/blob/beheim/bugFixes/src/QsCompiler/Transformations/Monomorphization.cs#L54).

src/QsCompiler/Transformations/MonomorphizationValidation.cs
ScottCarda-MS: I see that `Length` is intrinsic, which means it won't trigger this exception, which is good. It being intrinsic also means it probably shouldn't appear in the explicit exemption list in this same file.

src/QsCompiler/Core/Dependencies.fs
swernli: Oh, I think I understand better now. So because Length is intrinsic it will already be skipped and doesn't need to be explicitly listed for skipping again. That seems reasonable to me, and probably explains why it works without this.

bettinaheim: Given the name "RewriteStepDependencies" I think it makes sense to list all dependencies, independent on whether they are intrinsic or not. That also has the advantage that the monomorphization validation properly fails if there is a generic intrinsic dependency that is not properly handled.

src/QsCompiler/Transformations/MonomorphizationValidation.cs
bettinaheim: done

src/QsCompiler/Transformations/Monomorphization.cs
bettinaheim: done

src/QsCompiler/Transformations/Monomorphization.cs
bettinaheim: done

src/QsCompiler/Transformations/MonomorphizationValidation.cs
bettinaheim: See reply on the other comment.

src/QsCompiler/Transformations/MonomorphizationValidation.cs
bettinaheim: If an intrinsic callable is generic, then its type parameters can occur within expressions.

ScottCarda-MS: We had discussed this offline, and you had given the explanation that when generic intrinsics are called in the code, the type of that call contains type parameter types. Please add a comment to the code here with that expatiation to help make this clear for future readings of this.

bettinaheim: done

PR #428, opened 2020-05-06
Prevent entry point driver from being loaded as a QSC reference

samarsha:
This works around a problem caused by including the entry point driver assembly in the C# generation package. The SDK tries to load all of the assemblies, but this one has dependencies that are not included in the package, so loading fails.

Note: This is to feature/azure-quantum-preview instead of features/azure-quantum-preview to match the branch name on qsharp-runtime.

### DIFF
--- a/src/QuantumSdk/Tools/BuildConfiguration/Generate.cs
+++ b/src/QuantumSdk/Tools/BuildConfiguration/Generate.cs
@@ -2,15 +2,27 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Text;
 
 
 namespace Microsoft.Quantum.Sdk.Tools
 {
     public static partial class BuildConfiguration
     {
+        /// <summary>
+        /// Assembly names that are not compatible with the Q# compiler when loaded as a plugin.
+        /// </summary>
+        private static readonly IReadOnlyCollection<string> IncompatibleQscReferences = new[]
+        {
+            // TODO: This is to work around an assembly that is included with the C# generation package, but which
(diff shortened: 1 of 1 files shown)


src/QuantumSdk/Tools/BuildConfiguration/Generate.cs
```diff
@@ -2,15 +2,27 @@
...
 namespace Microsoft.Quantum.Sdk.Tools
 {
     public static partial class BuildConfiguration
     {
+        /// <summary>
+        /// Assembly names that are not compatible with the Q# compiler when loaded as a plugin.
+        /// </summary>
+        private static readonly IReadOnlyCollection<string> IncompatibleQscReferences = new[]
+        {
+            // TODO: This is to work around an assembly that is included with the C# generation package, but which
```
cesarzc, line 19: Is there already an issue that tracks this TODO?

bettinaheim: I don't think this is a todo per se. There are advantages and disadvantages to requiring package containing rewrite steps to self-subscribe. We haven't come to a conclusion on that yet. I'd suggest to just leave it as a comment.

samarsha: I don't think manually blacklisting assembly names is maintainable in the long-term. It is a short-term workaround. So I feel that there should be some TODO for removing this code at some point. The exact mechanism is not that important though. Would a TODO that does not mention a specific solution be better?

cesarzc: I think mentioning a specific solution is ok but the comment could clarify that the mentioned solution is not yet the definitive one.

samarsha: Added link to #435.

PR #431, opened 2020-05-14
Incremental build for Q# projects

anpaz-msft:
With these changes, the QsharpCompile step on Q# projects is only invoked if there are changes on the input files (*.qs) or the arguments sent to the compiler. Otherwise, the step to call the compiler is skipped improving build efficiency.

### DIFF
--- a/src/QuantumSdk/Sdk/Sdk.targets
+++ b/src/QuantumSdk/Sdk/Sdk.targets
@@ -40,11 +40,11 @@
     <Delete Files="@(FilesToRemove)" />
   </Target>  
 
-  <!-- Invokes the Q# command line compiler to build the project. -->
-  <Target Name="QsharpCompile" 
+  <!-- Creates the CommandArgsFile with all the parameters for the Q# compiler. -->
+  <Target Name="PrepareQsharpCompile" 
           Condition="'$(DesignTimeBuild)' != 'true'"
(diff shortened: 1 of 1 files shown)


src/QuantumSdk/Sdk/Sdk.targets
```diff
@@ -40,11 +40,11 @@
     <Delete Files="@(FilesToRemove)" />
   </Target>  
 
-  <!-- Invokes the Q# command line compiler to build the project. -->
-  <Target Name="QsharpCompile" 
+  <!-- Creates the CommandArgsFile with all the parameters for the Q# compiler. -->
+  <Target Name="PrepareQsharpCompile" 
           Condition="'$(DesignTimeBuild)' != 'true'"
           DependsOnTargets="ResolveAssemblyReferences;ResolveQscReferences;ResolveTargetPackage;BeforeQsharpCompile;_CopyFilesMarkedCopyLocal"
```
bettinaheim, line 46: I wonder if it makes more sense if BeforeQsharpCompile runs before PrepareQsharpCompile or after. 
To given context: The BeforeQsharpCompile is just a target that is rooted at a particular spot in the build process for the sole purpose of making it easy for people to plug in at that point - like BeforeCompile, and like BeforeCompile it does nothing by default.

PR #439, opened 2020-05-26
Replacing placeholder ExecutionTarget strings with partner targets

swernli:
This replaces the codename execution targets `Alfred`, `Bruno`, and `Clementine` with the execution target values supported by partners.


src/QuantumSdk/Sdk/Sdk.props
```diff
@@ -32,6 +32,7 @@
     <PackageReference Condition="$(IncludeQsharpCorePackages)" Include="Microsoft.Quantum.Standard" Version="0.10.2001.2831" />
     <PackageReference Condition="'$(ResolvedQsharpOutputType)' == 'QsharpExe'" Include="System.CommandLine" Version="2.0.0-beta1.20213.1" />
     <!-- Packages included for specific execution targets. -->
+    <PackageReference Condition="'$(ResolvedQsharpExecutionTarget)' == 'IonQProcessor'" Include="Microsoft.Quantum.Providers.IonQ" Version="0.10.2001.2831" />
```
swernli, line 35: @bettinaheim I used the same version number here for the provider package as the surrounding references, but I don't think that's correct. What is the expectation around versions in this file? Is it appropriate for me to update all of them, or should I just use a newer one here?

bettinaheim: We should set it to something that exists, and we can update the other version numbers at the same time.

PR #445, opened 2020-05-31
Removing .g.cs files before invoking the Q# compiler

bettinaheim:
The change is needed to fix issues when deleting or renaming files, as well as when retargeting executables.

### DIFF
--- a/src/QuantumSdk/Sdk/Sdk.targets
+++ b/src/QuantumSdk/Sdk/Sdk.targets
@@ -101,10 +101,15 @@
   <!-- Invokes the Q# command line compiler to build the project. -->
   <Target Name="QsharpCompile" 
           Condition="'$(DesignTimeBuild)' != 'true'"
-          Inputs="@(QsharpCompile);@(QscCommandArgsFile)"
+          Inputs="@(QscCommandArgsFile)"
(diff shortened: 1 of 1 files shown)


src/QuantumSdk/Sdk/Sdk.targets
```diff
@@ -101,10 +101,15 @@
   <!-- Invokes the Q# command line compiler to build the project. -->
   <Target Name="QsharpCompile" 
           Condition="'$(DesignTimeBuild)' != 'true'"
-          Inputs="@(QsharpCompile);@(QscCommandArgsFile)"
+          Inputs="@(QscCommandArgsFile)"
           Outputs="$(GeneratedFilesOutputPath)$(PathCompatibleAssemblyName).bson"
           DependsOnTargets="PrepareQsharpCompile"
           BeforeTargets="PrepareCsharpCompileAfterCsharpGeneration">
+    <ItemGroup>
+      <FilesToRemove Include="$(GeneratedFilesOutputPath)**/*.g.cs" />
```
anpaz-msft, line 109: maybe use `_FilesToRemove` to make sure it doesn't collide with some other target?

bettinaheim: agreed.

PR #449, opened 2020-06-05
Prevent result equality comparisons for QPRGen0 runtime capabilities

samarsha:
* To carry the runtime capabilities to the type checker, I created a new record type that contains the symbol tracker (which is what was originally passed to the type checker), the capabilities, and a couple of other things that I moved out of the symbol tracker because it didn't need them.
* Tests include both QPRGen0 and QPRGen1 test cases. Failing QPRGen1 tests are skipped.


src/QsCompiler/Tests.Compiler/TestCases/CapabilityVerification.qs
```diff
@@ -0,0 +1,81 @@
...
+        } else {
+            return true;
+        }
+    }
+
+    operation ResultAsBoolOpSetIf(result : Result) : Bool {
+        mutable b = false;
+        if (result == One) {
+            set b = true;
+        }
```
swernli, line 50: One quirk just to be aware of: somewhere in the compiler, the if-statements get transformed into an if-else on a comparison to zero. So 
```
if (result == One) {
    set b = true;
}
```
will effectively become:
```
if (result == Zero) {}
else {
    set b = true;
}
```
I don't think this invalidates this testing mechanism you've written, and in fact it makes sense to write the Q# code as the user would, since this transformation is a compiler detail and may change over time. Just calling it out so you can have that in the back of your mind for the future!

samarsha: There are operations for control flow like [`ApplyIfOne`](https://github.com/microsoft/qsharp-runtime/blob/master/src/Simulation/QsharpCore/ClassicalControl.qs#L130) that are used by the [classical control transformation](https://github.com/microsoft/qsharp-compiler/blob/master/src/QsCompiler/Transformations/ClassicallyControlled.cs#L408). I thought that is what the compiler would turn those into?

swernli: It does that classical rewrite step only when an `ExecutionTarget` is specified, and when it does, those calls still seem to get inserted so that it is logically equivalent to an if-else with comparison against `Zero`. I'm still not sure exactly where that happens though, as I haven't taken the time to dig into that part of the compiler yet. That's what I meant by it will "effectively become" the other statement, since it doesn't literally get transformed into that Q#. The only reason I know this is from stepping through and being surprised that a sample with `let res == One; if (res == One) { <do something> }` ended up in an else clause handler.

ScottCarda-MS: Having the different test cases is still useful here as these tests should be agnostic to the classical control conversions.

samarsha: OK, makes sense, thanks for the explanation. I agree that it's good to test all the cases since the classical control details can change.

src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
```diff
@@ -23,6 +25,28 @@ open Microsoft.Quantum.QsCompiler.Transformations.QsCodeOutput
...
+      /// True if the parent callable for the current scope is an operation.
+      IsInOperation : bool
+      /// The return type of the parent callable for the current scope.
+      ReturnType : ResolvedType
+      /// The runtime capabilities for the compilation unit.
+      Capabilities : RuntimeCapabilities }
+
+/// <summary>Creates a resolution context for the specialization.</summary>
+/// <exception cref="Exception">Thrown if the specialization's parent does not exist.</exception>
+let CreateResolutionContext (nsManager : NamespaceManager) capabilities (spec : SpecializationDeclarationHeader) =
```
bettinaheim, line 41: I would go with a static member instead of just another function in the module: 

    static member Create (nsManager : NamespaceManager) capabilities (spec : SpecializationDeclarationHeader) : ResolutionContext<'a> =
        match nsManager.TryGetCallable spec.Parent (spec.Parent.Namespace, spec.SourceFile) with
        | Found declaration ->
            { Symbols = SymbolTracker (nsManager, spec.SourceFile, spec.Parent)
              IsInOperation = declaration.Kind = Operation
              ReturnType = StripPositionInfo.Apply declaration.Signature.ReturnType
              Capabilities = capabilities }
        | _ -> failwith "The specialization's parent does not exist."

src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
```diff
@@ -23,6 +25,28 @@ open Microsoft.Quantum.QsCompiler.Transformations.QsCodeOutput
 
 // utils for verifying types in expressions
 
+/// The context used for symbol resolution and type checking.
+type ResolutionContext<'a> =
```
bettinaheim, line 29: Given that this is used also for verifying statements, it it a bit odd to me that it would be found within the Expression module and it looks pretty much like a nested class from within C#. 
I would instead suggest rename the namespace in SymbolTracker.fs to be just Microsoft.Quantum.QsCompiler.SyntaxProcessing and move the ResolutionContext into that file (and possibly rename that file). That seems like the better structure to me.

src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
```diff
@@ -23,6 +25,28 @@ open Microsoft.Quantum.QsCompiler.Transformations.QsCodeOutput
...
+      Symbols : SymbolTracker<'a>
+      /// True if the parent callable for the current scope is an operation.
+      IsInOperation : bool
+      /// The return type of the parent callable for the current scope.
+      ReturnType : ResolvedType
+      /// The runtime capabilities for the compilation unit.
+      Capabilities : RuntimeCapabilities }
+
+/// <summary>Creates a resolution context for the specialization.</summary>
+/// <exception cref="Exception">Thrown if the specialization's parent does not exist.</exception>
```
bettinaheim, line 40: I would make this an ArgumentException rather than just a generic exception. 
```suggestion
/// <exception cref="ArgumentException">
/// Thrown if the given namespace manager does not contain all resolutions, 
/// or if the specialization's parent does not exist in the given namespace manager.
/// </exception>
```

src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
```diff
@@ -23,6 +25,28 @@ open Microsoft.Quantum.QsCompiler.Transformations.QsCodeOutput
...
+    { /// The symbol tracker for the parent callable.
+      Symbols : SymbolTracker<'a>
+      /// True if the parent callable for the current scope is an operation.
+      IsInOperation : bool
+      /// The return type of the parent callable for the current scope.
+      ReturnType : ResolvedType
+      /// The runtime capabilities for the compilation unit.
+      Capabilities : RuntimeCapabilities }
+
+/// <summary>Creates a resolution context for the specialization.</summary>
```
bettinaheim, line 39: As much as I avoid any kind of redundancy in code, I would not do that for doc comments. The doc comment for the symbol tracker would be good to have here as well, since it directly applies to this method: 

 ```suggestion
/// <summary>
/// Creates a resolution context for the specialization.
/// The symbol tracker initialized as part of the context does *not* make a copy of the given namespace manager, 
/// but instead will throw an InvalidOperationException upon accessing its content if that content has been modified 
/// (i.e. if the version number of the namespace manager has changed). 
/// </summary>
```

File: src/QsCompiler/SyntaxProcessor/VerificationTools.fs [hunk 6]
```diff
@@ -74,14 +74,6 @@ type ResolvedType with
             | _ -> false
         this.Exists condition
 
-    /// If the given type supports equality comparison, 
-    /// returns the type of an equality comparison expression as Some (which is always Bool).
-    /// Returns None otherwise.
-    member this.supportsEqualityComparison =
```
bettinaheim, line 80: What was the reason for moving this into the ExpressionVerification.fs? Given that supportsArithmetic, supportsConcatenation, and supportsIteration are here, I'd rather keep it here as a property of the type, and then make the distinction whether an *expression* can contain an equality comparison depending on the execution target as part of the expression verification.

src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
```diff
@@ -246,11 +270,16 @@ let private VerifyConcatenation parent addError (lhsType : ResolvedType, lhsRang
...
-    VerifyIsOneOf expected (ErrorCode.InvalidTypeInEqualityComparison, [baseType |> toString]) addError (baseType, rhsRange) |> ignore
+let private VerifyEqualityComparison context addError (lhsType, lhsRange) (rhsType, rhsRange) =
+    // NOTE: this may not be the behavior that we want (right now it does not matter, since we don't support equality
+    // comparison for any derived type).
+    let argumentError = ErrorCode.ArgumentMismatchInBinaryOp, [toString lhsType; toString rhsType]
+    let baseType = CommonBaseType addError argumentError context.Symbols.Parent (lhsType, lhsRange) (rhsType, rhsRange)
+    match baseType.Resolution with
+    | Result when context.Capabilities = RuntimeCapabilities.QPRGen0 ->
+        addError (ErrorCode.UnsupportedResultComparison, [context.Capabilities.ToString ()]) rhsRange
+    | BigInt | Bool | Double | Int | Pauli | Qubit | Result | String -> ()
```
bettinaheim, line 281: This does not properly treat invalid/missing types. I'd rather go with this: 
```suggestion
    | _ -> 
        let expected (t : ResolvedType) = t.supportsEqualityComparison
        VerifyIsOneOf expected (ErrorCode.InvalidTypeInEqualityComparison, [baseType |> toString]) addError (baseType, rhsRange) |> ignore
```

src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
```diff
@@ -547,10 +576,10 @@ type QsExpression with
     /// recursively computes the corresponding typed expression for a Q# expression.
     /// Calls addDiagnostic on each diagnostic generated during the resolution. 
     /// Returns the computed typed expression. 
-    member this.Resolve (symbols : SymbolTracker<_>) addDiagnostic : TypedExpression =
-        
+    member this.Resolve ({ Symbols = symbols } as context) addDiagnostic : TypedExpression =
```
bettinaheim, line 579: I never quite realized how far F# pattern matching goes! Pretty neat!

src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -543,7 +544,8 @@ type DiagnosticItem =
             | ErrorCode.ExpectingIterableExpr                     -> "The type {0} does not support iteration. Expecting an expression of array type or of type Range."
             | ErrorCode.ExpectingCallableExpr                     -> "The type of the expression must be a function or operation type. The given expression is of type {0}." 
             | ErrorCode.UnknownIdentifier                         -> "No identifier with the name \"{0}\" exists."
-                                                            
+            | ErrorCode.UnsupportedResultComparison               -> "Measurement results cannot be compared because the execution target supports only {0} runtime capabilities."
```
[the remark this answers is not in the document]
samarsha: My philosophy on error messages is that they should be as specific as possible and contain actionable suggestions if possible. Here is my ideal but pretty long error message:

> Measurement results cannot be compared because the execution target is {0}, which supports only {1} runtime capabilities. Choose an execution target with QPRGen1 capabilities to compare results within an if-condition, or Unknown capabilities to compare results anywhere.

I want to include specific details about the project settings in case the user does not remember what execution target they have set (or is not aware that they set one - e.g. if the execution target was set by a project template). At minimum, this would be either the target ID or its runtime capabilities (or both).

The name of the execution target isn't available where the error message is generated right now, so I would have to figure out a way to get it there. Do you think it's worth it?

src/QsCompiler/CompilationManager/TypeChecking.cs
```diff
@@ -473,44 +473,46 @@ internal static void ImportGlobalSymbols(this FileContentManager file, Compilati
...
-            diagnostics.AddRange(messages.Select(msg => Diagnostics.Generate(symbolTracker.SourceFile.Value, msg, statementPos)));
+            var (statement, messages) = build(location, context);
+            diagnostics.AddRange(messages.Select(msg => Diagnostics.Generate(context.Symbols.SourceFile.Value, msg, statementPos)));
             return statement;
         }
 
         /// <summary>
         /// If the current tree node of the given iterator is a using-block intro,
-        /// builds the corresponding using-statement updating the given symbolTracker in the process,
+        /// builds the corresponding using-statement updating the given symbol tracker in the process,
```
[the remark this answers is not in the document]
samarsha: I was conflicted about this because `ResolutionContext` is immutable - so it doesn't really update that per se, but it updates the symbol tracker inside the resolution context.

src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
samarsha: I agree that that's a better place to put it. I'm not sure what to rename the file to - I want to call it `ResolutionContext.fs`, since `ResolutionContext` contains `SymbolTracker` so it feels like the more important type. But that is kind of confusing since `ContextVerification.fs`, which is a different kind of context, is right below it.

src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
samarsha: Thanks for catching this. Do you know where I can add a test for this? Are there existing tests for invalid/missing type handling?

src/QsCompiler/SyntaxProcessor/VerificationTools.fs
samarsha: The reason was because I wanted all of the equality checking to be in the same place (checking supported types, and checking supported capabilities). But now I agree keeping those separate makes more sense. I reverted this change.

src/QsCompiler/DataStructures/Diagnostics.fs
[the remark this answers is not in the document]
swernli: I agree that actionable error messages are best, but like Bettina said the "QPRGen1" term might just be more confusing until we have better documentation for it. That makes it a bit of an annoying thing, because once we do have docs who is going to remember to come back and update these messages? Maybe a middle ground for now would be "Measurement results cannot be compared because the execution target {0} does have that capability. Choose an execution target with additional capabilities or avoid usage of measurement comparison." Then pair it with a nice todo comment and/or issue in the backlog to improve the error message once we know how we are going to document what capabilities exist and which providers have which capabilities.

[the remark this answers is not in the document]
bettinaheim: I am happy both with exposing the execution target, or also remain the runtime capabilities to something like "NoControlFlow" and "LimitedControlFlow". Just the QPRGen* is not really something that should be exposed.

src/QsCompiler/CompilationManager/TypeChecking.cs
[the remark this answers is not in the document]
bettinaheim: I think mentioning the symbol tracker when there is no argument symbolTracker is more confusing. I tend to think about things as "updateable" when they have any kind of handles that can be updated, not necessarily just the top-level ones.

src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
bettinaheim: Ah, naming... For maximal confusion you can name it VerificationContext.fs...
Maybe ResolutionState.fs? or ContextTracking.fs?

src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
bettinaheim: You could add it anywhere to the LocalVerification tests. These tests will check the error message generated for different scenarios, and would thus easily catch e.g. that when something is already invalid, no additional error will be generated by this one.

src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
samarsha: I renamed `ResolutionContext` to `ScopeContext` and `SymbolTracker.fs` to `ScopeTools.fs`.

src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
samarsha: I added tests for this in #465.

src/QsCompiler/DataStructures/Diagnostics.fs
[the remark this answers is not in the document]
samarsha: I added a TODO for including the runtime capabilities in the error message. The message is now:

> The execution target {0} does not support comparing measurement results. Choose an execution target with additional capabilities or use a controlled operation instead.

src/QsCompiler/LanguageServer/EditorState.cs
```diff
@@ -105,6 +105,7 @@ internal bool QsProjectLoader(Uri projectFile, out ProjectInformation info)
             var targetFile = projectInstance.GetPropertyValue("TargetFileName");
             var outputPath = Path.Combine(projectInstance.Directory, outputDir, targetFile);
 
+            var executionTarget = projectInstance.GetProperty("ResolvedQsharpExecutionTarget")?.EvaluatedValue;
```
bettinaheim, line 108: Purely for the sake of using everywhere the same thing: Is there a reason not to use GetPropertyValue?

src/QsCompiler/SyntaxProcessor/ScopeTools.fs
```diff
@@ -280,3 +279,41 @@ type SymbolTracker<'P>(globals : NamespaceManager, sourceFile, parent : QsQualif
...
+    { /// The symbol tracker for the parent callable.
+      Symbols : SymbolTracker<'a>
+      /// True if the parent callable for the current scope is an operation.
+      IsInOperation : bool
+      /// The return type of the parent callable for the current scope.
+      ReturnType : ResolvedType
+      /// The runtime capabilities for the compilation unit.
+      Capabilities : RuntimeCapabilities
+      /// The name of the execution target for the compilation unit.
+      ExecutionTarget : QsNullable<string> }
```
bettinaheim, line 294: The convention right now is indeed that if something is a string  - opposed to a NonNullable<string> - then it is allowed to be null. You could hence go with just string here instead, but either works for me. A pass to get rid of the NonNullable and QsNullable types replacing them with the latest language features would indeed be good.

src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -545,7 +545,11 @@ type DiagnosticItem =
             | ErrorCode.ExpectingIterableExpr                     -> "The type {0} does not support iteration. Expecting an expression of array type or of type Range."
             | ErrorCode.ExpectingCallableExpr                     -> "The type of the expression must be a function or operation type. The given expression is of type {0}." 
             | ErrorCode.UnknownIdentifier                         -> "No identifier with the name \"{0}\" exists."
-            | ErrorCode.UnsupportedResultComparison               -> "Measurement results cannot be compared because the execution target supports only {0} runtime capabilities."
+            | ErrorCode.UnsupportedResultComparison               ->
+                // TODO: When the names of the runtime capabilities are finalized, they can be included in the error
+                // message.
+                "The execution target {0} does not support comparing measurement results. " +
+                "Choose an execution target with additional capabilities or use a controlled operation instead."
```
bettinaheim, line 552: I would get rid of this second half of the message - using controlled (a coherent if) is just usually not used with the same purpose as branching based on measurement result, and the suggestion hence seems odd to me.

src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -544,7 +545,12 @@ type DiagnosticItem =
             | ErrorCode.ExpectingIterableExpr                     -> "The type {0} does not support iteration. Expecting an expression of array type or of type Range."
             | ErrorCode.ExpectingCallableExpr                     -> "The type of the expression must be a function or operation type. The given expression is of type {0}." 
             | ErrorCode.UnknownIdentifier                         -> "No identifier with the name \"{0}\" exists."
-                                                            
+            | ErrorCode.UnsupportedResultComparison               ->
+                // TODO: When the names of the runtime capabilities are finalized, they can be included in the error
+                // message.
+                "The execution target {0} does not support comparing measurement results. " +
+                "Choose an execution target with additional capabilities or use a controlled operation instead."
```
bettinaheim, line 552: I would get rid of this second half of the message - using controlled (a coherent if) is just usually not used with the same purpose as branching based on measurement result, and the suggestion hence seems odd to me.

src/QsCompiler/DataStructures/DataTypes.fs
```diff
@@ -53,6 +53,14 @@ type QsNullable<'T> = // to avoid having to include the F# core in the C# part o
         | Some v -> Value v
         | None -> Null
 
+/// Factories for <see cref="QsNullable"/>.
+module QsNullable =
```
bettinaheim, line 57: Is this specifically only to mark the string for the execution target as nullable? Since we already have NonNullable, I'd suggest to drop this.

src/QsCompiler/DataStructures/Diagnostics.fs
samarsha: It can't be used in all cases (e.g. RUS loops), but `if (M(q1) == One) { X(q2); }` can be replaced with `Controlled X([q1], q2)`, for example.

src/QsCompiler/LanguageServer/EditorState.cs
samarsha: `GetPropertyValue` returns:

> The value of the property with the given name, or an empty string if no property exists with that name.

I used `GetProperty` so that `null` would be the default value instead.

src/QsCompiler/SyntaxProcessor/ScopeTools.fs
samarsha: I think I changed my mind and I would prefer the default value to be "Unspecified" like the Quantum SDK uses, instead of null, so I'll change this to be non-nullable.

src/QsCompiler/DataStructures/Diagnostics.fs
swernli: This came up in another thread, and Chris gave a really good rundown of why the controlled gates are actually quite different than conditionals:
https://github.com/microsoft/qsharp-runtime/pull/231#issuecomment-639585799
It may still be better than nothing in some scenarios, but it really can have rather major affects on overall system state.

samarsha: The intermediate state will be different - but I'm thinking of the [deferred measurement principle](https://en.wikipedia.org/wiki/Deferred_Measurement_Principle), which says that delaying measurements to the end of the circuit (using controlled gates) is equivalent to using classical controls with measurement.

swernli: Oh, that’s interesting! Good point, and likely the best option for folks on platforms that don’t have conditional support, so I can see why you are recommending it. How does that affect what operations you can perform on the qubits in the interim between the controlled gate and the measurement? Just curious for my own edification, as I wonder what other limitations or concerns this pattern comes with.

src/QsCompiler/DataStructures/DataTypes.fs
samarsha: I removed it.

src/QsCompiler/LanguageServer/EditorState.cs
samarsha: Since I decided to make `ScopeContext.ExecutionTarget` non-nullable, I changed this to `GetPropertyValue`.

src/QsCompiler/DataStructures/Diagnostics.fs
bettinaheim: I still don't think having that part of the error message is the right thing to do. For one, as you say, it is not applicable in all situations - if it was Q# would be a whole lot less useful! Then, the people who understand it and will do the right thing don't need it, and for those who don't know that already I don't think the message is actionable since it does require some quantum understanding.

bettinaheim:
@samarsha I realized this PR is going into master - could you please either retarget it to a feature branch or merge in the corresponding tests first, so that any merge to master always also includes the tests? Thanks!

samarsha:
The tests for this feature are in this PR. #465 added tests for equality comparisons in general (not testing runtime capabilities), but I merged that now.

src/QsCompiler/DataStructures/Diagnostics.fs
samarsha: How about:

> The execution target {0} does not support comparing measurement results. Choose an execution target with additional capabilities or avoid result comparisons.

My reasoning is: I want to mention the possibility of changing the execution target, so users know this limitation isn't inherent to Q#. I also want to mention the alternative (not comparing results), so that it doesn't look like the error message is saying the *only* solution is to change the execution target.

samarsha: > How does that affect what operations you can perform on the qubits in the interim between the controlled gate and the measurement? Just curious for my own edification, as I wonder what other limitations or concerns this pattern comes with.

I'm not aware of any limitations on the gates you can apply, but deferred measurement can increase the total number of qubits needed (measured qubits can be reused, but unmeasured qubits need to preserve their state for the whole circuit), and as mentioned loops aren't really possible. It can also increase errors since the unmeasured qubits need to stay coherent for longer.

bettinaheim: The message "The execution target {0} does not support comparing measurement results. Choose an execution target with additional capabilities or avoid result comparisons." sounds good to me.

src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
```diff
@@ -246,11 +246,17 @@ let private VerifyConcatenation parent addError (lhsType : ResolvedType, lhsRang
 /// If a common base type exists, verifies that this base type supports equality comparison, 
 /// adding the corresponding error otherwise.
 /// If one of the given types is a missing type, also adds the corresponding ExpressionOfUnknownType error(s).
-let private VerifyEqualityComparison parent addError (lhsType : ResolvedType, lhsRange) (rhsType : ResolvedType, rhsRange) =
-    // NOTE: this may not be the behavior that we want (right now it does not matter, since we don't support equality comparison for any derived type)
-    let baseType = CommonBaseType addError (ErrorCode.ArgumentMismatchInBinaryOp, [lhsType |> toString; rhsType |> toString]) parent (lhsType, lhsRange) (rhsType, rhsRange)
-    let expected (t : ResolvedType) = t.supportsEqualityComparison
-    VerifyIsOneOf expected (ErrorCode.InvalidTypeInEqualityComparison, [baseType |> toString]) addError (baseType, rhsRange) |> ignore
+let private VerifyEqualityComparison context addError (lhsType, lhsRange) (rhsType, rhsRange) =
+    // NOTE: this may not be the behavior that we want (right now it does not matter, since we don't support equality
```
bettinaheim, line 250: I am not sure what you mean by that actually? What behavior do you think should be different if we were to support equality comparison for operations?

samarsha: I believe that you wrote this comment originally. :) I only reformatted the lines.

bettinaheim: Hu, that comment has just enough information that something may be good thinking about without actually biasing the reader regarding what should be considered... 
While I need to write better comments, let's keep it - I've found in the past that often when I ignored a remark about something needing to be considered, I'd better not. Past me when writing the code tends to have better insights into it than current me.

PR #454, opened 2020-06-10
Call Graph from Entry Points

ScottCarda-MS:
When Entry Points are present in the compilation, the Call Graph Walker should build a call graph solely off of those entry points and their dependencies, recursively. This will result in smaller, more manageable call graphs that contain info on only the callables that are used.

Other changes:
- Part of the call graph data structure was updated to use a lookup rather than a dictionary as it is more in line with how that component is used and this simplifies some of the logic.
- ~Copy constructors are added for the node and edge classes.~
- Unit tests covering the new functionality of the entry point logic on the call graph.


src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -213,16 +225,41 @@ public static CallGraphEdge CombineEdges(CallGraphNode targetNode, params CallGr
...
+        {
+            if (specialization == null) throw new ArgumentNullException(nameof(specialization));
+            Initialize(specialization.Parent, specialization.Kind, specialization.TypeArguments);
+        }
+
+        /// <summary>
+        /// Copy constructor for CallGraphNode objects.
+        /// Throws an ArgumentNullException if node is null.
+        /// </summary>
+        public CallGraphNode(CallGraphNode node)
```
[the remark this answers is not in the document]
samarsha: I think that the `private set` vs `readonly` doesn't affect whether copying is needed; as long as the private set is only used by the constructor then the class is effectively immutable. Although it is nicer if the fields aren't settable. I did have a couple other ideas for making the fields readonly without using expression `throw`s though:

* It looks like there is a way to do it by [passing the fields to `Initialize` as `ref` parameters](https://stackoverflow.com/a/6795985), but it looks kind of weird.
* Alternatively, if you `#nullable enable` in this file, you can remove the null argument checks and make `Initialize` an alternate constructor called with `: this(...)`.

src/QsCompiler/Tests.Compiler/TestCases/TypeParameterResolution.qs
```diff
@@ -157,3 +157,110 @@ namespace Microsoft.Quantum.Testing.TypeParameterResolution {
...
+
+        adjoint(...) { }
+    }
+
+    operation Bar() : Unit { }
+}
+
+// =================================
+
+// Not Called With Entry Point
```
irinayat-MS, line 223: >Not Called With Entry Point [](start = 3, length = 27)

Isn't this test checking the same case as "Unrelated To Entry Point"?

src/QsCompiler/Tests.Compiler/TestCases/TypeParameterResolution.qs
```diff
@@ -157,3 +157,110 @@ namespace Microsoft.Quantum.Testing.TypeParameterResolution {
 
 // Foo.A -> Int
 // Bar.B -> Int
+
+// =================================
+
+// Basic Entry Point
```
irinayat-MS, line 163: Does this test adds value? The "unrelated" test below will check that @EntryPoint does trigger call tree creation and the tests without @EntryPoint already validate various call structures. No?

src/QsCompiler/Tests.Compiler/TestCases/TypeParameterResolution.qs
```diff
@@ -157,3 +157,110 @@ namespace Microsoft.Quantum.Testing.TypeParameterResolution {
...
+    }
+
+    operation Foo() : Unit { }
+
+    operation NotCalled() : Unit { }
+}
+
+// =================================
+
+// Not Called Without Entry Point
```
irinayat-MS, line 238: I'd group it with the other tests that aren't concerned with entry points. Maybe, there is already coverage for the case of disconnected callgraphs? And, if not, should review what other kinds of callgraph topologies aren't covered (e.g. take "Y" topology, when there are multiple roots that eventually call the same method, is it covered? do we care to test it?).

src/QsCompiler/Tests.Compiler/TestCases/TypeParameterResolution.qs
```diff
@@ -157,3 +157,110 @@ namespace Microsoft.Quantum.Testing.TypeParameterResolution {
 
 // Foo.A -> Int
 // Bar.B -> Int
+
+// =================================
```
irinayat-MS, line 161: As a general note on systematic approach to test case generation:
1. Identify interesting callgraph topology equivalence classes without entry point attribute being involved and create corresponding test cases for each.
2. Consider each of these cases and think whether adding the attribute somewhere _inside_ the callgraph creates a new test case. E.g. right now you are not covering the case when there is a caller of the method with entry point attribute. If the compiler supports user code like this, should this caller of the entry point be included into the tree or not?
3. Hopefully, multiple entry point attributes aren't allowed, but if they were, would need to test that as well.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -123,39 +150,51 @@ public override ExpressionKind OnIdentifier(Identifier sym, QsNullable<Immutable
...
+                        else if (SharedState.HasControlledDependency)
                         {
                             kind = QsSpecializationKind.QsControlled;
                         }
 
-                        SharedState.graph.AddDependency(SharedState.spec, global.Item, kind, typeArgs, typeParamRes);
+                        SharedState.Graph.AddDependency(
+                            SharedState.CurrentSpecialization,
+                            global.Item, kind, typeArgs,
+                            typeParamRes);
```
irinayat-MS, line 177: Do we have a style preference for how to split function arguments across multiple lines? E.g. some style guides require to either have all arguments on one line or split one argument per line.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -28,11 +28,37 @@ public static class BuildCallGraph
...
+                    // The current request must be added before it is processed to prevent
+                    // self-references from duplicating on the stack.
+                    walker.SharedState.ResolvedCallableSet.Add(currentRequest);
+
+                    walker.Namespaces.OnCallableDeclaration(currentCallable);
+                }
             }
-            return walker.SharedState.graph;
+            else
+            {
```
irinayat-MS, line 54: code inside `then/else` clauses of this `if` statement differs drastically and is kind of at different level of abstraction. It would potentially create a breeding ground for bugs: what if `OnNamespace` does something that is relevant to the subtree of entrypoints as well? Maybe, not right now, but what if someone adds such functionality in the future -- would they remember to update the code in `then` clause as well?

Could the driving logic of the two graph building strategies be more unified?

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -153,6 +153,18 @@ public CallGraphEdge(TypeParameterResolutions paramResolutions)
...
+        /// <summary>
+        /// Copy constructor for CallGraphEdge objects.
+        /// Throws ArgumentNullException if edge is null or its ParamResolutions field is null
+        /// </summary>
+        public CallGraphEdge(CallGraphEdge edge)
+        {
+            if (edge == null || edge.ParamResolutions == null) throw new ArgumentNullException(nameof(edge));
+
+            // Remove position info from type parameter resolutions
+            ParamResolutions = edge.ParamResolutions;
```
irinayat-MS, line 165: >ParamResolutions = edge.ParamResolutions [](start = 12, length = 40)

The new CallGraphEdge and `edge` now share ParamResolutions object, right? Is that OK? Is ParamResolutions field ever modified by the owning CallGraphEdge? If not, and sharing of the parameter resolutions is expected to be common, would it make sense to add object equality comparison into the Equal before comparing as sets?

irinayat-MS:
>         /// to ensure that the same type parameters will have the same hash.

You are not providing a hash function for the CallGraphEdge anymore, but I presume the position info will still be in the way of `Equals`, right? 

---
Refers to: src/QsCompiler/Transformations/CallGraph.cs:144 in 4a7c6bb. [](commit_id = 4a7c6bbf4c5389cd787508bf93060f89e3a4b0ed, deletion_comment = False)

irinayat-MS:
>         public static CallGraphEdge CombineEdges(CallGraphNode targetNode, params CallGraphEdge[] edges)

From the summary comment I'd name this method something like `CombinePathIntoSingleEdge`. On the other hand, if for correctness of this method it's required to provide a continuous path of edges in `edges` array, why would you need to filter the type resolutions by target? Aren't all type resolutions on a given edge specific to its target call node?

---
Refers to: src/QsCompiler/Transformations/CallGraph.cs:206 in 4a7c6bb. [](commit_id = 4a7c6bbf4c5389cd787508bf93060f89e3a4b0ed, deletion_comment = False)

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -563,8 +617,6 @@ private void RecordDependency(CallGraphNode callerKey, CallGraphNode calledKey,
             QsQualifiedName calledName, QsSpecializationKind calledKind, QsNullable<ImmutableArray<ResolvedType>> calledTypeArgs,
             TypeParameterResolutions typeParamRes)
         {
-            if (typeParamRes == null) throw new ArgumentNullException(nameof(typeParamRes));
```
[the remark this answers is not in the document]
ScottCarda-MS: Yah that null check is redundant of the null check done in the Call Graph Edge constructor.

src/QsCompiler/Transformations/CallGraph.cs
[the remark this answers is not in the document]
ScottCarda-MS: The code now uses the `this` syntax to consolodate constructor logic, rather than using an Initialize method. Using `readonly` keyword instead of `private set` syntax. @samarsha I will want to look more into the `#nullable enable` feature later to see if that can be used to clean up a lot of the null checks in the file. I will open a separate conversation on this PR to track that.

ScottCarda-MS:
Look into using the `#nullable enable` feature to clean up null checks across this feature. @samarsha for the idea.

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -2,11 +2,9 @@
 // Licensed under the MIT License.
```
ScottCarda-MS, line 2: Look into using the #nullable enable feature to clean up null checks across this feature. @samarsha for the idea.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -123,39 +150,51 @@ public override ExpressionKind OnIdentifier(Identifier sym, QsNullable<Immutable
                 {
                     // Type arguments need to be resolved for the whole expression to be accurate
                     // ToDo: this needs adaption if we want to support type specializations
-                    var typeArgs = tArgs;
+                    var typeArgs = QsNullable<ImmutableArray<ResolvedType>>.Null;
```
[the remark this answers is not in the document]
bettinaheim: Here would be an implementation for populating the type argument handles that are currently unused: https://github.com/microsoft/qsharp-compiler/pull/456
The idea would be that the cycle check would be run for each library and the source code separately, and then the full call graph containing the full type information can be build and stored with the compilation (just watch out to keep what is stored with the compilation immutable).

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -507,6 +544,23 @@ bool populateCycles(int currNode)
         private readonly Dictionary<CallGraphNode, Dictionary<CallGraphNode, ImmutableArray<CallGraphEdge>>> _Dependencies =
             new Dictionary<CallGraphNode, Dictionary<CallGraphNode, ImmutableArray<CallGraphEdge>>>();
 
+        /// <summary>
+        /// Represents an empty dependency for a node.
+        /// </summary>
+        private readonly ILookup<CallGraphNode, CallGraphEdge> _EmptyDependency =
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -507,6 +544,23 @@ bool populateCycles(int currNode)
...
+            ImmutableArray<KeyValuePair<CallGraphNode, CallGraphEdge>>.Empty
+            .ToLookup(kvp => kvp.Key, kvp => kvp.Value);
+
+        /// <summary>
+        /// A collection of the nodes in the call graph.
+        /// </summary>
+        public ICollection<CallGraphNode> Nodes => _Dependencies.Keys.ToImmutableHashSet();
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
```diff
@@ -123,16 +127,31 @@ type CallGraphTests (output:ITestOutputHelper) =
         let isMatch = sameLength && expected |> List.forall (fun res1 -> given |> List.exists (fun res2 -> CheckResolutionMatch res1 res2))
         Assert.True(isMatch, "Given resolutions did not match the expected resolutions.")
 
-    let AssertExpectedDepencency name (given : Dictionary<_,ImmutableArray<_>>) expected =
-        let opName = { Namespace = NonNullable<_>.New Signatures.TypeParameterResolutionNS; Name = NonNullable<_>.New name }
+    let AssertExpectedDepencency nameFrom nameTo (given : ILookup<_,_>)expected =
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/QsCompiler/Transformations/CallGraph.cs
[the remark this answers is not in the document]
ScottCarda-MS: The comment was copied from the other constructor. It has been removed.

src/QsCompiler/Transformations/CallGraph.cs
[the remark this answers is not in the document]
ScottCarda-MS: They should be fine accepting `QsNullable<_>.Null` for the `QsNullable` arguments. All other arguments must not be `null`. That is the message I was trying to convey.

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -153,6 +153,18 @@ public CallGraphEdge(TypeParameterResolutions paramResolutions)
                 kvp => StripPositionInfo.Apply(kvp.Value));
         }
 
+        /// <summary>
+        /// Copy constructor for CallGraphEdge objects.
+        /// Throws ArgumentNullException if edge is null or its ParamResolutions field is null
+        /// </summary>
+        public CallGraphEdge(CallGraphEdge edge)
```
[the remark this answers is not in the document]
ScottCarda-MS: So long as the classes are fully immutable, then they don't need copy constructors. The classes could be marked as `sealed` to ensure that any container is not actually dealing with a derived class that is not immutable. Do you want me to remove the copy constructors? And if so, do you want me to make them sealed classes?

src/QsCompiler/Tests.Compiler/TestCases/TypeParameterResolution.qs
ScottCarda-MS: The two Not Called tests cover how the call graph walker will respond to an operation that is never called. This is subtly different from a pair of operations that are called but are not descendants of an Entry Point. The expected behavior for these two situations is the same, but the situations are distinct.

src/QsCompiler/Tests.Compiler/TestCases/TypeParameterResolution.qs
ScottCarda-MS: This tests that the call graph is generated as expected with an entry point. It tests multiple directed descendants as well as indirect descendants. You are correct in pointing out that these two things are already checked by tests that do not use entry points, but this test makes sure that those behaviors work with entry points as well.

ScottCarda-MS: Separated From Entry Point by Specialization tests a special case for entry points. Only when entry points are in the graph do we care about the question of if two nodes are related to each other. This test answers the question of does that relation persist across different specializations of the same callable.

src/QsCompiler/Tests.Compiler/TestCases/TypeParameterResolution.qs
ScottCarda-MS: The category for these tests is "Populate Call Graph", making sure that the call graph gets all the expected nodes and edges across a variety of callable relationships, so making sure that unrelated subgraphs are included in the call graph when there is no entry point seemed consistent with that category to me.

ScottCarda-MS: This comment is similar to my comment in a similar thread on the is PR. The two Not Called tests cover how the call graph walker will respond to an operation that is never called. This is subtly different from a pair of operations that are called but are not related to the rest of the graph. Without an entry point present, these two situations are actually expected to behave differently. A lone node with no relationships should not be present in the call graph.

src/QsCompiler/Transformations/CallGraphWalker.cs
ScottCarda-MS: I am not sure if there is a style rule for this. I have just been going by what looks good and consistent to me. If you can find something in the C# style guide, please point it out to me.

src/QsCompiler/Transformations/CallGraphWalker.cs
ScottCarda-MS: I will take another look.

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: The ParamResolutions is a readonly Immutable Dictionary on the CallGraphEdge class, so I believe it cannot be changed and sharing should be fine. I'm not sure I understand your comment about the object equality. Could you dive into more detail about that?

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -136,7 +134,7 @@ internal static bool TryCombineTypeResolutions(out TypeParameterResolutions comb
     /// </summary>
     public class CallGraphEdge
```
ScottCarda-MS, line 135: In reference to `CombineEdges` method.

ScottCarda-MS: Copied from post by @irinayat-MS: From the summary comment I'd name this method something like `CombinePathIntoSingleEdge`. On the other hand, if for correctness of this method it's required to provide a continuous path of edges in `edges` array, why would you need to filter the type resolutions by target? Aren't all type resolutions on a given edge specific to its target call node?

ScottCarda-MS: Copied from post by @irinayat-MS: Oh! I think I'm getting it! The goal is to select the subset of type resolutions on the path that actually reach the target. Correct?

ScottCarda-MS: Right. Many different unrelated resolutions may happen between node `A` and node `Z`. But this method is just supposed to get an abstraction of the relationship between `A` and `Z`, so it needs to filter out all the unrelated resolutions and combine the related resolutions.

ScottCarda-MS: I like your name suggestion better. It draws attention to the fact that this acts more on a path than random edges.

ScottCarda-MS: I've renamed the method to `CombinePathIntoSingleEdge`, per your suggestion.

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -136,12 +134,12 @@ internal static bool TryCombineTypeResolutions(out TypeParameterResolutions comb
...
     {
-        public TypeParameterResolutions ParamResolutions;
+        public readonly TypeParameterResolutions ParamResolutions;
 
         /// <summary>
         /// Constructor for CallGraphEdge objects.
         /// Strips position info from given resolutions before assigning them to ParamResoluitons
-        /// to ensure that the same type parameters will have the same hash.
+        /// to ensure that the same type parameters will compare as equal.
```
ScottCarda-MS, line 142: @irinayat-MS I've updated the summary.

src/QsCompiler/Tests.Compiler/TestCases/TypeParameterResolution.qs
ScottCarda-MS: For 3: I can't currently write tests that have more that one entry point in them as that is disallowed. It doesn't seem like we are going to allow multiple entry points in the future, but if we do then we will address the changes that will entail then.

For 2: That is a good idea for some tests. I will look into adding such tests.

src/QsCompiler/Tests.Compiler/TestCases/TypeParameterResolution.qs
irinayat-MS: My understanding is that `Bar` in "Unrelated To Entry Point" is not called, similar to `NotCalled` here. Thus, "Unrelated" test covers strictly bigger subset of functionality compared to this one, no? 

---
In reply to: [439689809](https://github.com/microsoft/qsharp-compiler/pull/454#discussion_r439689809) [](ancestors = 439689809)

src/QsCompiler/Tests.Compiler/TestCases/TypeParameterResolution.qs
irinayat-MS: All of the tests prior to this PR can be thought of as "Populate Call Graph in absence of an specified entry point". Consider this collection of tests from the point of view of someone who reads it six months later not as an incremental change but as a whole suite -- does the structure and descriptions of the tests make it clear what is covered and what inputs belong to the same equivalence classes?

---
In reply to: [439693813](https://github.com/microsoft/qsharp-compiler/pull/454#discussion_r439693813) [](ancestors = 439693813)

src/QsCompiler/Transformations/CallGraph.cs
irinayat-MS: I meant that if it's common for object to share ParamResolutions you could compare the objects for equality of pointers first before comparing them as sets.

---
In reply to: [439695463](https://github.com/microsoft/qsharp-compiler/pull/454#discussion_r439695463) [](ancestors = 439695463)

src/QsCompiler/Tests.Compiler/TestCases/TypeParameterResolution.qs
ScottCarda-MS: Because `NotCalled` is never called by anything nor calls anything, it represents a different situation from `Bar` which calls things and would be in the call graph except that it is not related to an Entry Point.

src/QsCompiler/Tests.Compiler/TestCases/TypeParameterResolution.qs
ScottCarda-MS: The previous tests in this file are for testing the uses of the GetAllDependencies method of the call graph. This can be seen from the F# code associated with each test case.

src/QsCompiler/Tests.Compiler/TestCases/TypeParameterResolution.qs
irinayat-MS: Do I understand it correctly, that "NotCalled" won't be in the callgraph even if the entry point wasn't specified, because it has no dependencies in or out? Do you have a test for this among the no-entry-point cases? If this understanding is correct, this test seems to be even more superfluous. That is, unless the logic of building the callgraph from entry points has nothing to do with the logic of building the full graph, in which case I'd expect most of the tests to be duplicated.
However, I'd add a test case with NotCalled marked as entry point to highlight that in this situation the callgraph should remain empty (again, assuming my understanding above is correct).

src/QsCompiler/Tests.Compiler/TestCases/TypeParameterResolution.qs
irinayat-MS: I don't follow. The tests might be invoked differently but isn't the ultimate goal to validate that the correct callgraph has been created? In which case you systemically list the equivalence classes for the possible callgraphs (including multiple connected components, cycles, isolated nodes, duplicated edges, etc.) and then identify which of these cases would yield new equivalence classes if an entry point is added at different positions inside the graphs... This way you can convince yourself and future readers of the code that the test coverage is good and also document the behaviour.

---
In reply to: [440344069](https://github.com/microsoft/qsharp-compiler/pull/454#discussion_r440344069) [](ancestors = 440344069)

src/QsCompiler/Tests.Compiler/TestCases/TypeParameterResolution.qs
ScottCarda-MS: I've added some more tests related to when entry points are called from other operations.

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: Ah I see. That makes for a nice short-circuit. I've added this check to the equality comparison.

src/QsCompiler/Tests.Compiler/TestCases/TypeParameterResolution.qs
ScottCarda-MS: This goal of having a correct call graph is only the goal for the tests under the category "Populating Call Graph". It is not the goal of the test in the category "Get Dependencies". The goal of "Get Dependencies" tests is to check the correct behavior of the GetAllDependencies method.

src/QsCompiler/Tests.Compiler/TestCases/TypeParameterResolution.qs
ScottCarda-MS: Yes your understanding is correct. This is a good suggestion for a test, and I have now added such a test.

src/QsCompiler/Transformations/CallGraphWalker.cs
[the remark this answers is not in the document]
ScottCarda-MS: I don't understand what this comment is addressing or what the linked implementation does. What is the issue you are trying to address here?

src/QsCompiler/Transformations/CallGraphWalker.cs
ScottCarda-MS: (Sorry for the long comment) I have two thoughts on this: 1. My general thinking on concerns of the form "If someone changes X, then they will also have to change Y" is that if someone is changing things, then they will have to be responsible for make that change correctly and adjusting all the appropriate code. We can try to anticipate likely future changes and try to structure the code such that that future change is made easier and less error prone, but we can't plan for all possible future changes. Also planning for future changes that are not likely can make our code base worse, as complicating our code in preparation for a change that will never happen can make it more difficult to implement changes that *do* happen. So my general thinking on these kinds of concerns is: let's keep it simple for now and cross that bridge when we get to it.

ScottCarda-MS: 2. My second thought is that this method does two different things depending on a piece of information found inside the data structure of its argument, which is probably bad design. I think, rather than trying to unify this logic, it would make more sense to just have two separate methods for using this walker: one for libraries, and another for executables.

src/QsCompiler/Transformations/CallGraph.cs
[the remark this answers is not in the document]
samarsha: In that case, should this line be removed from the doc comment?

> Throws ArgumentNullException if any of the non-nullable arguments are null.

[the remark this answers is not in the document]
samarsha: Oh, nevermind, I see what you mean, the constructor will throw so `AddDependency` would throw transitively...

src/QsCompiler/Transformations/CallGraphWalker.cs
irinayat-MS: This was my main concern: the logic in the two branches of the `if` statement is too different and essentially unrelated so it makes the code hard to maintain correctly. If you can handle these two cases in a more encapsulated way, I believe that would improve maintainability of the code.

As for the philosophical discussion re "If someone changes X, then they will also have to change Y". The question is, how will our future selves discover that they need to change Y when they embark on changing X? Does the architecture of the codebase guides them towards a complete change? Does a reasonable search for references to X finds Y so the references can be reviewed and updated if needed? Does the test coverage help with the discoverability of the related bits that must be changed together?  


---
In reply to: [441202023](https://github.com/microsoft/qsharp-compiler/pull/454#discussion_r441202023) [](ancestors = 441202023)

ScottCarda-MS: I've separated the two branches out into separate methods. The original Apply method now just contains the discrimination logic between the two cases.

src/QsCompiler/Transformations/CallGraph.cs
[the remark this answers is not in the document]
ScottCarda-MS: @samarsha Here is the conversation tracking the use of the `#nullable enable` feature: https://github.com/microsoft/qsharp-compiler/pull/454#discussion_r439683548

src/QsCompiler/Transformations/CallGraphWalker.cs
irinayat-MS: I don't see anything about this in https://ms-quantum.visualstudio.com/Quantum%20Program/_wiki/wikis/Quantum-OKRs.wiki/487/Coding-Style either. I suppose, it's a personal call then.

src/QsCompiler/Transformations/CallGraph.cs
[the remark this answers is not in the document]
ScottCarda-MS: I've updated the doc comments for these to try to make that a bit clearer.

src/QsCompiler/Transformations/CallGraph.cs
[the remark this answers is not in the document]
ScottCarda-MS: I've removed the copy constructors and made the classes sealed to prevent mutable fields from being introduced through derived classes.

src/QsCompiler/Transformations/CallGraphWalker.cs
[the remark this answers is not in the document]
bettinaheim: The todo's in the code.

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: I've added the `#nullable enable` feature to the call graph file, although it wasn't able to reduce any of the null checks in the file because public facing members still need to have the null checks.

src/QsCompiler/Transformations/CallGraphWalker.cs
```diff
@@ -50,15 +94,16 @@ public BuildGraph() : base(new TransformationState())
...
+            internal bool IsInCall = false;
+            internal bool HasAdjointDependency = false;
+            internal bool HasControlledDependency = false;
+            internal QsSpecialization CurrentSpecialization;
+            internal CallGraph Graph = new CallGraph();
+            internal IEnumerable<TypeParameterResolutions> TypeParameterResolutions = new List<TypeParameterResolutions>();
+
+            internal bool IsLimitedToEntryPoints = false;
+            internal Stack<QsQualifiedName> RequestStack = null;
+            internal HashSet<QsQualifiedName> ResolvedCallableSet = null;
```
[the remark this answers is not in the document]
ScottCarda-MS: Yes that's right. I've added some comments there to explain their function and this limitation. Thanks!

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -688,6 +735,15 @@ public List<ImmutableArray<CallGraphNode>> GetCallCycles()
             var cycles = new JohnsonCycleFind().GetAllCycles(graph);
             return cycles.Select(cycle => cycle.Select(index => indexToNode[index]).ToImmutableArray()).ToList();
         }
+
+        /// <summary>
+        /// Returns true if the given node is found in the call graph, false otherwise.
+        /// </summary>
+        public bool ContainsNode(CallGraphNode node)
+        {
+            if (node == null) throw new ArgumentNullException(nameof(node));
```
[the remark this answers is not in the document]
ScottCarda-MS: Oh thanks I missed that one. I've added that to the documentation.

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
```diff
@@ -675,14 +719,193 @@ type CallGraphTests (output:ITestOutputHelper) =
...
+            "Baz", [ ]
+        ]
+        |> List.map (fun x -> AssertExpectedDirectDependencies (fst x) (snd x) graph)
+        |> ignore
+
+    [<Fact>]
+    [<Trait("Category","Populate Call Graph")>]
+    member this.``Entry Point No Descendants`` () =
+        let graph = CompileTypeParameterResolutionTestWithExe 17
+        Assert.True(graph.Count = 0, "Expected call graph to be empty.")
```
samarsha, line 871: Why is this an empty graph instead of a graph with one node?

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
```diff
@@ -675,14 +719,193 @@ type CallGraphTests (output:ITestOutputHelper) =
...
+        ]
+        |> List.map (fun x -> AssertExpectedDirectDependencies (fst x) (snd x) graph)
+        |> ignore
+
+        let BarNode =
+            CallGraphNode(
+                { Namespace = NonNullable<_>.New Signatures.TypeParameterResolutionNS; Name = NonNullable<_>.New "Bar" },
+                QsSpecializationKind.QsBody, QsNullable<ImmutableArray<ResolvedType>>.Null)
+
+        Assert.False(graph.ContainsNode(BarNode), "Expected call graph to not include Bar.")
```
samarsha, line 908: Can this use `AssertNotInGraph`?

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
```diff
@@ -675,14 +719,193 @@ type CallGraphTests (output:ITestOutputHelper) =
...
+
+        let AssertExpectedDirectDependencies nameFrom nameToList (givenGraph : CallGraph) =
+            let strToNode name =
+                let nodeName = { Namespace = NonNullable<_>.New Signatures.TypeParameterResolutionNS; Name = NonNullable<_>.New name }
+                CallGraphNode(nodeName, QsSpecializationKind.QsBody, QsNullable<ImmutableArray<ResolvedType>>.Null)
+            let dependencies = givenGraph.GetDirectDependencies (strToNode nameFrom)
+            for nameTo in nameToList do
+                let expectedNode = strToNode nameTo
+                Assert.True(dependencies.Contains(expectedNode) && dependencies.[expectedNode].Any(),
+                    sprintf "Expected %s to take dependency on %s." nameFrom nameTo)
```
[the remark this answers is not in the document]
ScottCarda-MS: What is the name of the other function? I think the link just points to the same function.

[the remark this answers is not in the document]
ScottCarda-MS: Oh I see the difference now...

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
ScottCarda-MS: Call graphs only contain relationships between callables. If a callable is not referenced and is not being referenced anywhere, it can't be in the graph. Here the entry point does not have any relationships, and so cannot be in the graph.

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
ScottCarda-MS: Good thinking! Done.

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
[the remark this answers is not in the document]
ScottCarda-MS: Yes this is a duplicate, it has been removed here.

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
```diff
@@ -123,16 +152,31 @@ type CallGraphTests (output:ITestOutputHelper) =
...
         AssertExpectedResolutionList expected given
 
+    let AssertExpectedDirectDependencies nameFrom nameToList (givenGraph : CallGraph) =
+        let strToNode name =
+            let nodeName = { Namespace = NonNullable<_>.New Signatures.TypeParameterResolutionNS; Name = NonNullable<_>.New name }
+            CallGraphNode(nodeName, QsSpecializationKind.QsBody, QsNullable<ImmutableArray<ResolvedType>>.Null)
+        let dependencies = givenGraph.GetDirectDependencies (strToNode nameFrom)
+        for nameTo in nameToList do
+            let expectedNode = strToNode nameTo
+            Assert.True(dependencies.Contains(expectedNode) && dependencies.[expectedNode].Any(),
```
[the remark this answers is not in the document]
ScottCarda-MS: Lookups can't be created with keys that map to empty enumerations, so the second check here is redundant. It must have been left over from when this was a dictionary. It has been removed.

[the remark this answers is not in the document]
samarsha: OK, that makes sense. There is also an assert for `Any()` on line 788 if you want to remove that one too.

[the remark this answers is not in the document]
ScottCarda-MS: Ah good catch. Done

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
samarsha: Based on our conversation, it looks like an ambiguity in the design doc, so I'll move this comment to the design doc PR.

src/QsCompiler/Transformations/CallGraphWalker.cs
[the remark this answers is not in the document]
ScottCarda-MS: I'm going to wait to make changes on this topic until it can be worked into the design doc for the call graph. I will create a separate task for this and its implementation will be a separate PR. For now, this PR needs to be merged in so that work can continue on the feature.

PR #463, opened 2020-06-15
Some tools for adding attributes

bettinaheim:
I was looking into a random project on the weekend and ended up writing some tools to add attributes to callables in a compilation. Seems like something handy to take in.


src/QsCompiler/CompilationManager/CompilationUnit.cs
```diff
@@ -44,7 +44,7 @@ public class Headers
                     ?? ImmutableArray<(SpecializationDeclarationHeader, SpecializationImplementation)>.Empty;
             }
 
-            internal Headers(NonNullable<string> source, IEnumerable<QsNamespace> syntaxTree) : this (
+            public Headers(NonNullable<string> source, IEnumerable<QsNamespace> syntaxTree) : this (
```
cgranade, line 47: If reasonable, I'd suggest an API documentation comment here would be helpful now that this constructor is public.

src/QsCompiler/CompilationManager/DiagnosticTools.cs
```diff
@@ -19,7 +19,7 @@ public static class DiagnosticTools
         /// Returns the line and character of the given position as tuple without verifying them.
         /// Throws an ArgumentNullException if the given position is null.
         /// </summary>
-        internal static Tuple<int, int> AsTuple(Position position) => 
+        public static Tuple<int, int> AsTuple(Position position) =>
```
cgranade, line 22: Perhaps a bigger change than this PR, but since this method enforces non-nullability of `position`, it may be helpful to add a `#nullable enable` line at the top of the file.

src/QsCompiler/Compiler/PluginInterface.cs
```diff
@@ -82,6 +84,25 @@ public struct Diagnostic
             /// The position is null if the diagnostic is not caused by a piece of source code. 
             /// </summary>
             public Tuple<int, int> End { get; set; }
+
+            /// <summary>
+            /// Initializes a new diagnostic. 
+            /// If a diagnostic generated by the Q# compiler is given as argument, the values are initialized accordingly. 
+            /// </summary>
+            public static Diagnostic Create(VS.Diagnostic d = null, Stage stage = Stage.Unknown) =>
```
cgranade, line 92: Here too, since `d` is intended to be nullable, it may be helpful to denote that by using `#nullable enable` with a nullable reference annotation:
```suggestion
            public static Diagnostic Create(VS.Diagnostic? d = null, Stage stage = Stage.Unknown) => 
```

src/QsCompiler/Compiler/PluginInterface.cs
```diff
@@ -82,6 +84,25 @@ public struct Diagnostic
...
+            /// If a diagnostic generated by the Q# compiler is given as argument, the values are initialized accordingly. 
+            /// </summary>
+            public static Diagnostic Create(VS.Diagnostic d = null, Stage stage = Stage.Unknown) => 
+                d == null ? new Diagnostic() : new Diagnostic
+                {
+                    Severity = 
+                        d.Severity == VS.DiagnosticSeverity.Error ? DiagnosticSeverity.Error :
+                        d.Severity == VS.DiagnosticSeverity.Warning ? DiagnosticSeverity.Warning :
+                        d.Severity == VS.DiagnosticSeverity.Information ? DiagnosticSeverity.Info :
+                        DiagnosticSeverity.Hidden,
```
cgranade, line 99:
```suggestion
                    Severity = d.Severity switch
                    {
                        VS.DiagnosticSeverity.Error => DiagnosticSeverity.Error,
                        VS.DiagnosticSeverity.Warning => DiagnosticSeverity.Warning,
                        VS.DiagnosticSeverity.Information => DiagnosticSeverity.Info,
                        _ => DiagnosticSeverity.Hidden
                    },
```

src/QsCompiler/Transformations/Attributes.cs
```diff
@@ -0,0 +1,78 @@
...
+using Microsoft.Quantum.QsCompiler.DataTypes;
+using Microsoft.Quantum.QsCompiler.SyntaxTree;
+
+
+namespace Microsoft.Quantum.QsCompiler.Transformations
+{
+    using CallablePredicate = Func<QsCallable, bool>;
+    using AttributeId = QsNullable<UserDefinedType>;
+    using QsRangeInfo = QsNullable<Tuple<QsPositionInfo, QsPositionInfo>>;
+    using AttributeSelection = IEnumerable<(QsDeclarationAttribute, Func<QsCallable, bool>)>;
```
cgranade, line 14: If I understand right, the shared state here is a collection of filters that describe when to apply each new attribute to a given callable?

src/QsCompiler/Transformations/Attributes.cs
```diff
@@ -0,0 +1,75 @@
...
+
+
+    public class Attributes
```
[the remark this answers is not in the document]
bettinaheim: You are right - I completely forgot! Thanks!

src/QsCompiler/Transformations/Attributes.cs
```diff
@@ -0,0 +1,75 @@
...
+            this.Namespaces = new NamespaceTransformation(this);
+            this.Statements = new Core.StatementTransformation<AttributeSelection>(this, Core.TransformationOptions.Disabled);
+            this.StatementKinds = new Core.StatementKindTransformation<AttributeSelection>(this, Core.TransformationOptions.Disabled);
+            this.Expressions = new Core.ExpressionTransformation<AttributeSelection>(this, Core.TransformationOptions.Disabled);
+            this.ExpressionKinds = new Core.ExpressionKindTransformation<AttributeSelection>(this, Core.TransformationOptions.Disabled);
+            this.Types = new Core.TypeTransformation<AttributeSelection>(this, Core.TransformationOptions.Disabled);
+        }
+
+
+        // public static methods
```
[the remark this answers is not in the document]
bettinaheim: Yes, I think that would make sense here - no point in exposing the transformation itself I think. Let me do that.

src/QsCompiler/CompilationManager/CompilationUnit.cs
bettinaheim: Yes, agreed.

src/QsCompiler/CompilationManager/DiagnosticTools.cs
bettinaheim: Indeed we should make a bigger pass (and breaking change) to switch to using the nullable feature instead of some of our custom typing/checks. I'd rather address that then across the entire code base rather than just here.

src/QsCompiler/Compiler/PluginInterface.cs
bettinaheim: Same here.

src/QsCompiler/Transformations/Attributes.cs
bettinaheim: Correct.

src/QsCompiler/CompilationManager/DiagnosticTools.cs
samarsha: This case doesn't look like a breaking change since the `position` parameter doesn't use either `QsNullable` or `NonNullable`, so its type would stay the same.

`#nullable enable` seems to be made for cases like this, so that you can incrementally convert an existing codebase to the new nullable system without doing it all at once. We will need a full pass too but that pass will be easier if there are already files that have been converted earlier.

cgranade: As @samarsha notes, adding `#nullable enable` is by design non-breaking aside from the possible introduction of new warnings; that statement will not introduce new errors so as to allow for progressively adopting nullability checks.

src/QsCompiler/Transformations/Attributes.cs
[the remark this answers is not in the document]
bettinaheim: Hm, the full name is Microsoft.Quantum.QsCompiler.Transformation.Attributes and correspondingly it should be a fair expectation that it is related to contain attribute related tools for transforming an existing compilation. I am not sure what name you had in mind. Do you have a concrete suggestion?

src/QsCompiler/Transformations/Attributes.cs
```diff
@@ -0,0 +1,75 @@
...
+namespace Microsoft.Quantum.QsCompiler.Transformations
+{
+    using CallablePredicate = Func<QsCallable, bool>;
+    using AttributeId = QsNullable<UserDefinedType>;
+    using QsRangeInfo = QsNullable<Tuple<QsPositionInfo, QsPositionInfo>>;
+    using AttributeSelection = IEnumerable<(QsDeclarationAttribute, Func<QsCallable, bool>)>;
+
+
+    public class Attributes
+    : Core.SyntaxTreeTransformation<AttributeSelection>
```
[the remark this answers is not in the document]
bettinaheim: I made a TransformationState class.

src/QsCompiler/Transformations/Attributes.cs
```diff
@@ -42,37 +30,66 @@ private Attributes(IEnumerable<(QsDeclarationAttribute, CallablePredicate)> attr
...
+            new AddAttributes(attributes.Select(att => (att, (CallablePredicate)null))).Namespaces.OnNamespace(ns);
 
-        // helper classes
 
-        private class NamespaceTransformation
-        : Core.NamespaceTransformation<AttributeSelection>
+        /// <summary>
+        /// Transformation to add attributes to an existing compilation.
+        /// </summary>
+        private class AddAttributes
```
ScottCarda-MS, line 54: This is a more descriptive name than `Attributes` for the top-level class. Since this class is private it is less important that this has a descriptive name than for the public class. I would suggest swapping the names for this and the top-level class.

src/QsCompiler/Transformations/Attributes.cs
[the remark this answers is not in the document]
ScottCarda-MS: It is hard for me to come up with a better name for this class when I am still unclear as to the use for its public static members. If you could add a description for each of those, it might help me have a better picture for what this class is for. My concern with `Attributes` is that it is fairly generic and non-descriptive about what the class is in relation to the concept of Q# Attributes.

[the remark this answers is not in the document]
ScottCarda-MS: In lieu of better understanding of the static methods, perhaps `AttributeUtils`?

src/QsCompiler/Transformations/Attributes.cs
```diff
@@ -0,0 +1,95 @@
...
+            new AddAttributes(attributes.Select(att => (att, (CallablePredicate)null))).Apply(compilation);
+
+        public static QsNamespace AddToCallables(QsNamespace ns, QsDeclarationAttribute attribute, CallablePredicate predicate = null) =>
+            new AddAttributes(new[] { (attribute, predicate) }).Namespaces.OnNamespace(ns);
+
+        public static QsNamespace AddToCallables(QsNamespace ns, params (QsDeclarationAttribute, CallablePredicate)[] attributes) =>
+            new AddAttributes(attributes).Namespaces.OnNamespace(ns);
+
+        public static QsNamespace AddToCallables(QsNamespace ns, params QsDeclarationAttribute[] attributes) =>
+            new AddAttributes(attributes.Select(att => (att, (CallablePredicate)null))).Namespaces.OnNamespace(ns);
```
cgranade, line 48: It looks like each of these add attributes based on callable predicates? Based on the usecase at microsoft/qsharp-runtime#253, though, it may help to have some static methods or extension methods for make it easier to add attributes to particular callables, given C# values representing literal inputs to those attributes.

For instance, the following LINQ call is used to add a sequence of attributes each of signature `(String, String)` given an `IDictionary<string, string>`:

```C#
        internal static IAttributeBuilder<T> WithDocumentationAttributesFromDictionary<T>(
            this IAttributeBuilder<T> builder, string attributeName, IDictionary<string, string> items
        ) =>
            items
            .Aggregate(
                builder,
                (acc, item) => acc.WithDocumentationAttribute(
                    attributeName,
                    // The following populates all of the metadata needed for a
                    // Q# literal of type (String, String).
                    new TypedExpression(
                        QsExpressionKind<TypedExpression, Identifier, ResolvedType>.NewValueTuple(
                            ImmutableArray.Create(
                                item.Key.AsLiteralExpression(),
                                item.Value.AsLiteralExpression()
                            )
                        ),
                        ImmutableArray<Tuple<QsQualifiedName, NonNullable<string>, ResolvedType>>.Empty,
                        ResolvedType.New(
                            QsTypeKind<ResolvedType, UserDefinedType, QsTypeParameter, CallableInformation>.NewTupleType(
                                ImmutableArray.Create(
                                    ResolvedType.New(QsTypeKind<ResolvedType, UserDefinedType, QsTypeParameter, CallableInformation>.String),
                                    ResolvedType.New(QsTypeKind<ResolvedType, UserDefinedType, QsTypeParameter, CallableInformation>.String)
                                )
                            )
                        ),
                        new InferredExpressionInformation(false, false),
                        QsNullable<Tuple<QsPositionInfo, QsPositionInfo>>.Null
                    )
                )
            );
```

The vast majority of that expression is constructing a `TypedExpression` from two C# strings, suitable for use in constructing a `QsDeclarationAttribute`, such that new SyntaxGeneration methods may help as well.

src/QsCompiler/Transformations/Attributes.cs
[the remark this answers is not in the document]
bettinaheim: That works.

src/QsCompiler/Transformations/Attributes.cs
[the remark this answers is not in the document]
bettinaheim: added a description

src/QsCompiler/CompilationManager/DiagnosticTools.cs
bettinaheim: It still seems a bit orthogonal to this PR, since it is not related to any change. I think that's a good thing to do, but it would still make sense to make a pass on a larger scale.

src/QsCompiler/Transformations/Attributes.cs
bettinaheim: I added a method to build arguments with arbitrary many strings.

PR #467, opened 2020-06-17
Symbols are restricted to include only valid characters

DmitryVasilevsky:


src/QsCompiler/TextProcessor/ParsingPrimitives.fs
```diff
@@ -38,6 +38,25 @@ let internal isSymbolStart c = isAsciiLetter c || c = '_'
 /// returns true if the given char is a valid symbol continuation - i.e. if it is an ascii letter, a digit, or an underscore
 let internal isSymbolContinuation c = isAsciiLetter c || isDigit c || c = '_'
 
+/// returns true if the given char is a valid start symbol for an identifier: Underscore or a Unicode character of classes Lu, Ll, Lt, Lm, Lo, Nl.
+/// https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/lexical-structure#identifiers
+let internal isValidIdentifierStart c =
```
bettinaheim, line 43: Taking a closer look, I think it would actually be good to just replace isSymbolStart (in this file) with this implementation.

src/QsCompiler/TextProcessor/ParsingPrimitives.fs
```diff
@@ -38,6 +38,25 @@ let internal isSymbolStart c = isAsciiLetter c || c = '_'
...
+    c = '_' ||
+    System.Char.IsLetter(c) || // Covers Lu, Ll, Lt, Lm, Lo
+    System.Char.GetUnicodeCategory(c) = System.Globalization.UnicodeCategory.LetterNumber // Nl
+
+/// returns true if the given char is a valid part symbol for an identifier: Unicode character of classes Lu, Ll, Lt, Lm, Lo, Nl; Nd; Pc; Mn, Mc; Cf.
+/// https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/lexical-structure#identifiers
+let internal isValidIdentifierPart c =
```
bettinaheim, line 50: Taking a closer look, I think it would actually be good to just replace isSymbolContinuation (in this file) with this implementation.

src/QsCompiler/TextProcessor/SyntaxBuilder.fs
```diff
@@ -335,7 +335,10 @@ let internal symbolNameLike errCode =
         let isReserved = name.StartsWith "__" && name.EndsWith "__" || InternalUse.CsKeywords.Contains name
         let isCsKeyword = SyntaxFacts.IsKeywordKind (SyntaxFacts.GetKeywordKind name)
         let moreThanUnderscores = name.TrimStart('_').Length <> 0
+        let isValidStart = name.Chars 0 |> isValidIdentifierStart
+        let isValidPart = name |> Seq.forall( isValidIdentifierPart )
         if isCsKeyword || isReserved then buildError (preturn range) ErrorCode.InvalidUseOfReservedKeyword >>% None
+        else if not isValidStart || not isValidPart then buildError (preturn range) ErrorCode.InvalidCharacterInIdentifier >>% None
```
bettinaheim, line 341: To still get the nice descriptive error message despite using the more restrictive versions more widely (my previous two comments), I would suggest to do something like this: 
```
    let identifier = 
        let id = IdentifierOptions(isAsciiIdStart = isSymbolStart, isAsciiIdContinue = isSymbolContinuation) |> identifier
        let permissiveId = 
            let permissiveSymbolStart c = isAsciiLetter c || c = '_'
            let permissiveSymbolContinuation c = isAsciiLetter c || isDigit c || c = '_'
            IdentifierOptions(isAsciiIdStart = permissiveSymbolStart, isAsciiIdContinue = permissiveSymbolContinuation) |> identifier
        let idLike = (id .>>. preturn false) <|> (permissiveId .>>. preturn true) 
        getPosition .>>. idLike .>>. getPosition |>> fun ((p1, name), p2) -> name, (p1,p2)
    let whenValid ((name : string, containsInvalidChar), range) = 
        ...
```
and then this line simply becomes
```suggestion
        elif containsInvalidChar then buildError (preturn range) ErrorCode.InvalidCharacterInIdentifier >>% None
```

bettinaheim: As you pointed out, the isAsciiStart and isAsciiContinuation indeed only apply to ascii characters... So no changes should be required here.

src/QsCompiler/TextProcessor/ParsingPrimitives.fs
DmitryVasilevsky: Yes, this should work.

src/QsCompiler/TextProcessor/ParsingPrimitives.fs
DmitryVasilevsky: Yes, this should work.

DmitryVasilevsky:
Looks like I no longer need the new error I introduced. I'll remove it.

src/QsCompiler/TextProcessor/ParsingPrimitives.fs
```diff
@@ -32,12 +32,25 @@ let private buildQsExpression kind (range : Position * Position) = (kind, range)
...
+/// https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/lexical-structure#identifiers
+let internal isSymbolStart c =
+    c = '_' ||
+    System.Char.IsLetter(c) || // Covers Lu, Ll, Lt, Lm, Lo
+    System.Char.GetUnicodeCategory(c) = System.Globalization.UnicodeCategory.LetterNumber // Nl
 
+/// returns true if the given char is a valid part symbol for an identifier: Unicode character of classes Lu, Ll, Lt, Lm, Lo, Nl; Nd; Pc; Mn, Mc; Cf.
+/// https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/lexical-structure#identifiers
+let internal isSymbolContinuation c =
+    System.Char.IsLetter(c) || // Covers Lu, Ll, Lt, Lm, Lo
```
bettinaheim, line 45: Do you have any idea how expensive System.Char.GetUnicodeCategory is? 
Since this is used a lot during processing, depending on that it may possibly make sense to split out "c ='_'" as a separate case. I suppose all the other checks like e.g. whether c is a whitespace character are not cheaper to split out. But I haven't looked in detail what is quick and what isn't.

src/QsCompiler/TextProcessor/SyntaxBuilder.fs
```diff
@@ -329,7 +329,11 @@ let internal buildTupleItem validSingle bundle errCode missingCode fallback cont
 /// In order to guarantee correct whitespace management, the name needs to be parsed as a term.
 let internal symbolNameLike errCode = 
     let identifier = 
-        let id = IdentifierOptions(isAsciiIdStart = isSymbolStart, isAsciiIdContinue = isSymbolContinuation) |> identifier
+        let id = IdentifierOptions(
```
bettinaheim, line 332: Ah, so this actually worked; by adding it to the precheck?

DmitryVasilevsky: Yes. Unsupported symbols aren't treated as identifiers. If you have unsupported symbol where identifier is supposed to be you'll get "expected identifier".

---
In reply to: [442525499](https://github.com/microsoft/qsharp-compiler/pull/467#discussion_r442525499) [](ancestors = 442525499)

src/QsCompiler/TextProcessor/ParsingPrimitives.fs
DmitryVasilevsky: I will do some tests in C# - this is calling same framework functions. I wouldn't expect GetUnicodeCategory to be too expensive - it is likely checking ranges.

---
In reply to: [442521582](https://github.com/microsoft/qsharp-compiler/pull/467#discussion_r442521582) [](ancestors = 442521582)

DmitryVasilevsky: A revelation: char.IsLetter() calls into the same function to get Unicode category as char.GetUnicodeCategory does. So char.IsLetter is slower. GetUnicodeCategory consists of 3 table clever lookups. It's only 2.5 times slower than this FAKE code:
                if (char.IsLetter(c)) {
                    result = System.Globalization.UnicodeCategory.NonSpacingMark;
                } else {
                    result = System.Globalization.UnicodeCategory.SpacingCombiningMark;
                }
I believe that there's no need to split c='_'.

---
In reply to: [442521582](https://github.com/microsoft/qsharp-compiler/pull/467#discussion_r442521582) [](ancestors = 442521582)

bettinaheim: Thanks for looking into this!

PR #468, opened 2020-06-18
Use Uri.LocalPath instead of Uri.AbsolutePath for file IDs

### LINKED ISSUE #333: Bug found when cloned repo is in folders with space in their names
**Describe the bug**

Bug found while trying to build library files when cloned repo is in folders with space in their names. 

**To Reproduce**

1. Edit a .qs file in the QuantumLibraries/Standard/src that's cloned into local folders with space in their names. In this case, "Quantum Computing" and "Doc PR" both have spaces. 
2. dotnet build
3. See error

image

samarsha:
For a file path containing spaces, like `C:\Foo Bar\Baz.qs`, that was converted to a URI `file://C:/Foo%20Bar/Baz.qs`:

* `AbsolutePath` returns `C:/Foo%20Bar/Baz.qs`
* `LocalPath` returns the original `C:\Foo Bar\Baz.qs`

Since the file ID is used as a path in C# generation with the `#line` annotations, using `AbsolutePath` was causing Q# projects to fail to build if `EmbedAllSources` was set to true in the project file, because the C# compiler would follow those annotations. The path given by `AbsolutePath` is not equivalent to the original file path, so the file couldn't be found. I think using `LocalPath` for all file IDs makes more sense here.

A corresponding PR microsoft/qsharp-runtime#263 is needed at the same time. The tests for this change are also there. Fixes #333.


src/QsCompiler/TestTargets/Libraries/Library1/Library1.csproj
```diff
@@ -6,7 +6,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Quantum.Simulators" Version="0.11.2006.411-beta" />
+    <PackageReference Include="Microsoft.Quantum.Simulators" Version="0.11.2006.1803-alpha" />
```
ScottCarda-MS, line 9: We probably don't want this version going into the repo, right? This is your local Nuget isn't it?

samarsha: [I responded here.](https://github.com/microsoft/qsharp-runtime/pull/263#discussion_r442381629)

filipw:
this is a great idea. I started doing the exact same thing as part of https://github.com/microsoft/qsharp-compiler/pull/356 but then abandoned it as I wasn't sure what would I break in the runtime ☺️

samarsha:
@filipw Sorry about that, I didn't realize your PR was so similar, or I would've mentioned you. :)

Out of curiosity, is testing changes/making PRs across multiple repos something you are interested in? It is currently difficult for non-Microsoft contributors to do this because our multi-repo integration repo is private at the moment (which is what I used to make the runtime packages that I updated in this PR). I'm trying to advocate for making this repo public sooner rather than later to help open-source contributors. :)

filipw:
> Out of curiosity, is testing changes/making PRs across multiple repos something you are interested in? It is currently difficult for non-Microsoft-Quantum contributors to do this because our multi-repo integration repo is private at the moment, and I believe that only employees can publish packages to the alpha NuGet feed (which is how I updated the runtime package versions in this PR). I'm trying to advocate for making this integration repo/packaging process public sooner rather than later to help open-source contributors. :)

yes, that would extremely helpful

samarsha:
> yes, that would extremely helpful

Great, thanks for the feedback! I'll pass this along.

samarsha:
I have updated this PR:

* Add a new public method `NonNullable<string> CompilationUnitManager.GetFileId(Uri uri)` as an alternative to `TryGetFileId`. This is so microsoft/iqsharp#175 can use `GetFileId` instead of directly accessing `Uri.LocalPath` - if we're treating file IDs as an opaque string generated by the compiler, downstream projects shouldn't need to know how it was generated.
* Deprecated `TryGetFileId` with the `[Obsolete]` attribute.

Here is my reasoning for deprecating `TryGetFileId`:

* After looking at how `TryGetFileId` was used in practice, it seemed like the majority of cases simply threw an exception if `TryGetFileId` returned false, which defeats the purpose of a `Try`-style method. In other cases, they defaulted to `uri.AbsolutePath`, which would throw if `uri` is not an absolute URI anyway.
* The `Try` pattern only makes sense if both branches occur in practice at runtime (for example, if user input decides whether parsing a string succeeds or not). But I think in most or all cases of `TryGetFileId`, the URI is known to be of the right type (an absolute file URI). This is something the programmer should be able to control, so if `TryGetFileId` fails, it's a programmer error rather than a user error, and I think an exception makes more sense in that case.

src/QsCompiler/CompilationManager/CompilationUnitManager.cs
```diff
@@ -141,14 +141,31 @@ public void Dispose()
...
+        /// </summary>
+        /// <returns>True if converting the URI to a file ID succeeded.</returns>
+        [Obsolete("Use GetFileId instead after ensuring that the URI is an absolute file URI.")]
+        public static bool TryGetFileId(Uri uri, out NonNullable<string> fileId)
         {
-            id = NonNullable<string>.New("");
-            if (uri == null || !uri.IsFile || !uri.IsAbsoluteUri) return false;
-            id = NonNullable<string>.New(uri.LocalPath);
-            return true;
+            if (uri.IsFile && uri.IsAbsoluteUri)
```
ScottCarda-MS, line 159: You got rid of the null check on the URI?

samarsha: Thanks, I think I'm too used to working with `#nullable enable`...

src/QsCompiler/Tests.Compiler/SerializationTests.fs
```diff
@@ -261,7 +260,8 @@ module SerializationTests =
         let specs = attrs.Specializations |> Seq.map (fun s -> (s.ToTuple() |> fst).ToJson()) |> Seq.toList
         let AssertEqual (expected : string list) (got : _ list) = 
             Assert.Equal(expected.Length, got.Length)
-            expected |> List.iteri (fun i ex -> Assert.Equal (ex.Replace("%%%", dllId.Value), got.[i]))
+            expected |> List.iteri (fun i ex ->
+                Assert.Equal (ex.Replace("%%%", HttpUtility.JavaScriptStringEncode dllId.Value), got.[i]))
```
bettinaheim, line 264: Why is the java script string encoding needed here?

src/QsCompiler/CompilationManager/CompilationUnitManager.cs
```diff
@@ -141,14 +141,34 @@ public void Dispose()
...
+                ? throw new ArgumentNullException(nameof(uri))
+                : TryGetFileId(uri, out var fileId)
+                ? fileId
+                : throw new ArgumentException("The URI is not an absolute file URI.", nameof(uri));
+
+        /// <summary>
+        /// Converts a URI into the file ID used during compilation if the URI is an absolute file URI. 
+        /// </summary>
+        /// <returns>True if converting the URI to a file ID succeeded.</returns>
+        [Obsolete("Use GetFileId instead after ensuring that the URI is an absolute file URI.")]
```
bettinaheim, line 159: What is the reason for the API change?

src/QsCompiler/Tests.Compiler/SerializationTests.fs
samarsha: The `%%%` markers occur inside of a JSON-serialized syntax tree. Some characters in the source file path need to be escaped (for example `\` should become `\\`).

src/QsCompiler/CompilationManager/CompilationUnitManager.cs
samarsha: [See here.](https://github.com/microsoft/qsharp-compiler/pull/468#issuecomment-647790459)

bettinaheim: ok, thanks.

Issue #458, opened 2019-03-13
Escape sequences in string literals are not interpreted


jclapis:
According to [the documentation for strings](https://docs.microsoft.com/en-us/quantum/language/expressions?view=qsharp-preview#string-expressions) in Q#, we should be able to write whitespace characters like \r\n and \t by using a backslash as the escape character, just like C# strings. I tried writing a few of these strings and called [Message()](https://docs.microsoft.com/en-us/qsharp/api/prelude/microsoft.quantum.primitive.message) with them:

```
operation StringEscapeTest () : Unit {
    Message("Normal line");
    Message("\tTab line");
    Message("\nNew line");
    Message("\r\nCRLF");
}
```
Produces this output:
```
Normal line
\tTab line
\nNew line
\r\nCRLF
```
It looks like the special characters aren't getting escaped properly.

PR #469, opened 2020-06-19
Fixing escape handling in string parsing

swernli:
This change adds proper escape character handling for literal strings.
Fixes Escape sequences in string literals are not interpreted #458


src/QsCompiler/Tests.Compiler/SyntaxTests.fs
```diff
@@ -38,12 +38,17 @@ let ``String parser tests`` () =
...
-    testChar 0 '\t' "\"\t\""
-    testChar 0 '\n' "\"\r\"" // carriage returns will get replace by \n
-    testChar 0 '\n' "\"\n\""
-    testChar 3 '\t' "$\"{0}\t\""
-    testChar 3 '\n' "$\"{0}\r\"" 
-    testChar 3 '\n' "$\"{0}\n\""
+    testChar 0 '\t' "\"\\t\""
+    testChar 0 '\r' "\"\\r\""
+    testChar 0 '\n' "\"\\n\""
+    // testChar 0 '\"' "\"\\\"\""
```
swernli, line 44: I added test cases for all the escape characters I think we are supposed to support (based on [our docs](https://docs.microsoft.com/en-us/quantum/user-guide/language/expressions?view=qsharp-preview#string-expressions)), but had to comment out the cases for `\\` and `\"` which still fail for some reason, even though `\{` in interpolated strings works fine. Anyone have any ideas why those characters still don't work correctly?

src/QsCompiler/TextProcessor/SyntaxBuilder.fs
```diff
@@ -182,16 +182,27 @@ let internal getStringContent interpolArg =
     let startDelimiter = pstring "\""
     let endDelimiter = notPrecededBySlash >>. startDelimiter
     let interpolatedString = 
-        let contentDelimiter = endDelimiter <|> (notPrecededBySlash >>. lCurly)
-        let nonInterpol = manyChars (notFollowedBy contentDelimiter >>. anyChar)
+        let interpolCharSnippet = manySatisfy (fun c-> c<> '\\' && c <> '"' && c <> '{')
+        let escapedChar = pstring "\\" >>. (anyOf "\\\"nrt{" |>> function // Also supports escapting '{'
```
swernli, line 186: This pattern for escape handling is based on the documentation for [FParsec.CharParsers.stringsSepBy](http://www.quanttec.com/fparsec/reference/charparsers.html#members.stringsSepBy), and as mentioned in the other comment works for the `\n`, `\r\`, `\t`, and `\{` but somehow not for `\\` and `\"`. I'm going to keep tweaking it to see if I can figure out why, but I would've thought that the `| c -> string c` case would work for those characters. It already seems like that case is correctly handling `\{` so I wonder if there is something in one of the other parsers used here that is preventing it from working for `\\` and `\"`.

src/QsCompiler/Tests.Compiler/SyntaxTests.fs
ScottCarda-MS: If you precede the string with a `@` character, it will be a verbatim string. That may help with readability here.

src/QsCompiler/Tests.Compiler/SyntaxTests.fs
```diff
@@ -38,12 +38,17 @@ let ``String parser tests`` () =
             Assert.Equal(offset + 1, parsed.Length)
             Assert.Equal(char, parsed.[offset])
         | _ -> Assert.True(false, "failed to parse")
-    testChar 0 '\t' "\"\t\""
-    testChar 0 '\n' "\"\r\"" // carriage returns will get replace by \n
-    testChar 0 '\n' "\"\n\""
-    testChar 3 '\t' "$\"{0}\t\""
-    testChar 3 '\n' "$\"{0}\r\"" 
-    testChar 3 '\n' "$\"{0}\n\""
+    testChar 0 '\t' "\"\\t\""
```
ScottCarda-MS, line 41: If you precede the string with a @ character, it will be a verbatim string. That may help with readability here.

swernli: That's a good tip, I'll try that!

src/QsCompiler/Tests.Compiler/SyntaxTests.fs
swernli: Hmm... not sure what changed, but I was able to get `\"` working. I still don't know why `\\` is giving me grief.

src/QsCompiler/Tests.Compiler/SyntaxTests.fs
ScottCarda-MS: Hmm, I may have led you astray. The `@` does indeed prevent escaping, removing the need for double backslashes. But you still need a way to embed the `"` character in strings. I believe F# supports strings delimited with single quotes, so you could do something like `@'"\t"'`.

ScottCarda-MS: https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/strings#triple-quoted-strings

ScottCarda-MS: Looks like it doesn't support the single quote strings, but I linked an alternative solution with F#'s triple quote syntax.

src/QsCompiler/TextProcessor/SyntaxBuilder.fs
bettinaheim: Looking at the test, it looks like it's not that it's producing the wrong string, but rather the parsing itself is failing. I haven't spotted the issue yet either though.

src/QsCompiler/Tests.Compiler/SyntaxTests.fs
swernli: Yeah, the verbatim strings helped some but the quotes get messy. I'll try the linked syntax.

swernli: Unfortunately, I don't think triple-quoted strings ends up being that much better:
`testChar 0 '\t' """"\t" """`
I end up needing to have a space between the last quote in the literal string and the triple-quote that ends the string, otherwise the string isn't terminated properly. Any character will work, but something other space starts to look significant. On the other hand, I can't have a space at the beginning because it confuses the stringLiteral parser itself, which expects the string to start with `"`.

Maybe it's years of working on Windows having warped my mind, but I'm partial to going back to traditional string escaping, since my brain has an easier time with this:
`testChar 0 '\t' "\"\\t\""`

ScottCarda-MS: Yah I think that's probably the right call. The triple quoting is more messy than helpful here, and the `@` doesn't get the right results for nested quotes.

src/QsCompiler/TextProcessor/SyntaxBuilder.fs
swernli: I see why it's failing to parse: `endDelimiter` parser looks for the string termination by parsing for a `"` that is not preceded by `\` but the test case is the literal string `"\"` which means the terminating quote IS preceded by slash. This means the current set of parsers would never be able to parse a string literal that ends in a slash, but at the same time the tests as written always look for the special character as the last character in the string literal. While I'm still not sure how to fix this, I don't think changing the test is the right thing here; a string literal should be allowed to end in an escaped slash.

swernli: Oh, I get it now! Since the `escapedChar` parser correctly handles the escape characters including escaped quotes, the `endDelimiter` parser no longer needs to handle the slash, making it identical to the `startDelimiter`. I've collapsed them into just `delimiter` and switched to using the [FParsec.CharParsers.between](http://www.quanttec.com/fparsec/reference/primitives.html#members.between) to get a slightly more optimized version of `delimiter >>. content ,>> delimiter`. Now all the tests pass :)

bettinaheim: Well done figuring this out!

src/QsCompiler/Tests.Compiler/SyntaxTests.fs
```diff
@@ -32,18 +32,26 @@ let ``String parser tests`` () =
     |> List.iter (fun s -> Assert.True(simpleParseString (stringLiteral .>> eof) s, "Failed to parse string " + s))
 
     // testing whether tabs etc in strings are processed correctly
+    // Note: the below strings use triple-quote syntax where the
```
bettinaheim, line 35: This comment here no longer seems to be accurate.

swernli: Whoops! I forgot to remove that when I went back to standard escaped strings.

PR #474, opened 2020-06-22
WIP: Initial scaffolding of Yeoman templates for VS Code Extension

ricardo-espinoza:
Initial commit for Yeoman-based templates for VS Code Extension.


src/VSCodeExtension/src/generator-qsharp/.editorconfig
```diff
@@ -0,0 +1,11 @@
...
+
+[*]
+indent_style = space
+indent_size = 2
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
+trim_trailing_whitespace = false
```
cgranade, line 11: If you're adding an editor config file, I'd suggest doing so at the root of the repo rather than in a subfolder.

src/VSCodeExtension/src/generator-qsharp/.eslintignore
```diff
@@ -0,0 +1,2 @@
+coverage
+**/templates
```
cgranade, line 2: With the move away from `generator-generator` to TypeScript in order to integrate better with the existing extension, I'd suggest dropping this file and adding an equivalent tslint configuration.

src/VSCodeExtension/src/generator-qsharp/.travis.yml
```diff
@@ -0,0 +1,9 @@
+language: node_js
+node_js:
+  - v14
+  - v12
+  - v10
+  - v8
+  - v6
+  - v4
+after_script: cat ./coverage/lcov.info | coveralls
```
cgranade, line 9: The `generator-generator` template makes this for use with Travis CI, but we use a mix of Azure Pipelines and GitHub Actions for our CI; I'd suggest dropping this file and creating an equivalent pipeline step or GitHub action instead.

src/VSCodeExtension/src/generator-qsharp/app/index.ts
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
cgranade, line 1:
```suggestion
// Copyright (c) Microsoft Corporation.
```

Updated guidance no longer suggests "All rights reserved" for new code.

src/VSCodeExtension/src/generator-qsharp/app/index.ts
```diff
@@ -0,0 +1,41 @@
...
+// Licensed under the MIT License.
+
+'use strict';
```
cgranade, line 4: Strict mode shouldn't be required when working with TypeScript.
```suggestion
```

src/VSCodeExtension/src/generator-qsharp/app/index.ts
```diff
@@ -0,0 +1,41 @@
...
+const Generator = require('yeoman-generator');
+const chalk = require('chalk');
+const yosay = require('yosay');
```
cgranade, line 7: I'd recommend replacing calls to require with the use of TypeScript's import statements, as t hat will be more portable to mixing different module systems.

src/VSCodeExtension/src/generator-qsharp/app/index.ts
```diff
@@ -0,0 +1,41 @@
...
+
+module.exports = class extends Generator {
```
cgranade, line 9: Similarly here, I'd recommend using TypeScript's `export` modifier keyword to directly setting `module.exports`. You can  also use the `default` modifier to make it the only exported definition.

I think this is right, but you'll want to double-check.
```suggestion
export default class extends Generator {
```

src/VSCodeExtension/src/generator-qsharp/app/templates/library/library.csproj.txt.v.template
```diff
@@ -0,0 +1,7 @@
+<Project Sdk="Microsoft.Quantum.Sdk/#NUGET_VERSION#">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.1</TargetFramework>
```
cgranade, line 4: It may be possible to consolidate this with the other two templates by adding an EPL variable for what the TargetFramework should be, reducing the number of files that need to be maintained.

src/VSCodeExtension/src/generator-qsharp/tests/app.ts
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
cgranade, line 1:
```suggestion
// Copyright (c) Microsoft Corporation.
```

src/VSCodeExtension/src/generator-qsharp/.editorconfig
ricardo-espinoza: Actually, this came from the generator-generator scaffolding, so I didn't actually intend to add it. I'll remove it.

src/VSCodeExtension/src/generator-qsharp/.eslintignore
ricardo-espinoza: Sounds good!

src/VSCodeExtension/src/generator-qsharp/app/templates/library/library.csproj.txt.v.template
ricardo-espinoza: Sounds right, and I considered it. But since the other two sets of templates have separate files for app/lib/test, the pros of keeping them in separate file is that it might be easier to match with them. If at some point in the future something changes in the library project for example, we would update the VS, CLI and yeoman files for that project type only, instead of a common parametrized file for yeoman but separate ones for VS/CLI. What do you think?

src/VSCodeExtension/src/generator-qsharp/app/index.ts
ricardo-espinoza: Sounds good. As I'm adding the next round of changes in index.ts (to replace the scaffolding by the real thing) I'll fix the imports.

src/VSCodeExtension/src/generator-qsharp/app/templates/library/library.csproj.txt.v.template
cgranade: Makes sense; let's maybe defer, then, any action either way until the rest of the scaffolding is up and working. That would be a pretty easy refactor at the end, in any case.

Issue #475, opened 2020-06-22
Classical control transformation unconditionally executes `elif` conditions


swernli:
When transforming an if-elif pattern into `ApplyIfElseR` with nested `ApplyIf*` the compiler produces code that will end up evaluating the elif condition regardless of the result of the original if. This can change the behavior of code completely. Take for example this Q#:
```
    @EntryPoint()
    operation Test() : Unit {
        for (i in 0..10) {
            using (qs = Qubit[2]) {
                H(qs[1]);
                if (M(qs[0]) == Zero) {}
                elif (M(qs[1]) == Zero) {}
                H(qs[1]);
                let r = M(qs[1]);
                Message($"{r}");
                ResetAll(qs);
            }
        }
    }
```
Because qs[0] is left in the ground state, the if-statement will always be true, so the elif does not get evaluated, meaning the sequence of operations on qs[1] is always H, H, M. This should result in an output of "Zero" every time, and the untargeted simulator confirms this behavior. If you add an ExecutionTarget to the project that triggers the conditional rewrite, the resulting transformed code is effectively performing H, _M_, H, M on qs[1], so the output is a mix of "One" and "Zero" instead.

ScottCarda-MS:
This will require a redesign of the feature. The issue currently is the lifting for the if contents happens first, then the reshaping of the if structures happens. So by the time we have a nested if-statement to lift, the lifting is already done.

PR #478, opened 2020-06-23
Classically Controlled Short Circuit Bug Fix

ScottCarda-MS:
Rearranged the order of steps in the Classically Controlled transformation so that the restructuring of if-statements is performed first. This addresses an issue where all conditional expressions were always evaluated even when they should have been short-circuited (#475)

### DIFF
--- a/src/QsCompiler/Transformations/ClassicallyControlled.cs
+++ b/src/QsCompiler/Transformations/ClassicallyControlled.cs
@@ -29,9 +29,177 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.ClassicallyControlled
     {
         public static QsCompilation Apply(QsCompilation compilation)
         {
+            compilation = RestructureConditions.Apply(compilation);
             compilation = LiftConditionBlocks.Apply(compilation);
-
             return ConvertConditions.Apply(compilation);
+        }
+
+        private class RestructureConditions : SyntaxTreeTransformation
+        {
+            public new static QsCompilation Apply(QsCompilation compilation)
+            {
+                var filter = new RestructureConditions();
+
+                return new QsCompilation(compilation.Namespaces.Select(ns => filter.Namespaces.OnNamespace(ns)).ToImmutableArray(), compilation.EntryPoints);
+            }
+
+            private RestructureConditions() : base()
+            {
+                this.Namespaces = new NamespaceTransformation(this);
+                this.Statements = new StatementTransformation(this);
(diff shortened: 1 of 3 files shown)


cesarzc:
@ScottCarda-MS if I understand correctly, the basics of this change are the following:
`ReshapeConditionals` has to be executed before `LiftConditionBlocks.Apply` or otherwise all conditional would be evaluated instead of being short-circuited.

In order to do that, another syntax tree transformation (`RestructureConditions`) was created to do so. Now, `ReshapeConditionals` is called as part of this new transformation. Does this explanation make sense?

src/QsCompiler/Transformations/ClassicallyControlled.cs
```diff
@@ -18,22 +18,190 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.ClassicallyControlled
...
-    /// 2nd Pass: On the way down the tree, reshape conditional statements to replace Elif's and
-    /// top level OR and AND conditions with equivalent nested if-else statements. One the way back up
-    /// the tree, convert conditional statements into interface calls, where possible.
-    /// This relies on anything having type parameters must be a global callable.
+    /// This transformation works in three passes.
+    /// 1st Pass: Reshape conditional statements to replace Elif's and top level OR and AND conditions
+    /// with equivalent nested if-else statements.
+    /// 2st Pass: Lift the contents of conditional statements into separate operations, where possible.
+    /// 3nd Pass: Convert conditional statements into interface calls, where possible.
+    /// This relies on global callables being the only things that having type parameters.
```
cesarzc, line 26:
```suggestion
    /// This relies on global callables being the only things that have type parameters.
```

ScottCarda-MS:
> @ScottCarda-MS if I understand correctly, the basics of this change are the following:
> `ReshapeConditionals` has to be executed before `LiftConditionBlocks.Apply` or otherwise all conditional would be evaluated instead of being short-circuited.
> 
> In order to do that, another syntax tree transformation (`RestructureConditions`) was created to do so. Now, `ReshapeConditionals` is called as part of this new transformation. Does this explanation make sense?

Yes, all that is correct and makes sense. It may be worth adding that before the changes in this PR the reshaping conditional logic was part of the `ConvertConditions` transformation that happened after the `LiftConditionBlocks` transformation.

src/QsCompiler/Transformations/ClassicallyControlled.cs
ScottCarda-MS: Thanks for finding this. It's been fixed.

PR #480, opened 2020-06-24
Adapt layout for vscode extension to keep an independent package for the Yeoman generator

ricardo-espinoza:
[See original PR #479 ]

This PR adds a lerna configuration to the vscode extension build workflow to allow for separating generator-qsharp into its own package and to allow for injecting that package into the vscode extension as an ordinary dependencies property. This PR also demonstrates the new layout by registering generator-qsharp as a stub with yeoman-environment from within the vscode extension.


src/VSCodeExtension/packages/extension/src/commands.ts
```diff
@@ -9,7 +9,8 @@ import { DotnetInfo, findIQSharpVersion } from './dotnet';
 import { IPackageInfo } from './packageInfo';
 import * as semver from 'semver';
 import { promisify } from 'util';
-import { oc } from 'ts-optchain';
+
+//import * as yeoman from 'yeoman-environment';
```
cgranade, line 13: A comment explaining why these are currently disabled would maybe be helpful.

PR #489, opened 2020-06-26
Add NoOp to Precondition Dependencies for Classically Controlled

ScottCarda-MS:
The NoOp operation is added to the list of operations that the precondition for the Classically Controlled rewrite step checks is in the compilation before allowing the step to proceed.


src/QsCompiler/Compiler/RewriteSteps/ClassicallyControlled.cs
```diff
@@ -44,6 +44,8 @@ public bool PreconditionVerification(QsCompilation compilation)
                 .ToHashSet();
             var requiredBuiltIns = new HashSet<QsQualifiedName>()
             {
+                BuiltIn.NoOp.FullName,
```
swernli, line 47: Do we also need to add `IndexRange` to this list, or is that already handled as a dependency somewhere else?

ScottCarda-MS: The Classically Controlled logic doesn't rely on `IndexRange`, so it doesn't need to be listed here.

Issue #342, opened 2020-02-26
Consider using Coding Style and .editorconfig


rstm-sf:
Hello!

Please, consider using Coding Style and [.editorconfig](https://editorconfig.org/). Because it’s unusual to see, for example, the following code

https://github.com/microsoft/qsharp-compiler/blob/9bdc21409409b8e74cde1e0a24c49fc80bb602f3/src/QsCompiler/Compiler/CompilationLoader.cs#L408-L409

marshallsa:
Thanks for the suggestion! We've discussed standardizing the coding style recently. An .editorconfig file sounds like a good idea. It looks like a lot of editors have support for it, including [Visual Studio](https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference?view=vs-2019) and [Visual Studio Code](https://github.com/editorconfig/editorconfig-vscode).

We're also interested in tools to automatically check code style, for example as part of the automated builds for PRs. It looks like [dotnet-format](https://github.com/dotnet/format) can do this, but we're open to other suggestions.

filipw:
In C# extension in VS Code, editorconfig support is an opt in flag for OmniSharp (https://www.strathweb.com/2019/07/editorconfig-support-in-omnisharp-and-c-extension-vs-code/), we hide most of these newer features behind an opt-in flag on order to keep the base featureset is as lightweight as possible. but yes it definitely makes sense to me too!

For enforcing code style at build time you can also have a look at `Microsoft.CodeAnalysis.CSharp.CodeStyle` prerelease package (only exists on Roslyn CI feed). It ships support for whitespace (IDE0055) violations (same as those covered by dotnet-format) in a regular analyzer package and reads the settings from editorconfig file.

cgranade:
We actually have an [editorconfig for the samples repo](https://github.com/microsoft/Quantum/blob/master/.editorconfig) for precisely this reason. We've looked into using dotnet-format with precommit before to help with code style, but to limited success.

marshallsa:
I'm starting to work on an [.editorconfig file in the marshallsa/editorconfig branch](https://github.com/microsoft/qsharp-compiler/blob/marshallsa/editorconfig/.editorconfig). Some mixed results so far.

First, it seems like the only way to reformat @rstm-sf's `foreach` example is to set `csharp_preserve_single_line_blocks = false`. However, this also reformats auto-properties that are defined on one line:

```
public int Foo { get; set; }
```

becomes

[...]

PR #496, opened 2020-06-30
Add StyleCop to CommandLineTool project and reformat source code; add .editorconfig for all projects

samarsha:
This is a first step to resolving #342. I added the `StyleCop.Analyzers` package to the `CommandLineTool` project to try out StyleCop's formatting rules. I plan to add StyleCop to the rest of the projects in qsharp-compiler, but I wanted to make a smaller PR with just this project to get feedback on the style choices I made.

StyleCop's default rules are extensive, and highly opinionated, so I ended up disabling several rules that seemed too strict for us. Those disabled rules are in `src/Common/CodeAnalysis.ruleset`.

I also added an `.editorconfig` file to override Visual Studio's default behavior and to enforce a style rule that StyleCop doesn't have:

`dotnet_style_qualification_for_{event,field,method,property} = true`
By default, Visual Studio suggests replacing `this.Foo` with `Foo`. @bettinaheim and I decided that we want to require `this.` in all cases. This is already the default for StyleCop, but Visual Studio's behavior needs to be set separately.

`dotnet_diagnostic.IDE0002.severity = warning`
This promotes Visual Studio's code action for simplifying qualified names to a warning. I added this to simplify references to static members within the same class. So if class `Foo` has a static method `Bar`, for example, I think we should call that method from within `Foo` by writing `Bar()` instead of `Foo.Bar()`. I think we should use this style because:

* Normally, with unqualified references, it's ambiguous whether it's an instance member or a static member. Since we require instance members to be prefixed with `this.`, we don't need to prefix static members - we know any unprefixed reference is static.
* Static members are analogous to members of a module in F# - we don't prefix references to other things in the same module in F#, so I don't think we should do that in C#, either.
* Prefixes are necessary when the static member is in a *different* class - adding an unnecessary prefix can create the impression that the reference is to a different class, which can be confusing.


src/QsCompiler/CommandLineTool/CompilationTracker.cs
```diff
@@ -28,64 +26,69 @@ private class CompilationTask
...
             /// <summary>
             /// Contains the duration of the task in milliseconds.
             /// </summary>
             public long? DurationInMs;
+
             /// <summary>
             /// Stopwatch used to measure the duration of the task.
             /// </summary>
-            private readonly Stopwatch Watch;
+            private readonly Stopwatch watch;
```
bettinaheim, line 53: What is the rule here - private things are lower case, public things are upper case? What about internal things? Is the idea to match this in F#?

src/QsCompiler/CommandLineTool/Options.cs
```diff
@@ -136,32 +187,31 @@ internal void UpdateSetIndependentSettings(Options updates, bool overwriteNonDef
...
+        // routines related to logging
 
         /// <summary>
-        /// If a logger is given, logs the options as CommandLineArguments Information before returning the printed string. 
+        /// If a logger is given, logs the options as CommandLineArguments Information before returning the printed string.
         /// </summary>
         public string[] Print(ILogger logger = null)
         {
-            string value(PropertyInfo p)
+            string Value(PropertyInfo p)
```
bettinaheim, line 197: How come private members are lower case but local functions are upper case? It would seem more consistent to me if those where lower case then, too.

src/Common/CodeAnalysis.ruleset
```diff
@@ -0,0 +1,38 @@
+﻿<?xml version="1.0"?>
```
filipw, line 1: rulesets are actually deprecated from Visual Studio 16.3+, the way to control diagnostic severities is now exclusively via editorconfig. in fact there is no way to create a ruleset anymore in VS and using the "suppress" feature in VS automatically updates editorconfig too

you can move it all into editorconfig using the `dotnet_diagnostic. SA1122.severity = none` syntax
I believe that would be much nice and cleaner

cgranade:
If you're adding an `.editorconfig` file, may I suggest adding a `.vscode/extensions.json` file that suggests the EditorConfig extension for VS Code as well? That won't have any effect other than to notify a contributor that the repo supports EditorConfig, but even that may be a helpful signal to provide.

src/Common/CodeAnalysis.ruleset
```diff
@@ -0,0 +1,38 @@
...
+    <Rule Id="SA1503" Action="None" />      <!-- Braces should not be omitted -->
+    <Rule Id="SA1512" Action="None" />      <!-- Single-line comments should not be followed by blank line -->
+    <Rule Id="SA1513" Action="None" />      <!-- Closing brace should be followed by blank line -->
+  </Rules>
+
+  <Rules AnalyzerId="StyleCop.Analyzers" RuleNamespace="StyleCop.Analyzers.DocumentationRules">
+    <Rule Id="SA1600" Action="None" />      <!-- Elements should be documented -->
+    <Rule Id="SA1602" Action="None" />      <!-- Enumeration items should be documented -->
+    <Rule Id="SA1611" Action="None" />      <!-- Element parameters should be documented -->
+    <Rule Id="SA1615" Action="None" />      <!-- Element return value should be documented -->
```
ScottCarda-MS, line 32: I'm not in favor of these two rules. Documenting every parameter and every return value is tedious and often unhelpful as documentation. We haven't been documenting along these rules so far, so I don't think we should introduce these now, in this PR.

src/QsCompiler/CommandLineTool/Commands/Build.cs
```diff
@@ -89,9 +112,8 @@ internal static bool IncorporateResponseFiles(BuildOptions options, out BuildOpt
             }
         }
 
-
```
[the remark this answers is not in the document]
ScottCarda-MS: I actually like trimming the double-lines. I've never been a big fan of them as, to me, they add a jaggedness to the formatting of a file.

src/Common/CodeAnalysis.ruleset
samarsha: These rules are disabled with `Action="None"`. :) Sorry, this file is pretty confusing, because the rule descriptions are phrased like they are being enabled, but all the rules in this file are being disabled - they are normally on by default in StyleCop.

src/QsCompiler/CommandLineTool/CompilationTracker.cs
ScottCarda-MS: The convention as I know it has public members as PascalCase and private members as camelCase (sometimes with an underscore prefix). I believe internal members are treated the same a public: https://stackoverflow.com/a/46777930/5399520

src/QsCompiler/CommandLineTool/Commands/Build.cs
```diff
@@ -24,47 +23,70 @@ public static IEnumerable<Example> UsageExamples
...
+                "perf",
+                Required = false,
+                SetName = CODE_MODE,
+                HelpText = "Destination folder where the output of the performance assessment will be generated.")]
             public string PerfFolder { get; set; }
 
-
             /// <summary>
-            /// Reads the content of all specified response files and processes it using FromResponseFiles. 
+            /// Reads the content of all specified response files and processes it using FromResponseFiles.
```
ScottCarda-MS, line 86: I really appreciate this removing trailing whitespace!

src/Common/CodeAnalysis.ruleset
filipw: I forgot to attach the relevant links:
 - https://docs.microsoft.com/en-us/visualstudio/code-quality/use-roslyn-analyzers?view=vs-2019#convert-an-existing-ruleset-file-to-editorconfig-file `Starting in Visual Studio 2019 version 16.5, ruleset files are deprecated in favor of the EditorConfig file for analyzer configuration for managed code (...)`
 - https://docs.microsoft.com/en-us/visualstudio/code-quality/use-roslyn-analyzers?view=vs-2019#set-rule-severity-in-the-rule-set-file (see the "note" box)

samarsha: Yes, this seems better, thanks. :)

src/QsCompiler/CommandLineTool/Commands/Build.cs
samarsha: I am really tempted to set `dotnet_diagnostic.SA1028.severity = error`, which would make trailing whitespace a compile error. :)

src/QsCompiler/CommandLineTool/CommandLineTool.csproj
```diff
@@ -31,8 +31,13 @@
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.3.1" />
     <PackageReference Include="Microsoft.CodeAnalysis.Features" Version="3.3.1" />
     <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.Common" Version="3.3.1" />
+    <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.164" />
```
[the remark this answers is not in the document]
samarsha: From what I can tell, it isn't necessary. StyleCop is an analyzer, and [`PrivateAssets` by default includes `Analyzers`](https://docs.microsoft.com/en-us/dotnet/core/tools/csproj#includeassets-excludeassets-and-privateassets).

src/QsCompiler/CommandLineTool/Commands/Build.cs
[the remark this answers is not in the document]
samarsha: This is [SA1507](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1507.md). I don't have a strong preference, but I slightly prefer not having double lines (like @ScottCarda-MS). IMO, if a file is large enough that it would benefit from sections separated by double lines, it might be better to create separate files...

src/QsCompiler/CommandLineTool/CompilationTracker.cs
samarsha: That's [SA1306](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1306.md). It only applies to private fields. Internal or public fields or private methods start with an uppercase letter.

F# private fields are `let foo = ...`. I think those should also start with a lowercase letter also, like most F# names (with the exception of things that can be seen by C#, which are capitalized - I think that's the rule?).

src/QsCompiler/CommandLineTool/Options.cs
samarsha: It's only private fields that are lowercase - private methods (and even private properties) are capitalized.

samarsha:
@filipw Do you know what the best way is to show Roslyn analyzer warnings in OmniSharp (for VS Code support)? It was not showing them by default, so I added this `omnisharp.json` file:

```
{
    "RoslynExtensionsOptions": {
        "enableAnalyzersSupport": true
    }
}
```

but now it shows all StyleCop warnings, even the ones that were disabled by `.editorconfig`.

filipw:
editorconfig has to be opted into separately in OmniSharp, add this to `omnisharp.json` too, next to the section you already have to enable analyzers:

```
    "FormattingOptions": {
        "enableEditorConfigSupport": true
    }
```

samarsha:
Thanks!

src/QsCompiler/CommandLineTool/Commands/Build.cs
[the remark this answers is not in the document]
cesarzc: I also have a slight preference for trimming double lines. As @samarsha mentioned, if the large is long enough for sections to be separated by double lines, it might be a good idea to create a separate file.

src/QsCompiler/CommandLineTool/Commands/Build.cs
```diff
@@ -74,7 +96,8 @@ internal static bool IncorporateResponseFiles(BuildOptions options, out BuildOpt
                     try
                     {
                         var fromResponseFiles = FromResponseFiles(options.ResponseFiles);
-                        if (fromResponseFiles == null) return false;
+                        if (fromResponseFiles == null)
```
cesarzc, line 99: Is this because an instruction can't be in-lined with an `if/else if/else`?

If that is the case, I think it would be better to always enclose instructions after an `if/else if/else` in curly brakets `{ }` because otherwise it is more likely that an error is introduced.

samarsha: Hmm, I think at some point having the statement on the same line as an `if` wasn't allowed, but now it seems like it is... Maybe I changed some StyleCop settings since then. However, I can't disallow the two-line version without also disallowing the one-line version. (There is an open issue for that: DotNetAnalyzers/StyleCopAnalyzers#2252.)

I think that omitting braces is unlikely to cause errors, because

```
if (foo)
    Bar();
    Baz();
```

will produce a warning from [SA1137: Elements Should Have the Same Indentation](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1137.md).

samarsha: Although the two-line version is allowed with the current style rules, I reverted all of them back to one-line versions to avoid making unnecessary formatting changes in this PR. Are we OK with this ambiguity in the style rules (either one-line or two-line braceless `if`s are allowed)?

ScottCarda-MS: My preference is to allow single-line and braced multi-line, but disallow unbraced two-line. However, I am fine with allowing the unbraced two-line if statements if people really want them.

samarsha: The problem is that StyleCop's rule settings are not precise enough, so we can't allow unbraced one-line without also allowing unbraced two-line, even if no one wants them.

ScottCarda-MS: Oh I see the problem. In that case I can see the desire to not want to have a rule for this and allow all the options.

src/QsCompiler/CommandLineTool/CommandLineTool.csproj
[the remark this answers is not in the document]
bettinaheim: Good to know!

src/QsCompiler/CommandLineTool/Commands/Build.cs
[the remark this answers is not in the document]
bettinaheim: works for me.

src/QsCompiler/CommandLineTool/CompilationTracker.cs
bettinaheim: I think it would be good to set it up in a way that people don't need to know whether the source was F# or C#, meaning if we could get public F# fields to be uppercase as well.

samarsha: This naming convention only applies to private fields. By definition, private fields are not visible from other projects, so there are no interop concerns.

src/QsCompiler/CommandLineTool/Commands/Build.cs
cesarzc: Given that the SA1137 warning will be produced, I think it is ok to have this ambiguity in the style rules.

src/QsCompiler/CommandLineTool/CompilationTracker.cs
samarsha: Sorry, I think I misunderstood you, what I mean is: As far as I know, most/all internal/public C# fields in the compiler were already capitalized, so I think the style change would mainly be limited to private fields. I believe most/all of the public F# fields (`member val`) are already capitalized as well - the only lowercase ones I found were in a private type in a test.

src/QsCompiler/CommandLineTool/Commands/Build.cs
samarsha: I figured out why this was not allowed earlier: the change from a ruleset file to an editorconfig file meant that StyleCop no longer realized that SA1503 (braces) was disabled, which meant that SA1501 (statement must be on single line) was behaving differently - it doesn't produce its own warning in this case because it thinks the braces warning will appear, but that was suppressed by editorconfig.

I would actually like to go back to that behavior (by re-adding a ruleset just to disable SA1503) and require that ifs be written on at least two lines in all cases. While I'm fine with short lines like

```
if (foo is null) throw new ArgumentNullException(nameof(foo));
```

I think it's easy for this to get out of hand. For example, these lines are from DiagnosticTools:

```
if (relativeRange == null) throw new ArgumentNullException(nameof(relativeRange));
if (LargerThan(relativeRange.Item1, relativeRange.Item2)) throw new ArgumentException("invalid range", nameof(relativeRange));
```

This code is very dense and the `throw`s do not line up, so it's hard to scan the code and understand the structure of it, IMO. I think that writing it like this is easier to read:

```
if (relativeRange == null)
    throw new ArgumentNullException(nameof(relativeRange));
if (LargerThan(relativeRange.Item1, relativeRange.Item2))
    throw new ArgumentException("invalid range", nameof(relativeRange));
```

Braces would be OK too, but when there are a lot of throws at the top of a method, having braces makes everything twice as long which also has its readability problems, I think. Is everyone OK with this?

ScottCarda-MS: Preventing single line if statements seems fine to me. This style rule seems fairly common too, base on my experiences.

cesarzc: I slightly prefer having braces all the time but I would be ok with not requiring them all the time of most people prefer it that way.

src/QsCompiler/CommandLineTool/CompilationTracker.cs
bettinaheim: sounds good - thanks for checking

src/QsCompiler/CommandLineTool/Commands/Build.cs
bettinaheim: either one is fine by me

samarsha: It seems like requiring braces is most people's first choice. It's also the default setting for StyleCop. So requiring braces seems like the option with the least friction overall. I'll change the settings to use that way.

PR #499, opened 2020-06-30
upgraded Roslyn to 3.6.0

filipw:
I noticed that Roslyn 3.3.1, which is almost 1 year old, is still in use. This PR updates to 3.6.0 which aligns with VS 2019 16.6


bettinaheim:
Thanks, @filipw!
This PR needs coordination with the runtime repo; Don't merge until we have confirmed the package update there.

PR #500, opened 2020-06-30
Use correct versions for both VS and VSCode VSIX

rmshaffer:


bootstrap.ps1
```diff
@@ -9,7 +9,10 @@ param(
     $NuGetVersion = $Env:NUGET_VERSION,
 
     [string]
-    $VsixVersion = $Env:VSIX_VERSION
+    $VsCodeVsixVersion = $Env:VSCODEVSIX_VERSION,
```
cgranade, line 12: VS Code has a hard requirement that its VSIX packages are SemVer 2.0 compliant. Given that, we possibly may not want a separate parameter for VS Code VSIX parameters.

rmshaffer: Thanks, you're right, this is redundant.

PR #503, opened 2020-06-30
Distinguishing the processor architecture and the execution target

bettinaheim:
While we are not ready for a precise definition of different processor architectures, we need to start distinguishing what will become the architecture handle in the future vs what is the execution target to invoke/submit to. 
This PR hence renames execution target to processor architecture for the usages within the compiler, and adds a separate entry for the execution target and the processor architecture to the dictionary of assembly constants. We also rename the keys of that dictionary to not distinguish resolved and unresolved properties.


src/QsCompiler/SyntaxProcessor/ScopeTools.fs
```diff
@@ -291,7 +291,7 @@ type ScopeContext<'a> =
       /// The runtime capabilities for the compilation unit.
       Capabilities : RuntimeCapabilities
       /// The name of the execution target for the compilation unit.
```
samarsha, line 293:
```suggestion
      /// The name of the processor architecture for the compilation unit.
```

src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
```diff
@@ -253,7 +253,7 @@ let private VerifyEqualityComparison context addError (lhsType, lhsRange) (rhsTy
     let baseType = CommonBaseType addError argumentError context.Symbols.Parent (lhsType, lhsRange) (rhsType, rhsRange)
     match baseType.Resolution with
     | Result when context.Capabilities = RuntimeCapabilities.QPRGen0 ->
-        addError (ErrorCode.UnsupportedResultComparison, [context.ExecutionTarget.Value]) rhsRange
+        addError (ErrorCode.UnsupportedResultComparison, [context.ProcessorArchitecture.Value]) rhsRange
```
samarsha, line 256: We should update the diagnostic message; it says "The execution target {0}" instead of "The processor architecture {0}." 

Also, do we want to show the execution target, the processor architecture, or both in the error message? The processor architecture is more important, but the user might not recognize it if they only specified an execution target in the project. IMO, ideally the string in the error message would be something the user could search for in their project to know what they would need to change to increase the runtime capabilities.

bettinaheim: Since the string doesn't actually list what processor architectures are capable of executing the code, I am not sure the latter is possible; even if the user finds the property in the project file to edit, they wouldn't know what to set it to. I hence would suggest to edit the error message to say "{0} does not support ... . Update the execution target or the processor architecture to ..."

samarsha: The current message "{0} does not support comparing measurement results." works for me.

bettinaheim: Shortened to "The target {0} does not support comparing measurement results." per our discussion.

PR #509, opened 2020-07-08
Remove version and copyright properties from Q# grammar file

### LINKED ISSUE #504: Official Q# syntax file not recognized by GitHub's Linguist
I'm working on adding Q# support to GitHub's Linguist. In order to add syntax highlighting I need to provide an up-to-date syntax file for the language. I am aware of the one [in the VSCode extension directory](https://github.com/microsoft/qsharp-compiler/blob/master/src/VSCodeExtension/syntaxes/qsharp.tmLanguage.json.v.template), however, it's not recognized by the utility that Linguist uses to scrape syntax files from repositories. The issue is that the extension it uses is not valid, and thus the tool cannot find it.

I'd like the syntax file in the repository to use one of the extensions that Linguist can recognize, for instance ".tmLanguage" or ".json" (see [this function](https://github.com/github/linguist/blob/b24ff11d9cf60dd996156a718f83024c9436e5ea/tools/grammars/compiler/loader.go#L106) for reference).

- Using another repository as a source for the syntax file - I've found [one](https://github.com/ivangabriele/qsharp-tmLanguage), but it's outdated. For instance, the "internal" keyword is still illegal in this one. I'd rather have the syntax come from the official repository for the language.
- Changing the way Linguist scans repositories for syntax files - it probably makes more sense for the language to adapt to GitHub's standards and not the other way around.

**Additional context**
I've submitted a pull request (github/linguist#4901) to Linguist to add Q# as a recognized language. The syntax is not strictly necessary at the time of adding a language, but it'd be nice to have GitHub highlight Q# source files from day one.

I've seen a reference to `syntaxes/qsharp.tmLanguage.json` in the `.gitignore`, so I guess the file is generated locally after installing the extension. However, the .template file is a valid syntax file as well ([link to Lightshow](https://github-lightshow.herokuapp.com/?utf8=✓&scope=from-url&grammar_format=auto&grammar_url=https%3A%2F%2Fgithub.com%2Fmicrosoft%2Fqsharp-compiler%2Fblob%2Fmaster%2Fsrc%2FVSCodeExtension%2Fsyntaxes%2Fqsharp.tmLanguage.json.v.template&grammar_text=&code_source=from-url&code_url=https%3A%2F%2Fraw.githubusercontent.com%2Fmicrosoft%2FQuantum%2Fmaster%2Fsamples%2Fnumerics%2FCustomModAdd%2FCustomModAdd.qs&code=)), so just changing the extension to one that's recognized by Linguist would most likely be enough.

samarsha:
The `.v.template` file extension for the Q# grammar file means that GitHub Linguist did not recognize it, which means that syntax highlighting for Q# cannot be added to GitHub (see issue #504 from @JakuJ). I'm removing the `#SEMVER_VERSION#` template variable from the grammar file so the template extension can be removed. Here are my reasons for removing it:

* The [starter file for VS Code extension grammars](https://github.com/microsoft/vscode-generator-code/blob/master/generators/app/templates/ext-language/syntaxes/new.tmLanguage.json) does not include a `version` property.
* Syntax highlighting in the VS Code extension seems to work fine without the `version` property.
* The `version` property is not actually part of the [`tmlanguage.json` schema](https://raw.githubusercontent.com/martinring/tmlanguage/master/tmlanguage.json) that the file uses.
* The version number is already specified for the extension itself; including it in the grammar seems redundant.

Since I don't know what the original reason was for including the version number, I'm not sure if this will break something that I haven't thought of. @bettinaheim @ricardo-espinoza?

Closes #504.


tcNickolas:
I think I might've been the first person to put that field in this file, and that was back in 2017... One thing that *might* break (I can't find a paper trail for this so it's from memory) is that VS is really good at caching the grammars (in `%localappdata%\Microsoft\VisualStudio\`), so it only updates the cached grammar files if the file version is newer, otherwise the user will be stuck with the highlight they get from the 1st installation. (We're sharing this file for VS and VS Code both.) I'm not sure this behavior is still there, or that the user is not stuck with the first highlight they get even with versions - probably needs to be checked.

src/VSCodeExtension/syntaxes/qsharp.tmLanguage.json
```diff
@@ -2,7 +2,6 @@
   "copyright": "Copyright ⓒ 2017-2019 Microsoft Corporation. All Rights Reserved.",
```
JakuJ, line 2:
```suggestion
```
Hmm, apparently the script Linguist uses to add grammars also complains about an "unexpected" field, `copyright`. The [starter file](https://github.com/microsoft/vscode-generator-code/blob/master/generators/app/templates/ext-language/syntaxes/new.tmLanguage.json) doesn't include that field, and if I remove it here, together with the version number (see: [my fork](https://github.com/JakuJ/qsharp-compiler/blob/3b1d2015ab7b45e223d0acb2ad6a9946fe0d4cce/src/VSCodeExtension/syntaxes/qsharp.tmLanguage.json#L2)), only then does it get detected properly.

samarsha:
> I think I might've been the first person to put that field in this file, and that was back in 2017... One thing that _might_ break (I can't find a paper trail for this so it's from memory) is that VS is really good at caching the grammars (in `%localappdata%\Microsoft\VisualStudio\`), so it only updates the cached grammar files if the file version is newer, otherwise the user will be stuck with the highlight they get from the 1st installation. (We're sharing this file for VS and VS Code both.) I'm not sure this behavior is still there, or that the user is not stuck with the first highlight they get even with versions - probably needs to be checked.

I tried this out:

1. Build VSIX and install it.
2. Open Q# file to look at syntax highlighting.
3. Change `qsharp.tmLanguage.json` (I added `hello` as a new highlighted keyword).
4. Update version number in `source.extension.vsixmanifest`, rebuild VSIX and install it.
5. Open Q# file and check that the changes were applied (`hello` was now highlighted blue).

It worked for me without a `version` property in the grammar file. I am assuming that whatever caching Visual Studio does is based on the extension version rather than the grammar file `version` property.

PR #510, opened 2020-07-08
Enabling Yeoman templates as part of VS Code extension.

ricardo-espinoza:


src/VSCodeExtension/packages/extension/package.json.v.template
```diff
@@ -97,7 +97,7 @@
         ]
     },
     "scripts": {
-        "vscode:prepublish": "npm install && npm run compile",
+        "vscode:prepublish": "npm install ../generator-qsharp/ && npm install && npm run compile",
```
cgranade, line 100: That's one way to do it, should definitely work around most of the issues we ran into with lerna and similar approaches!

PR #513, opened 2020-07-09
Add StyleCop to all C# projects and reformat code

samarsha:
This is the follow-up to #496, and is part of #342. All C# projects use StyleCop and have been reformatted. I disabled a few more default StyleCop rules; see `.editorconfig` and `src/Common/stylecop.json`.

Also, I decided to leave some code in place that violates the default style rules because it requires non-trivial fixes or fixes that are breaking changes (due to renaming publicly accessible things). Normally, I would disable the rules if I didn't want to fix them, but I think the rules are good to have for new code.

Style violations that would have breaking change fixes:
* [SA1306: Field Names Must Begin With Lower Case Letter](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1306.md). There's only one protected field in a public class that was violating this rule. The fix is to make it lowercase.
* [SA1310: Field Names Must Not Contain Underscore](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1310.md). Several violations with public fields that used `UPPER_CASE` names. They should use `PascalCase` instead.

Full disclosure: I did rename some parameters in public methods that started with an uppercase letter (all parameter names should be `camelCase`), which is technically breaking if you use named parameters, but the risk to renaming parameters seems lower compared to renaming fields.

Other style violations that need non-trivial (but not breaking) fixes:
* [SA1124: Do Not Use Regions](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1124.md). I decided not to remove the regions right now, since removing regions might mean other organization strategies are needed instead.
* [SA1600: Elements Must Be Documented](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1600.md). All public elements should have at least a `<summary>` or an `<inheritdoc/>` (for overrides). We have lots of undocumented public things, but I think we should keep the warnings as a reminder to fix them and to write documentation for new code.


ScottCarda-MS:
It looks like a fair amount of the changes in this PR are the result of applying the new style rules to existing code. Could you please point out the changes that cause the new style rules to be implemented so we can separate out cause from effect in the code changes?

ScottCarda-MS:
Could you please make tasks for some of the easier to fix violations listed in the description? The SA1600 violations might be too much work for a single task, but the others seem like they could be addressed by someone without taking too much time. The ones that are breaking changes would need to be discussed and documented properly before changing them, but I think we should at least create the task to get the ball rolling on that.

samarsha:
> It looks like a fair amount of the changes in this PR are the result of applying the new style rules to existing code. Could you please point out the changes that cause the new style rules to be implemented so we can separate out cause from effect in the code changes?

[`.editorconfig`](https://github.com/microsoft/qsharp-compiler/blob/samarsha/stylecop2/.editorconfig) and [`src/Common/stylecop.json`](https://github.com/microsoft/qsharp-compiler/blob/samarsha/stylecop2/src/Common/stylecop.json) determine the style rules by modifying [StyleCop's default rule set](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/rulesets/StyleCopAnalyzersDefault.ruleset). StyleCop has been added to each C# project with two lines:

* `<PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.164" PrivateAssets="All" />`
* `<AdditionalFiles Include="..\..\Common\stylecop.json" Link="stylecop.json" />`

samarsha:
> Could you please make tasks for some of the easier to fix violations listed in the description? The SA1600 violations might be too much work for a single task, but the others seem like they could be addressed by someone without taking too much time. The ones that are breaking changes would need to be discussed and documented properly before changing them, but I think we should at least create the tasks to get the ball rolling on that.

My understanding of the coding style guidelines for the team is that we shouldn't fix style issues in existing code unless it's part of other changes to a particular area of the code. Of course, this PR violates that rule, but I think it's an exception because the current coding style can be hard to read and is applied inconsistently, so doing a one-time restyling will help readability and productivity.

For the remaining style violations, my plan was to use the style warnings as TODOs for things to fix if those parts of the code are being changed later, instead of creating separate tasks. However, the extra build warnings in the meantime are not ideal, so I could suppress the warnings for the existing code if that would be better.

ScottCarda-MS:
> > Could you please make tasks for some of the easier to fix violations listed in the description? The SA1600 violations might be too much work for a single task, but the others seem like they could be addressed by someone without taking too much time. The ones that are breaking changes would need to be discussed and documented properly before changing them, but I think we should at least create the tasks to get the ball rolling on that.
> 
> My understanding of the coding style guidelines for the team is that we shouldn't fix style issues in existing code unless it's part of other changes to a particular area of the code. Of course, this PR violates that rule, but I think it's an exception because the current coding style can be hard to read and is applied inconsistently, so doing a one-time restyling will help readability and productivity.
> 
> For the remaining style violations, my plan was to use the style warnings as TODOs for things to fix if those parts of the code are being changed later, instead of creating separate tasks. However, the extra build warnings in the meantime are not ideal, so I could suppress the warnings for the existing code with `#pragma warning disable` lines, if that would be better.

For the code that is not public facing, I think I might like the warnings as a kind of ToDo list. I would think the guidelines on not fixing style issues wouldn't apply to correcting inconsistent style in public facing API's like the ones that cause the breaking violations. It just seems to be a matter of presenting professionalism in our product to have the API style consistent, and I believe that needs special attention.

samarsha:
I should clarify what the naming issues are:

1. [Protected field `SubSelector`](https://github.com/microsoft/qsharp-compiler/blob/samarsha/stylecop2/src/QsCompiler/Transformations/BasicTransformations.cs#L238) in `SelectByFoldingOverExpressions.StatementTransformation` should start with a lowercase letter.
2. In [`LanguageServer/Program.cs`](https://github.com/microsoft/qsharp-compiler/blob/samarsha/stylecop2/src/QsCompiler/LanguageServer/Program.cs#L17), the protected fields `CONNECTION_VIA_SOCKET` and `CONNECTION_VIA_PIPE` should use `PascalCase`.
3. In [`CommandLineTool/Options.cs`](https://github.com/microsoft/qsharp-compiler/blob/samarsha/stylecop2/src/QsCompiler/CommandLineTool/Options.cs) and [`CommandLineTool/Program.cs`](https://github.com/microsoft/qsharp-compiler/blob/samarsha/stylecop2/src/QsCompiler/CommandLineTool/Program.cs), there are several protected/public constants that use `UPPER_CASE`.

The first one is really minor, and I'm also not sure why StyleCop decided that protected fields should start with a lowercase letter.

The last two are only technically breaking changes because they're public - I don't think the public API of the language server and command line executables is really intended to be used. I'm not sure why these fields are public, either - it seems like they could simply be made private/internal.

src/QsCompiler/CompilationManager/ProcessingQueue.cs
```diff
@@ -4,142 +4,191 @@
...
+            if (processing == null)
+            {
+                throw new ArgumentNullException(nameof(processing));
+            }
+            this.QueueForExecution(
+                () =>
+                {
+                    processing();
+                    return new object();
+                }, out _);
```
ScottCarda-MS, line 90: Is this the correct style for lambdas moving forward? Does the `() =>` have to be on its own line? And the last parameter is _not_ on its own line?

samarsha: The lambda arrow is on a new line because of [SA1116: Split Parameters Must Start On Line After Declaration](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1116.md), which seems to be triggered since `QueueForExecution` has more than one argument. If `QueueForExecution` had only one argument, you could write:

```csharp
this.QueueForExecution(() =>
{
    processing();
    return new object();
});
```

I'm not sure why `out _` is allowed to be on the same line as the closing brace. Normally you can't have more than one argument on the same line unless all of them are on the same line. Maybe StyleCop considers the closing brace in `}, out _` to not count as part of the previous argument, but I agree it looks weird.

src/QsCompiler/LanguageServer/FileSystemWatcher.cs
```diff
@@ -84,51 +90,62 @@ private System.IO.FileSystemWatcher GetWatcher(string folder, string pattern, No
...
 
             var filters = globPatterns.Select(p => (p, NotifyFilters.FileName | NotifyFilters.LastWrite));
-            if (subfolders) filters = filters.Concat(new (string, NotifyFilters)[] { (String.Empty, NotifyFilters.DirectoryName) });
+            if (subfolders)
+            {
+                filters = filters.Concat(new (string, NotifyFilters)[] { (string.Empty, NotifyFilters.DirectoryName) });
+            }
 
-            return this.Processing.QueueForExecutionAsync(() =>
+            return this.processing.QueueForExecutionAsync(() =>
```
ScottCarda-MS, line 140: This is an example of the style I prefer for lambdas.

samarsha: I think this style is allowed as long as there are no arguments after the lambda.

src/QsCompiler/LanguageServer/FileSystemWatcher.cs
```diff
@@ -231,13 +258,18 @@ private void RecurDeleted(string fullPath)
...
-            if (this.WatchedDirectories.TryGetValue(new Uri(fullPath), out var _)) { } // nothing to do here
-            else this.OnChangedFile(fullPath);
+            if (this.watchedDirectories.TryGetValue(new Uri(fullPath), out _))
+            {
+                // nothing to do here
+            }
+            else
+            {
+                this.OnChangedFile(fullPath);
+            }
```
ScottCarda-MS, line 268: The condition here could be negated, removing the need for an empty code block. Although, doing changes like this might be out of scope for the PR.

src/QsCompiler/LanguageServer/FileSystemWatcher.cs
ScottCarda-MS: Ah okay, I may try to order parameters for functions I write in the future appropriately, so that lambdas are last. That may be a good rule in general.

src/QsCompiler/CompilationManager/ProcessingQueue.cs
ScottCarda-MS: I see. I may try to avoid this in the future by being careful about the parameters I make for functions, because this looks pretty rough to me.

src/QsCompiler/LanguageServer/FileSystemWatcher.cs
samarsha: I agree, but yes, I'd like to avoid doing any changes in this PR that aren't strictly necessary - it's too tempting to go overboard. For example, `TryGetValue` could also be replaced with `ContainsKey` here. :)

src/QsCompiler/CompilationManager/ProcessingQueue.cs
samarsha: If it helps, it is not required to have the argument on the same line as the closing brace. You can also write:

```csharp
this.QueueForExecution(
    () =>
    {
        processing();
        return new object();
    },
    out _);
```

ScottCarda-MS: That does help a bit, but it is still confusing to have the first argument, the lambda, take up multiple lines on the same indentation level because its argument list is separated from its body. At first glace it almost looks like there are 3 arguments to the method.

PR #517, opened 2020-07-14
Replacing deprecated IconUrl with Icon and updating the assets for NuGet packages.

ricardo-espinoza:


src/QuantumSdk/QuantumSdk.nuspec
```diff
@@ -24,5 +24,6 @@
     <file src="ProjectSystem\**\*" target="ProjectSystem" exclude="**\*.v.template"/>
     <file src="Tools\BuildConfiguration\bin\$Configuration$\netcoreapp3.1\publish\**" target="tools\utils" exclude="**\*.pdb" />
     <file src="..\QsCompiler\CommandLineTool\bin\$Configuration$\netcoreapp3.1\publish\**" target="tools\qsc" exclude="**\*.pdb" />
+    <file src="MS_Quantum_Spot_Dev.png" target="images" />
```
anpaz-msft, line 27: would it be better to have it on a single location, to avoid having the same file duplicated?

src/QsCompiler/Compiler/Compiler.nuspec.template
```diff
@@ -13,9 +13,12 @@
...
     <releaseNotes>See: https://docs.microsoft.com/en-us/quantum/relnotes/</releaseNotes>
     <projectUrl>https://github.com/microsoft/qsharp-compiler</projectUrl>
-    <iconUrl>https://secure.gravatar.com/avatar/bd1f02955b2853ba0a3b1cdc2434e8ec.png</iconUrl>
+    <icon>images\MS_Quantum_Spot_Dev.png</icon>
 
     <copyright>$copyright$</copyright>
     <tags>Quantum Q# Qsharp </tags>
   </metadata>
+  <files>
+    <file src="MS_Quantum_Spot_Dev.png" target="images" />
```
anpaz-msft, line 22: `MS_Quantum_Spot_Dev.png` seems like a moutfull. Maybe just `qdk.png` or something simple like that?

src/QuantumSdk/QuantumSdk.nuspec
ricardo-espinoza: I considered it. However, following the example of the extensions, and since the file is small, I opted for keeping a copy together with each package so they are self contained, and also so we could:
 1) Update them independently if needed.
 2) Avoid the dependency on the path or an external folder in case they get moved.
I'm happy to update this to a single location, if you think it would be better....

src/QsCompiler/Compiler/Compiler.nuspec.template
ricardo-espinoza: Renamed as qdk-nuget-icon.png

src/QuantumSdk/QuantumSdk.nuspec
ricardo-espinoza: Per our conversation, I'm moving this to a shared location that will be common across repos.

PR #519, opened 2020-07-14
enabled decompilation for omnisharp

filipw:
since `omnisharp.json` is now included in the repo, I would suggest to enable decompilation support.
we [shipped it recently](https://www.strathweb.com/2020/05/decompilation-support-in-omnisharp-and-c-extension-for-vs-code/) and it improves the quality of working with external DLLs a lot. For these dependencies ILSpy decompiled source is shown (when using VS Code code navigation commands) rather than basic metadata only


omnisharp.json
```diff
@@ -3,6 +3,7 @@
         "enableEditorConfigSupport": true
     },
     "RoslynExtensionsOptions": {
-        "enableAnalyzersSupport": true
+        "enableAnalyzersSupport": true,
+        "enableDecompilationSupport": true
```
samarsha, line 7: Thanks, this looks useful - although I'm curious why it's an option for the repository instead of in a local config file for the user?

filipw: it is a local setting too, just like the other settings you enable via omnisharp.json (analyzers and editorconfig) already 

since you have the file in the repo - and the reason for that is to provide the most optimal editor experience - I thought it would make sense to leverage it to enhance the contributor experience even further 🙂

samarsha: OK, that makes sense. :)

PR #520, opened 2020-07-14
Type Parameter Combine

ScottCarda-MS:
Refactoring the algorithm for combining type parameter resolution dictionaries. Added support for combining nested types.


src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -55,19 +56,78 @@ internal static bool TryCombineTypeResolutionsForTarget(QsQualifiedName target,
...
+                        foreach (var temp3 in GetTypeParameters.Apply(kvp.Value))
+                        {
+                            temp2.Add(new KeyValuePair<Tuple<QsQualifiedName, NonNullable<string>>, Tuple<QsQualifiedName, NonNullable<string>>>(kvp.Key, temp3));
+                        }
+                    }
+                }
+                var mayBeReplaced = temp2
+                    .ToLookup(
+                        kvp => kvp.Value,
+                        kvp => kvp.Key);
```
ScottCarda-MS, line 94: I may need some help untangling this and making it a clean LINQ statement.

ScottCarda-MS: Basically, I have a dictionary that I want to first filter with the `ContainsTypeParameter.Apply`, then each of the remaining KeyValuePairs needs to be mapped through the `GetTypeParameters.Apply` to produce 1 to many new KeyValuePairs per input KeyValuePair. Then I need to take all the resulting KeyValuePairs and create a lookup that swaps keys for values in the pairs.

ScottCarda-MS: What would be even better is if I could just use `GetTypeParameter.Apply` as the filter, because it returns an empty set if there are no type parameters. Then I could get rid of the `ContainsTypeParamter` transformation which is almost a duplicate of the `GetTypeParameter` transformation. Yes, that would be good.

ScottCarda-MS: This has been done.

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -8,6 +8,7 @@
 using Microsoft.Quantum.QsCompiler.DataTypes;
 using Microsoft.Quantum.QsCompiler.SyntaxTokens;
 using Microsoft.Quantum.QsCompiler.SyntaxTree;
+using Microsoft.Quantum.QsCompiler.Transformations.Core;
```
cesarzc, line 11: Regarding TypeParameterResolutions, it would be great to have a comment in the code regarding the meaning of the key Tuple, and maybe a type definition for that would also be helpful.

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -54,6 +55,72 @@ internal static bool TryCombineTypeResolutionsForTarget(QsQualifiedName target,
             return success;
         }
 
+        /// <summary>
+        /// Reversed the dependencies of type parameters resolving to other type parameters in the given
```
cesarzc, line 59:
```suggestion
        /// Reverses the dependencies of type parameters resolving to other type parameters in the given
```

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -54,6 +55,72 @@ internal static bool TryCombineTypeResolutionsForTarget(QsQualifiedName target,
...
+        /// dictionary to create a lookup whose keys are type parameters and whose values are all the type
+        /// parameters that can be updating by knowing the resolution of the lookup's associated key.
```
cesarzc, line 61:
```suggestion
        /// parameters that can be updated by knowing the resolution of the lookup's associated key.
```

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -54,6 +55,72 @@ internal static bool TryCombineTypeResolutionsForTarget(QsQualifiedName target,
...
+        /// </summary>
+        private static ILookup<Tuple<QsQualifiedName, NonNullable<string>>, Tuple<QsQualifiedName, NonNullable<string>>> GetReplaceable(TypeParameterResolutions.Builder typeParamResolutions)
+        {
+             return typeParamResolutions
+                .Select(kvp => (kvp.Key, GetTypeParameters.Apply(kvp.Value)))   // Get any type parameters in the resolution type.
```
cesarzc, line 66: Comments for every part of the query are very useful!

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -54,6 +55,72 @@ internal static bool TryCombineTypeResolutionsForTarget(QsQualifiedName target,
...
+                    kvp => kvp.value,
+                    kvp => kvp.Key);
+        }
+
+        /// <summary>
+        /// Applies all the type parameter resolutions in a dictionary to the whole dictionary,
+        /// resulting in a resolution dictionary that has type parameter keys that are not
+        /// referenced in the values of the dictionary.
+        /// </summary>
+        internal static bool TryCombineTypeResolutionDictionary(out TypeParameterResolutions combined, TypeParameterResolutions resolutionDictionary)
```
cesarzc, line 78: I think it would make it a bit clearer if the parameters' names were `combinedResolutions` and `detachedResolutions`/`independentResolutions`.

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -54,6 +55,72 @@ internal static bool TryCombineTypeResolutionsForTarget(QsQualifiedName target,
...
+             return typeParamResolutions
+                .Select(kvp => (kvp.Key, GetTypeParameters.Apply(kvp.Value)))   // Get any type parameters in the resolution type.
+                .SelectMany(tup => tup.Item2.Select(value => (tup.Key, value))) // For each type parameter found, match it to the dictionary key.
+                .ToLookup(                                                      // Reverse the keys and resulting type parameters to make the lookup.
+                    kvp => kvp.value,
+                    kvp => kvp.Key);
+        }
+
+        /// <summary>
+        /// Applies all the type parameter resolutions in a dictionary to the whole dictionary,
```
cesarzc, line 74:
```suggestion
        /// Adds all the type parameter resolutions in an independent dictionary to a combined one,
```

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -54,6 +55,72 @@ internal static bool TryCombineTypeResolutionsForTarget(QsQualifiedName target,
...
+        /// resulting in a resolution dictionary that has type parameter keys that are not
+        /// referenced in the values of the dictionary.
```
cesarzc, line 76:
```suggestion
       /// referenced in its values.
```

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: LINQ queries can be a bit difficult to read sometimes. Especially this one. I thought it would be useful to add a little extra explanation. Glad to hear it helps!

ScottCarda-MS:
Unfortunately this PR just became a bit more difficult to review as I've had to add in changes to align unrelated parts of the call graph walker to the new style rules. Sorry about that!

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -54,6 +55,72 @@ internal static bool TryCombineTypeResolutionsForTarget(QsQualifiedName target,
...
+                // if a certain type parameter is resolved by the currently processed dictionary.
+                var mayBeReplaced = GetReplaceable(combinedBuilder);
+
+                // Check that we are not constricting a type parameter to another type parameter of the same callable
+                // both before and after updating the current value with the resolutions processed so far.
+                success = success && !IsConstrictiveResolution(typeParam, paramRes);
+                var resolvedParamRes = ResolvedType.ResolveTypeParameters(combinedBuilder.ToImmutable(), paramRes);
+                success = success && !IsConstrictiveResolution(typeParam, resolvedParamRes);
+
+                // Create a dictionary with just the current resolution in it.
```
cesarzc, line 108: The logic from here to the end of the foreach is the same than the one in a section of `TryCombineTypeResolutions`. It might be worthy to create a helper method that takes `singleResolution` and the `IEnumerable` as parameters.

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: This is a good suggestion and I've applied it here as well as to the similar `TryCombineTypeResolutions` method.

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: I've tried to reword the description here. I want to make sure to reflect that the individual resolutions are being applied to the other resolutions in the same dictionary. Let me know if the new description is any clearer.

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: I've added a typedef (in the form of a using statement) for type parameter names and a comment to explain it. Hopefully that makes it a bit clearer.

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: I've added such a method, `UpdatedReplaceableResolutions`, but I'm not sure it was worth it. I needed to add in a lot of parameters to to give it all the context it needs, and it updates the combinedBuilder as a side effect, the return value being the success flag. It just seems a bit awkward. What do you think?

ScottCarda-MS: I could have passed in `singleResolution` and the result from `mayBeReplaced[typeParam]` as parameters, removing the need for `typeParam` and `mayBeReplaced` as parameters, but that is the same number of parameters and less of the shared logic encapsulated. I don't think that would have made the logic any less complex.

src/QsCompiler/Transformations/CallGraph.cs
cesarzc: It is much clearer, thanks!

src/QsCompiler/Transformations/CallGraph.cs
cesarzc: Thanks!

src/QsCompiler/Transformations/CallGraph.cs
cesarzc: The rewording is better than these small changes :)

src/QsCompiler/Transformations/CallGraph.cs
cesarzc: I think a bit of the awkwardness can be improved by removing the success flag as a parameter and just return the success that is local to the helper method. The lines in which the helper method is invoked have to be updated to something like the following:
```
success = success && UpdatedReplaceableResolutions(mayBeReplaced, combinedBuilder, typeParam, resolvedParamRes);
```

I think this makes things a bit less awkward since it is a pattern that is already present in other parts of the code.

Having said that, I think the helper method is worth it but I am flexible on whether you want to incorporate it. Maybe we can ask someone else's opinion and go from there?

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -16,10 +17,58 @@
 namespace Microsoft.Quantum.QsCompiler.DependencyAnalysis
 {
     using ResolvedTypeKind = QsTypeKind<ResolvedType, UserDefinedType, QsTypeParameter, CallableInformation>;
-    using TypeParameterResolutions = ImmutableDictionary<Tuple<QsQualifiedName, NonNullable<string>>, ResolvedType>;
+    // Type Parameters are frequently referenced by the callable of the type parameter followed by the name of the specific type parameter.
+    using TypeParameterName = Tuple<QsQualifiedName, NonNullable<string>>;
+    using TypeParameterResolutions = ImmutableDictionary</*TypeParamterName*/ Tuple<QsQualifiedName, NonNullable<string>>, ResolvedType>;
```
samarsha, line 22: Should `/*TypeParamterName*/` be uncommented?

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: If we don't pass in the success flag, we lose the efficiency gain from not having to traverse the resolved type when checking for constriction when we already know that the validation has failed. This would be simpler if we weren't trying to do the validation at the same time as the combining, but the two tasks are so similar it would be inefficient and needlessly redundant to do them separately.

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -68,64 +160,144 @@ internal static bool TryCombineTypeResolutionsForTarget(QsQualifiedName target,
...
+            public static HashSet<TypeParameterName> Apply(ResolvedType res)
+            {
+                var walker = new GetTypeParameters();
+                walker.OnType(res);
+                return walker.SharedState.TypeParams;
+            }
+
+            internal class TransformationState
+            {
+                public HashSet<TypeParameterName> TypeParams = new HashSet<TypeParameterName>();
```
samarsha, line 230: You could directly make this transformation a `TypeTransformation<HashSet<TypeParameterName>>`.

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -68,64 +160,144 @@ internal static bool TryCombineTypeResolutionsForTarget(QsQualifiedName target,
...
+            public static bool Apply(TypeParameterName typeParam, ResolvedType typeParamRes)
+            {
+                var walker = new ConstrictionCheck(typeParam);
+                walker.OnType(typeParamRes);
+                return walker.SharedState.IsConstrictive;
+            }
+
+            internal class TransformationState
+            {
+                public bool IsConstrictive = false;
```
samarsha, line 268: This could also be a `TypeTransformation<bool>`.

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -68,64 +160,144 @@ internal static bool TryCombineTypeResolutionsForTarget(QsQualifiedName target,
...
+            public new ResolvedType OnType(ResolvedType t)
+            {
+                // Short-circuit if we already know the type is constrictive.
+                if (!this.SharedState.IsConstrictive)
+                {
+                    base.OnType(t);
+                }
+
+                // It doesn't matter what we return because this is a walker.
+                return ResolvedType.New(ResolvedTypeKind.InvalidType);
```
samarsha, line 286: Since the return value doesn't matter, it might be simpler to just do `return t;`.

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -199,8 +378,15 @@ public ImmutableArray<CallGraphEdge> InsertEdge(ImmutableArray<CallGraphEdge> ed
         /// </summary>
         public static CallGraphEdge CombinePathIntoSingleEdge(CallGraphNode targetNode, params CallGraphEdge[] edges)
         {
-            if (targetNode == null) throw new ArgumentNullException(nameof(targetNode));
-            if (edges == null || edges.Any(e => e == null)) throw new ArgumentNullException(nameof(edges));
+            if (targetNode == null)
+            {
+                throw new ArgumentNullException(nameof(targetNode));
+            }
+
```
samarsha, line 385: I think the blank line after closing braces is optional, right? (Just checking to make sure StyleCop is configured correctly.)

src/QsCompiler/Transformations/CallGraph.cs
samarsha: I think that it is better to favor simple and clear code over efficient code, unless the code is demonstrably inefficient (i.e. a profiler or performance test shows that it is too slow to be usable).

Just wanted to make sure I'm understanding the behavior, but

```
success = success && !ConstrictionCheck.Apply(keyInCombined, paramRes);
combinedBuilder[keyInCombined] = ResolvedType.ResolveTypeParameters(singleResolution, combinedBuilder[keyInCombined]);
```

The first line does not always need to run, but the second line should always run? So

```
success = success && UpdatedReplaceableResolutions(mayBeReplaced, combinedBuilder, typeParam, resolvedParamRes);
```

would be incorrect?

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: Unfortunately C# doesn't allow previous `using` definitions to be used to define later `using` statements, so I can't actually use `TypeParameterName` in the definition of `TypeParameterResolutions` despite the fact that it should be. But I did want to put something in there so that people know that if they change `TypeParameterName`, they should also update `TypeParameterResolutions`. This is the solution I've used in the past when dealing with this issue.

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: It seems to be, the black line was automatically inserted when I automatically resolved the styling issues. When I remove the space between it looks at a glace like an `else if` rather that two unrelated `if` statements, so I prefer the blank line here.

src/QsCompiler/Transformations/CallGraph.cs
samarsha: I think the types are complicated enough at this point that it would be benefit from creating a new type for `Tuple<QsQualifiedName, NonNullable<string>>`.

```csharp
struct TypeParameterName
{
    QsQualifiedName Callable { get; }
    NonNullable<string> Name { get; }

    TypeParameterName(QsQualifiedName callable, NonNullable<string> name)
    {
        this.Callable = callable;
        this.Name = name;
    }
}
```

This would also make the `typeParamName.Item1` and `typeParamName.Item2` references more readable. (Also, when C# 9 comes out it'll be easier to define immutable records like this...)

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: I could skip over having an explicit TransformationState class, but then it is not as clear what is being used when referencing `this.SharedState` in the code of the transformation. `this.SharedState.TypeParams.Add(...)` is more clear to me than `this.SharedState.Add(...)`. In general I prefer to have a wrapper class for my state when writing and dealing with transformations.

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: I could, but please see my response to the similar comment.

ScottCarda-MS:
I've created a separate PR for the style changes: #525. Once that PR is complete, the diff here will no longer contain the style changes and will be more streamline to the purpose of the PR.

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: Yes, but `success` and the method call can be swapped to make it right.

ScottCarda-MS: I think we could also use the bit-wise `&` to prevent short-circuiting, but that seems a bit esoteric. 😁

cesarzc: > 
> 
> I think that it is better to favor simple and clear code over efficient code, unless the code is demonstrably inefficient (i.e. a profiler or performance test shows that it is too slow to be usable).
> 
> Just wanted to make sure I'm understanding the behavior, but
> 
> ```cs
> success = success && !ConstrictionCheck.Apply(keyInCombined, paramRes);
> combinedBuilder[keyInCombined] = ResolvedType.ResolveTypeParameters(singleResolution, combinedBuilder[keyInCombined]);
> ```
> 
> The first line does not always need to run, but the second line should always run? So
> 
> ```cs
> success = success && UpdatedReplaceableResolutions(mayBeReplaced, combinedBuilder, typeParam, resolvedParamRes);
> ```
> 
> would be incorrect?

Yes, it would be incorrect, sorry for the confusion, but as @ScottCarda-MS mentioned, it could be swapped to obtain the correct behavior.

@samarsha what do you think having the helper method? Do you think it is worth having it?

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: I'm not comfortable at this time to create a new type for this. I think the typedef does an adequate job of encapsulating and communicating what these objects are. If it were possible to add tuple field names to C# typedefs, I would do that for the extra clarity of having something other than `Item1` and `Item2`, but for now the language doesn't support that.

samarsha: I think that if transformation states with one field deserve a new type, then `TypeParameterName` does too. :) If C# 9 was here, we could do

```csharp

struct TypeParameterName
{
    QsQualifiedName Callable { get; init; }
    NonNullable<string> Name { get; init; }
}
```

which does a lot to make new types feel less heavyweight, but unfortunately that isn't possible in C# 8 without making the type mutable.

ScottCarda-MS: I really don't feel it's appropriate to make a new type here for this. The transformation state classes are part of the pattern that the transformations use. The typedef for TypeParameterName is a sufficient and more lightweight solution to encapsulate the tuple. I don't want the C# type checking the type parameter references throughout this code expecting a nuanced type where once the standard tuple was used to represent type parameters.

samarsha: I have never seen code that had a comment embedded inside of a type parameter before, and my first impression is that it was a mistake or typo. I agree that it would be better if C# let using aliases reference other using aliases, but since that is not possible, I think it is an indication that the type structure is too complex for aliases. Do you have any examples of similar code that uses nested aliases with embedded comments?

Also, there is a typo in the comment: `/*TypeParamterName*/` -> `/*TypeParameterName*/`

ScottCarda-MS: Ah thanks for catching the typo. It's fixed now.

ScottCarda-MS: Monomorphization is doing something similar with the using statements. https://github.com/microsoft/qsharp-compiler/blob/master/src/QsCompiler/Transformations/Monomorphization.cs#L18

src/QsCompiler/Transformations/CallGraph.cs
samarsha: I think the helper method is worth having. Talking with @ScottCarda-MS, we had some ideas for simplifying way the success state is threaded through the methods, but no simple solutions for now, so I'm OK with the way it is.

Issue #506, opened 2020-07-06
Simplify position and range handling for syntax tree nodes


samarsha:
The current API for source code position and ranges has some issues that make it hard to use:

* Use of zero- or one-based positions is inconsistent/unspecified.
* Use of named vs anonymous types is inconsistent, e.g. `int * int` vs `QsPositionInfo`.
* Naming is confusing, e.g. `QsPositionInfo * QsPositionInfo` is used to represent required ranges but `QsRangeInfo` is used to represent optional ranges.
* All position handling is done in DiagnosticTools in CompilationManager, but this is inconvenient because other projects need to generate diagnostics that requires some conversions on the ranges. Those conversions are currently delayed by storing more information in tuples and passing them back to the CompilationManager, which can be confusing.

It would be nice to clean up and refactor the way positions and ranges are handled.

samarsha:
@bettinaheim This is my idea so far:

## Types

```fsharp
// Zero-based line and column numbers.
type Position = { Line : int; Column : int }
type Range = { Start : Position; End : Position }
```

I think we can remove `QsLocation`. The name is confusing (why is "location" different than "position?"). We could store the offset and range separately in the syntax node. Or, since I think `QsLocation` is only used for declarations and statements, the range always starts at (0, 0) relative to the offset, anyway, so we could combine them into just the range of the fragment header. Then `Offset` would be `Range.Start`.

To accommodate generated syntax nodes, all nodes have optional position and range.

## Relative positions

Keep the same relative position hierarchy:

1. Start of file
2. Declaration
3. Statement

The only part I am concerned about is specializations vs. callables, and how the position of parameters in specializations are handled, where `Position` is null if they are declared in the parent and non-null if declared in the specialization:

```fsharp
type LocalVariableDeclaration<'Name> = {
    // ...
    Position : QsNullable<int*int>
    Range : QsPositionInfo * QsPositionInfo
}
```

I'm tempted to avoid this special case by saying that local variable positions are always relative to the specialization (either implicit or explicit). For parameters of an explicit specialization that were declared in the parent callable, their range can be negative, such that `Offset + Range` still gives the correct absolute position.

## FParsec and LSP

We should convert from FParsec positions as early as possible and convert to LSP positions as late as possible. We should try to use our position types instead, as much as possible.

PR #523, opened 2020-07-16
Unify position and range types in parser and syntax tree

samarsha:
I added two new types, `Position` and `Range`, replacing these types:

* `FParsec.Position`
* `int * int`
* `QsPositionInfo`
* `QsPositionInfo * QsPositionInfo`
* `QsRangeInfo`

The new `Position` type is zero-based, so it has a well-defined addition operator. I replaced FParsec's `getPosition` parser with one that returns the new zero-based `Position` type directly, so the compiler never has to deal with one-based positions anymore.

This is the first step in #506. Now that these types are defined, in the next PR I'll continue updating the diagnostics code to use them, replacing the LSP position and range types as much as possible.


samarsha:
C# generation tests are currently broken for an unknown reason (they pass on my computer, but not in the CI). Other than that, I think it's ready to review.

samarsha:
All tests passing now.

src/QsCompiler/CompilationManager/ContextBuilder.cs
```diff
@@ -31,7 +32,7 @@ internal static void VerifyTokenOrdering(IEnumerable<CodeFragment> tokens)
                 throw new ArgumentNullException(nameof(tokens));
             }
 
-            Position previousEnding = null;
+            Lsp.Position previousEnding = null;
```
ScottCarda-MS, line 35: Since the Protocol namespace is already opened, why do you need to qualify `Position` with `Lsp`? Is it conflicting with something? If it is just to make the code easier to read, then I might suggest adding the qualification to the `Diagnostic` type references in this file as well and removing line 13 which opens the namespace.

src/QsCompiler/DataStructures/DataTypes.fs
```diff
@@ -53,61 +53,118 @@ type QsNullable<'T> = // to avoid having to include the F# core in the C# part o
         | Some v -> Value v
         | None -> Null
 
+/// Operations for QsNullable<'T>.
+module QsNullable =
+    /// Applies 'f' to both nullable arguments if both have a Value. Returns Null otherwise.
+    let Map2 f a b =
```
ScottCarda-MS, line 59: nit: It may be useful to document the typing here, noting that Map2 returns a `QsNullable<'U>` for all `f : 'T * 'T -> 'U`.

src/QsCompiler/CompilationManager/ContextBuilder.cs
samarsha: Without `Lsp`, the reference is ambiguous, because `Microsoft.Quantum.QsCompiler.DataTypes.Position` is also in scope. I think I like the idea of qualifying all LSP types, even unambiguous ones like `Diagnostic`, though.

src/QsCompiler/DataStructures/DataTypes.fs
samarsha: The type is more general than that (`a` and `b` don't have to be the same type, so `f : 'a -> 'b -> 'c`), but it's rare to mention types in F# if they are inferred correctly. Do you mean as a type annotation or part of the doc comment?

src/QsCompiler/DataStructures/DataTypes.fs
```diff
@@ -53,61 +53,118 @@ type QsNullable<'T> = // to avoid having to include the F# core in the C# part o
...
+
+    /// The line number, where line zero is the first line in the document.
+    member this.Line = match this with Position (line, _) -> line
+
+    /// The column number, where column zero is the first column in the line.
+    member this.Column = match this with Position (_, column) -> column
+
+    /// Translates the first position by the amount of the second position. If the resulting position is on the same
+    /// line, the column numbers are added; otherwise, the second position's column number is used.
+    static member (+) (a : Position, b : Position) =
```
ScottCarda-MS, line 81: This seems like a confusing definition of addition. Why isn't it just `(x1, y1) + (x2, y2) = (x1 + x2, y1 + y2)`? Why does the behavior have to change when the line number of the second argument is zero?

src/QsCompiler/DataStructures/DataTypes.fs
```diff
@@ -53,61 +53,118 @@ type QsNullable<'T> = // to avoid having to include the F# core in the C# part o
...
+        let line = a.Line + b.Line
+        let column = if b.Line = 0 then a.Column + b.Column else b.Column
+        Position (line, column)
+
+    /// Returns true if the positions have the same line and column numbers.
+    static member op_Equality (a : Position, b : Position) = a = b
```
ScottCarda-MS, line 87: Isn't `op_Equality` a violation of the naming rules? Also, what does the `op` prefix refer to?

src/QsCompiler/CompilationManager/ContextBuilder.cs
samarsha: If I consistently apply the rule that `Microsoft.VisualStudio.LanguageServer.Protocol` should always be opened with an alias to all the files changed in this PR, though, there will be a lot of changes, so it is probably better to only use `Lsp.` for ambiguous types for now.

src/QsCompiler/DataStructures/DataTypes.fs
```diff
@@ -53,61 +53,118 @@ type QsNullable<'T> = // to avoid having to include the F# core in the C# part o
...
+
+    /// Returns true if the second position occurs after the first position. 
+    static member op_GreaterThan (a : Position, b) = (a :> Position IComparable).CompareTo b > 0
+
+    /// <summary>
+    /// Creates a position.
+    /// </summary>
+    /// <exception cref="ArgumentOutOfRangeException">Thrown if the line or column is negative.</exception>
+    static member Create line column =
```
ScottCarda-MS, line 96: Maybe this is just my inexperience with F#, but why have a private constructor only to have a public static Create method? Couldn't you have just made the constructor public?

src/QsCompiler/DataStructures/DataTypes.fs
```diff
@@ -53,61 +53,118 @@ type QsNullable<'T> = // to avoid having to include the F# core in the C# part o
...
+    /// </exception>
+    static member Create start ``end`` =
+        if start > ``end``
+        then ArgumentException "Range start cannot occur after range end." |> raise
+        Range (start, ``end``)
 
-and QsRangeInfo = 
-    QsNullable<QsPositionInfo * QsPositionInfo>
+    /// The empty range starting and ending at position zero.
+    static member Zero = Range (Position.Zero, Position.Zero)
```
ScottCarda-MS, line 132: Are Ranges with zero 'length' generally allowed? I see that negative 'length' ranges are rejected, but conceptually what would a zero 'length' range be used for at any given position?

src/QsCompiler/DataStructures/DataTypes.fs
samarsha: It's not possible to define a custom constructor for an F# DU type.

src/QsCompiler/DataStructures/DataTypes.fs
ScottCarda-MS, line 132: nit: Might be useful to have a member `Length` on Range. Could return the number of lines spanned and the number of columns spanned, maybe? Although I suppose columns spanned could be negative when lines spanned is greater than zero.

src/QsCompiler/CompilationManager/ContextBuilder.cs
ScottCarda-MS: Ah okay so it _does_ resolve an ambiguity. Makes sense.

src/QsCompiler/DataStructures/DataTypes.fs
ScottCarda-MS: I meant as part of the doc comment, but if it is abnormal to talk about types in F# then maybe it is just that I need to adjust my way of thinking. I am used to understanding code in terms of types, so when I approached this I was thinking, "What are the types?". But maybe that isn't the F# way?

samarsha: I think in terms of types too, and in Haskell this function would have an explicit type annotation, but the F# way seems to be to omit type information whenever possible. I would prefer not to include types in the doc comment because then the types mentioned would not be checked by the compiler and could be wrong/out of date, whereas type annotations are checked for correctness by the compiler.

src/QsCompiler/DataStructures/DataTypes.fs
ScottCarda-MS: Okay so the logic has to take the second position's col when when dealing with multiple lines because the col is reset every new line. But when there are no new lines, the col is relative instead of absolute so they need to be added in that case. This all adds up for me 😄.

src/QsCompiler/DataStructures/DataTypes.fs
samarsha: I think the most useful definition of length is the number of characters between the start and end, but I think it would need access to the original string to know that. Range length wasn't needed in any of the code I refactored, so I didn't add it.

src/QsCompiler/Tests.Compiler/SerializationTests.fs
```diff
@@ -58,7 +58,7 @@ module SerializationTests =
         Type = t |> ResolvedType.New
         InferredInformation = InferredExpressionInformation.New (false, false)
         Position = Null
-        Range = {Line = 1; Column = s}, {Line = 1; Column = e}
+        Range = Range.Create (Position.Create 1 s) (Position.Create 1 e)
```
ScottCarda-MS, line 61: Since we have gone from one-based to zero-based, should the literals in this be 0 instead of 1?

ScottCarda-MS: I guess I'm concerned that all the literals in this file are now one greater than they should be.

src/QsCompiler/DataStructures/DataTypes.fs
samarsha: The `op_` prefix is how overloaded operators are represented internally in .NET: https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/operator-overloading#overloaded-operator-names

I'm not sure why F# doesn't do this automatically, but I need to manually define `op_Equality` to get structural equality with `==` in C#. Otherwise, I would have to use `.Equals` in C#, which is less nice.

src/QsCompiler/Tests.Compiler/SerializationTests.fs
samarsha: I think you're right. It didn't matter for the test since it serializes to JSON and back based on the ranges given in the file, but updating the range positions seems reasonable.

src/QsCompiler/TextProcessor/SyntaxBuilder.fs
```diff
@@ -337,25 +332,26 @@ let internal buildTupleItem validSingle bundle errCode missingCode fallback cont
...
-    let identifier = 
-        let id = IdentifierOptions(
-                     isAsciiIdStart = isSymbolStart,
-                     isAsciiIdContinue = isSymbolContinuation,
-                     preCheckStart = isSymbolStart,
-                     preCheckContinue = isSymbolContinuation) |> identifier
-        getPosition .>>. id .>>. getPosition |>> fun ((p1, name), p2) -> name, (p1,p2)
-    let whenValid (name : string, range) = 
+let internal symbolNameLike errCode =
+    let id =
```
ScottCarda-MS, line 336: I think it was named `identifier` to avoid confusion with the built-in F# function `id`.

samarsha: That makes sense, although `identifier` is also the name of an FParsec parser that is used in the definition of `id`... I renamed this variable to `ident`.

src/QsCompiler/Tests.Compiler/SerializationTests.fs
ScottCarda-MS: All of the literals for HeaderRange and SymbolRange were reduced by one, but the literals on the Position fields were not. Is that okay here?

src/QsCompiler/DataStructures/DataTypes.fs
ScottCarda-MS: I'd be curious to find out if empty ranges include the single character or not. But I won't hold up the PR on that 😄.

src/QsCompiler/Tests.Compiler/SerializationTests.fs
samarsha: The position fields used to be of type `int * int`, while the ranges were `QsPositionInfo * QsPositionInfo`. I think `int * int` was zero-based and `QsPositionInfo` was one-based. (It's really confusing...) So I only subtracted 1 from the ones that used to be `QsPositionInfo`s.

PR #525, opened 2020-07-16
Fixed style in call graph and call graph walker.

ScottCarda-MS:
Updated the call graph and call graph walker to be in line with the new style rules.


samarsha:
Are these the same changes that are also in #520?

ScottCarda-MS:
> Are these the same changes that are also in #520?

Yah, @cesarzc suggested making a separate PR for the style fixes. Once this PR is in the diff on #520 will be more streamlined.

PR #526, opened 2020-07-17
Do not include dependencies with PrivateAssets="All" in the compiler package

samarsha:
`FindNuspecReferences.ps1` adds all dependencies to `Compiler.nuspec`, even ones marked with `PrivateAssets="All"`. That meant that the StyleCop analyzers for the Q# compiler source code were propagating to projects that have a reference to the Q# compiler package, which is a bit unwelcome for downstream projects.

### DIFF
--- a/src/QsCompiler/Compiler/FindNuspecReferences.ps1
+++ b/src/QsCompiler/Compiler/FindNuspecReferences.ps1
@@ -37,7 +37,7 @@ function Add-NuGetDependencyFromCsprojToNuspec($PathToCsproj)
 
         # Check if package already added as dependency, only add if new:
         $added = $dep.dependency | Where { $_.id -eq $id }
-        if (!$added) {
+        if (!$added -and ($_.PrivateAssets -ne "All")) {
             Write-Host "Adding $id"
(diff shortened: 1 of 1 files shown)


samarsha:
It seems that `PrivateAssets="All"` implies that something is a development-only (i.e. build-time only) reference, and shouldn't be a package dependency:

* https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#controlling-dependency-assets
* https://github.com/NuGet/Home/issues/4125#issuecomment-280787267 referring to `dotnet pack`:

  > pack should not emit a dependency when `<PrivateAssets>all</PrivateAssets>` is specified

filipw:
I just ran into this today. Thanks for fixing this, currently the compiler nuget package drags in Stylecop reference when installed and ruins the project's code analysis 😂 it would be good to release the new version out

samarsha:
Yes, sorry about that! I believe you can work around it for now by [adding this .editorconfig](https://gist.github.com/samarsha/f537b60a35ac80af68710a8a48d053ad) to your project, which should disable all StyleCop rules...

PR #529, opened 2020-07-21
Replace LSP positions with Q# positions

samarsha:
Replaces LSP position types with Q# position types in the compilation manager. Removed several functions that operated on LSP positions that are no longer necessary because the functionality is built into the Q# position type.

Because LSP ranges haven't been replaced yet, a lot of conversions like `range.Start.ToQSharp()` and `range.End.ToQSharp()` needed to be added temporarily. Those should go away when LSP ranges are replaced with Q# ranges that use Q# positions for their start and end points. (Probably in the next PR.)

Part of #506.

### DIFF
--- a/src/QsCompiler/CompilationManager/DiagnosticTools.cs
+++ b/src/QsCompiler/CompilationManager/DiagnosticTools.cs
@@ -35,36 +35,6 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
                 : new Lsp.Position(position.Line, position.Column);
 
         /// <summary>
-        /// Adds the Q# position to the language server protocol position, returning a new language server protocol
-        /// position.
-        /// </summary>
-        /// <param name="position1">
-        /// The language server protocol position. Null is equivalent to the zero position.
-        /// </param>
-        /// <param name="position2">The Q# position.</param>
-        /// <exception cref="ArgumentException">Thrown if <paramref name="position1"/> is invalid.</exception>
-        /// <exception cref="ArgumentNullException">Thrown if <paramref name="position2"/> is null.</exception>
-        internal static Lsp.Position GetAbsolutePosition(Lsp.Position position1, Position position2)
-        {
-            // TODO: This should be replaced with `position1 + position2` once both are Q# positions.
-            if (!Utils.IsValidPosition(position1))
-            {
-                throw new ArgumentException(nameof(position1));
-            }
-            if (position2 is null)
-            {
-                throw new ArgumentNullException(nameof(position2));
-            }
-            var absPos = position1?.Copy() ?? new Lsp.Position();
-            absPos.Line += position2.Line;
-            absPos.Character =
-                position2.Line > 0
-                    ? position2.Column
-                    : absPos.Character + position2.Column;
-            return absPos;
-        }
-
-        /// <summary>
         /// Adds the language server protocol position to the Q# compiler range.
         /// </summary>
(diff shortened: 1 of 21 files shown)


src/QsCompiler/CompilationManager/DiagnosticTools.cs
```diff
@@ -311,33 +249,29 @@ internal static bool SelectByEndLine(this Diagnostic m, int lowerBound, int uppe
...
         /// Returns true if the start position of the given diagnostic is larger or equal to lowerBound, and smaller than upperBound.
         /// </summary>
-        internal static bool SelectByStart(this Diagnostic m, Lsp.Position lowerBound, Lsp.Position upperBound)
-        {
-            return m?.Range?.Start?.Line == null ? false : lowerBound.IsSmallerThanOrEqualTo(m.Range.Start) && m.Range.Start.IsSmallerThan(upperBound);
-        }
+        internal static bool SelectByStart(this Diagnostic m, Position lowerBound, Position upperBound) =>
+            m?.Range?.Start?.Line != null
+            && lowerBound <= m.Range.Start.ToQSharp()
+            && m.Range.Start.ToQSharp() < upperBound;
```
ScottCarda-MS, line 261: Isn't there already a method, I think on the new Range type, that checks if a Position is within the Range? It seems like that logic should be centralized.

ScottCarda-MS: Hmm, I guess not. But maybe there should be?

src/QsCompiler/DataStructures/DataTypes.fs
```diff
@@ -83,12 +83,33 @@ type Position = private Position of int * int with
...
         Position (line, column)
 
+    /// <summary>
+    /// Translates the first position backwards by the amount of the second position. If the resulting position is on
+    /// the same line, the column numbers are subtracted; otherwise, the first position's column number is used.
+    /// </summary>
+    /// <exception cref="ArgumentOutOfRangeException">
+    /// Thrown if the resulting position has a negative line or column number.
+    /// </exception>
+    static member (-) (a : Position, b : Position) =
```
ScottCarda-MS, line 93: Similar to what you did with the `(+)` operator, can you provide an example for why this logic is what it is?

samarsha: It's the inverse of `+`, so if `absolute = offset + relative`, then `absolute - offset = relative`. I think the common rule is that the column number of position that occurs later is the one that's preserved. In `offset + relative`, `relative` occurs after `offset`, so `+` keeps the RHS column number. In `absolute - offset`, `absolute` occurs after `offset`, so `-` keeps the LHS column number.

src/QsCompiler/CompilationManager/DiagnosticTools.cs
samarsha: I added `Range.Contains` and rewrote these two functions.

src/QsCompiler/DataStructures/DataTypes.fs
ScottCarda-MS: Ah okay so the addition operator is not commutative. I didn't see that before. It could be good to put that in the doc for the `(+)`. It could also be good to give more descriptive names to the inputs like what you did in the example you just gave.

samarsha: Yes, sounds good, I updated docs and parameter names for `+` and `-`.

PR #535, opened 2020-07-27
Implement VS Code Formatter

theRoughCode:
This PR implements the skeleton for the VS Code formatter.  We have 2 basic rules for now:

1. Strip away any whitespace to the left of `namespace`. For example,
`   namespace Microsoft.Quantum {}` -> `namespace Microsoft.Quantum {}`.
2. Format operation definitions. For example:
`operation   Foo(q:Qubit, n   :Int)   :  Bool` -> `operation Foo (q : Qubit, n : Int) : Bool`


theRoughCode:
cc: @amrondonp, @swosti99

swernli:
Apologies for the delay on the review; I thought I had hit submit but had it sitting open in a browser instead. Great work so far!

swernli:
Looks like there was a failure in the testing of the VSCode extension. Is it possible something in these changes is showing up as a unit test when it shouldn't?
https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=12399&view=logs&j=aa540828-7369-5cca-9a86-6547727f16ff&t=5c10f11e-357e-5f9b-e70d-e2e6470f089e&l=1055

swernli:
Actually, looks like quite a few commits were added between my original review and now. I'm going to take a look back through the files again. Thanks for adding testing!

build/test.ps1
```diff
@@ -38,7 +38,17 @@ function Test-One {
     }
 }
 
+function Test-VSCode {
+    Param([string] $path);
+
+    Write-Host "##[info]Testing VSCode Extension inside $path..."    
+    Push-Location (Join-Path $PSScriptRoot $path)
+        npm run unittest
```
swernli, line 46: I think this should fix the break in the test pipeline.
```suggestion
        npm test
```

build/test.ps1
```diff
@@ -38,7 +38,17 @@ function Test-One {
...
+    Pop-Location
+}
```
swernli, line 48: You'll need to have something like this to make sure the failed tests get reported as expected.
```suggestion
    if  ($LastExitCode -ne 0) {
        Write-Host "##vso[task.logissue type=error;]Failed to test VSCodeExtension at $path."
        $script:all_ok = $False
    }
}
```

build/test.ps1
```diff
@@ -38,7 +38,17 @@ function Test-One {
...
+
 Test-One '../QsCompiler.sln'
+Test-VSCode '../src/VSCodeExtension'
```
swernli, line 51: To match the behavior the build.ps1 uses, you should skip executing VSCodeExtension tests if the environment variable is set to false.
```suggestion
if ($Env:ENABLE_VSIX -ne "false") {
    Test-VSCode '../src/VSCodeExtension'
} else {
    Write-Host "##vso[task.logissue type=warning;]VSIX building skipped due to ENABLE_VSIX variable."
}
```

src/VSCodeExtension/src/tests/unit/control-structure.test.ts
```diff
@@ -0,0 +1,107 @@
...
+    const code = "if(1 == 1){H(qs[0]);}";
+    const expectedCode = "if(1 == 1){H(qs[0]);}";
+
+    assert.equal(formatter(code, []), expectedCode);
+  });
+
+  describe("space after if rule", () => {
+    it("adds space after if statement", () => {
+      const code = "if(1 == 1){H(qs[0]);}";
+      const expectedCode = "if (1 == 1){H(qs[0]);}";
```
swernli, line 17: Some of these test cases will need to be updated once other rules are implemented (such as requiring a space before and after open curly brace, ie `if (1 == 1) { H(qs[0]); }`. But that's ok to leave this as-is for now and expect those updates to be done by later rule authors.

src/VSCodeExtension/src/formatter/operation-rules.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Formats operation declarations.
+ * 
+ * @param code incoming code
+ * 
+ * @example `operation   Foo(q:Qubit, n   :Int)   :  Bool` ->
+ * `operation Foo (q : Qubit, n : Int) : Bool`
+ */
+const argsRule = (code: string): string => {
+    const operationMatcher: RegExp = /operation\s*(\w+)\s*\((.*)\)\s*:\s*(\w+)/g;
```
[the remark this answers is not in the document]
theRoughCode: What's a generic umbrella term for operations and functions?

build/test.ps1
theRoughCode: I think this is failing because I forgot to edit the template `package.json`.

src/VSCodeExtension/src/tests/unit/control-structure.test.ts
theRoughCode: cc: @amrondonp

build/test.ps1
swernli: Ah, I see the corresponding change below that adds `unittest` as one of the targets.

src/VSCodeExtension/src/formatter/operation-rules.ts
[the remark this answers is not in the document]
swernli: That's a good question, and I don't think there really is one. But if the intent of the file is for checking operation and function declarations, maybe a decent name would be declaration-rules.ts? Then it could also handle user-defined type declarations (`newtype` as documented [here](https://docs.microsoft.com/en-us/quantum/user-guide/language/types#user-defined-types)). That file name change can be made in following PRs, no need to gate this PR on the name change.

src/VSCodeExtension/src/formatter/format-document.ts
```diff
@@ -0,0 +1,33 @@
...
+): vscode.TextEdit[] => {
+    const lines: vscode.TextLine[] = Array.from(
+        Array(document.lineCount),
+        (_, lineNo) => document.lineAt(lineNo)
+    );
+
+    const formattedLines: vscode.TextEdit[] = lines
+        .map((line) => {
+            const formattedLine: string = formatter(line.text, rules);
+            return formattedLine !== line.text
```
cgranade, line 27: Does this constrain all formatting rules to work one line at a time?

src/VSCodeExtension/src/formatter/rules/control-structure.ts
```diff
@@ -0,0 +1,17 @@
+import { FormatRule } from "../formatter";
+
+/**
+ * Laves just one space after an if statement
```
cgranade, line 4:
```suggestion
 * Leaves just one space after an if statement.
```

src/VSCodeExtension/src/formatter/rules/control-structure.ts
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * @param code input string
+ *
+ * @example `if                              (1 == 1){H(qs[0]);}` ->
+ * `if (1 == 1){H(qs[0]);}`
+ *
+ * more examples in the unit tests
+ */
+export const spaceAfterIf: FormatRule = (code: string): string => {
+    return code.replace(/(^| +|;|})if[ \n]*/, "$1if ");
```
cgranade, line 14: Using the `*` operator could make the regex match on zero instances of `[ \n]`, such that variable names starting with `if` could be affected. Do you want `+` here to match [1, ∞) copies of `[ \n]`?
```suggestion
    return code.replace(/(^| +|;|})if[ \n]+/, "$1if ");
```

src/VSCodeExtension/src/formatter/rules/indent.ts
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * Removes whitespace from left of `namespace`.
+ *
+ * @param code input string
+ *
+ * @example `   namespace   Foo   {}` ->
+ * `namespace Foo {}`
+ */
+export const namespaceRule: FormatRule = (code: string) => {
+    const namespaceMatcher: RegExp = /^\s*namespace\s*(\w+)\s*(\S|\S.*\S)\s*$/g;
```
cgranade, line 15: Does `\w+` catch dotted names like `Microsoft.Quantum.Arrays`?

src/VSCodeExtension/src/formatter/rules/operation.ts
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+/**
+ * Formats operation declarations.
+ *
+ * @param code incoming code
+ *
+ * @example `operation   Foo(q:Qubit,  n   :Int)   :  Bool` ->
+ * `operation Foo (q : Qubit, n : Int) : Bool`
```
cgranade, line 10: I'd generally suggest not including a space between the operation name and its inputs.

+@swernli for style guide discussion.

src/VSCodeExtension/src/tests/unit/control-structure.test.ts
```diff
@@ -0,0 +1,98 @@
+import "mocha";
```
cgranade, line 1: I'd suggest a unit test to ensure that `if` occuring in the beginning, middle, or end of an identifier name doesn't get replaced.

src/VSCodeExtension/src/tests/unit/indent.test.ts
```diff
@@ -0,0 +1,51 @@
+import "mocha";
```
cgranade, line 1: I'd suggest unit tests to make sure dotted names are preserved as namespace names.

src/VSCodeExtension/src/formatter/format-document.ts
theRoughCode: Yup. I'm not sure how to match the whole document at once. Will definitely have to change this to handle multi-line matching.

src/VSCodeExtension/src/formatter/rules/operation.ts
swernli: Yup! That's one of the open questions, and actually the next one I think we could tackle.

src/VSCodeExtension/src/formatter/rules/control-structure.ts
theRoughCode: We also want to match none in the case of `if(...)` -> `if (...)`. I would suggest the following
```suggestion
    return code.replace(/(^| +|;|})if[ |\n]*\(/, "$1if ");
```

cc: @amrondonp who wrote this

src/VSCodeExtension/src/formatter/operation-rules.ts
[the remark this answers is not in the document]
cgranade: FYI, we've generally used "callable" to mean "function or operation."

src/VSCodeExtension/src/formatter/rules/indent.ts
theRoughCode: Good catch. What do you think of this:

```javascript
/^\s*namespace\s+(\w+(?:\.\w+)*)\s*({|{.*})\s*/$
```

src/VSCodeExtension/src/tests/unit/control-structure.test.ts
theRoughCode: cc: @amrondonp

src/VSCodeExtension/src/formatter/rules/operation.ts
theRoughCode: Deferring to teams channel for discussion.

src/VSCodeExtension/src/tests/unit/control-structure.test.ts
amrondonp: The intention of this rule is only to handle the space after the if statement, We obviously need integration tests that test several rules at one but this specific rule only is concerned about the spacing after the if

src/VSCodeExtension/src/formatter/format-document.ts
amrondonp: we could just go lines.join(""); counldn't we?

src/VSCodeExtension/src/formatter/rules/control-structure.ts
amrondonp: Good suggestion, @theRoughCode thanks!

src/VSCodeExtension/src/formatter/rules/indent.ts
amrondonp: Please make a test for every edge case you encounter

src/VSCodeExtension/src/tests/unit/control-structure.test.ts
amrondonp: Sure, Will do

src/VSCodeExtension/src/formatter/format-document.ts
theRoughCode: It would be a bit more involved than that. We need to return `vscode.TextEdit[]` instead of just the final formatted string.

PR #536, opened 2020-07-27
Integrate space after if, with tests to formatter branch

amrondonp:
* Added 'space after if' rule
* Added unit test for it
* Simple rearrangement of files, created rules folder, separate the formatter implementation from the format file that imports vscode for testing propouses


src/VSCodeExtension/src/formatter/rules/operation.ts
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
theRoughCode, line 1: I prefer the original structure where we expose only an array of rules. So, this `operation.ts` file would expose a set of `rules` that includes `argsRule`. We would then import them and unroll them out, so that we don't clutter with the potentially huge number of rules.

src/VSCodeExtension/src/formatter/rules/space-after-if.ts
```diff
@@ -0,0 +1,5 @@
+import { FormatRule } from "../formatter";
```
theRoughCode, line 1: I would rename this file to something like `controls.ts` and have this as one of the `rules`. My proposal is that each file within `./rules` is a logical grouping of related rules.

src/VSCodeExtension/src/formatter/format-document.ts
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import * as vscode from "vscode";
+import { FormatRule, formatter } from "./formatter";
+import { namespaceRule } from "./rules/indent";
+import { argsRule } from "./rules/operation";
+import { spaceAfterIf } from "./rules/space-after-if";
```
theRoughCode, line 8: Instead of extracting each individual rule, I think it's better to expose just an array of rules to make it cleaner when there could be potentially many rules.

src/VSCodeExtension/src/formatter/rules/indent.ts
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Removes whitespace from left of `namespace`.
+ *
+ * @param code input string
+ *
+ * @example `   namespace ... {}` ->
+ * `namespace ... {}`
+ */
+export const namespaceRule = (code: string): string => {
```
theRoughCode, line 12:
```suggestion
const namespaceRule: FormatRule => {
```

src/VSCodeExtension/src/formatter/rules/space-after-if.ts
```diff
@@ -0,0 +1,5 @@
+import { FormatRule } from "../formatter";
+
+export const spaceAfterIf: FormatRule = (code: string): string => {
```
theRoughCode, line 3: Add a quick doc here similar to the other rules.

src/VSCodeExtension/src/tests/unit/formatter.test.ts
```diff
@@ -0,0 +1,107 @@
...
+    assert.equal(formatter(code, []), expectedCode);
+  });
+
+  describe("space after if rule", () => {
+    it("adds space after if statement", () => {
+      const code = "if(1 == 1){H(qs[0]);}";
+      const expectedCode = "if (1 == 1){H(qs[0]);}";
+
+      assert.equal(formatter(code, [spaceAfterIf]), expectedCode);
+    });
```
theRoughCode, line 20: Nit:
```suggestion
    const rules = [spaceAfterIf];
    it("adds space after if statement", () => {
      const code = "if(1 == 1){H(qs[0]);}";
      const expectedCode = "if (1 == 1){H(qs[0]);}";

      assert.equal(formatter(code, rules), expectedCode);
    });
```

And re-use `rules` throughout test.

src/VSCodeExtension/src/formatter/format-document.ts
amrondonp: If you do that, then you need to have all rules or none of the file? you would need to manually go and "unimport" the ones you don't want?

amrondonp: I think we can do both, export rules indivdually or all of them to an array :) I'll do that

src/VSCodeExtension/src/tests/unit/formatter.test.ts
amrondonp: I don't think that would make any noticieable difference, I think it is better to have the rule explicitly set, i.e. reading [spaceAfterIf] gives more information than rules. in the latter, you would potentially need to go and check for the definition for a simple concept?

src/VSCodeExtension/src/formatter/format-document.ts
amrondonp: I left the option of importing separated rules as well as packing them into an array for ease of use :)

src/VSCodeExtension/src/formatter/rules/space-after-if.ts
amrondonp: remaned to control-structures.ts

src/VSCodeExtension/src/formatter/format-document.ts
theRoughCode: Ah I see, yea, that works. Once we get to implementing optional rules, we might change this further. Maybe passing a list of strings of rules that we don't want and then returning the resulting array of rules.

PR #537, opened 2020-07-28
Minor fixes and match functions in args formatting rule

theRoughCode:
This PR fixes/improves stuff from #535 and adds functions tot the args formatting rule.


src/VSCodeExtension/src/tests/unit/declaration.test.ts
```diff
@@ -52,4 +52,11 @@ describe("arguments rule", () => {
         code = "operation Foo (q : Qubit,   n : Int) : Bool";
         assert.equal(formatter(code, [argsRule]), expectedCode);
     });
+
+    it("match functions", () => {
```
cgranade, line 56:
```suggestion
    it("matches functions", () => {
```

src/VSCodeExtension/src/formatter/rules/declaration.ts
```diff
@@ -1,33 +1,33 @@
...
+                firstArgMatcher,
+                (match: string, group: string, variable: string, type: string) =>
+                    `(${variable} : ${type}`
+            )
+            .replace(
+                restArgMatcher,
+                (match: string, group: string, variable: string, type: string) =>
+                    `, ${variable} : ${type}`
+            );
+        return `${keyword} ${opName} ${args} : ${retType}`;
```
cgranade, line 29: As mentioned on the other PR, I'd suggest no space between `opName` and `args` here.

src/VSCodeExtension/src/tests/unit/declaration.test.ts
```diff
@@ -1,24 +1,24 @@
 import "mocha";
 import * as assert from "assert";
 import { formatter } from "../../formatter/formatter";
-import { argsRule } from "../../formatter/rules/operation";
+import { argsRule } from "../../formatter/rules/declaration";
 
 describe("arguments rule", () => {
-    it("no error", () => {
+    it("has no error", () => {
         const code = "operation Foo (q : Qubit, n : Int) : Bool";
```
amrondonp, line 8: This could be done in a separated PR, but did you consider comments?

what if the input if

// operation Foo (q : Qubit, n : Int) : Bool

?

theRoughCode: Comments are an unhandled case for all rules. Will be addressing this in a future PR.

src/VSCodeExtension/src/formatter/rules/declaration.ts
theRoughCode: Since the decision hasn't been finalized, we'll address it in a separate PR for now :)

swernli:
Given that the formatter doesn't yet support multiple lines, I think some of the declaration matching rules will get confused by declarations that span multiple lines. If the mocha test infra supports it, it would be good to add a skipped test that is known to fail on multi-line callable declaration so that known case is captured.

theRoughCode:
My next PR will focus on multiline support!

PR #538, opened 2020-07-28
Manage more edge cases for space after 'if' rule

amrondonp:
*Handle case when 'if' is present in names of identifiers
*Handle case when there is a line break just after the if
*Comments management to be implemented as a separate PR because it is a cross rule concern


amrondonp:
@theRoughCode @swernli @cgranade

src/VSCodeExtension/src/tests/unit/control-structure.test.ts
```diff
@@ -95,4 +95,36 @@ describe("space after if rule", () => {
 
     assert.equal(formatter(code, [spaceAfterIf]), expectedCode);
   });
+
+  it("does not formats space after if statement if there is an identifier that starts with if", () => {
```
theRoughCode, line 99: nit:
```suggestion
  it("does not format space after if statement if there is an identifier that starts with if", () => {
```

src/VSCodeExtension/src/tests/unit/control-structure.test.ts
```diff
@@ -95,4 +95,36 @@ describe("space after if rule", () => {
...
+    const code = "ifunctionWithParams(parameter);"
+    const expectedCode = "ifunctionWithParams(parameter);"
+
+    assert.equal(formatter(code, [spaceAfterIf]), expectedCode);
+  });
+
+  it("does not formats space after if statement if there is an identifier that ends with if", () => {
```
theRoughCode, line 106: nit: same as above

src/VSCodeExtension/src/tests/unit/control-structure.test.ts
```diff
@@ -95,4 +95,36 @@ describe("space after if rule", () => {
...
+(1==1){Message("1==1");}`;
+
+    const expectedCode =
+      'mutable bits = new Result[0];if (1==1){Message("1==1");}';
+
+    assert.equal(formatter(code, [spaceAfterIf]), expectedCode);
+  });
+
+  // Activate this test once a strategy for managing comments is implemented
+  xit("leaves comments unchanged", () => {
```
theRoughCode, line 124: thanks for adding this!

src/VSCodeExtension/src/tests/unit/control-structure.test.ts
swernli: Minor: Should say "format" instead of "formats"

src/VSCodeExtension/src/tests/unit/control-structure.test.ts
```diff
@@ -95,4 +95,36 @@ describe("space after if rule", () => {
...
+
+  it("does not formats space after if statement if there is an identifier that starts with if", () => {
+    const code = "ifunctionWithParams(parameter);"
+    const expectedCode = "ifunctionWithParams(parameter);"
+
+    assert.equal(formatter(code, [spaceAfterIf]), expectedCode);
+  });
+
+  it("does not formats space after if statement if there is an identifier that ends with if", () => {
+    const code = "functionWithParamsif= 2;"
```
swernli, line 107: Can this case (or an additional one) also check if the behavior is correct when it's a invocation? As in, `functionWithParamsif(2);`

src/VSCodeExtension/src/tests/unit/control-structure.test.ts
```diff
@@ -95,4 +95,36 @@ describe("space after if rule", () => {
...
+(1==1){Message("1==1");}`;
+
+    const expectedCode =
+      'mutable bits = new Result[0];if (1==1){Message("1==1");}';
+
+    assert.equal(formatter(code, [spaceAfterIf]), expectedCode);
+  });
+
+  // Activate this test once a strategy for managing comments is implemented
+  xit("leaves comments unchanged", () => {
```
swernli, line 124: Nice! Good to know there is an easy way to do disabled test cases.

src/VSCodeExtension/src/tests/unit/control-structure.test.ts
amrondonp: Thar case is already implemented :) line 99

amrondonp: And line 55 https://github.com/microsoft/qsharp-compiler/pull/538/files/0c175803b781b5895195f7889433164fbef6a672#diff-a737a0ce651d7ed7689329bc1b5a97d5R55

swernli: Ah, I see that one on 55 now! Because it wasn't changed it was skipped in the diff. Thanks for pointing it out!

PR #539, opened 2020-07-28
Format whole document instead of line-by-line

theRoughCode:
This PR implements formatting the whole document instead of line-by-line.
Added a test case for namespaces to show that this multi-line implementation is working.

### DIFF
--- a/src/VSCodeExtension/src/formatter/format-document.ts
+++ b/src/VSCodeExtension/src/formatter/format-document.ts
@@ -16,18 +16,11 @@ const rules: FormatRule[] = [
 export const formatDocument = (
     document: vscode.TextDocument
 ): vscode.TextEdit[] => {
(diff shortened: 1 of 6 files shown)


theRoughCode:
@amrondonp

swernli:
I'm a little bit confused... I see how the regexp for namespaces was updated to correctly handle multiple lines, but how does the if-statement rule work now? Will it correctly handle multiple if-statements in the file or does that one also need to be updated? It's fine if that is planned for a later update.

theRoughCode:
The tests for the `if` statements still work, but I haven't looked too closely at the cases it covers. cc-ing @amrondonp on this

theRoughCode:
@swernli  I took a look at the code and I think you're right, we have to add `/g` flag at the end to catch multiple.

theRoughCode:
@swernli added a fix for it and a test to catch it

PR #543, opened 2020-07-28
Type Parameter Resolution Combinations

ScottCarda-MS:
Adding a utility class, ~~TypeResolutionSummary~~ TypeResolutionCombination, that can be used to ~~summarize~~ combine the effect of multiple type parameter resolution dictionaries into one resolution dictionary. This ~~summarizing~~ combination is calculated during the construction of these ~~summary~~ objects and the resulting ~~summaries~~ combinations are immutable. Validation for the given resolutions is also done during construction and can be check by referencing the ~~IsValidSummary~~ IsValid flag.

There are two constructors for the class. The first one takes an arbitrary list of resolution dictionaries that is expected be given in a specific order. It will construct a different ~~summary~~ combination if the input is in an unexpected order and validation is likely to fail. This constructor is internal only as it is likely that only the call graph will make use of it. The second constructor takes a TypedExpression object and collects all the relevant type parameter resolutions by walking the expression. It can then just use the logic from the first constructor to build the ~~summary~~ combination. This constructor is public as it will be used in the runtime repo.

In total there are 4 public members to the combination object:
- IndependentResolutionDictionaries: a readonly immutable array of the input resolutions to the combination.
- CombinedResolutionDictionary: the merged resolution for all the type parameters found in the IndependentResolutionDictionaries.
- IsValid: a bool that can tell the user if there were any invalid scenarios encountered during the merging of the resolutions.
- The public constructor, which takes a TypedExpression and collects all the appropriate resolutions from the expression and sub-expressions as input into the merging.

### DIFF
--- a/src/QsCompiler/DataStructures/SyntaxTree.fs
+++ b/src/QsCompiler/DataStructures/SyntaxTree.fs
@@ -67,8 +67,8 @@ type QsQualifiedName = {
     /// the declared name of the namespace element
     Name : NonNullable<string>
 }
-    with 
-    override this.ToString () = 
+    with
+    override this.ToString () =
         sprintf "%s.%s" this.Namespace.Value this.Name.Value
 
 
@@ -344,8 +344,8 @@ type InferredExpressionInformation = {
 type TypedExpression = {
     /// the content (kind) of the expression
     Expression : QsExpressionKind<TypedExpression, Identifier, ResolvedType>
-    /// contains all type arguments implicitly or explicitly determined by the expression, 
-    /// i.e. the origin, name and concrete type of all type parameters whose type can either be inferred based on the expression, 
+    /// contains all type arguments implicitly or explicitly determined by the expression,
+    /// i.e. the origin, name and concrete type of all type parameters whose type can either be inferred based on the expression,
     /// or who have explicitly been resolved by provided type arguments
     TypeArguments : ImmutableArray<QsQualifiedName * NonNullable<string> * ResolvedType>
     /// the type of the expression after applying the type arguments
@@ -361,13 +361,18 @@ type TypedExpression = {
 
     /// Contains a dictionary mapping the origin and name of all type parameters whose type can either be inferred based on the expression,
     /// or who have explicitly been resolved by provided type arguments to their concrete type within this expression
-    member this.TypeParameterResolutions = 
+    member this.TypeParameterResolutions =
         this.TypeArguments.ToImmutableDictionary((fun (origin, name, _) -> origin, name), (fun (_,_,t) -> t))
+
+    /// Given a dictionary containing the type resolutions for an expression,
+    /// returns the corresponding ImmutableArray to initialize the TypeArguments with.
+    static member AsTypeArguments (typeParamResolutions : ImmutableDictionary<_,_>) =
+        typeParamResolutions |> Seq.map (fun kv -> fst kv.Key, snd kv.Key, kv.Value) |> ImmutableArray.CreateRange
 
     /// Returns true if the expression is a call-like expression, and the arguments contain a missing expression.
     /// Returns false otherwise.
     static member public IsPartialApplication kind =
-        let rec containsMissing ex = 
+        let rec containsMissing ex =
             match ex.Expression with
             | MissingExpr -> true
             | ValueTuple items -> items |> Seq.exists containsMissing
@@ -564,7 +569,7 @@ and QsStatementKind =
 | QsRepeatStatement      of QsRepeatStatement
 | QsConjugation          of QsConjugation
 | QsQubitScope           of QsQubitScope // includes both using and borrowing scopes
-| EmptyStatement 
+| EmptyStatement
 
 
 and QsStatement = {
@@ -645,7 +650,7 @@ type QsSpecialization = {
     /// Contains the location information for the declared specialization.
     /// The position offset represents the position in the source file where the specialization is declared,
     /// and the range contains the range of the corresponding specialization header.
-    /// For auto-generated specializations, the location is set to the location of the parent callable declaration. 
+    /// For auto-generated specializations, the location is set to the location of the parent callable declaration.
     Location : QsNullable<QsLocation>
     /// contains the type arguments for which the implementation is specialized
     TypeArguments : QsNullable<ImmutableArray<ResolvedType>>
--- a/src/QsCompiler/Tests.Compiler/TestCases/TypeParameter.qs
+++ b/src/QsCompiler/Tests.Compiler/TestCases/TypeParameter.qs
@@ -1,7 +1,44 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-// Partial Resolution
+
+// Identifier Resolution
+namespace Microsoft.Quantum.Testing.TypeParameter {
+
+    operation Main() : Unit {
+        Foo<_, Int, String>(1.0, 2, "Three");
+    }
+
+    operation Foo<'A, 'B, 'C>(a : 'A, b : 'B, c : 'C) : Unit { }
+}
+
+// =================================
+
+// Adjoint Application Resolution
+namespace Microsoft.Quantum.Testing.TypeParameter {
+
+    operation Main() : Unit {
+        (Adjoint (Foo(1.0, 2, _)))("Three");
+    }
+
+    operation Foo<'A, 'B, 'C>(a : 'A, b : 'B, c : 'C) : Unit is Adj { }
+}
+
+// =================================
+
+// Controlled Application Resolution
+namespace Microsoft.Quantum.Testing.TypeParameter {
+
+    operation Main(qs : Qubit) : Unit {
+        (Controlled (Foo(1.0, 2, _)))([qs], "Three");
+    }
+
+    operation Foo<'A, 'B, 'C>(a : 'A, b : 'B, c : 'C) : Unit is Ctl { }
+}
+
+// =================================
+
+// Partial Application Resolution
 namespace Microsoft.Quantum.Testing.TypeParameter {
 
     operation Main() : Unit {
@@ -13,3 +50,28 @@ namespace Microsoft.Quantum.Testing.TypeParameter {
 
 // =================================
 
+// Sub-call Resolution
+namespace Microsoft.Quantum.Testing.TypeParameter {
+
+    operation Main() : Unit {
+        (Foo(1, 2, 3))(4);
+    }
+
+    operation Foo<'A, 'B, 'C>(a : 'A, b : 'B, c : 'C) : ('C => Unit) { return Bar<'A, 'B, 'C>(a, b, _); }
+
+    operation Bar<'A, 'B, 'C>(a : 'A, b : 'B, c : 'C) : Unit { }
+}
+
+// =================================
+
+// Argument Sub-call Resolution
+namespace Microsoft.Quantum.Testing.TypeParameter {
+
+    operation Main() : Unit {
+        Foo(1.0, Bar(2), "Three");
+    }
+
+    operation Foo<'A, 'B, 'C>(a : 'A, b : 'B, c : 'C) : Unit { }
+
+    operation Bar<'A>(a : 'A) : 'A { return a; }
+}
--- a/src/QsCompiler/Tests.Compiler/TestCases/TypeParameter.qs
+++ b/src/QsCompiler/Tests.Compiler/TestCases/TypeParameter.qs
@@ -1,0 +1,15 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// Partial Resolution
+namespace Microsoft.Quantum.Testing.TypeParameter {
+
+    operation Main() : Unit {
+        (Foo(_, 3, "Hi"))(1.0);
+    }
+
+    operation Foo<'A, 'B, 'C>(a : 'A, b : 'B, c : 'C) : Unit { }
+}
+
+// =================================
+
--- a/src/QsCompiler/Transformations/ClassicallyControlled.cs
+++ b/src/QsCompiler/Transformations/ClassicallyControlled.cs
@@ -250,29 +250,6 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.ClassicallyControlled
                 }
 
                 /// <summary>
-                /// Get the combined type resolutions for a pair of nested resolutions,
-                /// resolving references in the inner resolutions to the outer resolutions.
-                /// </summary>
-                private TypeArgsResolution GetCombinedTypeResolution(TypeArgsResolution outer, TypeArgsResolution inner)
-                {
-                    var outerDict = outer.ToDictionary(x => (x.Item1, x.Item2), x => x.Item3);
-                    return inner.Select(innerRes =>
-                    {
-                        if (innerRes.Item3.Resolution is ResolvedTypeKind.TypeParameter typeParam &&
-                            outerDict.TryGetValue((typeParam.Item.Origin, typeParam.Item.TypeName), out var outerRes))
-                        {
-                            outerDict.Remove((typeParam.Item.Origin, typeParam.Item.TypeName));
-                            return Tuple.Create(innerRes.Item1, innerRes.Item2, outerRes);
-                        }
-                        else
-                        {
-                            return innerRes;
-                        }
-                    })
-                    .Concat(outerDict.Select(x => Tuple.Create(x.Key.Item1, x.Key.Item2, x.Value))).ToImmutableArray();
-                }
-
-                /// <summary>
                 /// Checks if the scope is valid for conversion to an operation call from the conditional control API.
                 /// It is valid if there is exactly one statement in it and that statement is a call like expression statement.
                 /// If valid, returns true with the identifier of the call like expression and the arguments of the
@@ -289,25 +266,26 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.ClassicallyControlled
                         && !TypedExpression.IsPartialApplication(expr.Item.Expression)
                         && call.Item1.Expression is ExpressionKind.Identifier)
                     {
-                        // We are dissolving the application of arguments here, so the call's type argument
-                        // resolutions have to be moved to the 'identifier' sub expression.
-
-                        var callTypeArguments = expr.Item.TypeArguments;
-                        var idTypeArguments = call.Item1.TypeArguments;
-                        var combinedTypeArguments = this.GetCombinedTypeResolution(callTypeArguments, idTypeArguments);
-
-                        // This relies on global callables being the only things that have type parameters.
                         var newCallIdentifier = call.Item1;
-                        if (combinedTypeArguments.Any()
-                            && newCallIdentifier.Expression is ExpressionKind.Identifier id
-                            && id.Item1 is Identifier.GlobalCallable global)
+                        var callTypeArguments = expr.Item.TypeParameterResolutions;
+
+                        // This relies on anything having type parameters must be a global callable.
+                        if (newCallIdentifier.Expression is ExpressionKind.Identifier id
+                            && id.Item1 is Identifier.GlobalCallable global
+                            && callTypeArguments.Any())
                         {
+                            // We are dissolving the application of arguments here, so the call's type argument
+                            // resolutions have to be moved to the 'identifier' sub expression.
+                            var combination = new TypeResolutionCombination(expr.Item);
+                            var combinedTypeArguments = combination.CombinedResolutionDictionary.Where(kvp => kvp.Key.Item1.Equals(global.Item)).ToImmutableDictionary();
+                            QsCompilerError.Verify(combination.IsValid, "failed to combine type parameter resolution");
+
                             var globalCallable = this.SharedState.Compilation.Namespaces
                                 .Where(ns => ns.Name.Equals(global.Item.Namespace))
                                 .Callables()
                                 .FirstOrDefault(c => c.FullName.Name.Equals(global.Item.Name));
 
-                            QsCompilerError.Verify(globalCallable != null, $"Could not find the global reference {global.Item.Namespace.Value + "." + global.Item.Name.Value}");
+                            QsCompilerError.Verify(globalCallable != null, $"Could not find the global reference {global.Item}.");
 
                             var callableTypeParameters = globalCallable.Signature.TypeParameters
                                 .Select(x => x as QsLocalSymbol.ValidName);
@@ -319,8 +297,8 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.ClassicallyControlled
                                     id.Item1,
                                     QsNullable<ImmutableArray<ResolvedType>>.NewValue(
                                         callableTypeParameters
-                                        .Select(x => combinedTypeArguments.First(y => y.Item2.Equals(x.Item)).Item3).ToImmutableArray())),
-                                combinedTypeArguments,
+                                        .Select(x => combinedTypeArguments[Tuple.Create(global.Item, x.Item)]).ToImmutableArray())),
+                                TypedExpression.AsTypeArguments(combinedTypeArguments),
                                 call.Item1.ResolvedType,
                                 call.Item1.InferredInformation,
                                 call.Item1.Range);
--- a/src/QsCompiler/Transformations/TypeResolutionCombination.cs
+++ b/src/QsCompiler/Transformations/TypeResolutionCombination.cs
@@ -22,8 +22,7 @@ namespace Microsoft.Quantum.QsCompiler
     /// Combines a series of type parameter resolution dictionaries, IndependentResolutionDictionaries,
     /// into one resolution dictionary, CombinedResolutionDictionary, containing the ultimate type
     /// resolutions for all the type parameters found in the dictionaries. Validation is done on the
-    /// resolutions, which can be checked through the IsValid, CombinesOverConflictingResolution, and
-    /// CombinesOverParameterConstriction flags.
+    /// resolutions, which can be checked through the IsValid flag.
     /// </summary>
     public class TypeResolutionCombination
     {
@@ -75,19 +74,19 @@ namespace Microsoft.Quantum.QsCompiler
         /// and a type parameter being assigned to a type referencing a different type parameter of
         /// the same callable. Has value true if no invalid scenarios were encountered.
         /// </summary>
-        public bool IsValid { get => !this.CombinesOverConflictingResolution && !this.CombinesOverParameterConstriction; }
+        public bool IsValid { get => !this.combinesOverConflictingResolution && !this.combinesOverParameterConstriction; }
 
         /// <summary>
         /// Flag for if, at any time in the creation of the combination, there was a type parameter that
         /// was assigned conflicting type resolutions. Has value true if a conflict was encountered.
         /// </summary>
-        public bool CombinesOverConflictingResolution { get; private set; } = false;
+        private bool combinesOverConflictingResolution = false;
 
         /// <summary>
         /// Flag for if, at any time in the creation of the combination, there was a type parameter that
         /// was assigned a type resolution referencing a different type parameter of the same callable.
         /// </summary>
-        public bool CombinesOverParameterConstriction { get; private set; } = false;
+        private bool combinesOverParameterConstriction = false;
 
         // Constructors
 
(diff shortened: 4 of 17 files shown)


src/QsCompiler/Transformations/TypeResolutionSummary.cs
```diff
@@ -0,0 +1,390 @@
...
+                   kvp => kvp.value,
+                   kvp => kvp.Key);
+        }
+
+        // Fields and Properties
+
+        /// <summary>
+        /// Array of all the type parameter resolution dictionaries that are summarized in this summary.
+        /// </summary>
+        public readonly ImmutableArray<TypeParameterResolutions> ResolutionsInSummary;
```
cesarzc, line 63: I liked the previous name that we arrived to in the previous PR. Something similar to `IndependentResolutionsDictionary` would be great to convey what this field is about.

src/QsCompiler/Transformations/TypeResolutionSummary.cs
```diff
@@ -0,0 +1,390 @@
...
+        /// type parameter resolutions are relevant to the given expression's type parameter resolutions
+        /// are considered.
+        /// </summary>
+        public TypeResolutionSummary(TypedExpression expression) : this(GetTypeParameterResolutions.Apply(expression))
+        {
+        }
+
+        /// <summary>
+        /// Creates a type parameter resolution summary from independent type parameter resolution dictionaries.
+        /// The given resolutions are expected to be ordered such that dictionaries containing type parameters resolutions that
```
cesarzc, line 106: Can this be enforced? Maybe iterating over the list of dictionaries and if a dependency is out of order raise an exception?

src/QsCompiler/Transformations/TypeResolutionSummary.cs
```diff
@@ -0,0 +1,390 @@
...
+        /// <summary>
+        /// Updates the SummarizesOverParameterConstriction flag. If the flag is already set to true,
+        /// nothing will be done. If not, the given type parameter will be checked against the given
+        /// resolution for type parameter constriction, which is when one type parameter is dependent
+        /// on another type parameter of the same callable.
+        /// </summary>
+        private void UpdateConstrictionFlag(TypeParameterName typeParamName, ResolvedType typeParamResolution)
+        {
+            this.SummarizesOverParameterConstriction = this.SummarizesOverParameterConstriction
+                || ConstrictionCheck.Apply(typeParamName, typeParamResolution);
```
cesarzc, line 136: Is it ok to short circuit the constriction check?

src/QsCompiler/Transformations/TypeResolutionSummary.cs
ScottCarda-MS: This is enforced through the public constructor, which is the preferred way to construct summaries since it can automatically get all the relevant resolutions in the correct order. This other constructor is used mostly for the Call Graph logic. The way to tell if the order is correct is by checking the validation flags, rather than by throwing an exception.

src/QsCompiler/Transformations/TypeResolutionSummary.cs
ScottCarda-MS: The ConstrictionCheck is only there to set the `SummarizesOverParameterConstriction` flag. If we already know that there is a constriction violation, there is no need to continue to look for further constriction violations.

src/QsCompiler/Transformations/TypeResolutionSummary.cs
```diff
@@ -0,0 +1,390 @@
...
+            this.CombinedTypeParameterResolutions = combinedBuilder.ToImmutable();
+        }
+
+        // Nested Classes
+
+        /// <summary>
+        /// Walker that collects all of the type parameter references for a given ResolvedType
+        /// and returns them as a HashSet.
+        /// </summary>
+        private class GetTypeParameters : TypeTransformation<GetTypeParameters.TransformationState>
```
cesarzc, line 246: I missed this in the last PR but since this is a class, maybe a name like `TypeParametersGetter` might be a name that conveys that its is a class?

src/QsCompiler/Transformations/TypeResolutionSummary.cs
```diff
@@ -0,0 +1,390 @@
...
+                this.SharedState.TypeParams.Add(AsTypeResolutionKey(tp));
+                return base.OnTypeParameter(tp);
+            }
+        }
+
+        /// <summary>
+        /// Walker that checks a given type parameter resolution to see if it constricts
+        /// the type parameter to another type parameter of the same callable.
+        /// </summary>
+        private class ConstrictionCheck : TypeTransformation<ConstrictionCheck.TransformationState>
```
cesarzc, line 280: Similar feedback regarding the name of this class. Maybe `ConstrictionChecker` would better signal that it is a class.

src/QsCompiler/Transformations/TypeResolutionSummary.cs
```diff
@@ -0,0 +1,390 @@
...
+        /// </summary>
+        public TypeParameterResolutions CombinedTypeParameterResolutions { get; private set; }
+
+        /// <summary>
+        /// Flag for if there were any invalid scenarios encountered while creating the summary.
+        /// Invalid scenarios include a type parameter being assigned to multiple conflicting types
+        /// and a type parameter being assigned to a type referencing a different type parameter of
+        /// the same callable. Has value true if no invalid scenarios were encountered.
+        /// </summary>
+        public bool IsValidSummary { get => !this.SummarizesOverConflictingResolution && !this.SummarizesOverParameterConstriction; }
```
cesarzc, line 78: I think if this was named `IsValid` it would be slightly better.

src/QsCompiler/Transformations/TypeResolutionSummary.cs
ScottCarda-MS: The standard is for transformation classes to have verbs for names. For examples you can look at the [Basic Transformations]( https://github.com/microsoft/qsharp-compiler/blob/master/src/QsCompiler/Transformations/BasicTransformations.cs): `GetSourceFiles`, `FilterBySourceFile`, `SelectByFoldingOverExpressions`, `SelectByAnyContainedExpression`, `SelectByAllContainedExpressions`, and `FoldOverExpressions`.

src/QsCompiler/Transformations/TypeResolutionSummary.cs
ScottCarda-MS: Since transformations are supposed to be verbs, I should actually probably rename this to something like `CheckForConstriction`.

src/QsCompiler/Transformations/TypeResolutionSummary.cs
```diff
@@ -0,0 +1,390 @@
...
+    using TypeParameterResolutions = ImmutableDictionary</*TypeParameterName*/ Tuple<QsQualifiedName, NonNullable<string>>, ResolvedType>;
+
+    /// <summary>
+    /// Summarizes a series of type parameter resolution dictionaries, ResolutionsInSummary, into one
+    /// resolution dictionary, CombinedTypeParameterResolutions, containing the ultimate type resolutions
+    /// for all the type parameters found in the dictionaries. Validation is done on the resolutions,
+    /// which can be checked through the IsValidSummary, SummarizesOverConflictingResolution, and
+    /// SummarizesOverParameterConstriction flags.
+    /// </summary>
+    public class TypeResolutionSummary
```
samarsha, line 28: How will this class be used? I'm not sure I understand the design and purpose of the properties it exposes without knowing what they are needed for.

cesarzc, line 28: I am not totally convinced about the use of the word summary since I think it is more commonly associated with brief. Maybe something like `TypeResolutionAggregate` or `TypeResolutionBundle` or `TypeResolutionCollection`. Maybe it could be more explicit like `CombinedTypeResolution` (which is my preferred one so far). Naming is hard but I think there might be better options than summary in this case.

src/QsCompiler/Transformations/TypeResolutionSummary.cs
cesarzc: Thanks for the explanation!

One more question regarding this constructor. Is there a reason to choose `internal` instead of `private` in this case?

src/QsCompiler/Transformations/TypeResolutionSummary.cs
ScottCarda-MS: That's disappointing to hear. I thought the summary narrative made this whole concept come together nicely and made things easier to understand. It certainly made all of this more cohesive for me. I am concerned that this class will be more confusing without understanding the `CombinedTypeParameterResolutions` field is just a summary over the `ResolutionsInSummary`. Are you sure I should get rid of this way of conceptualizing the class?

samarsha: I'm also not really familiar with this use of the word summary, and "summarizing over" something sounds strange to me - I had to read the doc comments to understand it. I would describe something as "summarizing" something or being a "summary of" something but I'm not sure that applies here.

I agree with @cesarzc that it seems like you could describe this as merging or combining separate type parameter mappings into one mapping.

src/QsCompiler/Transformations/TypeResolutionSummary.cs
ScottCarda-MS: Just that the call graph logic uses it.

src/QsCompiler/Transformations/TypeResolutionSummary.cs
ScottCarda-MS: I renamed it to `IndependentResolutionDictionaries` and also renamed `CombinedTypeParameterResolutions` to `CombinedResolutionDictionary` to match it.

src/QsCompiler/Transformations/TypeResolutionSummary.cs
ScottCarda-MS: Done.

src/QsCompiler/Transformations/TypeResolutionSummary.cs
ScottCarda-MS: I guess I won't use the conceptualization of it being a summary since that is not the direction we are going in anymore.
It takes a series of type parameter resolutions and combines them into one resolution mapping. While doing this, it sets some validation flags so the user can tell if something went wrong during the combining, though they may still want the combined resolution depending on the situation. Since getting the right resolutions for a TypedExpression can be tedious, the public constructor gets all the resolutions for you from the provided TypedExpression. This can be useful, for example, if you want to know what are all the type parameter resolutions for a given call-like expression.

ScottCarda-MS: The term 'summary' has been replaced with the term 'combination'.

samarsha: Is it something we expect users to interact with using an API or is it only for the call graph? What will users or the call graph do with information like `IsValid`, `CombinesOverConflictingResolution`, and `CombinesOverParameterConstriction`?

samarsha: I guess specifically I'm wondering:

* What is the difference between `IsValid` and the diagnostics generated by the compiler for invalid type parameters?
* It seems like `CombinesOverConflictingResolution` and `CombinesOverParameterConstriction` both imply `not IsValid`. When is it needed to know specifically that there was a conflicting resolution or a constricted type, versus looking at `IsValid` or the compiler diagnostics?

ScottCarda-MS: This is something that coders using the QDK can use to on TypedExpressions to get the complete picture of type parameter resolutions in the expression. When writing transformations for the QDK, users may need to create custom TypedExpressions from scratch or from other components acquired from the syntax tree. The validation flags can be used to check if the created TypedExpression has correct type parameter resolutions.

ScottCarda-MS: I missed your last message. To your first question, I think I've answered that to the best of my ability. Since it is a public class, I can't account for you here how any user of the QDK will use it. For your second question: Whatever diagnostics you are referring to must not happen on the level that this class deals with as they are not being raised when testing invalid scenarios. It is possible that the diagnostics are specific to individual resolution dictionaries, while this class checks for combinations between dictionaries. For your third question: the `IsValid` is derived from the combination of the other two flags. While `IsValid` will probably be used more frequently, when someone does have an invalid series of type parameters, or TypedExpression, they will likely want to know more about what is making it invalid. They can being this by checking the other two flags for what quality specifically was violated. And they will be able to look at the input resolution dictionaries with the `IndependentResolutionDictionaries` since they will have just started with a TypedExpression.

samarsha: To design a public API, to me it seems like it is important to have concrete use cases in mind. For example, are there any specific examples of transformations that a user would want to write that would be difficult to do without this tool? Have these examples come up in practice - is there a feature request for this? Is there a use case for this within the compiler or runtime, and if so can it be designed as an internal tool instead of public?

Sorry for all of the questions, but the reason I'm asking is because it seems like there are multiple ways to design this, and I think that the design should be informed by those use cases in order to best accomplish the task. Since I'm not sure what those use cases are, I'm not able to review the design in the PR.

ScottCarda-MS: Let's take this offline.

samarsha: To summarize our conversation: I think there is a lot of context here that I'm not familiar with, and I'm not sure how this PR fits in with the big picture in the compiler. So I'll decline to review the PR since I am not confident saying one way or the other what the design should be.

src/QsCompiler/Transformations/TypeResolutionCombination.cs
```diff
@@ -0,0 +1,389 @@
...
+    {
+        // Static Members
+
+        /// <summary>
+        /// Checks if the given type parameter directly resolves to itself. If so, the
+        /// resolution may be overridden without conflict.
+        /// </summary>
+        private static bool IsSelfResolution(TypeParameterName typeParam, ResolvedType res)
+        {
+            return res.Resolution is ResolvedTypeKind.TypeParameter tp
```
bettinaheim, line 37:
```suggestion
            return res?.Resolution is ResolvedTypeKind.TypeParameter tp
```

src/QsCompiler/Transformations/TypeResolutionCombination.cs
```diff
@@ -0,0 +1,389 @@
...
+                && tp.Item.Origin.Equals(typeParam.Item1)
```
bettinaheim, line 38:
```suggestion
                && tp.Item.Origin.Equals(typeParam?.Item1)
```

src/QsCompiler/Transformations/TypeResolutionCombination.cs
```diff
@@ -0,0 +1,389 @@
...
+                && tp.Item.TypeName.Equals(typeParam.Item2);
```
bettinaheim, line 39:
```suggestion
                && tp.Item.TypeName.Equals(typeParam?.Item2);
```

src/QsCompiler/Transformations/TypeResolutionCombination.cs
```diff
@@ -0,0 +1,389 @@
...
+    /// resolutions for all the type parameters found in the dictionaries. Validation is done on the
+    /// resolutions, which can be checked through the IsValid flag.
+    /// </summary>
+    public class TypeResolutionCombination
+    {
+        // Static Members
+
+        /// <summary>
+        /// Checks if the given type parameter directly resolves to itself. If so, the
+        /// resolution may be overridden without conflict.
```
bettinaheim, line 33: The second sentence seems irrelevant for what this method does, and I suggest to remove it.

src/QsCompiler/Transformations/TypeResolutionCombination.cs
```diff
@@ -0,0 +1,389 @@
...
+                && tp.Item.Origin.Equals(typeParam.Item1)
+                && tp.Item.TypeName.Equals(typeParam.Item2);
+        }
+
+        /// <summary>
+        /// Reverses the dependencies of type parameters resolving to other type parameters in the given
+        /// dictionary to create a lookup whose keys are type parameters and whose values are all the type
+        /// parameters that can be updated by knowing the resolution of the lookup's associated key.
+        /// </summary>
+        private static ILookup<TypeParameterName, TypeParameterName> GetReplaceable(TypeParameterResolutions.Builder typeParamResolutions)
```
bettinaheim, line 47:
```suggestion
        private static ILookup<TypeParameterName, TypeParameterName> GetReplaceable(IDictionary<Tuple<QsQualifiedName, NonNullable<string>>, ResolvedType> typeParamResolutions)
```

src/QsCompiler/Transformations/TypeResolutionCombination.cs
```diff
@@ -0,0 +1,389 @@
...
+        {
+            return typeParamResolutions
```
bettinaheim, line 49:
```suggestion
            return typeParamResolutions?
```

src/QsCompiler/Transformations/TypeResolutionCombination.cs
```diff
@@ -0,0 +1,389 @@
...
+        /// </summary>
+        public TypeParameterResolutions CombinedResolutionDictionary { get; private set; }
+
+        /// <summary>
+        /// Flag for if there were any invalid scenarios encountered while creating the combination.
+        /// Invalid scenarios include a type parameter being assigned to multiple conflicting types
+        /// and a type parameter being assigned to a type referencing a different type parameter of
+        /// the same callable. Has value true if no invalid scenarios were encountered.
+        /// </summary>
+        public bool IsValid { get => !this.combinesOverConflictingResolution && !this.combinesOverParameterConstriction; }
```
bettinaheim, line 77:
```suggestion
        public bool IsValid => !this.combinesOverConflictingResolution && !this.combinesOverParameterConstriction;
```

src/QsCompiler/Transformations/TypeResolutionCombination.cs
```diff
@@ -0,0 +1,389 @@
...
+
+        /// <summary>
+        /// Creates a type parameter resolution combination from independent type parameter resolution dictionaries.
+        /// The given resolutions are expected to be ordered such that dictionaries containing type parameters resolutions that
+        /// reference type parameters in other dictionaries appear before those dictionaries containing the referenced type parameters.
+        /// I.e., dictionary A depends on dictionary B, so A should come before B. When using this method to resolve
+        /// the resolutions of a nested expression, this means that the innermost resolutions should come first, followed by
+        /// the next innermost, and so on until the outermost expression is given last.
+        /// </summary>
+        internal TypeResolutionCombination(params TypeParameterResolutions[] independentResolutionDictionaries)
```
bettinaheim, line 111: The doc comment should mention when something throws (it throws if any of the given dictionaries is null).

src/QsCompiler/Tests.Compiler/TestCases/TypeParameter.qs
```diff
@@ -0,0 +1,77 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+
+// Identifier Resolution
+namespace Microsoft.Quantum.Testing.TypeParameter {
+
+    operation Main() : Unit {
+        Foo<_, Int, String>(1.0, 2, "Three");
+    }
```
ScottCarda-MS, line 10:
```suggestion
    operation Main() : Unit? {
        Foo<_, Int, String>>(1.0, 2, "Three");
    }?
```

src/QsCompiler/Transformations/TypeResolutionCombination.cs
ScottCarda-MS: I'd rather handle the null mappings earlier on so I don't have to be concerned with null's throughout the algorithm. To that end I've removed null mappings and null resolution dictionaries earlier in the algorithm. These null checks here won't be necessary as a result.

src/QsCompiler/Transformations/TypeResolutionCombination.cs
ScottCarda-MS: I've combined this with the other null-check suggestion.

src/QsCompiler/Transformations/TypeResolutionCombination.cs
ScottCarda-MS: I've combined this with the other null-check suggestion.

src/QsCompiler/Transformations/TypeResolutionCombination.cs
ScottCarda-MS: Why do this?

src/QsCompiler/Transformations/TypeResolutionCombination.cs
ScottCarda-MS: Because this method is private, we have full control over how it is called. Since it is never called with arguments that can be null, this null-check seems unnecessary.

src/QsCompiler/Transformations/TypeResolutionCombination.cs
ScottCarda-MS: Done.

src/QsCompiler/Transformations/TypeResolutionCombination.cs
ScottCarda-MS: I've added logic to filter out any null dictionaries instead of throwing on them and updated the doc comment to include that.

src/QsCompiler/Transformations/TypeResolutionCombination.cs
```diff
@@ -0,0 +1,392 @@
...
+                   kvp => kvp.value,
+                   kvp => kvp.Key);
+        }
+
+        // Fields and Properties
+
+        /// <summary>
+        /// Array of all the type parameter resolution dictionaries that are combined in this combination.
+        /// </summary>
+        public readonly ImmutableArray<TypeParameterResolutions> IndependentResolutionDictionaries;
```
bettinaheim, line 61: The order of this should be documented - the order is relevant for interpretation.

ScottCarda-MS: Done

ScottCarda-MS:
> Now that this functionality is properly split out, you should replace the corresponding implementation in the ClassicallyControlled transformation and the Monomorphization (and later the CallGraph) with it. If you prefer to do this in a separate PR, then please retarget this to a feature branch - we shouldn't have redundant implementations.

The changes to ClassicallyControlled were already done in a branch under the CallGraph feature branch, so it is was easy and small enough to migrate those into this PR. I can't immediately see any changes to be done to the Monomorphization transformation with regards to this utility. The Monomorphization will need to be heavily changed once the call graph is ready anyway, so it might not be a big deal to leave it as it is for now.

src/QsCompiler/Transformations/TypeResolutionCombination.cs
ScottCarda-MS: As an additional precaution, the `#nullable enable` directive prevents null from being passed into private methods in this file.

ScottCarda-MS:
> > Now that this functionality is properly split out, you should replace the corresponding implementation in the ClassicallyControlled transformation and the Monomorphization (and later the CallGraph) with it. If you prefer to do this in a separate PR, then please retarget this to a feature branch - we shouldn't have redundant implementations.
> 
> The changes to ClassicallyControlled were already done in a branch under the CallGraph feature branch, so it is was easy and small enough to migrate those into this PR. I can't immediately see any changes to be done to the Monomorphization transformation with regards to this utility. The Monomorphization will need to be heavily changed once the call graph is ready anyway, so it might not be a big deal to leave it as it is for now.

The additions from the call graph branch for ClassicallyControlled includes a new static member of TypedExpression which I believe was originally added by @bettinaheim. Since this is a public-facing addition, I just wanted to point that out.

src/QsCompiler/Transformations/TypeResolutionCombination.cs
bettinaheim: Then you should document in the doc comment that this throws an an exception if they are null.

src/QsCompiler/Transformations/TypeResolutionCombination.cs
bettinaheim: The builder is an implementation detail of another method.

src/QsCompiler/Transformations/TypeResolutionCombination.cs
bettinaheim: sounds good.

bettinaheim:
> > > Now that this functionality is properly split out, you should replace the corresponding implementation in the ClassicallyControlled transformation and the Monomorphization (and later the CallGraph) with it. If you prefer to do this in a separate PR, then please retarget this to a feature branch - we shouldn't have redundant implementations.
> > 
> > 
> > The changes to ClassicallyControlled were already done in a branch under the CallGraph feature branch, so it is was easy and small enough to migrate those into this PR. I can't immediately see any changes to be done to the Monomorphization transformation with regards to this utility. The Monomorphization will need to be heavily changed once the call graph is ready anyway, so it might not be a big deal to leave it as it is for now.
> 
> The additions from the call graph branch for ClassicallyControlled includes a new static member of TypedExpression called `AsTypeArguments` which I believe was originally added by @bettinaheim. Since this is a public-facing addition, I just wanted to point that out.

Thanks for pointing it out! 
Ok, let's schedule an update to the monomorphization after the call graph is completed then. I am happy to have it go to master as long as something is using it and we don't add redundancy (before this functionality in the classically controlled logic was redundant with the same functionality needed for the monomorphizations).

src/QsCompiler/DataStructures/SyntaxTree.fs
```diff
@@ -361,13 +361,18 @@ type TypedExpression = {
 
     /// Contains a dictionary mapping the origin and name of all type parameters whose type can either be inferred based on the expression,
     /// or who have explicitly been resolved by provided type arguments to their concrete type within this expression
-    member this.TypeParameterResolutions = 
+    member this.TypeParameterResolutions =
         this.TypeArguments.ToImmutableDictionary((fun (origin, name, _) -> origin, name), (fun (_,_,t) -> t))
 
+    /// Given a dictionary containing the type resolutions for an expression,
+    /// returns the corresponding ImmutableArray to initialize the TypeArguments with.
+    static member AsTypeArguments (typeParamResolutions : ImmutableDictionary<_,_>) =
```
bettinaheim, line 369: You should replace the corresponding line in ConstructorExtensions.fs with a call to this.

ScottCarda-MS: Done

src/QsCompiler/Transformations/TypeResolutionCombination.cs
ScottCarda-MS: They can't be null because of the `#nullable enable` directive. Private methods cannot have null arguments unless explicitly specified.

src/QsCompiler/Transformations/TypeResolutionCombination.cs
ScottCarda-MS: This whole method is an implementation detail of another method 😁. It was just a convenient way of wrapping a somewhat complicated LINQ query. If it helps I could get rid of the method and just put the query back into the two methods that use it, but I thought this would be more readable this way.

src/QsCompiler/Transformations/ClassicallyControlled.cs
```diff
@@ -289,25 +266,26 @@ private TypeArgsResolution GetCombinedTypeResolution(TypeArgsResolution outer, T
...
+
+                        // This relies on anything having type parameters must be a global callable.
+                        if (newCallIdentifier.Expression is ExpressionKind.Identifier id
+                            && id.Item1 is Identifier.GlobalCallable global
+                            && callTypeArguments.Any())
                         {
+                            // We are dissolving the application of arguments here, so the call's type argument
+                            // resolutions have to be moved to the 'identifier' sub expression.
+                            var combination = new TypeResolutionCombination(expr.Item);
+                            var combinedTypeArguments = combination.CombinedResolutionDictionary.Where(kvp => kvp.Key.Item1.Equals(global.Item)).ToImmutableDictionary();
```
bettinaheim, line 280: How come the keys need to be filtered here? I.e. why isn't the CombinedResolutionsDictionary guaranteeing that all keys belong to the same parent? And if they are from the same parent, it would actually be nice to have a handle in the resolution that specifies that parent.

src/QsCompiler/Transformations/TypeResolutionCombination.cs
bettinaheim: Ok, either one works for me.

src/QsCompiler/Transformations/ClassicallyControlled.cs
ScottCarda-MS: I see the filtering done here as something that is situation-dependent and not something that fits into the narrow definition of combining type parameter resolutions. It seems better to me to return all the combinations, and if the caller wants to filter those by anything, they can do that themselves, rather than assuming the user only wants part of the data and filtering out the other parameters in the tool.

ScottCarda-MS: Another thought: If we find later that this specific kind of filtering is done quite often after getting the results from the tool, we can always add some functionality to it to do that automatically. We can always add onto the public-facing content. But if we add it now and it doesn't come up often, we can't remove it without a breaking change, so I would prefer to be conservative with regard to this until we can see lots of use cases that use the filter.

bettinaheim: Either one is ok for me; just as a question to think about: What would be a valid expression where the resolutions dictionary contains keys for more than one parent?

ScottCarda-MS: Thanks Bettina! It'll be good to keep in mind what kinds of varieties of inputs this tool can handle going forward. I will keep this in mind.

PR #545, opened 2020-07-28
Fixing pack script

swernli:
Fixing bug in pack script that kept failures in VS and VSCode extension packing from being correctly caught.


swernli:
@amrondonp FYI

ricardo-espinoza:
Let's notice that this PR will remain blocked by the checks until we clean the pre-exiting build issues in the VS Code extension that are exposed by taking this change.

PR #548, opened 2020-07-30
Swernli/qslint merge master

swernli:


swernli:
This change just merges master into the feature branch.

Issue #491, opened 2020-06-27
Position of InvalidReassignmentInApplyBlock error is incorrect


samarsha:
This Q# code is invalid because the `apply` block reassigns `foo`, which is used in the `within` block:

```
operation HelloQ() : Unit {
    mutable foo = false;
    within {
        let bar = foo;
    } apply {
        set foo = true;
    }
}
```

The error message should be on the `set foo = true;` line, but it's actually in the operation declaration:

![image](https://user-images.githubusercontent.com/33814365/85909609-adc98580-b7cf-11ea-8554-3de7b4c97ed5.png)

bettinaheim:
@samarsha Since you already know the one line that needs to be changed and how, do you mind taking this one?

PR #559, opened 2020-08-06
Add offsets to diagnostic ranges in Core and SyntaxProcessor instead of CompilationManager

samarsha:
This is the original problem that drove me to refactor the position and ranges types: I wanted to add the offset to the range directly in SyntaxProcessor, instead of returning a tuple to the CompilationManager. Now it's possible.

This simplifies the return types of several functions from `Position * QsCompilerDiagnostic[]` to `QsCompilerDiagnostic[]`, and eliminates the risk that CompilationManager will forget to add the offsets to the range (which happened before: #491).

Part of #531.

### DIFF
--- a/src/QsCompiler/CompilationManager/TypeChecking.cs
+++ b/src/QsCompiler/CompilationManager/TypeChecking.cs
@@ -425,8 +425,8 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
             var declDiagnostics = symbols.ResolveAll(BuiltIn.NamespacesToAutoOpen);
             var cycleDiagnostics = SyntaxProcessing.SyntaxTree.CheckDefinedTypesForCycles(symbols.DefinedTypes());
 
-            void AddDiagnostics(NonNullable<string> source, IEnumerable<Tuple<Position, QsCompilerDiagnostic>> msgs) =>
-                diagnostics.AddRange(msgs.Select(msg => Diagnostics.Generate(source.Value, msg.Item2, msg.Item1)));
+            void AddDiagnostics(NonNullable<string> source, IEnumerable<QsCompilerDiagnostic> msgs) =>
+                diagnostics.AddRange(msgs.Select(msg => Diagnostics.Generate(source.Value, msg)));
 
             if (fileName != null)
             {
--- a/src/QsCompiler/CompilationManager/TypeChecking.cs
+++ b/src/QsCompiler/CompilationManager/TypeChecking.cs
@@ -1053,12 +1053,8 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
                 var (ifStatement, ifDiagnostics) =
                     Statements.NewIfStatement(context, ifBlock.Item1, ifBlock.Item2, elifBlocks, elseBlock);
                 statement = ifStatement;
-                diagnostics.AddRange(ifDiagnostics.Select(item =>
-                {
-                    var (relativeOffset, diagnostic) = item;
-                    return Diagnostics.Generate(
-                        context.Symbols.SourceFile.Value, diagnostic, rootPosition + relativeOffset);
-                }));
+                diagnostics.AddRange(ifDiagnostics.Select(diagnostic =>
+                    Diagnostics.Generate(context.Symbols.SourceFile.Value, diagnostic, rootPosition)));
                 return true;
             }
             (statement, proceed) = (null, true);
--- a/src/QsCompiler/CompilationManager/TypeChecking.cs
+++ b/src/QsCompiler/CompilationManager/TypeChecking.cs
@@ -1604,8 +1604,7 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
             // verify that all paths return a value if needed (or fail)
             var (allPathsReturn, messages) = SyntaxProcessing.SyntaxTree.AllPathsReturnValueOrFail(implementation);
             var rootPosition = root.Fragment.Range.Start;
-            diagnostics.AddRange(messages.Select(msg =>
-                Diagnostics.Generate(sourceFile.Value, msg.Item2, rootPosition + msg.Item1)));
+            diagnostics.AddRange(messages.Select(msg => Diagnostics.Generate(sourceFile.Value, msg, rootPosition)));
             if (!(context.ReturnType.Resolution.IsUnitType || context.ReturnType.Resolution.IsInvalidType) && !allPathsReturn)
             {
                 var errRange = Parsing.HeaderDelimiters(root.Fragment.Kind.IsControlledAdjointDeclaration ? 2 : 1).Invoke(root.Fragment.Text);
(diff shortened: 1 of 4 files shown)


src/QsCompiler/CompilationManager/TypeChecking.cs
```diff
@@ -425,8 +425,8 @@ internal static void ResolveGlobalSymbols(NamespaceManager symbols, List<Diagnos
             var declDiagnostics = symbols.ResolveAll(BuiltIn.NamespacesToAutoOpen);
```
ScottCarda-MS, line 425: So before it was a position matched with the Range field of the QsCompilerDiagnostic, right? What did those to values before this change indicate? Now that it is just the Range field of the QsCompilerDiagnostic, has the conceptual definition of that field changed?

ScottCarda-MS: Discussed offline: the Range field is always the position of the diagnostic relative to a given reference frame. The reference frame may change through the code, but users should always have enough context to know what reference frame is being used.

src/QsCompiler/Core/SymbolTable.fs
```diff
@@ -1295,7 +1295,10 @@ and NamespaceManager
             let typeDiagnostics = this.CacheTypeResolution()
             let callableDiagnostics = this.CacheCallableResolutions()
             this.ContainsResolutions <- true
-            callableDiagnostics.Concat(typeDiagnostics).ToLookup(fst, snd)
+            callableDiagnostics
+                .Concat(typeDiagnostics)
+                .ToLookup(fst, fun (_, (position, diagnostic)) ->
+                    { diagnostic with QsCompilerDiagnostic.Range = position + diagnostic.Range })
```
DmitryVasilevsky, line 1301: Can you explain this in a few words? Does this mutate diagnostics? (I'm not familiar enough with F#).

src/QsCompiler/SyntaxProcessor/TreeVerification.fs
```diff
@@ -22,10 +22,10 @@ open Microsoft.Quantum.QsCompiler.SyntaxTree
 /// Such diagnostics include in particular warnings for all statement that will never be executed, and errors for misplaced returns statements
 /// Throws an ArgumentException if the statements contain no location information. 
 let AllPathsReturnValueOrFail body = 
-    let diagnostics = new List<Position * QsCompilerDiagnostic>()
+    let diagnostics = ResizeArray<_> ()
     let addDiagnostic diag (stm : QsStatement) = stm.Location |> function
         | Null -> ArgumentException "no location set for the given statement" |> raise
-        | Value loc -> (loc.Offset, loc.Range |> diag) |> diagnostics.Add
+        | Value loc -> loc.Offset + loc.Range |> diag |> diagnostics.Add
```
DmitryVasilevsky, line 28: The pattern offset+range seem to repeat quite a bit. What happens if offset isn't added to range, will it still compile? If it will, is there any way to prevent new code from forgetting to add offset?

src/QsCompiler/Core/SymbolTable.fs
ScottCarda-MS: The second argument of the ToLookup, the `fun` lambda, has the effect of mapping pairs of `position` and `diagnostic` to the same `diagnostic` but with the the `position` added to its Range field.

DmitryVasilevsky: Thanks. The question is - if the diagnostics object is modified or a copy is created. If it is modified, I assume the function can only be called once and process each diagnostics once.

ScottCarda-MS: Ah I understand. A new diagnostic is created. The `{ foo with bar }` syntax creates a new object that copies (deeply I think) the fields of the object `foo` and adds new fields or overrides fields for the new object from the expression `bar`. Here is an official reference on the subject: https://fsharpforfunandprofit.com/posts/records/#creating-records-from-other-records

samarsha: Yes, `QsCompilerDiagnostic` is an immutable record so it cannot be modified - it has to be copied. `with` is a shallow copy, but the fields in `QsCompilerDiagnostic` are also immutable, so it should be equivalent to a deep copy.

src/QsCompiler/SyntaxProcessor/TreeVerification.fs
samarsha: If you forget to add the offset (so you write `loc.Range` instead of `loc.Offset + loc.Range`), it will still compile because the type of both expressions is `Range`, but the range will be incorrect (i.e. the red squiggly for an error will show up in the wrong place in your text editor).

I would be very interested in a solution that could prevent this error at compile time. I can think of a couple ways, but I'm not sure if they are worth the overhead:

1. Add a phantom type variable to the `Position`, `Range` and `QsCompilerDiagnostic` types indicating what their positions are relative to (e.g. statement, specialization, or file). This will make the types more complicated and cumbersome.
2. Prevent direct access to QsLocation.Range, and make an accessor function that automatically adds the position. However, I don't think this would catch forgetting to add the position to an already-existing diagnostic range, such as:
   ```
   { diagnostic with QsCompilerDiagnostic.Range = position + diagnostic.Range }
   ```

samarsha: The other solution to preventing this class of errors would be to make the tests more comprehensive - currently I think the diagnostic tests check that an expected diagnostic is anywhere within a callable or type, but they could be more specific.

DmitryVasilevsky: F# is so proud of its type system. Can we make two incompatible Range types? Like let's say RelativeRange and AbsoluteRange? Like
type AbsoluteRange =
    | AbsoluteRange of Range
type RelativeRange =
    | RelativeRange of Range
And then accept only AbsoluteRange where errors are displayed? Then make + take position and relative range and return absolute range.
1. I have no knowledge of this code, names may be totally wrong.
2. I have no knowledge of F#, the above code may be incorrect.
3. It's up to you to decide if such increase in complexity is worth it.

samarsha: It's a good idea, one problem is that there are two kinds of relative ranges: (1) relative to start of statement and (2) relative to start of callable specialization. Both kinds of relative ranges, as well as an absolute range, can appear in diagnostics generated by the compiler, so I think we would need 3 variants of `QsCompilerDiagnostic`.

One way to deal with this is phantom type variables, so we would have

```fsharp
// 'a not used in the definition - only for creating distinct types.
type 'a Position = ...
type 'a Range = ...
type 'a QsCompilerDiagnostic = ...

QsStatement Position       // Relative to start of statement
QsSpecialization Position  // Relative to start of specialization
File Position              // Relative to start of file, i.e. absolute.

val (+) : QsSpecialization Position -> QsStatement Range -> QsSpecialization Range
val (+) : File Position -> QsSpecialization Range -> File Range
```

This might be worth it - I'll think a bit more about this...

samarsha: So I tried this, and it made the code very complex, because there were a lot of types/functions that needed to contain/return ranges of different types depending on where they are called from. That meant that they needed to add a type variable for the range type. The extra type variables spread throughout the code and it got out of hand pretty quickly.

I think I'm going to abandon this idea for now, but thanks for the suggestion @DmitryVasilevsky!

Issue #544, opened 2020-07-28
Extensible diagnostics in design time builds


samarsha:
**Is your feature request related to a problem? Please describe.**
The compiler has support for running arbitrary code on the Q# syntax tree during compilation in the form of the `IRewriteStep` interface. Rewrite steps can also generate arbitrary diagnostics that are displayed at build time. Unfortunately, rewrite steps don't run at design time when using the Q# language server, which means things like linters are limited to the command line, and can't provide live feedback in an editor.

**Describe the solution you'd like**
The compiler should provide some mechanism to run arbitrary diagnostic-generating code analysis in design-time builds, similar to [Roslyn analyzers](https://docs.microsoft.com/en-us/visualstudio/code-quality/roslyn-analyzers-overview?view=vs-2019).

**Describe alternatives you've considered**
No known alternatives that don't involve modifying the language server.

**Additional context**
This would have been useful for the runtime capabilities diagnostics (#449, #488). Those had to be built into the compiler, but they are not needed for compilation. Extensible diagnostics would make this kind of static analysis easier to add in a modular way.

swernli:
One important concern that was brought up about invoking too much of the `IRewriteStep` infrastructure is the overall impact on performance. We definitely want to keep the language server performant, as having the design time feedback be fast is key to providing a good IDE experience.

samarsha:
Yes, running the all of the full rewrite steps should probably be avoided. We could add a method that `IRewriteStep`s could implement that would only return diagnostics, which depending on the rewrite step could be implemented in a more efficient way. Another way is to make a separate plugin interface for analyzers only.

samarsha:
I can think of two hooks in the compiler that could be used to generate diagnostics:

1. When a specialization is recompiled
2. When global type checking runs

I feel like having a way to listen for just these two events could take care of a lot of diagnostics in a more modular way.

PR #561, opened 2020-08-11
Validate Cycles

ScottCarda-MS:
The method in this PR takes a cycle from the call graph, in the form of a sequence of Call Graph Edges, and determines if the cycle is valid or not. Invalid cycles are ones that map type parameters in the cycle to other type parameters of the same cycle (constricting the type parameter), i.e. Foo.A -> Foo.B, or to types that contain the same type parameter, i.e Foo.A -> Foo.A[].
Several unit tests are included.


src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -194,20 +194,91 @@ public override int GetHashCode()
...
-            return combination.IsValid && resolvedToConcrete;
-            //var isClosedCycle = validCycle && combined.Values.Any(res => res.Resolution is ResolvedTypeKind.TypeParameter tp && EqualsParent(tp.Item.Origin));
-            // TODO: check that monomorphization correctly processes closed cycles - meaning add a test...
+            if (!combination.IsValid)
+            {
+                return false;
+            }
+            var allResolutionsValid = combination.CombinedResolutionDictionary
+                .All(kvp => CheckTypeParameterResolutions.IsValidParamResolution(kvp.Key, kvp.Value));
+            return allResolutionsValid;
```
samarsha, line 214: `combination.IsValid` is true but `allResolutionsValid` could be false - it seems confusing that the combination could be valid but the resolutions inside the combination could be invalid?

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -194,20 +194,91 @@ public override int GetHashCode()
...
-        // This is the method that should be invoked to verify cycles of interest,
-        // i.e. where each callable in the cycle is type parametrized.
-        // It should probably generate diagnostics; I'll add the doc comment once its use is fully defined.
-        internal static bool VerifyCycle(CallGraphNode rootNode, params CallGraphEdge[] edges)
+        /// <summary>
+        /// Given a cycle of call graph edges, determines if the cycle is valid.
+        /// Invalid cycles are those that cause type parameters to be mapped to
+        /// other type parameters of the same callable (constricting resolutions)
+        /// or to a type containing a nested reference to the same type parameter,
+        /// i.e Foo.A -> Foo.A[].
```
samarsha, line 202: FYI, Haskell calls this an infinite type:

```
Prelude> f x = x : [f x]

<interactive>:12:1: error:
    * Occurs check: cannot construct the infinite type: t ~ [t]
      Expected type: t -> t
        Actual type: t -> [t]
    * Relevant bindings include
        f :: t -> t (bound at <interactive>:12:1)
```

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -194,20 +194,91 @@ public override int GetHashCode()
...
+        }
+
+        /// <summary>
+        /// Walker that checks a given type parameter resolution to see if it constricts
+        /// the type parameter to another type parameter of the same callable, or contains
+        /// a nested self-reference.
+        /// </summary>
+        private class CheckTypeParameterResolutions : TypeTransformation<CheckTypeParameterResolutions.TransformationState>
+        {
+            private readonly QsQualifiedName origin;
```
samarsha, line 224: It seems like it would make more sense to put all fields either in `CheckTypeParameterResolutions` or `TransformationState`, rather than splitting them. If `IsNestedSelfReference` is in `TransformationState` for consistency with multi-class transformations, then `origin` would also need to move into `TransformationState` if the transformation had multiple classes. If not, `IsNestedSelfReference` could also move into a field next to `origin`.

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -194,20 +194,91 @@ public override int GetHashCode()
...
-        internal static bool VerifyCycle(CallGraphNode rootNode, params CallGraphEdge[] edges)
+        /// <summary>
+        /// Given a cycle of call graph edges, determines if the cycle is valid.
+        /// Invalid cycles are those that cause type parameters to be mapped to
+        /// other type parameters of the same callable (constricting resolutions)
+        /// or to a type containing a nested reference to the same type parameter,
+        /// i.e Foo.A -> Foo.A[].
+        /// Returns true if the cycle is valid, false if invalid.
+        /// </summary>
+        internal static bool VerifyCycle(params CallGraphEdge[] edges)
```
samarsha, line 205: How does this relate to the diagnostic QS6213: "The given expression constrains the type parameter(s) 'a"? Will `VerifyCycle` be used to generate compile errors for the test cases?

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: Moved to the TransformationState class.

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: Looks like this hassle has to be dealt with in a lot of languages. 🙂

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: The first check, `combination.IsValid` comes from a tool that is agnostic to whether it is processing a cycle or a series of resolutions from a TypedExpression. Because of this it can't tell if the mapping `Foo.A -> Foo.A[]` is valid because that may be valid in a TypedExpression, if, for example the type parameter on the left is referring to a recursive reference while the type parameter on the left is referring to the calling context's type parameter. The second check is only valid in the case of cycle checking. In this case we know that `Foo.A -> Foo.A[]` is invalid, so we look for if that has occurred. Perhaps there is a better name for the variable for the second check? Or maybe we shouldn't capture it in a variable at all and just return the result of the expression, adding a comment to clarify what's being checked? What do you think?

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: I am hoping to separate out the logic that throws Diagnostic from the tooling for the QDK as much as possible. When this is put to use, we will see to what extent that can be accomplished and to what degree this code needs to throw Diagnostics.

samarsha: My question is mainly that if these test cases are invalid, why do they compile?

ScottCarda-MS: That's because the call graph and its features are not fully integrated into the language yet. 🙂 That will have to be another PR, or more likely a set of PR's as I think that will take quite a bit of work.

samarsha: OK, in the mean time will `VerifyCycle` be used for anything?

ScottCarda-MS: There is nothing in this PR that uses the VerifyCycle other than the unit tests.

samarsha: I meant: what was the reason for writing `VerifyCycle`? Is it only to generate diagnostics at some point in the future, or was there another use case?

ScottCarda-MS: Once we can integrate this with monomorphization, diagnostics will be generated based on this logic. I'm thinking in the precondition of the monomorphization rewrite step. So If there are invalid cycles in the call graph, we can throw an error diagnostic at that point.

samarsha: I think the rewrite step precondition would not be ideal, since diagnostics generated during the precondition won't show up automatically in the IDE (only when running `dotnet build`) - so the errors would not be noticed while typing. (Unless #544 is resolved.)

ScottCarda-MS: Okay, we can discuss that in more detail when the time comes.

src/QsCompiler/Transformations/CallGraph.cs
samarsha: Hmm. I understand this is not how type variables are represented currently, but it seems like the mapping `Foo.A -> Foo.A[]` should always be invalid, and if those `A`s are different, they should have different (maybe auto-generated?) names.

> Perhaps there is a better name for the variable for the second check?

My issue is not only with the variable name `allResolutionsValid`, but with the method name `IsValidParamResolution`.

ScottCarda-MS: Yes I agree, the type parameters need to have some way of differentiating between different instances of the same callable, but for now this is what we are stuck with. When that happens than `Foo.A -> Foo.A[]` where we are referring to the same type parameter would indeed always be not valid.

Yes I think I will remove the variable altogether and put a comment there, as well as rename the transformation and its static method.

ScottCarda-MS: I've renamed the `IsValidParamResolution` method to `ContainsNestedSelfResolution`.

bettinaheim: Yes, exactly; TypeResolutionCombination should not allow this - if it does, I'd consider that a bug.

PR #566, opened 2020-08-12
Using NuGet.ProjectModel from the install path to avoid version conflicts

ricardo-espinoza:
This change removes the dependency on NuGet.ProjectModel from the LanguageServer project and instead relies on the installed version of the binary. This change intends to avoid the dependency between the .Net Core SDK version and NuGet in MSBuildLocator. See issue [#86](https://github.com/microsoft/MSBuildLocator/issues/86) for additional reference.

### DIFF
--- a/src/QsCompiler/LanguageServer/Program.cs
+++ b/src/QsCompiler/LanguageServer/Program.cs
@@ -111,7 +111,6 @@ namespace Microsoft.Quantum.QsLanguageServer
                     {
                         return assemblyLoadContext.LoadFromAssemblyPath(path);
                     }
-
                     return null;
                 };
             }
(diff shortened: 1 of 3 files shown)


src/QsCompiler/LanguageServer/Program.cs
```diff
@@ -102,7 +103,16 @@ private static int Run(Options options)
             // This needs to be done before any MsBuild packages are loaded.
             try
             {
-                MSBuildLocator.RegisterDefaults();
+                VisualStudioInstance vsi = MSBuildLocator.RegisterDefaults();
```
anpaz-msft, line 106: I would add a comment explaining why you are doing this and pointing to the github issue where you found the workaround.

ricardo-espinoza: Sounds good. Adding.

PR #569, opened 2020-08-13
Master to Call Graph Walker

ScottCarda-MS:
master to feature branch

### DIFF
--- a/build/ci.yml
+++ b/build/ci.yml
@@ -12,8 +12,14 @@ variables:
   VSIX.Outdir: $(Drops.Dir)/vsix
   Blobs.Outdir: $(Drops.Dir)/blobs
 
+pr:
+- master
+- feature/*
+
 jobs:
 - job: "QsCompiler"
+  pool: 
+    vmImage: 'windows-latest'
   steps:
   - template: steps.yml
 


build/ci.yml
```diff
@@ -12,8 +12,14 @@ variables:
   VSIX.Outdir: $(Drops.Dir)/vsix
   Blobs.Outdir: $(Drops.Dir)/blobs
 
+pr:
+- master
+- feature/*
```
samarsha, line 17: I suspect this won't work because on qsharp-compiler, we have both `feature/*` and `features/*` branches. (I'd prefer we standardize to `feature/*`, like other repos, but there are existing branches with an `s`...)

ScottCarda-MS: Oh, you should defiantly bring that up with Andres as this was his fix. I would also prefer to use the `feature/*` format since that is what I have seen elsewhere.

ScottCarda-MS: Whatever the case is, it seems to have worked at least as far as this PR can tell. If it didn't we'd have the same problem as before where the build on ADO doesn't even start. Since this is just a master-to-feature PR I'm going to push forward with this. If you think this change is needed please bring it up with Andres and make a separate PR addressing this.

PR #570, opened 2020-08-13
Fix to build of faulty Q# program succeeds when calling `dotnet build` twice

cesarzc:
This change fixes an issue in which building a faulty Q# program succeeds when calling `dotnet build` twice.

This happens because a `bson` file is generated when the program is built the first time, even though compilation failed. When `dotnet build` is invoked for the second time, the `bson` file already exists and it is newer than the original `qs` file, so the incremental build skips invoking the compiler command line tool.

The fix is to delete the `bson` file when Q# compilation fails.

### DIFF
--- a/src/QuantumSdk/Sdk/Sdk.targets
+++ b/src/QuantumSdk/Sdk/Sdk.targets
@@ -114,7 +114,10 @@
     <PropertyGroup>
       <QscCommand>$(QscExe) build --format MsBuild $(_VerbosityFlag) --response-files @(QscCommandArgsFile)</QscCommand>
     </PropertyGroup>
-    <Exec Command="$(QscCommand)" IgnoreExitCode="false" /> 
+    <Exec Command="$(QscCommand)" IgnoreExitCode="false" > 
+      <Output TaskParameter="ExitCode" PropertyName="QscExitCode"/>
+    </Exec>
+    <Message Importance="high" Text="Qsc command exit code: $(QscExitCode)"/>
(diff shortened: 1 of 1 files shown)


src/QuantumSdk/Sdk/Sdk.targets
```diff
@@ -114,7 +114,12 @@
     <PropertyGroup>
       <QscCommand>$(QscExe) build --format MsBuild $(_VerbosityFlag) --response-files @(QscCommandArgsFile)</QscCommand>
     </PropertyGroup>
-    <Exec Command="$(QscCommand)" IgnoreExitCode="false" /> 
+    <Exec Command="$(QscCommand)" IgnoreExitCode="false" ContinueOnError="ErrorAndContinue"> 
+      <Output TaskParameter="ExitCode" PropertyName="QscExitCode"/>
+    </Exec>
+    <Message Importance="low" Text="Qsc command exit code: $(QscExitCode)"/>
+    <Delete Condition="'$(QscExitCode)' != '0'" Files="$(GeneratedFilesOutputPath)$(PathCompatibleAssemblyName).bson" />
+    <Error Text="Qsc command failed with exit code: $(QscExitCode)." Condition="'$(QscExitCode)' != '0'" />
```
anpaz-msft, line 122:
```suggestion
    <Error Text="Q# compiler failed with exit code: $(QscExitCode)." Condition="'$(QscExitCode)' != '0'" />
```

PR #578, opened 2020-08-20
Warning for underscore patterns that will be reserved in the future

bettinaheim:
This PR adds a warning when using double underscores in symbol names, or an underscore before a dot or at the end of a namespace name. We will reserve these patterns for internal use in the future.

### DIFF
--- a/src/QsCompiler/DataStructures/ReservedKeywords.fs
+++ b/src/QsCompiler/DataStructures/ReservedKeywords.fs
@@ -202,6 +202,9 @@ module InternalUse =
             "Borrow" 
             "Return"
 
+            "Data"
+            "Item"
+
             "QVoid"
             "Int64" 
             "BigInteger"
--- a/src/QsCompiler/DataStructures/ReservedKeywords.fs
+++ b/src/QsCompiler/DataStructures/ReservedKeywords.fs
@@ -202,9 +202,6 @@ module InternalUse =
             "Borrow" 
             "Return"
 
-            "Data"
-            "Item"
-
             "QVoid"
             "Int64" 
             "BigInteger"
--- a/src/QsCompiler/DataStructures/ReservedKeywords.fs
+++ b/src/QsCompiler/DataStructures/ReservedKeywords.fs
@@ -202,8 +202,8 @@ module InternalUse =
             "Borrow" 
             "Return"
 
-            "Data"
-            "Item"
+            "Data" // REL0920: can be removed once we give errors for underscore patterns
+            "Item" // REL0920: can be removed once we give errors for underscore patterns
 
             "QVoid"
             "Int64" 
--- a/src/QsCompiler/Tests.Compiler/CommandLineTests.fs
+++ b/src/QsCompiler/Tests.Compiler/CommandLineTests.fs
@@ -242,7 +242,7 @@ let ``generate docs`` () =
 let ``find path relative`` () =
     let fullPath = Path.Combine (Path.GetFullPath "alpha","beta","c","test-path.qs")
     let options = new BuildCompilation.BuildOptions()
-    let id = CompilationUnitManager.TryGetFileId (new Uri(fullPath)) |> snd
+    let id = CompilationUnitManager.GetFileId (new Uri(fullPath))
     let expected = Path.Combine (Path.GetFullPath "alpha","beta","c","test-path.g.cs")
     let actual = CompilationLoader.GeneratedFile(id, options.OutputFolder, ".g.cs")
     Assert.Equal(expected, actual)
@@ -252,7 +252,7 @@ let ``find path relative to outputfolder`` () =
     let fullPath = Path.Combine(Path.GetFullPath "alpha","beta","c","test-path.qs")
     let options = new BuildCompilation.BuildOptions()
     options.OutputFolder <- Path.Combine (pathRoot,"foo","bar")
-    let id = CompilationUnitManager.TryGetFileId (new Uri(fullPath)) |> snd
+    let id = CompilationUnitManager.GetFileId (new Uri(fullPath))
     let expected = Path.Combine (pathRoot, "foo", "bar", "alpha", "beta", "c", "test-path.g.cs")
     let actual = CompilationLoader.GeneratedFile(id, options.OutputFolder, ".g.cs")
     Assert.Equal(expected, actual)
@@ -262,7 +262,7 @@ let ``find path relative to relative outputfolder`` () =
     let fullPath = Path.Combine (Path.GetFullPath "alpha","beta","c","test-path.qs")
     let options = new BuildCompilation.BuildOptions()
     options.OutputFolder <- Path.Combine("..","foo","bar")
-    let id = CompilationUnitManager.TryGetFileId (new Uri(fullPath)) |> snd
+    let id = CompilationUnitManager.GetFileId (new Uri(fullPath))
     let expected = Path.Combine (parentDir,"foo","bar","alpha","beta","c","test-path.g.cs")
     let actual = CompilationLoader.GeneratedFile(id, options.OutputFolder, ".g.cs")
     Assert.Equal(expected, actual)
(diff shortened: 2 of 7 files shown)


src/QsCompiler/DataStructures/ReservedKeywords.fs
```diff
@@ -202,8 +202,8 @@ module InternalUse =
             "Borrow" 
             "Return"
 
-            "Data"
-            "Item"
+            "Data" // REL0920: can be removed once we give errors for underscore patterns
```
cgranade, line 205: Are we using `REL𝑚𝑚𝑦𝑦` to indicate deprecation timers? That seems like a really good practice to apply uniformly across repos, perhaps even to have a CI check for.

src/QsCompiler/Tests.Compiler/SyntaxTests.fs
```diff
@@ -17,7 +17,40 @@ open Xunit
...
+        ("mySymbol__", true , Some "mySymbol__", [Warning WarningCode.UseOfUnderscorePattern])
+        ("__my__sym" , true , Some "__my__sym" , [Warning WarningCode.UseOfUnderscorePattern])
+        ("my__sym__" , true , Some "my__sym__" , [Warning WarningCode.UseOfUnderscorePattern])
+        ("__mysym__" , true , None             , [Error ErrorCode.InvalidUseOfReservedKeyword])
+    ]
+    |> List.iter (testOne (symbolNameLike ErrorCode.InvalidIdentifierName))
+    [
+        ("a.b" , true , ([Some "a" ],Some "b"), [])
+        ("_a.b" , true , ([Some "_a" ],Some "b"), [])
+        ("a_.b" , true , ([Some "a_" ],Some "b"), [Warning WarningCode.UseOfUnderscorePattern])
```
cgranade, line 38: To clarify, are trailing single underscores reserved as well?

src/QsCompiler/TextProcessor/SyntaxBuilder.fs
```diff
@@ -345,17 +345,26 @@ let internal symbolNameLike errCode =
                      preCheckStart = isSymbolStart,
                      preCheckContinue = isSymbolContinuation) |> identifier
         getPosition .>>. id .>>. getPosition |>> fun ((p1, name), p2) -> name, (p1,p2)
-    let whenValid (name : string, range) = 
-        let isReserved = name.StartsWith "__" && name.EndsWith "__" || InternalUse.CsKeywords.Contains name
+    let whenValid ((isAfterDot, (name : string, range)), isBeforeDot) =
+        // REL0920: 
+        // The warning for reservedUnderscorePattern should be replace with an error in the future,
```
cgranade, line 350:
```suggestion
        // The warning for reservedUnderscorePattern should be replaced with an error in the future, 
```

src/QsCompiler/Tests.Compiler/SyntaxTests.fs
```diff
@@ -17,7 +17,40 @@ open Xunit
...
+        ("a._b" , true , ([Some "a" ],Some "_b"), [Warning WarningCode.UseOfUnderscorePattern])
```
cgranade, line 39: Similarly, are all leading underscores on operation and function names now banned? There's some places where that's still required to mark non-public APIs that can't yet be formally made internal due to needing internalsvisibleto or similar workarounds.

src/QsCompiler/DataStructures/ReservedKeywords.fs
bettinaheim: Yes, exactly. I started with TODO, and then though better of it and it seems useful to have tags ready that we can easily search for when we already know certain things need to be done with release X.

src/QsCompiler/Tests.Compiler/SyntaxTests.fs
bettinaheim: A trailing single underscore at the end of a callable/type name is fine, but not at the end of a namespace name (since the full name of a callable would then have that underscore be followed by a dot).

src/QsCompiler/Tests.Compiler/SyntaxTests.fs
bettinaheim: I update the PR to allow leading underscores.

src/QsCompiler/DataStructures/ReservedKeywords.fs
rmshaffer: FYI, I've copied this pattern here:
https://github.com/microsoft/iqsharp/blob/23ef42c4ea5b872b12a35dcd494845b4dc18bfb3/src/Core/Workspace/Project.cs#L233

In that case it's only a guess as to which future release, but it's definitely a better reminder than `TODO`.

src/QsCompiler/Tests.Compiler/SyntaxTests.fs
cgranade: Got it, thanks for clarifying. Is there somewhere that would make sense document that, perhaps?

bettinaheim: Right now it's just a warning. We actually currently don't document in our docs repo what valid identifiers in Q# are but expect to do that when we open the language repo. Hopefully that should be in time for the change next month.

src/QsCompiler/TextProcessor/QsFragmentParsing.fs
```diff
@@ -59,15 +59,31 @@ let private buildSymbolTuple (items, range : Position * Position) =
 let private expectedIdentifierDeclaration continuation = 
     expected localIdentifier ErrorCode.InvalidIdentifierDeclaration ErrorCode.MissingIdentifierDeclaration invalidSymbol continuation 
 
+/// Uses multiSegmentSymbol to parse the name of a namespace, 
+/// concatenates all path segments and the symbol with a dot, and returns a the concatenated string as Some.
+/// Returns None if the path contains segments that are None (i.e. invalid).
+/// Generates a suitable diagnostic if the namespace name ends in an underscore. 
+let internal namespaceName = // internal for testing purposes
+    let asNamespaceName ((path, sym : string option), range : Position * Position) = 
+        let names = [for segment in path do yield segment ] @ [sym]
```
samarsha, line 68: Not sure if path is a list or a seq, but wouldn't this be `path @ [sym]` or `Seq.toList path @ [sym]`?

src/QsCompiler/Tests.Compiler/CommandLineTests.fs
```diff
@@ -348,7 +348,7 @@ let ``find path absolute to outputfolder with spaces`` () =
 let ``find path relative to here with spaces`` () =
     let fullPath = Path.Combine (Path.GetFullPath "alpha","some beta","c","test 03.qs")
     let options = new BuildCompilation.BuildOptions()
-    let id = CompilationUnitManager.TryGetFileId (new Uri(fullPath)) |> snd
+    let id = CompilationUnitManager.GetFileId (new Uri(fullPath))
```
samarsha, line 351: Thanks for updating these!

src/QsCompiler/TextProcessor/SyntaxBuilder.fs
```diff
@@ -345,17 +345,25 @@ let internal symbolNameLike errCode =
                      preCheckStart = isSymbolStart,
                      preCheckContinue = isSymbolContinuation) |> identifier
         getPosition .>>. id .>>. getPosition |>> fun ((p1, name), p2) -> name, (p1,p2)
-    let whenValid (name : string, range) = 
-        let isReserved = name.StartsWith "__" && name.EndsWith "__" || InternalUse.CsKeywords.Contains name
+    let whenValid ((name : string, range), isBeforeDot) =
+        // REL0920: 
+        // The warning for reservedUnderscorePattern should be replaced with an error in the future, 
+        // and the first half of isReserved should be removed.
+        let reservedUnderscorePattern = name.Contains "__" || (isBeforeDot && name.EndsWith "_")
```
samarsha, line 352: I know this code is temporary - but maybe we can say `futureReservedUnderscores` in contrast with `isReserved` (which also contains a reserved underscore pattern)?

src/QsCompiler/Tests.Compiler/CommandLineTests.fs
bettinaheim: I try to actually pay attention to warnings as I see them. :)

src/QsCompiler/TextProcessor/SyntaxBuilder.fs
bettinaheim: Sure.

PR #579, opened 2020-08-20
Address review comments and merge from master for #531

samarsha:
See #531.


bettinaheim:
@samarsha Is there a reason this is still a draft PR?

samarsha:
I am waiting for your responses in #531 before I know which changes to make here (e.g. the naming of `TranslateLines`). I would like to address all the review comments in a single PR, so I was keeping this a draft until they could all be addressed.

PR #582, opened 2020-08-22
Enable temporary .csproj creation and associated language server tests

rmshaffer:
This PR re-enables temporary .csproj creation (essentially undoing #585) and also ensures that language server tests pass deterministically when opening both .qs files and non-.qs files.

There a few pieces to this fix, in addition to re-adding the code that was removed in #585:

1. Only create temporary projects for `.qs` files. Some `BasicFunctionality` tests load randomly-named source files without a `.qs` extension, and we want to allow those tests to operate as they do today.

1. Ensure that `MSBuildLocator` is registered for both the `BasicFunctionality` and `ProjectLoader` test classes via `AssemblyInitialize` attribute, rather than just a static constructor. With the static constructor approach, the order of execution was not deterministic, and if the `BasicFunctionality` tests were run prior to the `ProjectLoader` test class initialization, the proper version of MSBuild was not always being discovered. This actually led the tests to always _pass_, because the failure to create the temporary project was simply logged and the tests continued down their old path without a temporary project.

1. Change the temporary project folder naming scheme so that it is stable. The previous `GetHashCode()` approach was not even stable within a single test run, which occasionally caused errors. This is a good change from a user perspective as well, since it will ensure that the same temporary project is always used for code in a given folder.


src/QsCompiler/Tests.LanguageServer/Tests.cs
```diff
@@ -233,8 +238,8 @@ async Task RunTest(bool emptyLastLine)
                     Assert.AreEqual(expected, got);
                 }
             }
-            await RunTest(emptyLastLine: true);
-            await RunTest(emptyLastLine: false);
+            await RunTest(emptyLastLine: true, useQsExtension: false);
```
bettinaheim, line 241: It would be good to also run these tests when the files have a .qs extension.

rmshaffer: Yes, I agree. As I mentioned in the PR description, that works fine for me when running the tests locally but fails in the CI build for some reason that I haven't figured out yet. (If you look at the next-to-last commit of this PR you'll see what I tried.)

I will investigate and get that added ASAP, but since this unblocks the tests I was thinking to get this merged first. Let me know if you agree.

src/QsCompiler/LanguageServer/EditorState.cs
```diff
@@ -157,13 +157,13 @@ internal bool QsProjectLoader(Uri projectFile, out ProjectInformation info)
...
         {
-            projectUri = null;
-            info = null;
-
             var localFolderPath = Path.GetDirectoryName(sourceFileUri.LocalPath);
-            var projectFolderPath = Directory.CreateDirectory(Path.Combine(Path.GetTempPath(), "qsharp", $"{localFolderPath.GetHashCode():x8}")).FullName;
+            var projectFolderPath = Directory.CreateDirectory(Path.Combine(
+                Path.GetTempPath(),
+                "qsharp",
+                string.Join("_", localFolderPath.Split(Path.GetInvalidFileNameChars())))).FullName;
```
bettinaheim, line 166: Having just dealt with a bunch of edge case bugs related to naming conflicts; what is the reason to not just keep the folder structure relative to the workspace folder? I.e. if the workspace folder is /foo, and the file is /foo/subfolder/file.qs, then why not create a subfolder in the temp folder such that the project file is in tempfolder/subfolder/projFile.csproj?

src/QsCompiler/LanguageServer/EditorState.cs
```diff
@@ -244,27 +236,25 @@ internal bool QsTemporaryProjectLoader(Uri sourceFileUri, string sdkVersion, out
...
                 {
                     try
                     {
-                        if (this.QsTemporaryProjectLoader(textDocument.Uri, sdkVersion: null, out Uri projectUri, out _))
-                        {
-                            this.projects.ProjectChangedOnDiskAsync(projectUri, this.QsProjectLoader, this.GetOpenFile);
-                            this.onTemporaryProjectLoaded(projectUri);
-                            createdTemporaryProject = true;
-                        }
+                        var projectUri = this.QsTemporaryProjectLoader(textDocument.Uri, sdkVersion: null);
```
bettinaheim, line 254:
```suggestion
                        var projectUri = this.QsTemporaryProjectLoader(textDocument.Uri, sdkVersion: null); // null means the Sdk version will be set to the extension version
```

src/QsCompiler/LanguageServer/EditorState.cs
rmshaffer: What if a folder is open in multiple VS Code instances with different workspace roots? For example, if I open folder `/foo` in one instance of VS Code, and I open folder `/foo/subfolder` in another instance. Then `/foo/subfolder` has two temporary projects lying around. I'm not sure if that would actually cause any issue (I suspect not, in that case), but wouldn't it be more desirable to have a 1:1 mapping between folder and temporary .csproj?

rmshaffer:
@ricardo-espinoza @bettinaheim the tests are now passing reliably with the current set of changes, so I think is ready to go, pending your review. Thanks!

src/QsCompiler/LanguageServer/EditorState.cs
bettinaheim: I am not sure I see the benefits, since with two VS Code instances there would also be two servers running, each one compiling one project (the additional file doesn't bother me). 
In any case, there is also another way though to avoid naming conflicts without preserving a hierarchy inspired by url schemes, as pointed out by @samarsha: you could map both the path combinator and the underscore; e.g.  _ -> _x and / -> _y where x and y can be anything but need to be nontrivial.

rmshaffer: Yes, that approach makes sense. (I know it doesn't really matter much, but the additional file would bother me a little bit; I really don't want to leave more files than necessary on the user's machine. 😊)

Issue #583, opened 2020-08-22
Combining type parameter resolutions succeeds when it shouldn't


bettinaheim:
The following case and other similar cases (including direct recursions) should be indicated as invalid by TypeResolutionCombination:

    [<Fact>]
    [<Trait("Category","Type Resolution")>]
    member this.``Conflicting Resolution for type parameter in tuple`` () =
        let given = [|
            ResolutionFromParam [
                (FooA, BarA |> TypeParameter)
            ]
            ResolutionFromParam [
                (BarA, [Int; FooA |> TypeParameter] |> MakeTupleType)
            ]
        |]
        let expected = ResolutionFromParam [
            (BarA, [Int; FooA |> TypeParameter] |> MakeTupleType)
            (FooA, [Int; FooA |> TypeParameter] |> MakeTupleType)
        ]

        AssertCombinedResolutionFailure expected given

ScottCarda-MS:
Closed with #587.

PR #587, opened 2020-08-24
Nested Self Resolutions are Disallowed

ScottCarda-MS:
This Pr updates the TypeParameterCombinations tool so that nested self-resolutions for type parameters are considered constrictive and invalidate the combination.

Addresses #583.

### DIFF
--- a/src/QsCompiler/Tests.Compiler/TypeParameterTests.fs
+++ b/src/QsCompiler/Tests.Compiler/TypeParameterTests.fs
@@ -263,7 +263,7 @@ type TypeParameterTests () =
         AssertCombinedResolution expected given
 
     [<Fact>]
-    [<Trait("Category","Type resolution")>]
+    [<Trait("Category","Type Resolution")>]
     member this.``Multi-Stage Resolution of Multiple Resolutions to Type Parameter`` () =
 
         let given = [|
@@ -446,7 +446,7 @@ type TypeParameterTests () =
         AssertCombinedResolutionFailure expected given
 
     [<Fact>]
-    [<Trait("Category","Type resolution")>]
+    [<Trait("Category","Type Resolution")>]
     member this.``Inner Cycle Constrains Type Parameter`` () =
 
         let given = [|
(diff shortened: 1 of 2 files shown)


src/QsCompiler/Transformations/TypeResolutionCombination.cs
```diff
@@ -282,33 +282,46 @@ public override ResolvedTypeKind OnTypeParameter(QsTypeParameter tp)
...
-
             /// <summary>
             /// Walks the given ResolvedType, typeParamRes, and returns true if there is a reference
-            /// to a different type parameter of the same callable as the given type parameter, typeParam.
+            /// to a different type parameter of the same callable as the given type parameter, typeParam,
+            /// or the same type parameter but in a nested type.
             /// Otherwise returns false.
             /// </summary>
-            public static bool Apply(TypeParameterName typeParam, ResolvedType typeParamRes)
+            public static bool IsConstrictiveResolution(TypeParameterName typeParam, ResolvedType typeParamRes)
```
bettinaheim, line 296: This is a part of the public API and it hence cannot simply be renamed. Since this is a bug fix, it also does not change the behavior in a way that would warrant deprecating it in favor of a new name.

src/QsCompiler/Tests.Compiler/TypeParameterTests.fs
```diff
@@ -501,25 +501,39 @@ type TypeParameterTests () =
...
+        |]
+        let expected = ResolutionFromParam [
+            (FooA, [FooA |> TypeParameter; Int] |> MakeTupleType)
+        ]
+
+        AssertCombinedResolutionFailure expected given
+
+    [<Fact>]
+    [<Trait("Category","Type Resolution")>]
+    member this.``Constricting Indirect Nested Self Resolution`` () =
```
bettinaheim, line 519: It would be good to also add the tests for arrays.

src/QsCompiler/Transformations/TypeResolutionCombination.cs
ScottCarda-MS: This is part of a private class, so it isn't part of the public API.

src/QsCompiler/Tests.Compiler/TypeParameterTests.fs
ScottCarda-MS: I've added tests for nested self references in arrays in addition to the tests for nested self references in tuples.

src/QsCompiler/Transformations/TypeResolutionCombination.cs
bettinaheim: Sorry, I only looked at the diff.

PR #588, opened 2020-08-25
Cycle Verification Diagnostics 

ScottCarda-MS:
Adds a call graph field to the QsCompilation type. The ICallGraphEdge, ICallGraphNode, and ICallGraph interfaces are added. Uses the cycle verification to raise diagnostics.


src/QsCompiler/CompilationManager/CompilationUnit.cs
```diff
@@ -883,14 +889,32 @@ public QsCompilation Build()
...
         }
 
+        /// <summary>
+        /// Checks the cycles of the given callables and returns diagnostics for
+        /// any invalid cycles, along with the qualified name of the callable the
+        /// diagnostic should be placed.
+        /// </summary>
+        public IEnumerable<Tuple<QsQualifiedName, QsCompilerDiagnostic>> VerifyCycles(IEnumerable<QsCallable> callables)
+        {
+            this.CachedCallGraph = QsNullable<ICallGraph>.NewValue(Transformations.CallGraphWalker.BuildCallGraph.Apply(callables));
```
samarsha, line 913: Does this always need to rebuild the call graph, or can it reuse a cached version if it exists?

src/QsCompiler/CompilationManager/TypeChecking.cs
```diff
@@ -2015,6 +2015,12 @@ QsCallable GetCallable((QsQualifiedName, ImmutableArray<QsSpecialization>) specI
                 }
                 compilation.UpdateCallables(callables);
                 compilation.UpdateTypes(types);
+                foreach (var (parent, diag) in compilation.VerifyCycles(callables))
+                {
+                    var info = callableDeclarations[parent];
+                    var offset = info.Position is DeclarationHeader.Offset.Defined pos ? pos.Item : null;
+                    diagnostics.Add(Diagnostics.Generate(info.SourceFile.Value, diag, offset));
```
samarsha, line 2022: I am curious why the call graph can't make diagnostics with absolute ranges? In my experience, if you have a `QsCallable`, you should have enough information to get absolute positions.

src/QsCompiler/DataStructures/SyntaxTree.fs
```diff
@@ -789,11 +791,42 @@ type QsNamespace = {
...
+// Interface used to represent a call graph
+type ICallGraph = 
+    abstract member Count : int
+    abstract member Nodes : ImmutableHashSet<ICallGraphNode>
+    abstract member ContainsNode : ICallGraphNode -> bool
+    abstract member GetAllDependencies : ICallGraphNode -> ILookup<ICallGraphNode, ICallGraphEdge>
+    abstract member GetAllDependencies : QsSpecialization -> ILookup<ICallGraphNode, ICallGraphEdge>
+    abstract member GetDirectDependencies : ICallGraphNode -> ILookup<ICallGraphNode, ICallGraphEdge>
+    abstract member GetDirectDependencies : QsSpecialization -> ILookup<ICallGraphNode, ICallGraphEdge>
+    abstract member VerifyAllCycles : Unit -> IEnumerable<QsQualifiedName * QsCompilerDiagnostic>
```
samarsha, line 820: Can you move the documentation comments from the call graph to here?

src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -649,6 +650,7 @@ type DiagnosticItem =
             | ErrorCode.AttributeInvalidOnCallable                -> "Invalid attribute placement. The attribute {0} cannot be attached to a callable declaration."
             | ErrorCode.UnresolvedTypeParameterForRecursiveCall   -> "The type argument(s) for the recursive call could not be inferred. Please provide explicit type arguments, e.g. Op<Int, Double>(arg)."
             | ErrorCode.TypeParameterResConflictWithTypeArgument  -> "The type of the expression needs to match the defined type argument. Expecting an expression of type {0}."
+            | ErrorCode.InvalidCyclicTypeParameterResolution      -> "Call cycle results in ambiguous or conflicting type parameter resolution."
```
samarsha, line 653: Is it possible to have the diagnostic message include the specific type variables that are ambiguous or conflicting?

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -837,7 +837,7 @@ private QsCompilation ReplaceTargetSpecificImplementations(IEnumerable<string> p
                 LogError(ErrorCode.ConflictsInTargetSpecificDecompositions, Array.Empty<string>());
             }
 
-            var targetSpecificDecompositions = new QsCompilation(replacements, ImmutableArray<QsQualifiedName>.Empty);
+            var targetSpecificDecompositions = new QsCompilation(replacements, ImmutableArray<QsQualifiedName>.Empty, QsNullable<ICallGraph>.Null);
```
samarsha, line 840: I'm wondering if it's possible for QsCompilation to automatically build a call graph when it is constructed, if one isn't provided? It would be convenient if the call graph wasn't behind a QsNullable, are there cases where it is actually necessary to not build the call graph?

src/QsCompiler/CompilationManager/CompilationUnit.cs
ScottCarda-MS: The arguments to this method are the callables that are being verified against, so if someone calls this method with several callables, but there is already a cached call graph, we would't want to ignore the given callable in favor of the old call graph. Here we want to make a new call graph with the given callable, set it as the cache and verify if the cycles in the new graph are valid.

src/QsCompiler/CompilationManager/TypeChecking.cs
ScottCarda-MS: The QsCallables here are not defined with locations. The way I have added the offset here is the way it is done further up in the same method. I did it this way to maintain consistency with the other logic in this method. The reason why the callables being operated on don't have locations likely has something to do with them coming from a fragment tree rather than a proper syntax tree, but the details on that are a bit out of scope for what I needed to accomplish here.

src/QsCompiler/DataStructures/Diagnostics.fs
ScottCarda-MS: I had given some thought to how to do that, but wasn't able to come up with anything. I'd have to give this a lot more thought on what that would look like. I supposed it's not really a single type parameter causing the issue, just like it's not a single call that causes the issue. We would want to avoid having the cross product between all calls in the cycle and all type parameters in the cycle, but that might be doable. Unfortunately we are running short on time with this, so this is the simple solution for now, and I believe we will be able to update the error message later when we figure this out.

src/QsCompiler/Compiler/CompilationLoader.cs
ScottCarda-MS: When a call graph isn't provided, it is usually because one is not needed or wanted in that circumstance. For example, in the CompilationLoader's SerializeSyntaxTree method, we don't want to serialize a call graph along with the syntax tree, so we want a compilation that doesn't have a call graph.

samarsha: If we want to serialize the compilation without the call graph, I feel like that should be a separate data structure that only has the fields we want to serialize, rather than always serializing a null value. Are there other instances besides serialization where a null call graph is needed?

samarsha: To explain a little more, my main concern is for rewrite steps, transformations, etc. that work on the compilation, is the call graph always going to be present? If so, it would be more convenient to remove the nullability from it so they do not need to handle the case where the call graph is not there.

src/QsCompiler/DataStructures/Diagnostics.fs
samarsha: OK, that seems reasonable, thanks for the explanation.

src/QsCompiler/DataStructures/SyntaxTree.fs
ScottCarda-MS: I've copied over the docs as you suggested.

src/QsCompiler/Compiler/CompilationLoader.cs
ScottCarda-MS: By the time we are doing rewrite steps, the main compilation that gets pass from one rewrite step to the next always has a call graph. But there are some instances where secondary compilations are made that don't need the call graph, such as when we are merging in intrinsic decomposition.

samarsha: What happens when a rewrite step changes the compilation but doesn't update the call graph? Does the next rewrite step get an out-of-date call graph?

Since the call graph is

1. derived entirely from the other fields in the compilation
2. often invalidated/out-of-date
2. not always needed

I am sort of thinking it might make sense to have it be a computed property instead. Perhaps using `lazy` or [`System.Lazy`](https://docs.microsoft.com/en-us/dotnet/api/system.lazy-1?view=netcore-3.1), so the call graph is always up to date and always computed when needed, but never computed when it isn't needed?

ScottCarda-MS: I've updated the docs PR for the call graph with the solution to out-of-date call graphs. It is a strategy that involves comparing guids. If you'd like to discuss a different strategy, like your suggestion of using a lazy evaluation, we can do that offline. But either way that would be part of a different PR.

ScottCarda-MS: On second though, the `lazy` call graph idea might be easy enough to try out on this PR. Let me do some looking into it and get back to you.

samarsha: For this PR, my concern is primarily about QsNullable. It seems like the cases where the call graph is present outnumber the cases where it isn't - let me know if I misunderstood that. I am interested in making the common case convenient and type-safe, but with a QsNullable I am imagining that rewrite steps will either:

1. Assume that the call graph is always there and unsafely access QsNullable.Item. If the rewrite step is used in a way that means the call graph isn't there, it will crash.
2. Check if the call graph is there and handle both cases, potentially skipping the rewrite step or manually building the call graph if it is missing. However, if the call graph is always there, every rewrite step will need to do unnecessary checks, adding to the required boilerplate for rewrite steps that use the call graph

Please let me know if my assessment of the situation isn't accurate. :) Otherwise, is there a way to resolve this in this PR?

src/QsCompiler/CompilationManager/TypeChecking.cs
samarsha: Sorry, I didn't see that the callables above are created with a null location. Is there any way to avoid the duplicated code here and in GetCallable above? Maybe GetCallable can use the offset it finds (plus the range, which I think is also in the DeclarationHeader?) to fill in the callable location?

src/QsCompiler/Compiler/CompilationLoader.cs
ScottCarda-MS: Yah most of the time the main compilation will have a non-null call graph. I'm still looking into the `lazy` call graph approach to see if it'll work the way we need it to. One practical concern I have with that approach is that the QsCompiler object will need to have the ability to create call graphs, which means it would need access to the call graph walker. This might create a dependency issue.

samarsha: Hmm, yeah, this circular dependency between the call graph and the compilation is an issue...

What are the disadvantages to building it on-the-fly instead of storing it in `QsCompilation`? For performance, the compilation manager could still cache it, but if it wasn't stored, rewrite steps never need to worry about accidentally having an out-of-date call graph, and things that create a `QsCompilation` but don't care about the call graph don't need to do unnecessary work...

The one case where this would cause duplicate work that I can think of is:

1. A `QsCompilation` is passed from one place to another (so they couldn't share a cache),
2. without being modified (so the call graph wouldn't be out-of-date), and
3. both places need the call graph?

Is that common?

src/QsCompiler/DataStructures/SyntaxTree.fs
```diff
@@ -789,11 +791,91 @@ type QsNamespace = {
...
+    /// Returns all specializations that are used directly within the given caller, whether they are
+    /// called, partially applied, or assigned. Each key in the returned lookup represents a
+    /// specialization that is used by the caller. Each value in the lookup is an IEnumerable of edges
+    /// representing all the different ways the given caller specialization took a dependency on the
+    /// specialization represented by the associated key.
+    /// Throws ArgumentNullException if argument is null.
+    /// Returns an empty ILookup if the node was found with no dependencies or was not found in
+    /// the graph.
+    abstract member GetDirectDependencies : QsSpecialization -> ILookup<ICallGraphNode, ICallGraphEdge>
+    /// Given a call graph edges, finds all cycles and determines if each is valid.
```
samarsha, line 862: This is a minor nitpick but could you put a blank line between the members? This is a lot of text. :)

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -192,30 +170,344 @@ public override int GetHashCode()
...
+        /// Returns all specializations that are used directly within the given caller, whether they are
+        /// called, partially applied, or assigned. Each key in the returned lookup represents a
+        /// specialization that is used by the caller. Each value in the lookup is an IEnumerable of edges
+        /// representing all the different ways the given caller specialization took a dependency on the
+        /// specialization represented by the associated key.
+        /// Throws ArgumentNullException if argument is null.
+        /// Returns an empty ILookup if the node was found with no dependencies or was not found in
+        /// the graph.
+        /// </summary>
+        public ILookup<ICallGraphNode, ICallGraphEdge> GetDirectDependencies(ICallGraphNode callerNode)
```
samarsha, line 329: I think you could replace these doc comments with `/// <inheritdoc/>` since the interface has documentation now.

src/QsCompiler/Compiler/CompilationLoader.cs
ScottCarda-MS: With how often the call graph is going to be made out of date (every time a new QsCompilation is made), maybe it does make sense to not have it as a field on the compilation, and just have all the locations create one on-the-fly as needed. This would certainly simplify the changes to the QsCompilation (eliminate the changes, actually). Would we still want to keep around the interfaces for the call graph classes?

src/QsCompiler/CompilationManager/CompilationUnit.cs
samarsha: What happens in this situation:

1. `VerifyCycles` is called with a list of callables that is not the same as `this.GlobalSymbols.DefinedCallables()`. It sets the cached call graph.
2. `Build` is called afterwards. It checks `this.CachedCallGraph.IsNull` and does not regenerate the call graph.

Does the compilation returned by `Build` have an incorrect call graph?

ScottCarda-MS: If the callables given to VerifyCycles are not the same callables that are used in the Build() method, than the call graph would not be accurate to the returned syntax tree.

samarsha: I think without looking at the implementation of both `VerifyCycles` and `Build`, it is not obvious that that sequence of events causes `Build` to behave incorrectly, since `VerifyCycles` does not indicate that it changes the state of the compilation unit. It seems like it would be better to prevent this scenario from happening.

src/QsCompiler/Compiler/CompilationLoader.cs
samarsha: I think if the call graph isn't in `QsCompilation` (or some other type in the DataStructures project), then that probably removes the need to have the interfaces.

src/QsCompiler/CompilationManager/CompilationUnit.cs
ScottCarda-MS: This method has been removed because the QsCompilation no longer has a call graph field. RunTypeChecking will generate its own call graph for generating diagnostics.

src/QsCompiler/Compiler/CompilationLoader.cs
ScottCarda-MS: As per this discussion, the call graph field of the compilation has been removed in favor of the design where, if a piece of code needs a call graph, they generate it. The interfaces have been removed as they are no longer necessary.

src/QsCompiler/DataStructures/SyntaxTree.fs
ScottCarda-MS: These interfaces have been removed.

src/QsCompiler/Transformations/CallGraph.cs
ScottCarda-MS: The interfaces have been removed, so this is the only place where the doc comments live now.

src/QsCompiler/CompilationManager/TypeChecking.cs
ScottCarda-MS: I can add the location information to the callables and the specializations on the callables, but I don't really know why they were null before. It seems like it was an intentional choice. But by doing that I do get the correct position info for the diagnostic.

ScottCarda-MS: We decided offline that this change would require Compiler Diagnostics to support Source File information.

src/QsCompiler/CompilationManager/TypeChecking.cs
```diff
@@ -2015,6 +2015,13 @@ QsCallable GetCallable((QsQualifiedName, ImmutableArray<QsSpecialization>) specI
                 }
                 compilation.UpdateCallables(callables);
                 compilation.UpdateTypes(types);
+                var callGraph = Transformations.CallGraphWalker.BuildCallGraph.Apply(callables);
+                foreach (var (parent, diag) in callGraph.VerifyAllCycles())
+                {
+                    var info = callableDeclarations[parent];
+                    var offset = info.Position is DeclarationHeader.Offset.Defined pos ? pos.Item : null;
+                    diagnostics.Add(Diagnostics.Generate(info.SourceFile.Value, diag, offset));
```
samarsha, line 2023: I was testing this with the language server and the VS Code plugin, and I'm not sure this is sufficient to keep the diagnostics up-to-date. I was able to come up with this example:

```
    operation Foo<'a>() : Unit {
        Bar<'a[]>();
    }

    operation Bar<'a>() : Unit {
        Message("bar");
        // Foo<'a>();
    }
```

where if I uncommented `// Foo<'a>();`, it would not give a diagnostic for a bad cycle until I manually saved the file. Similarly, if I commented that line again, the error message on that line would go away, but the error message at the `Bar` call would stay until I saved the file.

samarsha: A related issue: if `Foo<'a>();` is uncommented, and I edit the source code of `Bar` without changing the call to `Foo`, the diagnostic for the `Foo` call will disappear.

src/QsCompiler/DataStructures/SyntaxTree.fs
```diff
@@ -6,8 +6,10 @@ namespace Microsoft.Quantum.QsCompiler.SyntaxTree
 open System
 open System.Collections.Immutable
 open System.Linq
+open Microsoft.Quantum.QsCompiler
 open Microsoft.Quantum.QsCompiler.DataTypes
 open Microsoft.Quantum.QsCompiler.SyntaxTokens
+open System.Collections.Generic
```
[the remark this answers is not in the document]
ScottCarda-MS: I removed the added directives as they are not used anymore.

ScottCarda-MS:
Closed in favor of #621.

Issue #586, opened 2020-08-24
Verify runtime capabilities of callables defined in referenced libraries


samarsha:
The compiler generates errors for use of result equality that isn't supported by a Q# executable's execution target, if the result equality occurs in the executable's source code (see #449 and #487). But it doesn't check result equality in libraries that the executable may be using. An unsupported result equality in a library may cause a runtime error for the executable, which we'd like to avoid.

Since libraries do not have an execution target (they can be used by an executable with any target), we need to analyze the minimum capabilities required by each callable and save that information in the AST. Then executable projects can verify that every callable used from a library does not have capabilities that are too high.

PR #593, opened 2020-08-27
Add callable-local runtime capability inference

samarsha:
Infers the minimum required runtime capabilities of each callable in the compilation, and adds the `@Microsoft.Quantum.Core.Capability` attribute to it, containing its capability ("QPRGen0", "QPRGen1", or "Unknown"). The inference is run as a rewrite step for libraries only, because executables have an execution target with a fixed runtime capability, so labeling is not needed. The attribute will be used later to verify that executables can't call library functions/operations with a capability that is too high.

The existing code for capability diagnostics was refactored, so both inference and verification can use the same code. The inference algorithm finds syntactic patterns in statements and expressions that imply a runtime capability. These patterns can be converted into either an inferred capability (by taking the max of all required capabilities), or diagnostics (by selecting patterns that imply a capability that is too high for the current execution target).

For now, inference only looks at the source code of callables individually - it doesn't find references or use the call graph, so the actual capability may be higher than what is currently inferred.

Part of issue #586.

### DIFF
--- a/src/QsCompiler/Compiler/PluginInterface.cs
+++ b/src/QsCompiler/Compiler/PluginInterface.cs
@@ -44,6 +44,12 @@ namespace Microsoft.Quantum.QsCompiler
         /// evaluates classical computations as much as possible.
         /// </summary>
         public const int EvaluationOfClassicalComputations = 100;
+
+        /// <summary>
+        /// Priority of the built-in transformation that infers the minimum runtime capabilities required by each
+        /// callable.
+        /// </summary>
+        public const int CapabilityInference = 90;
     }
 
     public interface IRewriteStep
--- a/src/QsCompiler/Compiler/PluginInterface.cs
+++ b/src/QsCompiler/Compiler/PluginInterface.cs
@@ -11,7 +11,8 @@ using VS = Microsoft.VisualStudio.LanguageServer.Protocol;
 namespace Microsoft.Quantum.QsCompiler
 {
     /// <summary>
-    /// Lists the priorities for built-in rewrite steps.
+    /// Lists the priorities for built-in rewrite steps. Steps with a larger priority number have higher priority and
+    /// will be executed first.
     /// </summary>
     public static class RewriteStepPriorities
     {
@@ -121,7 +122,7 @@ namespace Microsoft.Quantum.QsCompiler
 
         /// <summary>
         /// The priority of the transformation relative to other transformations within the same dll or package.
-        /// Steps with higher priority will be executed first.
+        /// Steps with a larger priority number have higher priority and will be executed first.
         /// </summary>
         public int Priority { get; }
 
--- a/src/QsCompiler/Compiler/RewriteSteps/CapabilityInference.cs
+++ b/src/QsCompiler/Compiler/RewriteSteps/CapabilityInference.cs
@@ -38,10 +38,11 @@ namespace Microsoft.Quantum.QsCompiler.BuiltInRewriteSteps
                 if (element is QsNamespaceElement.QsCallable callable)
                 {
                     Console.WriteLine(callable.Item.FullName);
-                    foreach (var inference in callable.Item.Specializations.SelectMany(SpecializationInferences))
+                    foreach (var pattern in callable.Item.Specializations.SelectMany(SpecializationPatterns))
                     {
-                        Console.WriteLine(inference);
+                        Console.WriteLine(pattern);
                     }
+                    Console.WriteLine();
                 }
             }
             transformed = compilation;
--- a/src/QsCompiler/Compiler/RewriteSteps/CapabilityInference.cs
+++ b/src/QsCompiler/Compiler/RewriteSteps/CapabilityInference.cs
@@ -32,7 +32,10 @@ namespace Microsoft.Quantum.QsCompiler.BuiltInRewriteSteps
 
         public bool Transformation(QsCompilation compilation, out QsCompilation transformed)
         {
-            // TODO
+            foreach (var inference in SyntaxProcessing.CapabilityInference.Inferences(compilation))
+            {
+                Console.WriteLine(inference);
+            }
             transformed = compilation;
             return true;
         }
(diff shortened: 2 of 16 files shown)


src/QsCompiler/Compiler/RewriteSteps/CapabilityInference.cs
```diff
@@ -0,0 +1,42 @@
...
+        public IEnumerable<IRewriteStep.Diagnostic> GeneratedDiagnostics { get; } =
+            Enumerable.Empty<IRewriteStep.Diagnostic>();
+
+        public bool ImplementsPreconditionVerification { get; } = false;
+
+        public bool ImplementsTransformation { get; } = true;
+
+        public bool ImplementsPostconditionVerification { get; } = false;
+
+        public bool PreconditionVerification(QsCompilation compilation) => throw new NotSupportedException();
```
cesarzc, line 32: How does the rewrite steps infrastructure handles `PreconditionVerification` and `PostconditionVerification` throwing not  supported exception?

src/QsCompiler/Compiler/PluginInterface.cs
```diff
@@ -44,6 +44,12 @@ public static class RewriteStepPriorities
         /// evaluates classical computations as much as possible.
         /// </summary>
         public const int EvaluationOfClassicalComputations = 100;
+
```
DmitryVasilevsky, line 47: Unrelated to this change, but it would probably be nice to update the comment on this class to say which are higher priorities and which are lower. I.e. higher number - higher priority or otherwise.

src/QsCompiler/Compiler/RewriteSteps/CapabilityInference.cs
samarsha: I think they should never be called because `ImplementsPreconditionVerification` and `ImplementsPostconditionVerification` are both false.

src/QsCompiler/SyntaxProcessor/CapabilityInference.fs
```diff
@@ -0,0 +1,240 @@
...
+
+/// Returns the required runtime capability of the pattern, given whether it occurs in an operation.
+let private toCapability inOperation = function
+    | ResultEqualityInCondition _ -> if inOperation then RuntimeCapabilities.QPRGen1 else RuntimeCapabilities.Unknown
+    | ResultEqualityNotInCondition _ -> RuntimeCapabilities.Unknown
+    | ReturnInResultConditionedBlock _ -> RuntimeCapabilities.Unknown
+    | SetInResultConditionedBlock _ -> RuntimeCapabilities.Unknown
+
+/// The runtime capability with the lowest level.
+let private baseCapability = RuntimeCapabilities.QPRGen0
```
cesarzc, line 39: It might be worth it to add a comment with the characteristics of each runtime capability in `ReservedKeywords.fs` where the names of the capabilities are defined. That way, when they are referenced, people can look at the definition and have a bit more context.

src/QsCompiler/Tests.Compiler/TestCases/CapabilityVerification.qs
```diff
@@ -191,4 +197,38 @@ namespace Microsoft.Quantum.Testing.CapabilityVerification {
...
+        }
+    }
+
+    @Capability("QPRGen0")
+    operation ResetOverrideLow(q : Qubit) : Unit {
+        if (M(q) == One) {
+            X(q);
+        }
+    }
+}
```
cesarzc, line 216: It would be good to have another test case in which capability is overridden with `QPRGen1`. Maybe even two cases, one in which the if-statement contains a return statement and one in which the if-statement contains a set-statement for a mutable variable declared outside that block.

src/QsCompiler/Tests.Compiler/CapabilityInferenceTests.fs
```diff
@@ -0,0 +1,63 @@
...
+
+/// Asserts that the inferred capability of the callable with the given name matches the expected capability.
+let private expect capability name =
+    let actual =
+        callables.[CapabilityVerificationTests.testName name].Attributes
+        |> QsNullable<_>.Choose BuiltIn.GetCapability
+    Assert.Equal<RuntimeCapabilities> ([ capability ], actual)
+
+[<Fact>]
+let ``Infers QPRGen0`` () =
```
cesarzc, line 26: Result comparison for `QPRGen0` not being supported only applies for operations (not for functions)?

samarsha: In QPRGen0, result comparison is forbidden everywhere, including both operations and functions.

samarsha: If you are asking about `ResultTuple` and `ResultArray` being inferred as QPRGen0, it's because at the moment tuples and arrays in Q# do not support equality, so no result equality comparison happens in those functions (they don't compile because equality is not supported). The test case is confusing, though...

cesarzc: Thanks for the explanation Sarah! Yes, I was a bit confused about the test case and the only thing I could see different is that it was a function rather than an operation.

src/QsCompiler/Tests.Compiler/TestCases/CapabilityVerification.qs
samarsha: I added tests for overriding QPRGen0 to QPRGen1, and for explicitly specifiying QPRGen1 on something that would already be inferred as QPRGen1.

> Maybe even two cases, one in which the if-statement contains a return statement and one in which the if-statement contains a set-statement for a mutable variable declared outside that block.

IMO it's not necessary to distinguish between these two cases in the override tests - there are other tests that make sure both of those cases are inferred as Unknown.

samarsha: Also just added a test for overriding Unknown to Gen1 (forgot about that case).

cesarzc: Thanks Sarah! I liked the new names of the operations :)

PR #595, opened 2020-08-30
Remove deprecated functions from syntax highlight

tcNickolas:


src/VSCodeExtension/syntaxes/qsharp.tmLanguage.json
```diff
@@ -96,7 +96,7 @@
         },
         {
           "name": "support.function.builtin.qsharp",
-          "match": "\\b(Message|Length|Assert|AssertProb|AssertEqual|Random|Floor|Float)\\b"
+          "match": "\\b(Message|Length|Floor)\\b"
```
samarsha, line 99: Removing deprecated functions makes sense, but the remaining functions confuse me. :) Why include Floor but not Ceiling, for example?

tcNickolas: That, I have no recollection of... I don't think it's ever been a built-in or needed any special treatment.
I'm happy to remove it as well?

samarsha: My guess for what "built-in" should mean is everything in [Microsoft.Quantum.Core](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.core), since that namespace is auto-opened... so Message and Floor would be out, but Length stays (plus other stuff like EntryPoint, Deprecated, etc. is included). Does that seem reasonable?

tcNickolas: We might need a separate syntax highlight for attributes, we'd need to check what other languages do. I don't have time to dig much deeper into this this week, sorry :-( I can drop `Floor` now (I'll admit I'm rather attached to `Message` highlights...) so that we can merge this, and file an issue to add highlights for attributes?

samarsha: It's OK, the PR is fine as-is, since Message and Floor were already there.

I was thinking of only highlighting the built-in attributes, and leaving other attributes with the default text color, but like you mentioned it might make more sense to highlight all attributes (built-in or not) the same color. Yes, we don't need to discuss that here, another issue is fine. :)

samarsha (approved):

Issue #573, opened 2020-08-17
Type Argument Inference Fails if the same Type Parametrized Callable is used Multiple Times


ScottCarda-MS:
This code highlights an issue with type argument inference:
```
operation` Foo<'T> (arg : 'T) : Unit { }
 
operation DoTwo<'A, 'B>(op1 : ('A => Unit), arg1 : 'A, op2: ('B => Unit), arg2 : 'B) : Unit {
    op1(arg1);
    op2(arg2);
}
 
operation CombineTwo<'A, 'B>(op1 : ('A => Unit), op2: ('B => Unit)) : (('A, 'B) => Unit) {
    return DoTwo(op1, _, op2, _);
}
 
operation Main() : Unit {
    (CombineTwo(Foo, Foo))("Hi", 3);
}
```
The expression in the Main operation has an issue here. The type parameter resolution data structure in the syntax tree doesn't explicitly contain any way of keeping track of which instance of a type parameter the resolution is for. It only contains the callable name, the type parameter name, and the resolving type. In the code above, in Main, the expression has two conflicting resolutions because it is trying to resolve `Foo.T` to `String` and its trying to resolve `Foo.T` to `Int`. It sees this conflict because it can't differentiate between the resolution for the first Foo and the resolution for the second Foo. As a result it throws an error complaining about ambiguity.

However if we instead have two different functions passed into `CombineTwo`:
```
operation Bar<'T> (arg : 'T) : Unit { }

operation Main() : Unit {
    (CombineTwo(Foo, Bar))("Hi", 3);
}
```
Now there is no more conflict because it sees `Foo.T` resolving to `String` and `Bar.T` resolving to `Int`. The fact that one of these scenarios is supported and not the other is the issue.

ScottCarda-MS:
I have been looking into how C# handles similar cases and found that they have a rule that protects against this. They don't allow type arguments to be inferred from how the method's return is used. This would actually prevent both of these cases from compiling, which is at least consistent, because in both cases the type parameters don't get resolved until the `("Hi", 3)` is applied to the returned operation of CombineTwo. While I think it would be neat if we could handle both cases, I think preventing type argument inference by use of the return makes sense and would be consistent with out rule against capturing unresolved generics.

[...]

Issue #29, opened 2019-07-11
Allow function overloading


Strilanc:
When writing arithmetic functions, it seems common to have variants specialized to `int`, `BigInt`, and `LittleEndian`. It's confusing to have to remember different names for these functions, instead of just having one standard name with disambiguation via the type system.

bettinaheim:
We have temporarility moved this issue to the compiler repository for tracking. 
This issue will migrate to the appropriate repository in the near future and will be processed there.
At this point we are not ready for contributions implementing this feature yet.

alan-geller:
I'm going to transfer this issue to the qsharp-language repo, since that's now the proper place for language suggestions.

PR #597, opened 2020-09-01
Call Graph Design Changes

ScottCarda-MS:
Major changes include:
1. Two kinds of call graphs now: SimpleCallGraph, and ConcreteCallGraph
2. Call Graph nodes are based on callables instead of specializations
3. GetAllDependencies and associated tests are removed


src/QsCompiler/Tests.Compiler/CallGraphTests.fs
```diff
@@ -76,19 +57,14 @@ type CallGraphTests (output:ITestOutputHelper) =
...
+        compilationDataStructures.BuiltCompilation
 
-    let CompileCycleValidationTest testNumber =
-        let graph = CompileTest testNumber "CycleValidation.qs"
+    let BuildTrimmedGraph (compilation : QsCompilation) = SimpleCallGraph(compilation, true)
 
-        graph.GetCallCycles()
-        |> Seq.map (fun x -> GetEdges graph x) |> Seq.collect (fun x -> CartesianProduct x)
+    let CompileCycleDetectionTest testNumber =
+        let SimpleCallGraph = CompileTest testNumber "CycleDetection.qs" |> SimpleCallGraph
```
samarsha, line 66: Are there also tests for `ConcreteCallGraph`? I didn't see them.

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
```diff
@@ -675,33 +387,3 @@ type CallGraphTests (output:ITestOutputHelper) =
...
-
-    [<Fact>]
-    [<Trait("Category","Cycle Validation")>]
-    member this.``Cycle with Mutated Forwarding`` () =
-        CompileCycleValidationTest 5 |> AssertInvalidCycleExists
-
-    [<Fact>]
-    [<Trait("Category","Cycle Validation")>]
-    member this.``Cycle with Multiple Concrete Resolutions`` () =
-        CompileCycleValidationTest 6 |> AssertValidCycles
```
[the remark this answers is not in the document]
ScottCarda-MS: Yep, they are part of that PR now.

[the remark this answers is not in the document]
ScottCarda-MS: #621

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
ScottCarda-MS: There are no tests specifically for concrete call graph. There will be a PR similar to #621 for the use case of the concrete call graph in monomorphization that will test the concrete call graph.

src/QsCompiler/Tests.Compiler/TestCases/TypeParameterResolution.qs
```diff
@@ -1,165 +1,6 @@
...
-
-    operation Foo<'A, 'B>(a : 'A, b : 'B) : Unit {
-        Bar(a);
-        Baz(b);
-    }
-
-    operation Bar<'X>(x : 'X) : Unit { }
-
-    operation Baz<'Y>(y : 'Y) : Unit { }
-}
```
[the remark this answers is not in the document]
ScottCarda-MS: No, these tests were testing GetAllDependencies's ability to combine type parameter resolutions across multiple edges, which is a feature that has been dropped and is not present in GetDirectDependencies.

src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -649,6 +650,7 @@ type DiagnosticItem =
             | ErrorCode.AttributeInvalidOnCallable                -> "Invalid attribute placement. The attribute {0} cannot be attached to a callable declaration."
             | ErrorCode.UnresolvedTypeParameterForRecursiveCall   -> "The type argument(s) for the recursive call could not be inferred. Please provide explicit type arguments, e.g. Op<Int, Double>(arg)."
             | ErrorCode.TypeParameterResConflictWithTypeArgument  -> "The type of the expression needs to match the defined type argument. Expecting an expression of type {0}."
+            | ErrorCode.InvalidCyclicTypeParameterResolution      -> "Call cycle results in ambiguous or conflicting type parameter resolution."
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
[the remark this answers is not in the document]
samarsha: Until #621 is merged, could you add a TODO comment about re-adding the cycle tests?

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
samarsha: Could you add a TODO comment for concrete call graph tests?

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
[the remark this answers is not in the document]
ScottCarda-MS: done

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
ScottCarda-MS: I'm not at this time foreseeing any tests specific to the concrete call graph. All the tests for the call graph work in general for call graphs, and the tests for monomorphization will be sufficient to test the specifics of the concrete graph.

src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -1,47 +1,95 @@
...
-    using ResolvedTypeKind = QsTypeKind<ResolvedType, UserDefinedType, QsTypeParameter, CallableInformation>;
-    // Type Parameters are frequently referenced by the callable of the type parameter followed by the name of the specific type parameter.
-    using TypeParameterName = Tuple<QsQualifiedName, NonNullable<string>>;
-    using TypeParameterResolutions = ImmutableDictionary</*TypeParameterName*/ Tuple<QsQualifiedName, NonNullable<string>>, ResolvedType>;
+    using TypeParameterResolutions = ImmutableDictionary<Tuple<QsQualifiedName, NonNullable<string>>, ResolvedType>;
+
+    /// <summary>
+    /// Base class for call graph edge types.
+    /// </summary>
+    public abstract class CallGraphEdgeBase : IEquatable<CallGraphEdgeBase>
```
samarsha, line 21: There's a lot happening in this file that makes it a bit hard for me to follow: several top-level classes from two different call graphs plus the base call graph, and several nested classes. Could this be split out into more than one file? The base call graph, simple call graph, and concrete call graph seem like they could go in 3 separate files.

ScottCarda-MS: The three call graphs are now in there own separate files.

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
samarsha: It's probably OK if it's only tested via monomorphization, but a TODO mentioning that the concrete call graph is currently not tested until then would be nice.

src/QsCompiler/Transformations/CallGraph/CallGraphBase.cs
```diff
@@ -0,0 +1,255 @@
...
+            }
+
+            this.FromCallableName = fromCallableName;
+            this.ToCallableName = toCallableName;
+            this.ReferenceRange = referenceRange;
+        }
+
+        /// <summary>
+        /// Determines if the object is the same as the given edge, ignoring the
+        /// ordering of key-value pairs in the type parameter dictionaries.
```
samarsha, line 62: This comment is referring to a type parameter dictionary that the base edges don't have.

src/QsCompiler/Transformations/CallGraph/CallGraphBase.cs
```diff
@@ -0,0 +1,255 @@
...
+        public bool Equals(CallGraphNodeBase other) =>
+            this.CallableName.Equals(other.CallableName);
+
+        /// <inheritdoc/>
+        public override int GetHashCode()
+        {
+            HashCode hash = default;
+            hash.Add(this.CallableName);
+            return hash.ToHashCode();
+        }
```
samarsha, line 110:
```suggestion
        public override int GetHashCode() => this.CallableName.GetHashCode();
```

src/QsCompiler/Transformations/CallGraph/CallGraphBase.cs
```diff
@@ -0,0 +1,255 @@
...
+    public abstract class CallGraphBase<TNode, TEdge>
+        where TNode : CallGraphNodeBase
+        where TEdge : CallGraphEdgeBase
+    {
+        // Static Elements
+
+        /// <summary>
+        /// Returns an empty dependency for a node.
+        /// </summary>
+        protected static ILookup<TNode, TEdge> EmptyDependency() =>
```
samarsha, line 125: This doesn't seem to be used outside the class:

```suggestion
        private static ILookup<TNode, TEdge> EmptyDependency() =>
```

src/QsCompiler/Transformations/CallGraph/CallGraphBase.cs
```diff
@@ -0,0 +1,255 @@
...
+            ImmutableArray<KeyValuePair<TNode, TEdge>>.Empty
+            .ToLookup(kvp => kvp.Key, kvp => kvp.Value);
+
+        // Member Fields
+
+        /// <summary>
+        /// This is a dictionary mapping source nodes to information about target nodes. This information is represented
+        /// by a dictionary mapping target node to the edges pointing from the source node to the target node.
+        /// </summary>
+        protected readonly Dictionary<TNode, Dictionary<TNode, ImmutableArray<TEdge>>> dependencies =
```
samarsha, line 135: Seems like it might be cleaner to make this private and only expose it via `GetDirectDependencies`. Otherwise, this field is effectively public, which exposes an implementation detail about the call graph that I'm not sure was intended.

src/QsCompiler/Transformations/CallGraph/CallGraphBase.cs
```diff
@@ -0,0 +1,255 @@
...
+                return EmptyDependency();
+            }
+        }
+
+        /// <summary>
+        /// Adds a dependency to the call graph using the two nodes and the edge between them.
+        /// The nodes are added to the graph if they are not already there. The edge is always added.
+        /// Throws ArgumentNullException if any of the arguments are null.
+        /// </summary>
+        protected void AddDependency(TNode fromNode, TNode toNode, TEdge edge)
```
samarsha, line 198: Is the call graph intended to be immutable or mutable?

ScottCarda-MS: I was trying to do the best of both worlds: I need parts of it to be mutable while I'm building it up with the walkers, but after it is built, it should be immutable. That is why the properties were immutable datatypes. Was there something specific to AddDependency regarding this, or is this just a general comment for the call graph?

ScottCarda-MS: Oh I see how it is related. It adds dependencies to the graph 😁!

src/QsCompiler/Transformations/CallGraph/CallGraphBase.cs
ScottCarda-MS: I especially like this idea! 😃

src/QsCompiler/Transformations/CallGraph/CallGraphBase.cs
```diff
@@ -0,0 +1,255 @@
...
+
+        /// <summary>
+        /// Name of the callable being referenced.
+        /// </summary>
+        public QsQualifiedName ToCallableName { get; }
+
+        /// <summary>
+        /// The range of the reference represented by the edge.
+        /// </summary>
+        public DataTypes.Range ReferenceRange { get; }
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/QsCompiler/Transformations/CallGraph/CallGraphBase.cs
ScottCarda-MS: Removed

src/QsCompiler/Transformations/CallGraph/CallGraphBase.cs
ScottCarda-MS: done

ScottCarda-MS:
The commit that addressed the PR comments were accidentally made on another branch. I will try and merge the changes over to this branch.

src/QsCompiler/Transformations/CallGraph/CallGraphBase.cs
ScottCarda-MS: The access level for this could be set to `private protected` which would mean it was only accessible in this assembly and only to its derived classes. But I'm not sure this is the thing to do, as anyone who wants to create their own call graph class derived from this class would need some way to being able to build that graph, which would be this method.

samarsha: You could use the builder pattern, where all the members that mutate the graph are private, and can only be accessed through a nested builder class that returns an immutable graph after you're done building it.

ScottCarda-MS: I guess that is kind of what the walkers are (the builder class), except they are not nested, and nesting them could become pretty messy.

samarsha: I just had in mind only exposing `AddDependency` and `AddNode` in the builder, not adding any additional logic. I guess they wouldn't need to be private methods - they could probably directly be public methods in the builder class.

ScottCarda-MS: The builder classes would have as methods on them the AddDependency and AddNode public methods, and they would have access to the private mutable fields of the call graph by virtue of the builder being a nested class. The walkers would then use the methods of the builder class to build the call graph. Is that correct? I'm not really sure what this does other than put all the mutating methods into a public nested class.

ScottCarda-MS: Ah so when you instantiate the builder class object, you have access to the mutator methods. But when you instantiate the regular call graph class, you don't have access to those anymore, so it is immutable, right?

samarsha: Yes, the builder would have a method that returns an immutable graph based on what has been added to the builder. The builder couldn't be used to mutate the returned graph. That enforces that the graph can only be mutated while it is being built.

samarsha (changes requested):
I think it'd be good to resolve the immutability issues before merging, see the discussion above.

src/QsCompiler/Transformations/CallGraph/CallGraphBase.cs
ScottCarda-MS: Removed the CallGraphBase as an inherited type and changed it to a call graph builder class. The Simple and Concrete call graph types have a private builder member that is used during construction.

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
bettinaheim: The monomorphization tests I believe don't current capture type argument information, do they? In any case, a separate set of tests for the additional information in the concrete call graph is warranted here.

ScottCarda-MS: I've added a ToDo for the concrete call graph tests. I don't want to add tests for it in this PR to keep it from getting any bigger, but I will look into some more specific tests for it in the PR that is specific to the concrete call graph.

samarsha (dismissed):
Thanks for addressing the immutability issue - that part looks good now. I have some additional comments about decreasing the size of the public API.

src/QsCompiler/Transformations/CallGraph/SimpleCallGraph.cs
```diff
@@ -0,0 +1,470 @@
...
+        /// other type parameters of the same callable (constricting resolutions)
+        /// or to a type containing a nested reference to the same type parameter,
+        /// i.e Foo.A -> Foo.A[].
+        /// Returns an enumerable of tuples for each edge of each invalid cycle found,
+        /// each tuple containing a diagnostic and the callable name where the diagnostic
+        /// should be placed.
+        /// </summary>
+        public IEnumerable<Tuple<QsCompilerDiagnostic, QsQualifiedName>> VerifyAllCycles()
+        {
+            var diagnostics = new List<Tuple<QsCompilerDiagnostic, QsQualifiedName>>();
```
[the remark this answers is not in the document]
ScottCarda-MS: I think in general in this code base it is standard to use System.Tuple because it has better interoperability with F#. This code in particular does not interact with F#, but as a matter of consistency I'd like to keep it as a System.Tuple unless there is a good reason to use the value tuple.

src/QsCompiler/Transformations/CallGraph/CallGraphBuilder.cs
```diff
@@ -0,0 +1,235 @@
...
+    {
+        /// <summary>
+        /// Name of the callable where the reference was made.
+        /// </summary>
+        public QsQualifiedName FromCallableName { get; }
+
+        /// <summary>
+        /// Name of the callable being referenced.
+        /// </summary>
+        public QsQualifiedName ToCallableName { get; }
```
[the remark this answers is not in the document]
ScottCarda-MS: We should maybe address this one in the PR that makes use of the VerifyAllCycles (#621) to see if we can remove the need to embed the FromCallableName and ToCallableName from the edge. Let's move this discussion there.

src/QsCompiler/Transformations/CallGraph/ConcreteCallGraph.cs
```diff
@@ -0,0 +1,144 @@
...
+
+        /// <summary>
+        /// The number of nodes in the call graph.
+        /// </summary>
+        public int Count => this.graphBuilder.Count;
+
+        /// <summary>
+        /// A hash set of the nodes in the call graph.
+        /// </summary>
+        public ImmutableHashSet<ConcreteCallGraphNode> Nodes => this.graphBuilder.Nodes;
```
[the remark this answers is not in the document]
samarsha: Discussed offline: `Nodes` is useful for finding which concrete implementations are part of the graph, but I think `Count` can be removed.

src/QsCompiler/Transformations/CallGraph/SimpleCallGraph.cs
```diff
@@ -0,0 +1,470 @@
...
+
+        /// <summary>
+        /// The number of nodes in the call graph.
+        /// </summary>
+        public int Count => this.graphBuilder.Count;
+
+        /// <summary>
+        /// A hash set of the nodes in the call graph.
+        /// </summary>
+        public ImmutableHashSet<SimpleCallGraphNode> Nodes => this.graphBuilder.Nodes;
```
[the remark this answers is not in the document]
samarsha: Discussed offline: `Nodes` is useful for determining which nodes were kept in the trimmed call graph, but I think `Count` can be removed.

src/QsCompiler/Transformations/CallGraph/CallGraphBuilder.cs
```diff
@@ -0,0 +1,235 @@
...
+            }
+
+            this.CallableName = callableName;
+        }
+
+        /// <inheritdoc/>
+        public override bool Equals(object obj)
+        {
+            return obj is CallGraphNodeBase && this.Equals((CallGraphNodeBase)obj);
+        }
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/QsCompiler/Transformations/CallGraph/ConcreteCallGraph.cs
```diff
@@ -0,0 +1,144 @@
...
+            this.ParamResolutions = paramResolutions.ToImmutableDictionary(
+                kvp => kvp.Key,
+                kvp => StripPositionInfo.Apply(kvp.Value));
+        }
+
+        /// <inheritdoc/>
+        public override bool Equals(object obj)
+        {
+            return obj is ConcreteCallGraphNode && this.Equals((ConcreteCallGraphNode)obj);
+        }
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/QsCompiler/Transformations/CallGraph/SimpleCallGraph.cs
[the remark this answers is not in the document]
samarsha: Yes, F# tuples use `System.Tuple`, but the standard tuple in C# is `System.ValueTuple`, which also has nice built-in syntax. That is used throughout the compiler codebase as well as `System.Tuple` when needed for F# interop.

src/QsCompiler/Transformations/CallGraph/CallGraphBuilder.cs
```diff
@@ -0,0 +1,235 @@
...
+        /// </summary>
+        private readonly Dictionary<TNode, Dictionary<TNode, ImmutableArray<TEdge>>> dependencies =
+            new Dictionary<TNode, Dictionary<TNode, ImmutableArray<TEdge>>>();
+
+        // Properties
+
+        /// <summary>
+        /// The number of nodes in the call graph.
+        /// </summary>
+        public int Count => this.dependencies.Count;
```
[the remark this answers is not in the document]
ScottCarda-MS: yes, removed

src/QsCompiler/Transformations/CallGraph/CallGraphBuilder.cs
```diff
@@ -0,0 +1,235 @@
...
+                return EmptyDependency();
+            }
+        }
+
+        /// <summary>
+        /// Adds a dependency to the call graph using the two nodes and the edge between them.
+        /// The nodes are added to the graph if they are not already there. The edge is always added.
+        /// Throws ArgumentNullException if any of the arguments are null.
+        /// </summary>
+        internal void AddDependency(TNode fromNode, TNode toNode, TEdge edge)
```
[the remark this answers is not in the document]
ScottCarda-MS: all members are public now

src/QsCompiler/Transformations/CallGraph/SimpleCallGraph.cs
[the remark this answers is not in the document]
ScottCarda-MS: Okay so in your experience with the code base System.Tuple is only used in the C# when necessary for F# interop? Alright then I can go for that. Value tuple _does_ have an awfully nice syntax. 😊

src/QsCompiler/Transformations/CallGraph/SimpleCallGraph.cs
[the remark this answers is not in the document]
ScottCarda-MS: Count has been removed.

src/QsCompiler/Transformations/CallGraph/ConcreteCallGraph.cs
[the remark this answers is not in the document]
ScottCarda-MS: Count has been removed.

src/QsCompiler/Transformations/CallGraph/SimpleCallGraph.cs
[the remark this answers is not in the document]
ScottCarda-MS: changed

src/QsCompiler/Transformations/CallGraph/CallGraphBuilder.cs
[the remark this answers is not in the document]
samarsha: I believe these properties were just added in this PR, is that correct? If so, I would prefer to address it in this PR. The idea that I had in mind would not require changing the signature of `VerifyAllCycles`. My idea was to change `GetEdges` to return `IEnumerable<IEnumerable<(SimpleCallGraphEdge, QsQualifiedName)>>` instead of simply `IEnumerable<IEnumerable<SimpleCallGraphEdge>>`.

src/QsCompiler/Transformations/CallGraph/SimpleCallGraph.cs
```diff
@@ -0,0 +1,465 @@
...
+    /// <summary>
+    /// A kind of call graph whose nodes represent Q# callables.
+    /// </summary>
+    public sealed class SimpleCallGraph
+    {
+        // Static Elements
+
+        private static IEnumerable<IEnumerable<SimpleCallGraphEdge>> CartesianProduct(IEnumerable<IEnumerable<SimpleCallGraphEdge>> sequences)
+        {
+            IEnumerable<IEnumerable<SimpleCallGraphEdge>> result = new[] { Enumerable.Empty<SimpleCallGraphEdge>() };
```
samarsha, line 84: You could make this polymorphic since it doesn't need anything from the edge:

```suggestion
        private static IEnumerable<IEnumerable<T>> CartesianProduct<T>(IEnumerable<IEnumerable<T>> sequences)
        {
            IEnumerable<IEnumerable<T>> result = new[] { Enumerable.Empty<T>() };
```

src/QsCompiler/Transformations/CallGraph/SimpleCallGraph.cs
```diff
@@ -0,0 +1,465 @@
...
+            foreach (var sequence in sequences)
+            {
+                result = sequence.SelectMany(item => result, (item, seq) => seq.Concat(new[] { item })).ToList();
```
samarsha, line 87: Is the `ToList()` necessary?

src/QsCompiler/Transformations/CallGraph/SimpleCallGraph.cs
```diff
@@ -0,0 +1,465 @@
...
+        public IEnumerable<(QsCompilerDiagnostic, QsQualifiedName)> VerifyAllCycles()
+        {
+            var diagnostics = new List<(QsCompilerDiagnostic, QsQualifiedName)>();
+
+            if (this.Nodes.Any())
+            {
+                var cycles = this.GetCallCycles().SelectMany(x => CartesianProduct(this.GetEdges(x).Reverse()));
+                foreach (var cycle in cycles)
+                {
+                    var combination = new TypeResolutionCombination(cycle.Select(edge => edge.ParamResolutions).ToArray());
```
samarsha, line 173: It's a little awkward that you need to call `ToArray()`, but `TypeResolutionCombination` calls `Where` which only requires an `IEnumerable`. Maybe `TypeResolutionCombination` could take an `IEnumerable` instead?

src/QsCompiler/Transformations/CallGraph/CallGraphWalker.cs
```diff
@@ -0,0 +1,507 @@
...
+                    {
+                        this.SharedState.AddDependency(global.Item);
+                    }
+
+                    return ExpressionKind.InvalidExpr;
+                }
+            }
+        }
+
+        private static class ConcreteCallGraphWalker
```
samarsha, line 261: Similar to `SimpleCallGraph` and `ConcreteCallGraph`, it might be cleaner to have the simple and concrete walkers in separate files as well.

src/QsCompiler/Transformations/CallGraph/CallGraphWalker.cs
```diff
@@ -0,0 +1,507 @@
...
+            where TNode : CallGraphNodeBase
+            where TEdge : CallGraphEdgeBase
+        {
+            public abstract class TransformationState
+            {
+                internal TNode CurrentNode;
+                internal readonly TGraph Graph;
+
+                // The type parameter resolutions of the current expression.
+                internal IEnumerable<TypeParameterResolutions> ExpTypeParamResolutions = new List<TypeParameterResolutions>();
```
samarsha, line 61: Maybe `Expr` instead of `Exp` is a more common abbreviation of expression:

```suggestion
                internal IEnumerable<TypeParameterResolutions> ExprTypeParamResolutions = new List<TypeParameterResolutions>();
```

src/QsCompiler/Transformations/CallGraph/CallGraphWalker.cs
```diff
@@ -0,0 +1,507 @@
...
+                {
+                }
+
+                public override QsStatement OnStatement(QsStatement stm)
+                {
+                    this.SharedState.CurrentStatementOffset = stm.Location.IsValue
+                        ? QsNullable<Position>.NewValue(stm.Location.Item.Offset)
+                        : QsNullable<Position>.Null;
+                    return base.OnStatement(stm);
+                }
```
samarsha, line 91: This looks like a recurring need in transformations: to keep track of the offset of the last seen statement in order to add it to an expression range. Capability inference needed to write a statement transformation for this as well. Maybe there is some common infrastructure we could make for this? @bettinaheim

src/QsCompiler/Transformations/CallGraph/CallGraphWalker.cs
```diff
@@ -0,0 +1,507 @@
...
+
+                    if (ex.TypeParameterResolutions.Any())
+                    {
+                        this.SharedState.ExpTypeParamResolutions = this.SharedState.ExpTypeParamResolutions.Prepend(ex.TypeParameterResolutions);
+                    }
+                    var rtrn = base.OnTypedExpression(ex);
+
+                    this.SharedState.CurrentExpressionRange = contextRange;
+
+                    return rtrn;
```
samarsha, line 114: For consistency with the other transformations below:

```suggestion
                    }

                    var result = base.OnTypedExpression(ex);
                    this.SharedState.CurrentExpressionRange = contextRange;
                    return result;
```

ScottCarda-MS: done

src/QsCompiler/Transformations/CallGraph/CallGraphWalker.cs
ScottCarda-MS: done

src/QsCompiler/Transformations/CallGraph/SimpleCallGraph.cs
ScottCarda-MS: done

src/QsCompiler/Transformations/CallGraph/SimpleCallGraph.cs
ScottCarda-MS: removed

src/QsCompiler/Transformations/CallGraph/SimpleCallGraph.cs
ScottCarda-MS: done

src/QsCompiler/Transformations/CallGraph/CallGraphWalker.cs
ScottCarda-MS: Here they are private members of a static class. I don't want to have to expose them.

src/QsCompiler/Transformations/CallGraph/CallGraphBuilder.cs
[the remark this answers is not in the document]
ScottCarda-MS: done

src/QsCompiler/Transformations/CallGraph/CallGraphBuilder.cs
```diff
@@ -0,0 +1,205 @@
...
+            {
+                throw new ArgumentException(nameof(callableName));
+            }
+
+            this.CallableName = callableName;
+        }
+
+        /// <inheritdoc/>
+        public override bool Equals(object obj) =>
+            obj is CallGraphNodeBase && this.Equals((CallGraphNodeBase)obj);
```
samarsha, line 70:
```suggestion
            obj is CallGraphNodeBase node && this.Equals(node);
```

src/QsCompiler/Transformations/CallGraph/ConcreteCallGraph.cs
```diff
@@ -0,0 +1,137 @@
...
+
+            // Remove position info from type parameter resolutions
+            this.ParamResolutions = paramResolutions.ToImmutableDictionary(
+                kvp => kvp.Key,
+                kvp => StripPositionInfo.Apply(kvp.Value));
+        }
+
+        /// <inheritdoc/>
+        public override bool Equals(object obj) =>
+            obj is ConcreteCallGraphNode && this.Equals((ConcreteCallGraphNode)obj);
```
samarsha, line 70:
```suggestion
            obj is ConcreteCallGraphNode node && this.Equals(node);
```

src/QsCompiler/Transformations/CallGraph/CallGraphWalker.cs
samarsha: They would only be exposed internally at most. The reason I'm asking is because there are 15 classes in this file at varying levels of nesting which makes it a little hard to follow the logic.

src/QsCompiler/Transformations/CallGraph/ConcreteCallGraph.cs
ScottCarda-MS: done

src/QsCompiler/Transformations/CallGraph/CallGraphBuilder.cs
ScottCarda-MS: done

src/QsCompiler/Transformations/CallGraph/CallGraphWalker.cs
ScottCarda-MS: done

src/QsCompiler/Transformations/CallGraph/CallGraphWalker.cs
bettinaheim: Yes, that would likely make sense. Not as part of this change though.

src/QsCompiler/Transformations/CallGraph/CallGraphWalkerBase.cs
```diff
@@ -0,0 +1,87 @@
...
+                }
+
+                public override TypedExpression OnTypedExpression(TypedExpression ex)
+                {
+                    var contextRange = this.SharedState.CurrentExpressionRange;
+                    this.SharedState.CurrentExpressionRange = ex.Range;
+
+                    if (ex.TypeParameterResolutions.Any())
+                    {
+                        this.SharedState.ExprTypeParamResolutions = this.SharedState.ExprTypeParamResolutions.Prepend(ex.TypeParameterResolutions);
```
samarsha, line 78: I might be misunderstanding this, but shouldn't the resolutions be removed from the list after this expression is processed, similar to how `CurrentExpressionRange` is reset, otherwise the list could carry over resolutions from unrelated expressions?

src/QsCompiler/Transformations/CallGraph/ConcreteCallGraphWalker.cs
```diff
@@ -0,0 +1,275 @@
...
+                    });
+
+                    if (!specArgMatches.Any(m => m.Item1 >= 0))
+                    {
+                        throw new ArgumentException($"Could not find a suitable {request.Kind} specialization for {request.CallableName}");
+                    }
+
+                    relevantSpecs = specArgMatches
+                        .OrderByDescending(match => match.Item1)
+                        .Select(match => match.spec);
```
samarsha, line 132: What is the ranking for - shouldn't there be one specialization that the call could refer to, rather than using a best match?

src/QsCompiler/Transformations/CallGraph/CallGraphWalkerBase.cs
ScottCarda-MS: The ExprTypeParamResolutions are supposed to accumulate resolutions from the top of a nested expression hierarchy to the bottom where an identifier will be found. At the identifier, the list is cleared after it is used. See the AddDependencies method of the other walker classes.

src/QsCompiler/Transformations/CallGraph/ConcreteCallGraphWalker.cs
ScottCarda-MS: Let's discuss that offline.

samarsha: Discussed offline: This is to rank possible overload resolutions based on the specializations `TypeArguments` property, currently a placeholder since #29 is not implemented.

src/QsCompiler/Transformations/CallGraph/CallGraphWalkerBase.cs
samarsha: Could you give an example of an expression where resolutions propagate down to a (global) identifier? I am wondering what happens if there is no identifier at the bottom (do the resolutions not get cleared?), or if there is more than one identifier adjacent to each other in the same expression.

ScottCarda-MS: Any expression where not all of the generic type parameters are not resolved on the identifier directly would provide an example of this. The syntax tree would be considered malformed and would have thrown an error before this if there were type parameter resolutions in an expression did not have an identifier at one of the leaves. If there are multiple identifiers in the expression hierarchy, the resolutions for the a given identifier would not be encountered until all the identifiers to the left are already processed, which is the order that the expression hierarchy is parsed by the transformation.

ScottCarda-MS: So your typical generic call:
```
Foo<'A>(a : 'A) : Unit { }
...
Bar() : Unit {
    Foo(12.0);
}
```
The call to Foo needs to carry the type parameter resolution from the call-like expression `Foo(12.0)` to the identifier sub-expression `Foo`.

An example with multiple parameters:
```
Foo<'A, 'B>(a : 'A, b : 'B) : Unit { }
...
Bar() : Unit {
    Foo<Int, _>(3, 12.0);
}
```
Here the B parameter is resolved to Double by the call-like expression `Foo<Int, _>(3, 12.0)` while the A parameter is resolved to Int by the identifier sub-expression `Foo<Int, _>`.

src/QsCompiler/Transformations/CallGraph/ConcreteCallGraphWalker.cs
```diff
@@ -0,0 +1,275 @@
...
+                    {
+                        if (typeParam is QsLocalSymbol.ValidName name)
+                        {
+                            if (request.ParamResolutions.TryGetValue(Tuple.Create(callable.FullName, name.Item), out var res))
+                            {
+                                return res;
+                            }
+                            else
+                            {
+                                throw new ArgumentException($"Couldn't resolve all type parameters for {callable.FullName}");
```
samarsha, line 85: I was able to get this exception to throw on a Q# program that compiles without errors:

```qsharp
function Foo<'a>(x : 'a) : Unit { }

function Bar<'a>(x : 'a) : 'a {
    return x;
}

@EntryPoint()
function Main() : Unit {
    Foo(Bar);
}
```

> System.ArgumentException: Couldn't resolve all type parameters for Graph.Bar

But I feel like this program shouldn't compile, actually...

samarsha: Simpler example:

```qsharp
function Foo<'a>() : Unit { }

@EntryPoint()
function Main() : Unit {
    let x = Foo();
}
```

> System.ArgumentException: Couldn't resolve all type parameters for Graph.Foo

src/QsCompiler/Transformations/CallGraph/ConcreteCallGraphWalker.cs
```diff
@@ -0,0 +1,275 @@
...
+                        else
+                        {
+                            throw new ArgumentException("Encountered invalid type parameter name during call graph construction");
+                        }
+                    }).ToImmutableArray());
+                    typeArgs = SpecializationBundleProperties.BundleId(typeArgs);
+                }
+                else if (relevantSpecs.Count() != 1)
+                {
+                    throw new ArgumentException($"Missing specialization {request.Kind} for {request.CallableName}");
```
samarsha, line 97: With this program:

```qsharp
function Foo() : Unit { }

@EntryPoint()
function Main() : Unit {
    let f = Foo;
    f();
}
```

I get:

> System.ArgumentException: Missing specialization QsControlledAdjoint for Graph.Foo

src/QsCompiler/Transformations/CallGraph/ConcreteCallGraphWalker.cs
```diff
@@ -0,0 +1,275 @@
...
+                        {
+                            return (-1, spec);
+                        }
+                        var matches = specTypeArgs.Where((tArg, idx) => !tArg.Resolution.IsMissingType && tArg.Resolution.Equals(typeArgs.Item[idx])).Count();
+                        return (matches, spec);
+                    });
+
+                    if (!specArgMatches.Any(m => m.Item1 >= 0))
+                    {
+                        throw new ArgumentException($"Could not find a suitable {request.Kind} specialization for {request.CallableName}");
```
samarsha, line 127: With this program:

```qsharp
function Foo<'a>(x : 'a) : 'a {
    return x;
}

function Bar<'a>() : ('a -> 'a) {
    return Foo<'a>;
}

@EntryPoint()
function Main() : Unit {
    (Bar<Unit>())();
}
```

I get:

> System.ArgumentException: Could not find a suitable QsControlledAdjoint specialization for Graph.Foo

src/QsCompiler/Transformations/CallGraph/ConcreteCallGraphWalker.cs
samarsha: Here's an example where the type parameter actually does resolve to something concrete (`Int` and `Int -> Int`), but the concrete call graph still throws "Couldn't resolve all type parameters":

```qsharp
function Foo<'a>(x : 'a) : 'a {
    return x;
}

@EntryPoint()
function Main() : Unit {
    let x = (Foo(Foo))(0);
}
```

ScottCarda-MS: My understanding is that none of these scenarios, even the last one, should be supported. It is odd to me that the call graph error is surfacing though, and maybe an issue should be filed to better protected against these cases. Either way that would be outside the scope of this PR.

src/QsCompiler/Transformations/CallGraph/ConcreteCallGraphWalker.cs
ScottCarda-MS: I see the issue. When referencing a callable without calling it, the logic is overly cautious and assumes a link for all functor specializations. Since Foo does not have a body for all functor specializations, this causes an error. What it should do is assume a link for all functor specializations _defined for that callable_. I believe I can get this information from the callable, but I'll take a closer look in a bit.

src/QsCompiler/Transformations/CallGraph/ConcreteCallGraphWalker.cs
ScottCarda-MS: This is actually the same error as above. It just gets thrown in a different place because the reference is generic. Let's consolidate this conversation on the other thread.

src/QsCompiler/Transformations/CallGraph/ConcreteCallGraphWalker.cs
bettinaheim: The first two are reasonable to forbid (they are both the same issue as you mention, namely that the return type of the callable is fully determined even when the callable is not concrete) - @samarsha Do you mind filing a bug? 
The last one however is perfectly possible to infer what the concrete callable(s) is/are. The third case I believe is fully supported throughout both the compiler and runtime, and we should keep it that way. However, the changes we are going to do to support [this task](https://github.com/microsoft/qsharp-compiler/issues/573) should actually make it easier to support this in the call graph. I am hence ok with adding a test for the third case that is currently skipped.

samarsha: > they are both the same issue as you mention, namely that the return type of the callable is fully determined even when the callable is not concrete

Is this the problem? There are useful functions with this property, for example `Length (a : 'T[]) : Int`.

ScottCarda-MS: Ah right the last scenario is something we are looking into. I have added a unit test for that scenario that is, for now, skipped.

ScottCarda-MS: It could be very useful to have a centralized place, like a design doc somewhere, to point people toward for information about generics in Q#, having information about what is currently supported, what we plan to support, what we have decided not to support, as well as having information about what scenarios are covered with type argument inference in Q#. @bettinaheim do we currently have something like this?

src/QsCompiler/Transformations/CallGraph/ConcreteCallGraphWalker.cs
ScottCarda-MS: I've added some unit tests to cover these scenarios. They currently don't pass because I'm still working on the bug.

ScottCarda-MS: fixed

src/QsCompiler/Transformations/CallGraph/ConcreteCallGraphWalker.cs
samarsha: The first one is probably OK to forbid as having an ambiguous type variable, since `Bar` is assigned to a parameter without its type being determined. I don't think we should forbid the second one, though, but maybe unused type variables should be removed from the AST before monomorphization.

samarsha (approved):

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
```diff
@@ -543,6 +268,139 @@ type CallGraphTests (output:ITestOutputHelper) =
 
         AssertNotInGraph graph "Bar"
 
+    [<Fact>]
+    [<Trait("Category","Populate Call Graph")>]
+    member this.``Concrete Graph has Concretizations`` () =
+        let graph = CompileTypeParameterResolutionTestWithExe 9 |> ConcreteCallGraph
+
+        let makeNode name resType =
+            let qalifiedName = { Namespace = NonNullable<_>.New Signatures.TypeParameterResolutionNS; Name = NonNullable<_>.New name }
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/QsCompiler/Transformations/CallGraph/ConcreteCallGraphWalker.cs
```diff
@@ -0,0 +1,288 @@
...
+                    this.Types = new TypeTransformation<TransformationState>(this, TransformationOptions.Disabled);
+                }
+            }
+
+            private class TransformationState : CallGraphWalkerBase<ConcreteGraphBuilder, ConcreteCallGraphNode, ConcreteCallGraphEdge>.TransformationState
+            {
+                public bool IsInCall = false;
+                public bool HasAdjointDependency = false;
+                public bool HasControlledDependency = false;
+                public Func<QsQualifiedName, IEnumerable<QsSpecializationKind>> GetSpecializationKinds = (_) => new QsSpecializationKind[] { };
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
```diff
@@ -543,6 +268,139 @@ type CallGraphTests (output:ITestOutputHelper) =
 
         AssertNotInGraph graph "Bar"
 
+    [<Fact>]
+    [<Trait("Category","Populate Call Graph")>]
+    member this.``Concrete Graph has Concretizations`` () =
+        let graph = CompileTypeParameterResolutionTestWithExe 9 |> ConcreteCallGraph
+
+        let makeNode name resType =
```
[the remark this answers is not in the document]
ScottCarda-MS: The tests now use a global MakeNode function.

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
```diff
@@ -543,6 +268,139 @@ type CallGraphTests (output:ITestOutputHelper) =
...
+        AssertNotInConcreteGraph graph Unused
+
+    [<Fact(Skip="Double reference resolution is not yet supported")>]
+    [<Trait("Category","Populate Call Graph")>]
+    member this.``Concrete Graph Double Reference Resolution`` () =
+        let graph = CompileTypeParameterResolutionTestWithExe 11 |> ConcreteCallGraph
+
+        let makeNode name resType =
+            let qalifiedName = { Namespace = NonNullable<_>.New Signatures.TypeParameterResolutionNS; Name = NonNullable<_>.New name }
+            let res = dict[(qalifiedName, NonNullable<_>.New "A"), ResolvedType.New resType].ToImmutableDictionary()
```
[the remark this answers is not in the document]
ScottCarda-MS: For consistency with the other tests, I've update the test to use uppercase "A" for the type parameter name.

src/QsCompiler/Transformations/CallGraph/ConcreteCallGraphWalker.cs
ScottCarda-MS: Removing unused type parameters before monomorphization could be a solution. However, let's take this discussion elsewhere as the details of this aren't in scope for this PR. @samarsha please file a github issue for this generics behavior you'd like to address, and we can continue this discussion there.

PR #600, opened 2020-09-02
Verify runtime capabilities of callables defined in referenced libraries

### LINKED ISSUE #586: Verify runtime capabilities of callables defined in referenced libraries

samarsha:
This merges `feature/capability-inference` to `main` (via a user branch).

TODO:

- [x] Infer capabilities in libraries based on local source code (#593)
- [x] Merge call graph into the capability inference branch (#598)
- [x] Infer capabilities in libraries based on call dependencies (#601)
- [x] Restrict inference to callables defined in the current library (#601)
- [x] Verify capabilities in executables based on previously inferred capability attributes (#620)
- [x] Merge the call graph to `main` and rebase this branch
- [x] Define the `Capability` attribute in the standard library (microsoft/qsharp-runtime#372)

Closes #586.


samarsha:
Build failing on qsharp-runtime:

```
D:\a\1\s\submodules\qsharp-runtime\src\Simulation\Simulators.Tests\TestProjects\IonQExe\MeasurementSupportTests.qs(26,13): error QS5027: The callable Reset requires runtime capabilities that are not supported by the target IonQProcessor. [D:\a\1\s\submodules\qsharp-runtime\src\Simulation\Simulators.Tests\TestProjects\IonQExe\IonQExe.csproj]
D:\a\1\s\submodules\qsharp-runtime\src\Simulation\Simulators.Tests\TestProjects\IonQExe\MeasurementSupportTests.qs(16,13): error QS5027: The callable Reset requires runtime capabilities that are not supported by the target IonQProcessor. [D:\a\1\s\submodules\qsharp-runtime\src\Simulation\Simulators.Tests\TestProjects\IonQExe\IonQExe.csproj]
D:\a\1\s\submodules\qsharp-runtime\src\Simulation\Simulators.Tests\TestProjects\IonQExe\MeasurementSupportTests.qs(15,13): error QS5027: The callable Reset requires runtime capabilities that are not supported by the target IonQProcessor. [D:\a\1\s\submodules\qsharp-runtime\src\Simulation\Simulators.Tests\TestProjects\IonQExe\IonQExe.csproj]
C:\Users\VssAdministrator\.nuget\packages\microsoft.quantum.sdk\0.12.200914858-pull\Sdk\Sdk.targets(118,5): error MSB3073: The command "dotnet "C:\Users\VssAdministrator\.nuget\packages\microsoft.quantum.sdk\0.12.200914858-pull\DefaultItems\../tools/qsc/qsc.dll" build --format MsBuild -v Normal --response-files obj/qsharp/config/qsc.rsp" exited with code -3. [D:\a\1\s\submodules\qsharp-runtime\src\Simulation\Simulators.Tests\TestProjects\IonQExe\IonQExe.csproj]
C:\Users\VssAdministrator\.nuget\packages\microsoft.quantum.sdk\0.12.200914858-pull\Sdk\Sdk.targets(123,5): error : Q# compiler failed with exit code: -3. [D:\a\1\s\submodules\qsharp-runtime\src\Simulation\Simulators.Tests\TestProjects\IonQExe\IonQExe.csproj]
```

This is actually working as intended, because `Reset` is implemented in QsharpCore as

```qsharp
    operation Reset (target : Qubit) : Unit {
        
        if (M(target) == One)
        {
            X(target);
        }
    }
```

which is not supported on IonQ. I suspect IonQ has a custom implementation of reset that does not use a measurement branch. It might need a capability override, or for the IonQExe test to use the IonQ-specific `Reset` implementation.

samarsha:
After adding a `@Capability("QPRGen0")` override to `Reset`, the QDK pipeline passes: https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=15395&view=results

samarsha:
The end-to-end build will fail until microsoft/qsharp-runtime#372 is merged to main, but the compiler-only build succeeds.

src/QsCompiler/CompilationManager/TypeChecking.cs
```diff
@@ -859,11 +859,7 @@ internal static void ImportGlobalSymbols(this FileContentManager file, Compilati
                     proceed = nodes.MoveNext();
                 }
 
-                var (ifStatement, ifDiagnostics) =
-                    Statements.NewIfStatement(context, ifBlock.Item1, ifBlock.Item2, elifBlocks, elseBlock);
-                statement = ifStatement;
```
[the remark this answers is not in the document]
samarsha: All capability diagnostics are generated for the whole scope in [`BuildUserDefinedImplementation`](https://github.com/microsoft/qsharp-compiler/blob/1336c21a08856b998d50f738a708a646f808c9a5/src/QsCompiler/CompilationManager/TypeChecking.cs#L1290).

src/QsCompiler/Compiler/PluginInterface.cs
```diff
@@ -12,10 +12,17 @@
...
+    /// Lists the priorities for built-in rewrite steps. Steps with a larger priority number have higher priority and
+    /// will be executed first.
     /// </summary>
     public static class RewriteStepPriorities
     {
+        /// <summary>
+        /// Priority of the built-in transformation that infers the minimum runtime capabilities required by each
+        /// callable.
+        /// </summary>
+        public const int CapabilityInference = 1200;
```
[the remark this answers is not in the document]
samarsha: It was originally last, but I think I moved it up because it was not possible to run capability inference on generated code due to bug #634. Since that bug has been fixed, I'll try moving it back again.

src/QsCompiler/SyntaxProcessor/CapabilityInference.fs
```diff
@@ -0,0 +1,376 @@
...
+            cache.[callable.FullName] <- capability
+            capability)
+
+    cachedCapability Set.empty
+
+/// Adds the given capability to the callable as an attribute.
+let private addAttribute callable capability =
+    let arg = capability.ToString () |> AttributeUtils.StringArgument
+    let attribute = AttributeUtils.BuildAttribute (BuiltIn.Capability.FullName, arg)
+    { callable with QsCallable.Attributes = callable.Attributes.Add attribute }
```
bettinaheim, line 359: An AddAttribute exists on QsCallable.

src/QsCompiler/SyntaxProcessor/ScopeTools.fs
```diff
@@ -285,19 +285,23 @@ type SymbolTracker(globals : NamespaceManager, sourceFile, parent : QsQualifiedN
...
+    { /// The namespace manager for global symbols.
+      Globals : NamespaceManager
+
+      /// The symbol tracker for the parent callable.
       Symbols : SymbolTracker
+
       /// True if the parent callable for the current scope is an operation.
       IsInOperation : bool
-      /// True if the current expression is contained within the condition of an if- or elif-statement.
-      IsInIfCondition : bool
```
bettinaheim, line 293: It would be good to formally deprecate things when reasonable. How much unused code do you need to keep to properly deprecate this? And along with it WithinIfCondition.

src/QsCompiler/Core/Dependencies.fs
```diff
@@ -94,6 +110,11 @@ type BuiltIn = {
         Kind = Attribute
     }
 
+    static member Capability = {
+        FullName = {Name = "Capability" |> NonNullable<string>.New; Namespace = BuiltIn.CoreNamespace}
```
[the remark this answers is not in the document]
samarsha: Based on our conversation offline, the name was changed to `RequiresCapability`.

src/QsCompiler/Compiler/RewriteSteps/CapabilityInference.cs
```diff
@@ -0,0 +1,42 @@
...
+        public IEnumerable<IRewriteStep.Diagnostic> GeneratedDiagnostics { get; } =
+            Enumerable.Empty<IRewriteStep.Diagnostic>();
+
+        public bool ImplementsPreconditionVerification { get; } = false;
+
+        public bool ImplementsTransformation { get; } = true;
+
+        public bool ImplementsPostconditionVerification { get; } = false;
+
+        public bool PreconditionVerification(QsCompilation compilation) => throw new NotSupportedException();
```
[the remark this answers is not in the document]
samarsha: It's not valid for a Q# library to have an execution target, right?

src/QsCompiler/SyntaxProcessor/ScopeTools.fs
samarsha: Not much unused code. Re-added and marked obsolete.

src/QsCompiler/Compiler/RewriteSteps/CapabilityInference.cs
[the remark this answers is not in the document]
bettinaheim: Yes, that is correct.

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -527,7 +529,7 @@ public CompilationLoader(SourceLoader loadSources, ReferenceLoader loadReference
 
             if (this.config.IsExecutable && this.CompilationOutput?.EntryPoints.Length == 0)
             {
-                if (this.config.RuntimeCapabilities == RuntimeCapabilities.Unknown)
+                if (this.config.RuntimeCapabilities == RuntimeCapabilities.FullComputation)
```
bettinaheim, line 532: I think it might make sense to distinguish the capabilities for a library from the capabilities of an executable targeted for execution on a simulator (we will e.g. monomorphize the executable but not the library). How about keeping Unknown for the library and having FullComputation be for an executable targeted for a backend that supports arbitrary computations?

src/QsCompiler/Compiler/RewriteSteps/CapabilityInference.cs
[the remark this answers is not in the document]
samarsha: Shouldn't rewrite steps be able to assume certain correctness properties about the compilation, such as that a library compilation can't have an execution target, rather than needing to check for it explicitly?

src/QsCompiler/Core/Dependencies.fs
```diff
@@ -66,6 +71,21 @@ type BuiltIn = {
         | Value tId -> tId.Namespace.Value = GeneratedAttributes.Namespace && tId.Name.Value = GeneratedAttributes.LoadedViaTestNameInsteadOf
         | Null -> false
 
+    /// Returns the required runtime capability if the attribute is a valid instance of the RequiresCapability
+    /// attribute.
+    static member RequiredCapability attribute =
```
bettinaheim, line 76: There are already a couple of very similar methods at the beginning of the SymbolResolution module. It would make sense to follow the same pattern for everything; following that pattern, there would be a static member MarksRequiredCapability and then a separate TryGetRequiredCapability that takes a sequence of attributes. Probably it would make sense to split out both of these for all attributes that are recognized by the compiler into a separate module. That is an API change and we need to properly deprecate it, but seeing as the list of attributes that require similar methods grows, I think that would be a good change to make.

src/QsCompiler/Compiler/RewriteSteps/CapabilityInference.cs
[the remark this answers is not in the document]
bettinaheim: The rewrite steps should certainly assume that the QsCompilation object is valid, and I think also assuming that the assembly constants passed by the compiler are accurate seems fair. But that is different from relying on knowledge about the relation that assembly constants should have (you did ask the question about whether or not a Q# library can be targeted for example). 
Aside from that, you may be right in that the capability inference should run even if it is a targeted compilation. So I am good with keeping it as is.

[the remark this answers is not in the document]
samarsha: I can see it being useful as a sanity test. I added this as a precondition, and also that the `RequiresCapability` attribute must be defined in the compilation.

src/QsCompiler/Compiler/CompilationLoader.cs
samarsha: In that case, I would prefer a name like `Any`, since libraries can be executed on any target, rather than on a particular target that we don't know about.

samarsha: > we will e.g. monomorphize the executable but not the library

There is already a flag to indicate whether something is an executable or a library, though? Why does it need to be reflected in the runtime capabilities too?

src/QsCompiler/Core/Dependencies.fs
samarsha: I can see the advantage of changing this method to take in a sequence of attributes rather than a single attribute. I do not see the benefit of a separate `MarksRequiredCapability` method, though. I don't think it would be used anywhere.

bettinaheim: The benefit would be to stick with the same pattern for everything, and I think in terms of API predictability there is value to that. The idea is to have roughly the same methods for all compiler recognized attributes. I don't think that all Marks* methods could be eliminated (that would be the other option, but I think the C# generation might also be using one of these).

src/QsCompiler/Compiler/CompilationLoader.cs
bettinaheim: I guess it just feels a bit awkward to me to rely on certain combinations meaning specific things. Just in terms of what the compilation looks like, there is a difference between something that hasn't been targeted yet and something that happens to be targeted to a backend that supports everything (everything here may exclude e.g. type parameterizations). But you are right in that the same information may be captured without that. The idea of "finishing" compilation of a partially compiled executable has come up in the past, but we also have no immediate need or plans to support that. I am ok with leaving it as is.

src/QsCompiler/Core/Dependencies.fs
samarsha: I added `TryGetRequiredCapability`.

>The idea is to have roughly the same methods for all compiler recognized attributes. I don't think that all Marks* methods could be eliminated (that would be the other option, but I think the C# generation might also be using one of these).

The `Marks*` methods are written in a very repetitive way. Why not define one function that takes in a `BuiltIn` (or a `QsQualifiedName`) and an attribute and returns whether they match? That would work for all attributes, instead of needing one per attribute.

src/QsCompiler/CommandLineTool/Options.cs
```diff
@@ -72,6 +72,13 @@ public class CompilationOptions : Options
             HelpText = "Specifies the classical capabilites of the runtime. Determines what QIR profile to compile to.")]
         public AssemblyConstants.RuntimeCapabilities RuntimeCapabilites { get; set; }
 
+        internal RuntimeCapability RuntimeCapability => this.RuntimeCapabilites switch
+        {
+            AssemblyConstants.RuntimeCapabilities.QPRGen0 => RuntimeCapability.BasicQuantumFunctionality,
+            AssemblyConstants.RuntimeCapabilities.QPRGen1 => RuntimeCapability.BasicMeasurementFeedback,
+            _ => RuntimeCapability.FullComputation
+        };
```
samarsha, line 80: I was not able to update the capability type in the command-line parser, since the new type is not an enum (CommandLineParser has special parsing for enums). I think this will take a little more time to parse the custom type with proper error handling.

src/QsCompiler/Compiler/CompilationLoader.cs
samarsha: Thinking about this a little more, I'm not opposed to having an `Unknown` capability, but it would take me a little bit of time to make sure I used it correctly (since not all FullComputation references should be replaced with Unknown). Do you still want to do that in this PR, or think about adding it later?

src/QsCompiler/CommandLineTool/Options.cs
bettinaheim: I'd suggest to just go with string and use the TryParse method that you added to the type. But I agree to make this a separate PR.

bettinaheim (approved):

src/QsCompiler/Core/Dependencies.fs
bettinaheim: I created a task to revise the API methods for compiler recognized attributes. Right now it is not very predictable.

src/QsCompiler/Compiler/CompilationLoader.cs
bettinaheim: Since we will already want a second PR to update the command line compiler to process the new capability names, let's add this to that PR instead.

PR #601, opened 2020-09-02
Infer runtime capabilities using callable dependencies

samarsha:
Updates capability inference to transitively consider the dependencies of a callable (callables that the callable references or calls). 

Dependencies are only followed within the source files of the current compilation unit. Once a library callable is encountered, it either needs to have been previously compiled with a `Capability` attribute, or its capability will default to level 0.

Part of PR #600.


src/QsCompiler/SyntaxProcessor/CapabilityInference.fs
```diff
@@ -193,48 +196,85 @@ let private scopePatterns scope = scope.Statements |> Seq.collect statementPatte
...
-    { callable with Attributes = callable.Attributes.Add attribute }
+    { callable with QsCallable.Attributes = callable.Attributes.Add attribute }
 
 /// Infers the capability of all callables in the compilation, adding the built-in Capability attribute to each
 /// callable.
-///
-/// TODO:
-/// This infers the capability of each callable individually, without looking at the capabilities of anything it
-/// calls or references. The inferred capability is only a lower bound.
 let InferCapabilities compilation =
```
cesarzc, line 267: How is this meant to be used once the capability inference feature is completed? For any given Q# project, will this be called for the generated QsCompilation?

samarsha: `InferCapabilities` is called automatically by a rewrite step for Q# libraries. It's the only place that I intend to call it from.

https://github.com/microsoft/qsharp-compiler/blob/370f6bf5052af25b28291905be4c7077bc6e62eb/src/QsCompiler/Compiler/RewriteSteps/CapabilityInference.cs#L34-L38

cesarzc: Thanks Sarah! Have you tried compiling the standard library with these changes using the `--perf` option? It would be good to double check that `OverallCompilation.RewriteSteps` does not take much more time after the change.

src/QsCompiler/SyntaxProcessor/CapabilityInference.fs
```diff
@@ -193,48 +196,85 @@ let private scopePatterns scope = scope.Statements |> Seq.collect statementPatte
...
     let transformation = SyntaxTreeTransformation ()
     transformation.Namespaces <- {
         new NamespaceTransformation (transformation) with
             override this.OnCallableDeclaration callable =
-                if callable.Attributes |> QsNullable<_>.Choose BuiltIn.GetCapability |> Seq.isEmpty
-                then addAttribute callable
+                let isMissingCapability =
+                    callable.Attributes |> QsNullable<_>.Choose BuiltIn.GetCapability |> Seq.isEmpty
+                if isMissingCapability && callable.SourceFile.Value.EndsWith ".qs"
+                then callableDependentCapability callables graph Set.empty callable |> addAttribute callable
```
ScottCarda-MS, line 277: If you could keep track of the visited callables as well as what capability is for visited callables here, you could check for the "new" version of callables, or at least to see the capability of visited callables. This would prevent the transformation from having to look down the same lineages of dependencies multiple times.

ScottCarda-MS: Perhaps instead of a set of QsQualified names, the data type could be a map of the visited names to the capability found for each.

src/QsCompiler/SyntaxProcessor/CapabilityInference.fs
samarsha: Thanks for the suggestion. Here is a run with capability inference enabled:

```
{
  "OverallCompilation": 58794,
  "OverallCompilation.Build": 38276,
  "OverallCompilation.OutputGeneration": 3332,
  "OverallCompilation.OutputGeneration.BinaryGeneration": 30,
  "OverallCompilation.OutputGeneration.DocumentationGeneration": 722,
  "OverallCompilation.OutputGeneration.SyntaxTreeSerialization": 2580,
  "OverallCompilation.ReferenceLoading": 2297,
  "OverallCompilation.RewriteSteps": 14782,
  "OverallCompilation.SourcesLoading": 73
}
```

and one with capability inference disabled:

```
{
  "OverallCompilation": 50164,
  "OverallCompilation.Build": 38713,
  "OverallCompilation.OutputGeneration": 3505,
  "OverallCompilation.OutputGeneration.BinaryGeneration": 26,
  "OverallCompilation.OutputGeneration.DocumentationGeneration": 733,
  "OverallCompilation.OutputGeneration.SyntaxTreeSerialization": 2745,
  "OverallCompilation.ReferenceLoading": 2008,
  "OverallCompilation.RewriteSteps": 5822,
  "OverallCompilation.SourcesLoading": 79
}
```

So, it does have a significant impact on the rewrite step time. That is expected since this algorithm is not very efficient and is intended to be replaced with a more efficient one later (since the more efficient algorithm I had in mind requires changes to the call graph). But I can try the ideas @ScottCarda-MSFT mentioned for increasing the performance to see if that works without changing the call graph.

samarsha: I updated the algorithm. This one is much faster. Here is a new run with capability inference enabled:

```
{
  "OverallCompilation": 53029,
  "OverallCompilation.Build": 40176,
  "OverallCompilation.OutputGeneration": 3792,
  "OverallCompilation.OutputGeneration.BinaryGeneration": 27,
  "OverallCompilation.OutputGeneration.DocumentationGeneration": 798,
  "OverallCompilation.OutputGeneration.SyntaxTreeSerialization": 2965,
  "OverallCompilation.ReferenceLoading": 2390,
  "OverallCompilation.RewriteSteps": 6535,
  "OverallCompilation.SourcesLoading": 81
}
```

So it's adding about 700ms to the rewrite steps (very roughly - I didn't bother to average over multiple runs).

cesarzc: It would be worth trying @ScottCarda-MS idea. Even though the current algorithm is not very efficient, making it a bit better would be great.

src/QsCompiler/SyntaxProcessor/CapabilityInference.fs
samarsha: It needs to add a name to `visited` at the start of the function, before the capability has been computed, to prevent infinite loops with cycles, so I don't think `visited` can be turned into a map. However, I did add a separate map to store computed capabilities. As a result of that I also needed to handle cycles separately. The new algorithm should avoid recomputing capabilities.

src/QsCompiler/SyntaxProcessor/CapabilityInference.fs
samarsha: Yes, I did update the algorithm based on @ScottCarda-MS's suggestion - I think it's asymptotically optimal now.

cesarzc: That's great! It looks much better now. Thanks for making the change.

ScottCarda-MS (approved):

cesarzc (approved):

PR #605, opened 2020-09-03
Improve generated documentation for deprecated callables and types

### LINKED ISSUE #592: Preserve API documentation on deprecated operations
Currently we remove the API documentation as soon as we mark the operation deprecated, so its page in the docs only says "use the new operation instead". A person who sees the deprecation notice and wants to switch the code to use the new operation might not remember exactly the details of what the old operation did, and it's fairly hard to dig them out at this point, especially if the new operation has different set of parameters. 

For example, [DrawRandomInt](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.random.drawrandomint) replaces [RandomIntPow2](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.math.randomintpow2) but requires figuring out the range of numbers generated.

It would be nice to keep API documentation for deprecated operations/functions/UDTs until they are removed completely. 

One can figure out the necessary parameters by diving into the libraries source code (in this example getting to the [source file with RandomIntPow2 definition](https://github.com/microsoft/QuantumLibraries/blob/26ae8b963f7f92d46c818b49c28105a22cac5be7/Standard/src/Math/Random.qs#L10) will allow to figure out the necessary parameters), but going to the libraries code requires fairly in-depth familiarity with our repos, which I'm afraid not all Q# users have.

samarsha:
* Use the short summary for items shown on the namespace page (`YamlMime:QSharpNamespace`).
* Preserve the long documentation string on the full page of a deprecated callable or type (`YamlMime:QSharpType`).

[Here is what it looks like on the docs site](https://review.docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.math.randomint?branch=refs%2Ftags%2Fbeta%2Fv0.12.2010.16021).

Resolves #592.


bettinaheim (approved):
Lgtm. I assume you checked the docs page?

samarsha:
Yes, an example of the rendered docs page is in the PR description.

PR #621, opened 2020-09-11
Cycle validation

ScottCarda-MS:
Adds cycle validation diagnostics to the type checking of the compilation. This will return design-time errors for cycles that are not valid.


samarsha:
Does this replace #588?

ScottCarda-MS:
> Does this replace #588?

Yah, this replaces that PR.

src/QsCompiler/CompilationManager/TypeChecking.cs
```diff
@@ -2015,6 +2015,16 @@ QsCallable GetCallable((QsQualifiedName, ImmutableArray<QsSpecialization>) specI
                 }
                 compilation.UpdateCallables(callables);
                 compilation.UpdateTypes(types);
+
+                var allCallableDeclarations = compilation.GlobalSymbols.DefinedCallables().ToImmutableDictionary(x => x.QualifiedName);
```
samarsha, line 2019: Instead of creating this dictionary, can you use `compilation.GetCallables()` instead?

src/QsCompiler/CompilationManager/TypeChecking.cs
```diff
@@ -2015,6 +2015,16 @@ QsCallable GetCallable((QsQualifiedName, ImmutableArray<QsSpecialization>) specI
...
+
+                var callGraph = new SimpleCallGraph(compilation.GetCallables().Values);
+                foreach (var (diag, parent) in callGraph.VerifyAllCycles())
+                {
+                    var info = allCallableDeclarations[parent];
+                    var offset = info.Position is DeclarationHeader.Offset.Defined pos ? pos.Item : null;
+                    diagnostics.Add(Diagnostics.Generate(info.SourceFile.Value, diag, offset));
+                }
```
samarsha, line 2027: Since this method is already pretty long can we split the cycle diagnostics out into a separate method?

src/QsCompiler/CompilationManager/TypeChecking.cs
```diff
@@ -2207,6 +2217,15 @@ internal static void UpdateTypeChecking(this FileContentManager file, Compilatio
                     : contentTokens);
 
                 diagnostics = QsCompilerError.RaiseOnFailure(() => RunTypeChecking(compilation, declarationTrees, CancellationToken.None), "error on running type checking");
+
+                var numRemovedCycleDiags = file.CurrentSemanticDiagnostics().Count(m => m.Code == ErrorCode.InvalidCyclicTypeParameterResolution.Code())
+                    - diagnostics.Count(m => m.Code == ErrorCode.InvalidCyclicTypeParameterResolution.Code());
```
samarsha, line 2222: Is it possible that some cycle diagnostics are removed and some are added since the last check, such that the difference is 0 and global type checking is not triggered?

src/QsCompiler/CompilationManager/TypeChecking.cs
```diff
@@ -2207,6 +2217,15 @@ internal static void UpdateTypeChecking(this FileContentManager file, Compilatio
...
+
+                if (numRemovedCycleDiags > 0 || diagnostics.Any(x => x.Source != file.FileName.Value))
+                {
+                    file.TriggerGlobalTypeChecking();
+                }
```
samarsha, line 2227: Similarly to the code added to `RunTypeChecking`, it could be nice if this was factored out into a separate function since this one is already fairly long.

src/QsCompiler/Transformations/CallGraph/SimpleCallGraph.cs
```diff
@@ -84,7 +84,7 @@ private static IEnumerable<IEnumerable<T>> CartesianProduct<T>(IEnumerable<IEnum
             IEnumerable<IEnumerable<T>> result = new[] { Enumerable.Empty<T>() };
             foreach (var sequence in sequences)
             {
-                result = sequence.SelectMany(item => result, (item, seq) => seq.Concat(new[] { item }));
+                result = sequence.SelectMany(item => result, (item, seq) => seq.Concat(new[] { item })).ToList();
```
samarsha, line 87: Why is `ToList()` needed?

ScottCarda-MS: The SelectMany is lazy, and since the thing being assigned, `result`, is used in the selector as well, the code get confused and fails if `result` is not forced to be evaluated after each iteration.

src/QsCompiler/CompilationManager/TypeChecking.cs
ScottCarda-MS: The info we want is the position and source file of each callable, which we are getting from the declaration headers. We might be able to get this off the callables if those fields are defined on them. I'll look into it.

src/QsCompiler/CompilationManager/TypeChecking.cs
ScottCarda-MS: I'm not able to produce this situation. You could toggle between a local invalid cycle and a separate global invalid cycle with a single character change, but even then the behavior is what I'd expect. The errors correctly switch between the local cycle and the global cycle.

ScottCarda-MS: I will admit this seems like a bit of a hacky way to check for this though. I'd appreciate any ideas you might have on how to make this check more robust.

samarsha: > You could toggle between a local invalid cycle and a separate global invalid cycle with a single character change

I think the language server can receive changes longer than one character (e.g. with copy and paste).

src/QsCompiler/CompilationManager/TypeChecking.cs
ScottCarda-MS: Looks like that works well.

src/QsCompiler/CompilationManager/TypeChecking.cs
ScottCarda-MS: Sure, but the effect is still the same: toggling between a local invalid cycle and a global invalid cycle.

src/QsCompiler/CompilationManager/TypeChecking.cs
ScottCarda-MS: done

src/QsCompiler/CompilationManager/TypeChecking.cs
ScottCarda-MS: done

src/QsCompiler/CompilationManager/TypeChecking.cs
ScottCarda-MS: Actually now I seem to be triggering an issue sometimes. It seems to be somewhat random if the errors get cleared or not.

ScottCarda-MS: fixed

src/QsCompiler/CompilationManager/TypeChecking.cs
```diff
@@ -2222,5 +2236,16 @@ internal static void UpdateTypeChecking(this FileContentManager file, Compilatio
                 compilation.ExitWriteLock();
             }
         }
+
+        private static void CheckForGlobalCycleChange(FileContentManager file, List<Diagnostic> diagnostics)
+        {
+            var numRemovedCycleDiags = file.CurrentSemanticDiagnostics().Count(m => m.Code == ErrorCode.InvalidCyclicTypeParameterResolution.Code())
+                - diagnostics.Count(m => m.Code == ErrorCode.InvalidCyclicTypeParameterResolution.Code());
+
+            if (numRemovedCycleDiags != 0 || diagnostics.Any(x => x.Source != file.FileName.Value))
```
samarsha, line 2245: If `numRemovedCycleDiags` can be negative, I think it'd be better to call it something like `numCycleDiagnosticsChange`.

src/QsCompiler/CompilationManager/TypeChecking.cs
```diff
@@ -2222,5 +2236,16 @@ internal static void UpdateTypeChecking(this FileContentManager file, Compilatio
...
+            {
+                file.TriggerGlobalTypeChecking();
+            }
+        }
```
samarsha, line 2249: I noticed a minor bug. When I have a cycle across two files:

File1.qs
```qsharp
namespace Example {
    operation Foo1<'A>() : Unit {
        Bar<'A>();
    }

    operation Bar<'A>() : Unit {
        Foo2<'A[]>();
    }
}
```

File2.qs
```qsharp
namespace Example {
    operation Foo2<'A>() : Unit {
        Bar<'A>();
    }
}
```

Then saving either of the files (e.g. with Ctrl+S) will cause a red squiggly to flicker in for a split second before disappearing. In the first file, the squiggly is under the call to `Bar`. In the second file, the squiggly is at the last character in the file.

src/QsCompiler/Transformations/CallGraph/SimpleCallGraph.cs
samarsha: Could you add a comment about that since it doesn't seem obvious to me from the code?

src/QsCompiler/CompilationManager/TypeChecking.cs
ScottCarda-MS: I've change the code to always trigger the global type checking. Can you try again?

samarsha: It seems like it still happens.

src/QsCompiler/Transformations/CallGraph/SimpleCallGraph.cs
ScottCarda-MS: done

src/QsCompiler/CompilationManager/TypeChecking.cs
ScottCarda-MS: done

src/QsCompiler/CompilationManager/TypeChecking.cs
ScottCarda-MS: The issue seems to have been fixed by not keeping errors for callables that are not fully loaded into what the editor considers.

samarsha: Thanks, it seems like it's fixed now.

samarsha (approved):

PR #622, opened 2020-09-12
Renaming master -> main

bettinaheim:
There were a couple of master left


bettinaheim:
Ech, the build doesn't like my branch name. I'll fix that later. For now using a different branch name.

PR #624, opened 2020-09-12
Add Edges Property

ScottCarda-MS:
The call graph doesn't have enough information on its public properties to be able to fully define it. To fix that I added an Edges property to match the Nodes property.

### DIFF
--- a/src/QsCompiler/Transformations/CallGraph.cs
+++ b/src/QsCompiler/Transformations/CallGraph.cs
@@ -262,7 +262,7 @@ namespace Microsoft.Quantum.QsCompiler.DependencyAnalysis
         // Member Fields
 
         /// <summary>
-        /// This is a dictionary mapping source nodes and target node tuples to the list
+        /// This is a dictionary mapping source node and target node tuples to the list
         /// of edges going from the source node to the target node.
         /// </summary>
         private readonly Dictionary<Tuple<TNode, TNode>, List<TEdge>> edges =
--- a/src/QsCompiler/Transformations/CallGraph.cs
+++ b/src/QsCompiler/Transformations/CallGraph.cs
@@ -255,46 +255,46 @@ namespace Microsoft.Quantum.QsCompiler.DependencyAnalysis
         /// <summary>
         /// Returns an empty dependency for a node.
         /// </summary>
-        protected static ILookup<TNode, TEdge> EmptyDependency() =>
+        private static ILookup<TNode, TEdge> EmptyDependency() =>
             ImmutableArray<KeyValuePair<TNode, TEdge>>.Empty
             .ToLookup(kvp => kvp.Key, kvp => kvp.Value);
 
         // Member Fields
 
         /// <summary>
-        /// This is a dictionary mapping source nodes to information about target nodes. This information is represented
-        /// by a dictionary mapping target node to the edges pointing from the source node to the target node.
+        /// This is a dictionary mapping source nodes and target node tuples to the list
+        /// of edges going from the source node to the target node.
         /// </summary>
-        protected readonly Dictionary<TNode, Dictionary<TNode, ImmutableArray<TEdge>>> dependencies =
-            new Dictionary<TNode, Dictionary<TNode, ImmutableArray<TEdge>>>();
+        private readonly Dictionary<Tuple<TNode, TNode>, List<TEdge>> edges =
+            new Dictionary<Tuple<TNode, TNode>, List<TEdge>>();
+
+        /// <summary>
+        /// The hash set of nodes in the call graph.
+        /// </summary>
+        private readonly HashSet<TNode> nodes = new HashSet<TNode>();
 
         // Properties
 
         /// <summary>
         /// The number of nodes in the call graph.
         /// </summary>
-        public int Count => this.dependencies.Count;
+        public int Count => this.nodes.Count;
 
         /// <summary>
-        /// A hash set of the nodes in the call graph.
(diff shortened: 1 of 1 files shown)


src/QsCompiler/Transformations/CallGraph.cs
```diff
@@ -255,46 +255,46 @@ public abstract class BaseCallGraph<TNode, TEdge>
...
+        /// A lookup for the edges in the call graph. Given a tuple of source node
+        /// and target node, returns the list of edges going from the source node
+        /// to the target node.
         /// </summary>
-        public ImmutableHashSet<TNode> Nodes => this.dependencies.Keys.ToImmutableHashSet();
-
-        // Member Methods
+        public ILookup<Tuple<TNode, TNode>, TEdge> Edges => this.edges
+            .SelectMany(kvp => kvp.Value.ToImmutableArray(), Tuple.Create)
+            .ToLookup(x => x.Item1.Key, x => x.Item2);
```
samarsha, line 290: > The call graph doesn't have enough information on its public properties to be able to fully define it.

Could you explain what is not fully defined? It looks like `Edges` has the same information as `GetDirectDependencies`.

ScottCarda-MS: It is similar. The main difference is that Edges is a property while GetDirectDependencies is a method. Properties are preserved in serialization while methods are not, so if we ever wanted to serialized/deserialize a call graph, this would be needed to reconstruct the graph.

samarsha: The method does not need to be serialized, just the `dependencies` field. It isn't public, but non-public fields can be serialized as well.

samarsha: What do you think of a different type for `Edges`?

For example, `ILookup<TNode, ILookup<TNode, TEdge>>` would be the same interface as `GetDirectDependencies`, so `GetDirectDependencies` could be removed.

Or a simpler type could be `ILookup<TNode, TEdge>`, so the equivalent to `GetDirectDependencies` would be `Edges[x].ToLookup(e => e.ToCallableName)`, I think.

ScottCarda-MS: `ILookup<TNode, ILookup<TNode, TEdge>>` would mean that each node maps to a list of lookups, which isn't quite right. Maybe something like `ILookup<TNode, Tuple<TNode, IEnumerable<TEdge>>>`, but I think `ILookup<Tuple<TNode, TNode>, TEdge> is cleaner and more efficient.

This typing also lets me easily get both the incoming and outgoing edges of a given node:
outgoing edges: `this.Edges.Where(g => g.Key.Item1.Equals(node)).Select(g => g.ToImmutableArray());`
incoming edges: `this.Edges.Where(g => g.Key.Item2.Equals(node)).Select(g => g.ToImmutableArray());`

ScottCarda-MS: The type `ILookup<TNode, TEdge>` is close to what I want, but doesn't easily let me query for incoming edges.

ScottCarda-MS: Perhaps we could have two properties, IncomingEdges and OutgoingEdges, both of type `ILookup<TNode, TEdge>`, but that seems like it would have a lot of redundant information between the two.

samarsha: Sorry, I meant `IReadOnlyDictionary<TNode, ILookup<TNode, TEdge>>` for the first one.

samarsha: Wanting symmetry for ingoing and outgoing edges makes sense, but doesn't it mean that `GetDirectDependencies` is O(n) now instead of O(1) before?

ScottCarda-MS: Yah it does slow down the method call. Hmm, do we need to have an optimized backing data structure for each use case? It would be fast, but also have a lot of extra redundant data.

ScottCarda-MS: Oh and I should mention that I was hoping to have all the methods on this use only public properties. This gets us close to Andres's comment about serialization on the design doc PR.

samarsha: I think at least one of the directions should be fast. Capability inference calls `GetDirectDependencies` repeatedly, and I think this change would increase its time complexity from linear to quadratic.

ScottCarda-MS: With the Edges property as `ILookup<Tuple<TNode, TNode>, TEdge>`, you would have the ability to walk backwards through the graph, which I believe was something you had said would speed up your algorithm quite a bit?

samarsha: I would have to think about it a little more to be sure, but I believe the current implementation of capability inference, using a hash map for memoizing previously computed capabilities, is asymptotically optimal. But if it isn't, I don't think `ILookup<Tuple<TNode, TNode>, TEdge>` would be faster, because looking up an outgoing edge is not constant time.

samarsha (changes requested):
(Requesting changes related to the conversation above about time complexity.)

bettinaheim:
@ScottCarda-MS Can this PR be closed?

ScottCarda-MS:
Closing as this we decided not to do this.

PR #626, opened 2020-09-17
Monomorphization with Call Graph

ScottCarda-MS:
Uses the concrete call graph to do monomorphization. The rewrite step is now only responsible for generating the callables based on the nodes in the graph.


ScottCarda-MS:
Will be adding unit tests for the concrete call graph in this PR.

ScottCarda-MS:
Unit tests for the concrete call graph were moved to #597 and are now in the feature branch.

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -568,6 +562,12 @@ public CompilationLoader(SourceLoader loadSources, ReferenceLoader loadReference
                 steps.Add((rewriteStep.Priority, () => this.ExecuteAsAtomicTransformation(rewriteStep, ref this.compilationStatus.PreEvaluation)));
             }
 
+            if (this.config.IsExecutable && !this.config.SkipMonomorphization)
```
cesarzc, line 565: Was this moved after `GenerateFunctorSupport` for some reason?

ScottCarda-MS: Yes Monomorphization is going to be the last rewrite step processed before external rewrite steps now. I guess the order in this file is not super important, as the priority number is actually the thing that determines the order, but having the order here match that I think makes things more consistent.

ScottCarda-MS: The reason Monomorphization has to be processed after functor support is that we don't have a way to deal with auto-generated functor specializations that have not been generated yet when building the concrete call graph, so the solution is to run functor generation first to make sure all of the functor specializations are generated and present.

cesarzc: Thanks for the explanation Scott! Yes, I like the consistency.

cesarzc (approved):
Looks good to me! The call graph really makes this simpler

PR #628, opened 2020-09-18
Enable C# nullable reference types for CommandLineTool project

samarsha:
We're planning to gradually enable null checking across the C# projects. This is the first project to get it.

### DIFF
--- a/src/QsCompiler/CommandLineTool/CommandLineTool.csproj
+++ b/src/QsCompiler/CommandLineTool/CommandLineTool.csproj
@@ -6,6 +6,8 @@
     <AssemblyTitle>Microsoft Q# compiler command line tool.</AssemblyTitle>
     <TargetFramework>netcoreapp3.1</TargetFramework>
     <OutputType>Exe</OutputType>
+    <Nullable>enable</Nullable>
+    <WarningsAsErrors>nullable</WarningsAsErrors>
   </PropertyGroup>
 
   <PropertyGroup>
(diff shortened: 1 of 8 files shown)


src/QsCompiler/CommandLineTool/CommandLineTool.csproj
```diff
@@ -6,6 +6,8 @@
     <AssemblyTitle>Microsoft Q# compiler command line tool.</AssemblyTitle>
     <TargetFramework>netcoreapp3.1</TargetFramework>
     <OutputType>Exe</OutputType>
+    <Nullable>enable</Nullable>
+    <WarningsAsErrors>nullable</WarningsAsErrors>
```
cgranade, line 10: Oh, very nice. Really exciting to see that elevated to an error!

cgranade (approved):

ScottCarda-MS (approved):
Very excited to see this change. Is the plan currently to enable the feature, but keep as many things nullable as possible right now by adding the `?`, then making things non-nullable as we work with the code?

samarsha:
My plan is to make as few things nullable as possible (just enough to fix the build errors) without spending too much time refactoring the code to potentially avoid a nullable.

samarsha:
I believe a test is failing because CommandLineParser has a bug: commandlineparser/commandline#136. It seems like it was fixed last month in Git, but that version is not available on NuGet yet.

PR #631, opened 2020-09-21
Bond schema translator

cesarzc:
This change implements translation between Bond schemas and compiler objects in order to update serialization/deserialization performance.


src/QsCompiler/BondSchemas/CompilerDataStructures.bond
```diff
@@ -0,0 +1,911 @@
...
+}
+
+// Completely represents F#'s QsExpressionKind.
+// N.B. Does not have an equivalent F# type.
+struct QsExpressionKindComposition<TExpression, TSymbol, TType>
+{
+    5: required QsExpressionKind Kind = InvalidExpr;
+
+    // Not null when Kind is Identifier.
+    10: nullable<QsExpressionKindIdentifier<TSymbol, TType>> Identifier = nothing;
```
swernli, line 171: `nullable` usually doesn't need to be used along with `nothing`. See [A Thorough Guide to Bond for C#](https://microsoft.github.io/bond/manual/bond_cs.html#default-value-of-nothing) which is actually not all that thorough, but does have some detail on `nullable` and `nothing`. When you specify that a given property is `nullable` it will already default to null when the type for that property is initialized.

src/QsCompiler/BondSchemas/CompilerDataStructures.bond
```diff
@@ -0,0 +1,911 @@
...
+    5: required bool IsMutable;
+
+    10: required bool HasLocalQuantumDependency;
+}
+
+struct TypedArgument
+{
+    5: required QsQualifiedName Callable;
+
+    10: required string Name;
```
swernli, line 467: You are making frequent use of `required` in a way that I think will not beneficial. In fact, the [Schema Evolution](https://microsoft.github.io/bond/manual/bond_cs.html#schema-evolution) section of the bond guide lists this as a best practice: `required` should be used sparingly and only with careful consideration.

The `required` attribute has two major effects: during deserialization if that field is missing it is treated as a deserialization error, and during serialization it will never be omitted when it is the default value. By contrast, bond fields are optional by default, meaning their absence during deserialization is not a failure and they may be omitted during serialization if they are equal to their default. The biggest place this makes a difference is with compatibility across versions, which is likely less of a concern here, but still with thinking through. I'm not certain this will help or hurt performance, so take this all as me thinking out loud and not necessarily as a recommendation.

Consider this case, a `string Name` property. If it is required, you are saying that the field must be present in order to pass deserialization. But an empty string is still a valid string, since that is the default value. So imagining a struct with one field that is `required string Name;` and deserializing from JSON, this would be considered invalid:
```json
{ }
```
But both of these are actually valid and would deserialize successfully:
```json
{ "Name": "foo" }
```
and 
```json
{ "Name": "" }
```
So `required` doesn't actually provide any additional protection against the empty string; you'd still need to check for it and reject it after deserializing the data. To match this, if your data structure is being serialized and has `Name` equal to empty string, it will serialize into `{ "Name": "" }`.

 If instead the `Name` property were not required, both `{ "Name": "" }` and `{ }` would deserialize into an object with Name equal to the empty string (which your code already has to handle), and would always serialize into `{ }`, saving space in the generated encoding. This tends to be better for both serialized data size and future compatibility.

Sorry, I know that was a long explanation, but hopefully that helps provide context to decide whether all these fields really need to be required or whether most or all of them can be switched to optional.

src/QsCompiler/BondSchemas/CompilerDataStructures.bond
```diff
@@ -0,0 +1,911 @@
...
+    InvalidExpr
+}
+
+// Used to encapsulate data when QsExpressionKind is Identifier.
+// N.B. Does not have an equivalent F# type.
+struct QsExpressionKindIdentifier<TSymbol, TType>
+{
+    5: required TSymbol Symbol;
+
+    10: nullable<vector<TType>> Types = nothing;
```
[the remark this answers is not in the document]
cesarzc: In this case the distinction between an empty vector and the absence of a vector is needed since the purpose is to represent the `QsExpressionKind` as `Identifier` which has the following definition: `Identifier            of 'Symbol * QsNullable<ImmutableArray<'Type>>`. 

The idea is to revise in more detail the compiler data structures to make them more efficient, but as a first step we want to be able to translate back and forth as accurately as possible because we currently don't know what side effects would arise from slight changes like this one.

src/QsCompiler/BondSchemas/CompilerDataStructures.bond
cesarzc: Thanks for pointing this out! I remember reading that part but at some point during my testing I was seeing that if `nothing` is not explicitly set and the field is omitted when building the object, the field is set to a default instance of the object type rather than null. 

I am not sure if this is because of some early set up that I had but I am not seeing the behavior that I described anymore. Having said that, I will remove the explicit sets to null unless I can reproduce the other behavior.

src/QsCompiler/BondSchemas/CompilerDataStructures.bond
cesarzc: Yes, this helps a lot. I will go back and be more careful about using required.

src/QsCompiler/BondSchemas/CompilerDataStructures.bond
cesarzc: All of the redundant `nothing` is now removed from the schemas.

src/QsCompiler/BondSchemas/CompilerDataStructures.bond
swernli: Sounds good! Let me know if there are any specific fields you’d like to discuss the trade-offs for. Reviewing the guidance is making me reevaluate the couple places where I’ve used it, so now I’m going back through those and rethinking them too. There is also a `required_optional` that splits the difference and is used for compatibility to allow gradually converting between the two.

swernli (approved):
Looks good! Thanks for working on this!

src/QsCompiler/BondSchemas/CompilerDataStructures.bond
cesarzc: I ended up removing the `required` from almost all fields. The only ones I left as `required` were the ones that are composite structs in which depending on the kind a different field is the one that contains specific data. I did this because if the kind is really required in those cases.

Thanks for the review! I think the schema is now much more flexible and it aligns with our purpose of making the data structures more efficient without breaking compatibility.

cesarzc:
> 
> 
> Looks good! Thanks for working on this!

Thanks for the review! It really helped to make the bond schema much better.

PR #632, opened 2020-09-21
pin .NET SDK to 3.1.300

filipw:
I noticed that tests fail when using .NET 5.0 SDK with the following errors (excuse my de locale)

> Could not load file or assembly 'System.Runtime, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. Das System kann die angegebene Datei nicht finden. Stellen Sie sicher, dass die <UsingTask>-Deklaration korrekt ist, die Assembly und alle zugehörigen Abhängigkeiten verfügbar sind und die Aufgabe eine öffentliche Klasse enthält, die Microsoft.Build.Framework.ITask implementiert.

I think it makes sense to pin it to 3.1.x. In fact, I noticed the CI build was using SDK to 3.1.300 so I upgraded it to 3.1.401 (the latest one).


global.json
```diff
@@ -0,0 +1,6 @@
+{
+    "sdk": {
+        "version": "3.1.401",
+        "rollForward": "latestPatch"
+    }
+}
```
samarsha, line 6: Looking at the [`global.json` documentation](https://docs.microsoft.com/en-us/dotnet/core/tools/global-json?tabs=netcore3x), it says:

> In general, you want to use the latest version of the SDK tools, so no *global.json* file is needed. In some advanced scenarios, you might want to control the version of the SDK tools, and this article explains how to do this.

Not sure if we fall into the "some advanced scenarios" part (I haven't looked at the failing test). If it's not too hard to fix the tests with .NET 5 in a way that continues to work with 3.1, would that be better?

samarsha:
It looks like the PR builds aren't working when the PR comes from a fork, I'll look into this.

anpaz-msft:
let's try again.

anpaz-msft:
the build should be working now. it's unclear why it's not triggered. @filipw do you mind pushing a new commit to your branch to see if that triggers it?

anpaz-msft (commented):
@filipw, never mind, figured out how to trigger the build with a comment.
As Sara pointed out, though, using a global.json might not be the best option, it would probably be better to figure out a way to fix the tests.
Also, we were having problems with 3.1.401 in the QDK's end-to-end build, which is why we need to stay on 3.1.300 for now for building. At runtime 3.1.401 works fine.

global.json
filipw: I think the failing test is a good example of the types of problems that might arise when the `global.json` is not pinned.

I believe what you want to have here, is a state of the repo that gives you deterministic building and testing conditions. Otherwise contributor A might build with `3.1.100`, contributor B might build with `3.1.401` and contributor C might build with `5.0.100-rc.1` and potentially get subtly (or not so subtly) different behaviors. There are different MSBuild versions in those SDKs (16.3 vs 16.7 vs 16.8) and from what I can say from the experience of building C# tooling they can have quite a lot of behavioral differences. Given that this project relies on a custom project SDK makes it even more important. 

We did the same thing in OmniSharp - where the SDK is pinned to the latest, guaranteeing the contributors are all on the same one https://github.com/OmniSharp/omnisharp-roslyn/blob/master/global.json 

It also gives you clear, unambiguous context: this is the only dotnet SDK version supported for contributing to the repo, and it will be changed to a different one only by a conscious version bump in `global.json`, so you are saving yourself and others time trying to figure out why something fails on a newer or older version.

filipw:
no problem!  I will rebase this once https://github.com/microsoft/qsharp-compiler/pull/648 is merrged

samarsha:
> Also, we were having problems with 3.1.401 in the QDK's end-to-end build, which is why we need to stay on 3.1.300 for now for building. At runtime 3.1.401 works fine.

@anpaz Out of curiosity what was the problem, and is there an ETA for when the QDK pipeline can be updated?

global.json
samarsha: Yeah, it makes sense, I agree that reproducible builds are important.

It looks like when the pinned version is not installed, the build fails rather than automatically downloading the right version, is that correct? That is one difference between pinning the SDK and pinning NuGet/NPM package versions, I think, is needing to manually install the right version rather than having the build automatically download it. That's not a reason not to use global.json necessarily, but just something I thought about.

@anpaz @bettinaheim Do you have any other thoughts on using global.json?

filipw: correct, the .NET SDK behavior is such that when the relevant version is missing, it fails with a message that the SDK is missing.

The behavior can be additionally controlled by `"rollForward"` setting:
 - setting it in `global.json` to `latestPatch` would allow any newer SDK that still fits 3.1.4xx to  be used too (but not e.g. 3.1.500)
 - setting it to `latestFeature` would allow any newer SDK that fits into `3.1.xxx` to be used as well (but not e.g. 5.0.0). 
 - setting it to `major` would allow any newer SDK to be used

It depends on the level of SDK predictability you want to have. For example in OmniSharp we chose to pin the exact version, while e.g. on Roslyn there is the lowest SDK defined and `rollForward` is set to `major`.

build/init.yml
```diff
@@ -27,4 +27,4 @@ steps:
     command: custom
     verbose: false
     customCommand: 'install -g vsce'
- 
\ No newline at end of file
+
```
samarsha, line 30: I think the blank lines were added when @anpaz was trying to re-trigger the CI build.

```suggestion
```

global.json
```diff
@@ -0,0 +1,5 @@
+{
+    "sdk": {
+        "version": "3.1.402"
```
samarsha, line 3: Since the SDK is currently pinned to 3.1.300 in init.yml (and according to @anpaz we can't update that version until QDK pipeline problems are sorted out), can we make that the minimum version here with a roll-forward?

filipw: yes, makes sense

samarsha (approved):
I just saw an issue where an older version of the SDK (3.1.200) had errors building the compiler due to differences in how nullable types were handled, and upgrading to 3.1.402 fixed the issue. Having a global.json would have given a less confusing error message. I'm convinced now that pinning the SDK version is a good idea. Thanks @filipw!

bettinaheim (approved):

PR #635, opened 2020-09-22
Added support for rewrite steps as Types and as IRewriteStep instances

filipw:
This PR is based on the feedback I submitted in Teams. Hopefully it will be useful.
Tagging @bettinaheim 

## Problem

At the moment the only way to provide rewrite steps into the compiler is to use a plugin DLL. The problem is that in this case the compiler arbitrarily discovers all `IRewriteStep` types there, initializes them and uses in the compilation pipeline. This works fine in command line  scenarios, but is severely limiting  when hosting the compiler in own application ("compiler-as-a-service").

## This PR

This PR adds a possibility to pass in `RewriteStepTypes` and `RewriteStepInstances` to the configuration of the `CompilationLoader`. To make things consistent, it also renames the existing `RewriteSteps` option to `RewriteStepAssemblies`, which is technically a breaking change (hopefully not a problem). It is technically not needed, but it makes sense IMHO.

The existing logic of loading assemblies, instantiating types and so on is 100% reused. Since the existing `RewriteSteps.cs` was already very big, adding more stuff to it was probably not the best idea, so I ended up splitting the logic into more (again, IMHO) manageable separate pieces.

## Benefits

Step instances can now be provided by the host application as types or instances, allowing the host to communicate with the step. One example use case would be capturing the rewritten C# code into in-memory representation, ratherthan having it output  into the files on disk.


samarsha:
Thanks for the PR! Just wanted to mention that it looks like the PR builds are broken on our end here as well (same problem as #632), we'll try to get that fixed soon...

filipw:
no problem! I will rebase this once https://github.com/microsoft/qsharp-compiler/pull/648 is merged

filipw:
I rebased on top of the latest `main`, due to changes in #648 there were quite a few conflicts 😀

samarsha:
Sorry about that. :)

src/QsCompiler/Compiler/RewriteSteps/IRewriteStepsLoader.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using System;
+using System.Collections.Immutable;
+using Microsoft.VisualStudio.LanguageServer.Protocol;
+
+namespace Microsoft.Quantum.QsCompiler
+{
+    internal interface IRewriteStepsLoader
+    {
+        ImmutableArray<LoadedStep> GetLoadedSteps(CompilationLoader.Configuration config);
```
bettinaheim, line 12: The only think that bothers me slightly with this setup is that each step gets the full configuration object; also the ExternalRewriteStepsManager is an IRewriteStepsLoader. It would be sufficient for the manager to get the configuration and then ensure that each loader gets the necessary arguments.

bettinaheim (commented):
This is much better structured. Thanks, @filipw!

src/QsCompiler/Compiler/RewriteSteps/AbstractRewriteStepsLoader.cs
```diff
@@ -0,0 +1,64 @@
...
+                    this.onDiagnostic?.Invoke(this.LoadWarning(target, WarningCode.FailedToLoadRewriteStepViaReflection, target.LocalPath));
+                }
+            }
+            catch (Exception ex)
+            {
+                this.onDiagnostic?.Invoke(this.LoadError(target, ErrorCode.CouldNotInstantiateRewriteStep, type.ToString(), target.LocalPath));
+                this.onException?.Invoke(ex);
+            }
+
+            return LoadedStep.Empty;
```
bettinaheim, line 60: Since this step would not really be usable, I would tend to actually make it more obvious that it is not a successfully loaded step by returning null.

src/QsCompiler/Compiler/LoadedStep.cs
```diff
@@ -0,0 +1,212 @@
...
+using Range = Microsoft.Quantum.QsCompiler.DataTypes.Range;
+
+namespace Microsoft.Quantum.QsCompiler
+{
+    /// <summary>
+    /// Concrete implementation of a rewrite steps with an additional property specifying the dll it was loaded from.
+    /// </summary>
+    internal class LoadedStep : IRewriteStep
+    {
+        internal static LoadedStep Empty = new LoadedStep();
```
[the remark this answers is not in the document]
bettinaheim: Nvm, found it. See my comment below.

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -175,7 +175,19 @@ public struct Configuration
             /// (i.e. classes implementing IRewriteStep) and the corresponding output folder.
             /// The contained rewrite steps will be executed in the defined order and priority at the end of the compilation.
             /// </summary>
-            public IEnumerable<(string, string?)>? RewriteSteps;
```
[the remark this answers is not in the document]
filipw: yep definitely makes sense, I will do it. I was not sure about the obsoletion strategy (given that the SDK is still at 0.x)

src/QsCompiler/Compiler/RewriteSteps/AbstractRewriteStepsLoader.cs
filipw: I actually had it like that in the beginning, but then nullability analysis was added to the project in a separate PR. I wasn't sure whether to make it nullable at that point anymore, so I decided to use the [null object pattern](https://en.wikipedia.org/wiki/Null_object_pattern)

filipw: done

src/QsCompiler/Compiler/CompilationLoader.cs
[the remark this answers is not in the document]
filipw: done

src/QsCompiler/Compiler/RewriteSteps/IRewriteStepsLoader.cs
filipw: sure I will change it, will need to think a bit how to reflect this best

filipw: done - I removed the interface altogether and simplified everything. hopefully it is more reasonable now 😀

bettinaheim (approved):
This is great! Thank you!

PR #636, opened 2020-09-23
Enable C# nullable reference types in CompilationManager

samarsha:


src/QsCompiler/CompilationManager/CompilationUnit.cs
```diff
@@ -1043,16 +1052,15 @@ internal LocalDeclarations TryGetLocalDeclarations(FileContentManager file, Posi
...
                 throw new ArgumentNullException(nameof(types));
             }
-            predicate ??= _ => true;
 
             // Assign a unique ID to each reference.
 
-            var decorator = new NameDecorator($"QsRef");
-            var ids =
+            NameDecorator decorator = new NameDecorator($"QsRef");
+            ImmutableDictionary<string, int> ids =
```
ScottCarda-MS, line 1059: Can these be `var`?

src/QsCompiler/CompilationManager/CompilationUnitManager.cs
```diff
@@ -862,7 +859,7 @@ T TryQueryFile(FileContentManager f)
         /// Note: this method waits for all currently running or queued tasks to finish
         /// before getting the file content by calling FlushAndExecute.
         /// </summary>
-        public QsFragmentKind[][] GetTokenization(TextDocumentIdentifier textDocument) =>
+        public QsFragmentKind?[][]? GetTokenization(TextDocumentIdentifier textDocument) =>
```
ScottCarda-MS, line 862: I'm curious what this syntax means: `QsFragmentKind?[][]?`

src/QsCompiler/CompilationManager/CompilationUnit.cs
samarsha: Unfortunately, no, because both of them are inferred as nullable types for some reason, which causes an error on line 1078 when `GetMappingForSourceGroup` tries to use them. I'm not sure why this happens.

src/QsCompiler/CompilationManager/CompilationUnitManager.cs
samarsha:
```
The outermost array is nullable (GetTokenization may return null).
                   v
QsFragmentKind?[][]?
              ^
The elements in the innermost array are nullable QsFragmentKinds.
```

src/QsCompiler/CompilationManager/CompilationUnit.cs
```diff
@@ -318,14 +319,15 @@ public void Dispose()
...
+            externals ??= References.Empty;
+            dependentLocks ??= Enumerable.Empty<ReaderWriterLockSlim>();
+
             this.syncRoot = new ReaderWriterLockSlim(LockRecursionPolicy.SupportsRecursion);
-            this.dependentLocks = dependentLocks == null
-                ? new HashSet<ReaderWriterLockSlim>()
-                : new HashSet<ReaderWriterLockSlim>(dependentLocks);
-            if (dependentLocks?.Contains(null) ?? false)
+            this.dependentLocks = new HashSet<ReaderWriterLockSlim>(dependentLocks);
+            if (dependentLocks.Contains(null!))
```
ScottCarda-MS, line 330: What does the `!` do here? Is that new syntax for the nullable references?

ScottCarda-MS: I guess this is just the null-forgiveness operator applied to a null value, because now references can't be null by default.

src/QsCompiler/CompilationManager/EditorSupport/EditorCommands.cs
```diff
@@ -40,10 +40,10 @@ public static SymbolInformation[] DocumentSymbols(this FileContentManager file)
         /// Returns the source file and position where the item at the given position is declared at,
         /// if such a declaration exists, and returns null otherwise.
         /// </summary>
-        public static Location DefinitionLocation(this FileContentManager file, CompilationUnit compilation, Position position)
+        public static Location? DefinitionLocation(this FileContentManager file, CompilationUnit compilation, Position? position)
         {
             var symbolInfo = file?.TryGetQsSymbolInfo(position, true, out CodeFragment _); // includes the end position
-            if (symbolInfo == null || compilation == null)
+            if (file is null || symbolInfo == null || compilation == null || position == null)
```
ScottCarda-MS, line 46: If the types for symbolInfo, compilation, and position do not overload the `==` operator, could we use `is null` for them for consistency?

ScottCarda-MS (approved):

src/QsCompiler/CompilationManager/CompilationUnitManager.cs
ScottCarda-MS: Oh that makes sense. I think my brain was trying to make it more complicated than it is 😁.

src/QsCompiler/CompilationManager/AssemblyLoader.cs
```diff
@@ -204,9 +212,10 @@ private static (StringHandle, StringHandle)? GetAttributeType(MetadataReader met
         // have a single constructor taking a single string argument.
         private static (string, string)? GetAttribute(MetadataReader metadataReader, CustomAttribute attribute)
         {
-            var attrType = GetAttributeType(metadataReader, attribute);
-            QsCompilerError.Verify(attrType.HasValue, "the type of the custom attribute could not be determined");
-            var (ns, name) = attrType.Value;
+            var attrType =
+                GetAttributeType(metadataReader, attribute)
+                ?? throw new Exception("the type of the custom attribute could not be determined");
```
cesarzc, line 217: Isn't this changing the current behavior?

samarsha: Sort of. It only changed what kind of exception is thrown. I changed it because the C# compiler did not realize that `QsCompilerError.Verify` would throw if the condition was false (so it thought `attrType` could still be null). However, I just learned that I can add `[<DoesNotReturnIf false>]` annotations to `Verify`, so I reverted this change.

PR #644, opened 2020-09-23
Enable C# nullable reference types in Transformations project

### LINKED ISSUE #634: AccumulateIdentifiers transformation throws NullReferenceException if AST is missing locations
**Describe the bug**

The [`AccumulateIdentifiers`](https://github.com/microsoft/qsharp-compiler/blob/main/src/QsCompiler/Transformations/SearchAndReplace.cs#L328) transformation assumes all locations in the AST are non-null, and will throw an exception if there is a missing location (e.g., from generated code).

**Expected behavior**

The requirement that locations are non-null should be documented, or ideally, the transformation should be updated to handle null locations because it may be useful to run it on generated code.

samarsha:
Closes #634.

### DIFF
--- a/src/QsCompiler/CompilationManager/AssemblyLoader.cs
+++ b/src/QsCompiler/CompilationManager/AssemblyLoader.cs
@@ -13,6 +13,7 @@ using Microsoft.Quantum.QsCompiler.CompilationBuilder;
 using Microsoft.Quantum.QsCompiler.ReservedKeywords;
 using Microsoft.Quantum.QsCompiler.Serialization;
 using Microsoft.Quantum.QsCompiler.SyntaxTree;
+using Microsoft.Quantum.QsCompiler.Transformations;
 using Newtonsoft.Json.Bson;
 
 namespace Microsoft.Quantum.QsCompiler
--- a/src/QsCompiler/Transformations/Utils.cs
+++ b/src/QsCompiler/Transformations/Utils.cs
@@ -7,23 +7,23 @@ namespace Microsoft.Quantum.QsCompiler.Transformations
     /// <summary>
     /// Helpful utility functions.
     /// </summary>
-    public static class Utils
+    internal static class Utils
     {
         /// <summary>
         /// Applies the action <paramref name="f"/> to the value <paramref name="x"/>.
         /// </summary>
-        public static void Apply<T>(this T x, Action<T> f) => f(x);
+        internal static void Apply<T>(this T x, Action<T> f) => f(x);
 
         /// <summary>
         /// Applies the function <paramref name="f"/> to the value <paramref name="x"/> and returns the result.
         /// </summary>
-        public static TOut Apply<TIn, TOut>(this TIn x, Func<TIn, TOut> f) => f(x);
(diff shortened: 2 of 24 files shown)


src/QsCompiler/Transformations/Utils.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// <summary>
+        /// Projects each element of a sequence into a new form and discards null values.
+        /// </summary>
+        /// <remarks>Overload for value types.</remarks>
+        public static IEnumerable<TResult> SelectNotNull<TSource, TResult>(
+            this IEnumerable<TSource> source, Func<TSource, TResult?> selector)
+            where TResult : struct =>
+            source.SelectMany(item =>
+                selector(item)?.Apply(result => new[] { result })
+                ?? Enumerable.Empty<TResult>());
```
ScottCarda-MS, line 42: Wrapping each result in a single element array and doing a SelectMany, should have opposite effects. Wouldn't you get the same behavior if you did a regular Select without the Apply, followed by a Where to filter out nulls?

src/QsCompiler/CompilationManager/AssemblyLoader.cs
```diff
@@ -13,6 +13,7 @@
 using Microsoft.Quantum.QsCompiler.ReservedKeywords;
 using Microsoft.Quantum.QsCompiler.Serialization;
 using Microsoft.Quantum.QsCompiler.SyntaxTree;
+using Microsoft.Quantum.QsCompiler.Transformations;
```
ScottCarda-MS, line 16: Why does this file now need to reference the Transformations project? Same question for CodeActions.cs, ProjectManager.cs, and TypeChecking.cs.

ScottCarda-MS (approved):

src/QsCompiler/Transformations/Utils.cs
samarsha: `Select` followed by `Where` has the same behavior, but a different type. The C# compiler is not able to figure out that the resulting enumerable is a non-nullable type in that case.

src/QsCompiler/CompilationManager/AssemblyLoader.cs
samarsha: It's for the Utils class defined in Transformations. I couldn't find a better place to put it that all the C# projects could access it from.

src/QsCompiler/Transformations/Utils.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Quantum.QsCompiler.Transformations
+{
+    /// <summary>
+    /// Helpful utility functions.
+    /// </summary>
+    public static class Utils
```
samarsha, line 10: @bettinaheim, can you review these additions to the public API?

samarsha: Based on our discussion, I'm keeping these utilities internal and using `InternalsVisibleTo` to share them with the CompilationManager project.

samarsha:
While we're still discussing if using `InternalsVisibleTo` is a good idea, I'm merging this and the other nullable PRs in the meantime. We can change the visibility of the utility functions later if needed.

PR #645, opened 2020-09-24
Enable C# nullable reference types in LanguageServer project

samarsha:


src/QsCompiler/CompilationManager/ProjectManager.cs
```diff
@@ -903,10 +904,10 @@ public ProjectManager(Action<Exception> exceptionLogger, Action<string, MessageT
             // TODO: allow to cancel this task via cancellation token?
             return this.load.QueueForExecutionAsync(() =>
             {
-                var loaded = projectLoader(projectFile, out ProjectInformation info);
+                var loaded = projectLoader(projectFile, out var info);
                 var existing = this.projects.TryRemove(projectFile, out Project current) ? current : null;
 
-                if (!loaded)
+                if (!loaded || info is null)
```
cesarzc, line 910: With the check that you added to the `Loader` delagate, is the `info is null` condition needed here?

samarsha: Looks like the problem is that C# loses track of the `NotNullWhen` attribute when the return value is saved in a variable. I put the `projectLoader` call directly in the `if` and was able to avoid the extra `info is null` check.

src/QsCompiler/CompilationManager/Utils.cs
```diff
@@ -183,7 +183,7 @@ public static (List<T>, List<T>) Partition<T>(this IEnumerable<T> collection, Fu
         /// Projects each element of a sequence into a new form and discards null values.
         /// </summary>
         /// <remarks>Overload for reference types.</remarks>
-        internal static IEnumerable<TResult> SelectNotNull<TSource, TResult>(
+        public static IEnumerable<TResult> SelectNotNull<TSource, TResult>(
```
cesarzc, line 186: Why were the `SelectNotNull` methods changed from `internal` to `public`?

samarsha: They're needed by the language server, although ultimately I think they need to live in Transformations since that is the one C# project that all other C# projects have a reference to (see https://github.com/microsoft/qsharp-compiler/pull/644#discussion_r494515713)

cesarzc (approved):
Thanks for doing this Sarah!

PR #646, opened 2020-09-24
Enable C# nullable reference types in DocumentationParser project

samarsha:


src/QsCompiler/DocumentationParser/Utils.cs
```diff
@@ -410,7 +410,7 @@ internal static void WriteYamlFile(YamlNode rootNode, string rootPath, string fi
         internal static void MergeYamlFile(YamlMappingNode map, string rootPath, string fileBaseName)
         {
             var yamlReader = new YamlStream();
-            YamlMappingNode fileMap = new YamlMappingNode();
+            YamlMappingNode? fileMap = new YamlMappingNode();
```
cgranade, line 413: Out of curiosity, why does this one need to be nullable?

cgranade (approved):

src/QsCompiler/DocumentationParser/Utils.cs
samarsha: In a lambda below, `fileMap` is reassigned using an `as` expression that may return null. Although for some reason the compiler is OK with `fileMap.Children` being accessed in the foreach loop without a null check. I'm not sure what should happen if `fileMap` is null there.

samarsha: I changed this function to avoid the null by using `?? fileMap` if the `as` fails.

ScottCarda-MS (approved):

Issue #575, opened 2020-08-17
Q# callables should not have same fully qualified name as another namespace


swernli:
This issue is copied from https://github.com/microsoft/qsharp-runtime/issues/245.
When a Q# callable has a fully qualified name that exactly matches the name of another namespace, it becomes very hard to disambiguate them. This specifically causes problems in C# generation, but is really more a generally bad thing to do. We should detect and flag this in the compiler so that developers know early that the given name will not work.

For example, this code is currently accepted by the compiler:
```
namespace Foo {
    function Bar() : Unit { }
}

namespace Foo.Bar {
    function Hello() : Unit { }
}
```
But the function `Bar` has the fully qualified name `Foo.Bar` which collides with the namespace of the same name. This should not be allowed.

cgranade:
```
namespace Baz {
    function Bar() : Unit { }
}

namespace Foo.Bar {
    open Baz as Foo; // is this bad as well?
    function Hello() : Unit { }
}
```

bettinaheim:
@cgranade No, that is fine, since everything is resolved to fully qualified names upon compilation the issue only arises if the clash is with the full name.

bettinaheim:
Fixed with #647.

PR #647, opened 2020-09-24
Compilation error for naming conflicts with namespace names

bettinaheim:
This should address https://github.com/microsoft/qsharp-compiler/issues/575.


bettinaheim:
@swernli I believe it should be sufficient to check that a declaration doesn't conflict with the full namespace name. Checking that it doesn't conflict with a (C#) partial namespace name should not be needed for the autogenerated code to work fine - correct me if you see an issue with that.

swernli (approved):
Yes, this change looks good! Thanks for fixing this!

PR #648, opened 2020-09-24
Enable C# nullable reference types in Compiler project

samarsha:


src/QsCompiler/CompilationManager/Utils.cs
```diff
@@ -155,12 +155,12 @@ internal static string GetTextChangedLines(FileContentManager file, TextDocument
         /// <summary>
         /// Applies the action <paramref name="f"/> to the value <paramref name="x"/>.
         /// </summary>
-        internal static void Apply<T>(this T x, Action<T> f) => f(x);
+        public static void Apply<T>(this T x, Action<T> f) => f(x);
```
ScottCarda-MS, line 158: These are the same methods defined in the Utils class in PR #644. They probably shouldn't be redefined for each project, right? Is there some common place we could put these where all project could get access to them?

ScottCarda-MS (approved):

src/QsCompiler/CompilationManager/Utils.cs
ScottCarda-MS: Ah I've seen your comment on #645 addressing this concern.

PR #649, opened 2020-09-24
Tracking for Specific Rewrite Steps

ScottCarda-MS:
Added performance tracking for specific rewrite steps based off the name field of the step.

### DIFF
--- a/src/QsCompiler/CompilationManager/PerformanceTracking.cs
+++ b/src/QsCompiler/CompilationManager/PerformanceTracking.cs
@@ -150,17 +150,17 @@ namespace Microsoft.Quantum.QsCompiler.Diagnostics
         /// <summary>
         /// Raises a task start event.
         /// </summary>
-        public static void TaskStart(Task task)
+        public static void TaskStart(Task task, string? details = null)
         {
-            InvokeTaskEvent(CompilationTaskEventType.Start, task);
+            InvokeTaskEvent(CompilationTaskEventType.Start, task, details);
         }
 
         /// <summary>
         /// Raises a task end event.
         /// </summary>
-        public static void TaskEnd(Task task)
+        public static void TaskEnd(Task task, string? details = null)
         {
-            InvokeTaskEvent(CompilationTaskEventType.End, task);
+            InvokeTaskEvent(CompilationTaskEventType.End, task, details);
         }
 
         /// <summary>
@@ -181,7 +181,7 @@ namespace Microsoft.Quantum.QsCompiler.Diagnostics
         /// Invokes a compilation task event.
         /// If an exception occurs when calling this method, the error message is cached and subsequent calls do nothing.
         /// </summary>
-        private static void InvokeTaskEvent(CompilationTaskEventType eventType, Task task)
+        private static void InvokeTaskEvent(CompilationTaskEventType eventType, Task task, string? details = null)
         {
             if (FailureOccurred)
             {
@@ -190,8 +190,17 @@ namespace Microsoft.Quantum.QsCompiler.Diagnostics
 
             try
             {
-                var parent = GetTaskParent(task);
-                CompilationTaskEvent?.Invoke(eventType, parent?.ToString(), task.ToString());
+                if (details is null)
+                {
+                    var parent = GetTaskParent(task);
+                    CompilationTaskEvent?.Invoke(eventType, parent?.ToString(), task.ToString());
+                }
+                else
+                {
+                    var parent = task.ToString();
+                    CompilationTaskEvent?.Invoke(eventType, parent?.ToString(), details.ToString());
+                }
+
             }
             catch (Exception ex)
             {
--- a/src/QsCompiler/CompilationManager/PerformanceTracking.cs
+++ b/src/QsCompiler/CompilationManager/PerformanceTracking.cs
@@ -65,7 +65,7 @@ namespace Microsoft.Quantum.QsCompiler.Diagnostics
             RewriteSteps,
 
             /// <summary>
-            /// A task for a specific rewrite step.
+            /// Task for a specific rewrite step.
             /// These tasks should be accompanied with details of which rewrite step it is specific to.
             /// </summary>
             SingleRewriteStep,
--- a/src/QsCompiler/CompilationManager/PerformanceTracking.cs
+++ b/src/QsCompiler/CompilationManager/PerformanceTracking.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Text.RegularExpressions;
 
 #nullable enable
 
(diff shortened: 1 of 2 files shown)


src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -500,53 +500,55 @@ public CompilationLoader(SourceLoader loadSources, ReferenceLoader loadReference
...
+                steps.Add((priority, rewriteStep.Name, Execute(j)));
             }
 
-            PerformanceTracking.TaskStart(PerformanceTracking.Task.RewriteSteps);
             SortRewriteSteps(steps, t => t.Item1);
-            foreach (var (_, rewriteStep) in steps)
+            PerformanceTracking.TaskStart(PerformanceTracking.Task.RewriteSteps);
+            foreach (var (_, name, rewriteStep) in steps)
             {
+                PerformanceTracking.TaskStart(PerformanceTracking.Task.RewriteSteps, name);
```
cesarzc, line 548: I would add a new item to the `PerformanceTracking.Task` enum for individual rewrite steps and use it here. Something like `IndividualRewriteStep` or `SingularRewriteStep`.

This way, every rewrite step would be hierarchically organized under the `RewriteSteps` task.

Remember to add the hierarchy specification in `TasksHierarchy`. Something like the following:
```csharp
{ Task.SingleRewriteStep, Task.RewriteSteps }
```

src/QsCompiler/CompilationManager/PerformanceTracking.cs
```diff
@@ -190,8 +190,17 @@ private static void InvokeTaskEvent(CompilationTaskEventType eventType, Task tas
...
+                {
+                    var parent = GetTaskParent(task);
+                    CompilationTaskEvent?.Invoke(eventType, parent?.ToString(), task.ToString());
+                }
+                else
+                {
+                    var parent = task.ToString();
+                    CompilationTaskEvent?.Invoke(eventType, parent?.ToString(), details.ToString());
+                }
+
```
cesarzc, line 203:
```suggestion
                var parent = GetTaskParent(task);
                var taskId =
                    task.ToString() +
                    (details is null ? string.Empty : $"-{details}");

                CompilationTaskEvent?.Invoke(eventType, parent?.ToString(), taskId);
```

cesarzc: This way we can keep the hierarchy protections and with the suggestion in my other comment it would make every rewrite step look like the following:
```
"OverallCompilation.RewriteSteps.SingleRewriteStep-FunctorGenerator"
"OverallCompilation.RewriteSteps.SingleRewriteStep-Monomorphization"
```

ScottCarda-MS: Currently the rewrite steps look like this in the perf output:
```
"OverallCompilation.RewriteSteps": 427,
"OverallCompilation.RewriteSteps.Classically Controlled": 375,
"OverallCompilation.RewriteSteps.Conjugation Inlining": 2,
"OverallCompilation.RewriteSteps.Functor Generation": 13,
"OverallCompilation.RewriteSteps.Monomorphization": 35,
```

You're wanting it changed to:
```
"OverallCompilation.RewriteSteps": 427,
"OverallCompilation.RewriteSteps.SingleRewriteStep-Classically Controlled": 375,
"OverallCompilation.RewriteSteps.SingleRewriteStep-Conjugation Inlining": 2,
"OverallCompilation.RewriteSteps.SingleRewriteStep-Functor Generation": 13,
"OverallCompilation.RewriteSteps.SingleRewriteStep-Monomorphization": 35,
```
What is the advantage to having another enum value, SingleRewriteStep, and prefacing the rewrite step output with "SingleRewriteStep-"?

src/QsCompiler/Compiler/CompilationLoader.cs
ScottCarda-MS: I've added a separate enum for SingleRewriteStep, although I'd like to learn more about why this is necessary. I think there is something about this setup I'm not quite clear on. Let's chat offline about it.

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -500,53 +500,55 @@ public CompilationLoader(SourceLoader loadSources, ReferenceLoader loadReference
...
+                steps.Add((priority, rewriteStep.Name, Execute(j)));
             }
 
-            PerformanceTracking.TaskStart(PerformanceTracking.Task.RewriteSteps);
             SortRewriteSteps(steps, t => t.Item1);
-            foreach (var (_, rewriteStep) in steps)
+            PerformanceTracking.TaskStart(PerformanceTracking.Task.RewriteSteps);
+            foreach (var (_, name, rewriteStep) in steps)
             {
+                PerformanceTracking.TaskStart(PerformanceTracking.Task.SingleRewriteStep, name);
```
cesarzc, line 548: It would be good to remove spaces from the name that is passed as details.

ScottCarda-MS: done

src/QsCompiler/CompilationManager/PerformanceTracking.cs
ScottCarda-MS: Discussed offline: Using an explicit enum value better supports the task hierarchy, as well as future work on this feature.

src/QsCompiler/Compiler/CompilationLoader.cs
ScottCarda-MS: Discussed offline: Using an explicit enum value better supports the task hierarchy, as well as future work on this feature.

src/QsCompiler/CompilationManager/PerformanceTracking.cs
```diff
@@ -64,6 +65,12 @@ public enum Task
             /// </summary>
             RewriteSteps,
 
+            /// <summary>
+            /// Task for a specific rewrite step.
+            /// These tasks should be accompanied with details of which rewrite step it is specific to.
+            /// </summary>
+            SingleRewriteStep,
```
cesarzc, line 72: I would add this to the end of the enum such that the current integer value of the Tasks that are already part of the enum.

src/QsCompiler/CompilationManager/PerformanceTracking.cs
```diff
@@ -191,7 +199,8 @@ private static void InvokeTaskEvent(CompilationTaskEventType eventType, Task tas
             try
             {
                 var parent = GetTaskParent(task);
-                CompilationTaskEvent?.Invoke(eventType, parent?.ToString(), task.ToString());
+                var taskId = task.ToString() + (details is null ? string.Empty : $"-{Regex.Replace(details, @"\s+", "")}");
```
cesarzc, line 202:
```suggestion
                var taskId = task.ToString() + (details is null ? string.Empty : $"-{Regex.Replace(details, @"\s+", string.Empty)}");
```

src/QsCompiler/CompilationManager/PerformanceTracking.cs
ScottCarda-MS: done

cesarzc (approved):
Thanks for adding this Scott!

PR #654, opened 2020-09-29
Call Graph to Main

ScottCarda-MS:
Adds the Call Graph data structures and several use cases for them, such as invalid cycle diagnostics, and cleaner monomorphization code.
New to this PR:
-  Some fixups for the nullable reference types
- Renamed SimpleCallGraph to CallGraph


src/QsCompiler/CompilationManager/DataStructures.cs
```diff
@@ -474,7 +475,7 @@ internal struct HeaderEntry<T>
             CodeFragment.TokenIndex tIndex,
             ImmutableArray<AttributeAnnotation> attributes,
             ImmutableArray<string> doc,
-            string? keepInvalid = null)
+            [AllowNull] string? keepInvalid = null)
```
samarsha, line 478: The `[AllowNull]` shouldn't be needed - `string?` should already allow null.

src/QsCompiler/CompilationManager/Diagnostics.cs
```diff
@@ -82,7 +83,7 @@ private static string Code(QsCompilerDiagnostic msg)
         /// Throws an ArgumentNullException if the Range of the given CompilerDiagnostic is null.
         /// Throws an ArgumentOutOfRangeException if the contained range contains zero or negative entries, or if its Start is bigger than its End.
         /// </summary>
-        internal static Diagnostic Generate(string filename, QsCompilerDiagnostic msg, Position? positionOffset = null)
+        internal static Diagnostic Generate(string filename, QsCompilerDiagnostic msg, [AllowNull] Position? positionOffset = null)
```
samarsha, line 86: Same here with `[AllowNull]`. `Position?` should already be a nullable type.

src/QsCompiler/CompilationManager/TypeChecking.cs
```diff
@@ -2245,7 +2245,7 @@ internal static void UpdateTypeChecking(this FileContentManager file, Compilatio
             }
         }
 
-        private static void CheckForGlobalCycleChange(FileContentManager file, List<Diagnostic> diagnostics)
+        private static void CheckForGlobalCycleChange(FileContentManager file, List<Diagnostic>? diagnostics)
```
samarsha, line 2248: It looks like maybe `Count` and `Any` haven't been updated to use nullable reference types, but it says they throw `ArgumentNullException` if the source enumerable is null, so I think `diagnostics.Count(...)` and `diagnostics.Any(...)` where `diagnostics` is nullable is not safe.

src/QsCompiler/CompilationManager/DataStructures.cs
ScottCarda-MS: I think I was getting a "fake" error. After merging in your changes with the nullable references and removing these, I am not getting the error I was getting before. I've removed this.

src/QsCompiler/CompilationManager/Diagnostics.cs
ScottCarda-MS: Removed

src/QsCompiler/CompilationManager/TypeChecking.cs
ScottCarda-MS: I've updated the method to have `diagnostics` be non-null here.

src/QsCompiler/DataStructures/SyntaxExtensions.fs
```diff
@@ -372,52 +371,57 @@ let private TupleItems<'T when 'T :> ITuple> (arg: 'T): 'T list option =
...
-let (| Missing | _ |) arg = 
+let (| Missing | _ |) arg =
     match TupleItems arg with
     | Some [] -> Some Missing
     | _ -> None
 
+// filter for type parameter resolution dictionaries
+
+[<Extension>]
+let FilterByOrigin (this : ImmutableDictionary<(QsQualifiedName * NonNullable<string>), ResolvedType>) origin =
```
samarsha, line 392: I think this function is a bit more generic:

```suggestion
let FilterByOrigin (this : KeyValuePair<_, _> seq) origin =
```

Although then it makes a little bit less sense as an extension method that returns a dictionary.

src/QsCompiler/Tests.Compiler/TestCases/CycleValidation.qs
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    operation Bar<'A,'B>() : Unit {
+        Baz<'A,'A>();
+    }
+
+    operation Baz<'A,'B>() : Unit {
+        Foo<'A,'B>();
+    }
+}
```
samarsha, line 55: Why are constricting cycles not allowed?

src/QsCompiler/CompilationManager/DataStructures.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Diagnostics.CodeAnalysis;
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/QsCompiler/CompilationManager/Diagnostics.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/QsCompiler/CompilationManager/TypeChecking.cs
```diff
@@ -2212,13 +2229,17 @@ internal static void UpdateTypeChecking(this FileContentManager file, Compilatio
...
                 {
-                    diagnostics?.Apply(file.AddAndFinalizeSemanticDiagnostics); // diagnostics have been cleared already for the edited callables (only)
-                }
-                else
-                {
-                    diagnostics?.Apply(file.ReplaceSemanticDiagnostics);
+                    CheckForGlobalCycleChange(file, diagnostics);
+                    if (sameImports)
+                    {
+                        diagnostics.Apply(file.AddAndFinalizeSemanticDiagnostics); // diagnostics have been cleared already for the edited callables (only)
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/QsCompiler/DataStructures/SyntaxExtensions.fs
ScottCarda-MS: The extension is supposed to operate specifically on type parameter resolutions dictionaries, filtering them out by their origins. The type of those dictionaries is `ImmutableDictionary<(QsQualifiedName * NonNullable<string>), ResolvedType>)`. Making it more permissive than it needs to be would have us supporting scenarios we don't intend to support right now. If in the future a situation comes up where we need to support different kinds of dictionaries with different type arguments, then the extension can be expanded at that point as needed without breaking code formed around it being specific to this type, but we would not be able to do the reverse without potentially breaking code.

src/QsCompiler/Tests.Compiler/TestCases/CycleValidation.qs
ScottCarda-MS: I believe this was a requirement I got earlier on from Bettina. @bettinaheim Could you comment on why constricting cycles are disallowed?

ScottCarda-MS: I can modify the TypeParamCombinations tool, which generally encapsulates the rules for how type parameter resolutions are combined, to allow constriction mappings, but then we allow constrictive type arguments within the same method.

ScottCarda-MS: ... or more accurately, _the tool can't be used to disallow_ constrictive type arguments within the same method.

src/QsCompiler/Transformations/CallGraph/CallGraphWalkerBase.cs
```diff
@@ -0,0 +1,87 @@
...
+            where TNode : CallGraphNodeBase
+            where TEdge : CallGraphEdgeBase
+        {
+            public abstract class TransformationState
+            {
+                internal TNode? CurrentNode;
+                internal readonly TGraph Graph;
+
+                // The type parameter resolutions of the current expression.
+                internal IEnumerable<TypeParameterResolutions> ExprTypeParamResolutions = new List<TypeParameterResolutions>();
```
bettinaheim, line 31: Since you are prepending, a List here seems like a bit of a suboptimal choice. You could go with a Stack instead.

src/QsCompiler/Transformations/CallGraph/CallGraphWalkerBase.cs
```diff
@@ -0,0 +1,87 @@
...
+                }
+
+                public override TypedExpression OnTypedExpression(TypedExpression ex)
+                {
+                    var contextRange = this.SharedState.CurrentExpressionRange;
+                    this.SharedState.CurrentExpressionRange = ex.Range;
+
+                    if (ex.TypeParameterResolutions.Any())
+                    {
+                        this.SharedState.ExprTypeParamResolutions = this.SharedState.ExprTypeParamResolutions.Prepend(ex.TypeParameterResolutions);
```
bettinaheim, line 78:
```suggestion
                        this.SharedState.ExprTypeParamResolutions.Push(ex.TypeParameterResolutions);
```

src/QsCompiler/Transformations/CallGraph/ConcreteCallGraphWalker.cs
```diff
@@ -0,0 +1,293 @@
...
+
+            private class NamespaceWalker : NamespaceTransformation<TransformationState>
+            {
+                public NamespaceWalker(SyntaxTreeTransformation<TransformationState> parent) : base(parent, TransformationOptions.NoRebuild)
+                {
+                }
+
+                public override QsGeneratorDirective OnGeneratedImplementation(QsGeneratorDirective directive)
+                {
+                    throw new ArgumentException("Encountered unresolved generated specialization while constructing concrete call graph.");
```
bettinaheim, line 244: Is this documented in the corresponding public method/class?

bettinaheim (commented):
All in all this has come together really nicely!

src/QsCompiler/Tests.Compiler/TestUtils/Signatures.fs
```diff
@@ -147,18 +149,10 @@ let public MonomorphizationSignatures =
             GenericsNs, "GenericCallsSpecializations", [|"Double"; "String"; "Double"|], "Unit"
             GenericsNs, "GenericCallsSpecializations", [|"String"; "Int"; "Unit"|], "Unit"
 
-            GenericsNs, "BasicGeneric", [|"Qubit[]"; "Qubit[]"|], "Unit"
```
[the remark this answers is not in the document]
ScottCarda-MS: The new method of trimming the syntax tree only keeps around specializations that are used. The old method would keep all the specializations of a callable if the callable was used. All of the signatures that are removed here are for concretizations that were associated with a specialization that was never actually used, and so are removed under the new trimming method.

src/QsCompiler/CompilationManager/FileContentManager.cs
```diff
@@ -196,6 +196,8 @@ private static void InvalidateOrUpdateBySyntaxCheckDelimeters(ManagedList<Diagno
             diagnostics.SyncRoot.EnterWriteLock();
             try
             {
+                // remove all cycle related diagnostics
+                diagnostics.RemoveAll(m => m.Code == ErrorCode.InvalidCyclicTypeParameterResolution.Code());
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/QsCompiler/Compiler/PluginInterface.cs
```diff
@@ -22,29 +22,29 @@ public static class RewriteStepPriorities
...
-        /// all type parametrized callables with concrete instantiations and drops any unused callables.
-        /// </summary>
-        public const int TypeParameterElimination = 1000;
-
         /// <summary>
         /// Priority of the built-in transformation that replaces
         /// all functor generation directives with the corresponding implementation.
         /// </summary>
-        public const int GenerationOfFunctorSupport = 600;
+        public const int GenerationOfFunctorSupport = 1000;
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/QsCompiler/Transformations/CallGraph/CallGraph.cs
```diff
@@ -0,0 +1,441 @@
...
+            {
+                // results needs to be evaluated before being assigned to avoid errors, so we use ToList
+                result = sequence.SelectMany(item => result, (item, seq) => seq.Concat(new[] { item })).ToList();
+            }
+            return result;
+        }
+
+        // Member Fields
+
+        private CallGraphBuilder<CallGraphNode, CallGraphEdge> graphBuilder = new CallGraphBuilder<CallGraphNode, CallGraphEdge>();
```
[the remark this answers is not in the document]
ScottCarda-MS: done, also added `readonly` to the concrete call graph counterpart.

src/QsCompiler/Transformations/CallGraph/CallGraph.cs
```diff
@@ -0,0 +1,441 @@
...
+        /// Constructor for CallGraphNode objects.
+        /// </summary>
+        public CallGraphNode(QsQualifiedName callableName)
+            : base(callableName)
+        {
+        }
+    }
+
+    /// <summary>
+    /// A kind of call graph whose nodes represent Q# callables.
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/QsCompiler/Transformations/CallGraph/CallGraphWalkerBase.cs
ScottCarda-MS: done

src/QsCompiler/Transformations/CallGraph/CallGraphWalkerBase.cs
ScottCarda-MS: done

src/QsCompiler/Transformations/CallGraph/ConcreteCallGraphWalker.cs
ScottCarda-MS: The documentation in the relevant places has been updated to make it clear that the concrete call graph constructor expects there to be no Generated Implementation directives, and will error if one is encountered.

src/QsCompiler/Transformations/CodeTransformations.cs
```diff
@@ -10,12 +10,12 @@
...
 using Microsoft.Quantum.QsCompiler.Transformations.SearchAndReplace;
 
-namespace Microsoft.Quantum.QsCompiler
+namespace Microsoft.Quantum.QsCompiler.Transformations
 {
     /// <summary>
     /// Static base class to accumulate the handles to individual syntax tree rewrite steps.
     /// </summary>
-    public static class CodeTransformations
+    public static class CodeTransformationExtensions
```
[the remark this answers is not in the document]
ScottCarda-MS: Since this is a static class that only contains extension methods, this class name never actually gets used because it is an extension class or "sponsor" class. Hence I don't believe it would be a breaking change to update the name because the actual name would never appear in any code. The reason it was updated is that in my experience sponsor classes end with "Extensions", although after trying to find a standard for that online, it seems there is not a standard way of naming sponsor classes. My thinking on this is that there should be something in the name that indicates it is a sponsor class. Regardless of what specifically we do, we should change the name somehow to indicate that it is a sponsor class, as that will help make this file more readable.
[General Guidelines for Extension Methods](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/extension-methods)
[Lack of Naming Convention](https://stackoverflow.com/a/1051278/5399520)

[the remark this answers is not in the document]
samarsha: You can technically call extension methods by using the name of their containing class, although it's rare.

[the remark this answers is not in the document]
ScottCarda-MS: I did this in the course of trying to make things a bit more readable for myself, however it is ultimately immaterial to the Call Graph feature, so if you are still opposed to it, it wouldn't be an issue to remove. Let me know what you think.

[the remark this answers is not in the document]
ScottCarda-MS: I didn't realize extension methods could be called by referencing the sponsor class. I've removed the rename.

src/QsCompiler/Tests.Compiler/TestCases/CycleValidation.qs
ScottCarda-MS: I believe one of the reasons for this restriction is that, currently, I don't believe the system has a way of differentiating between a type parameter reference from the previous instance of the callable in the cycle and the type parameters of the current instance of the callable, so there is not currently a way to differentiate between whether a callable is self-constricting or constricting due to the cycle, as far as type parameters are concerned. This is my understanding anyway. Because we don't allow self-constricting callables, we are conservative here and don't allow constriction due to cycles either. If in the future we are able to make this distinction and make sure any other requirements about cycles are still met, we may relax this restriction without making a breaking change, but, I believe, that would require some major changes elsewhere in the system. I'm not sure if there are other reasons constricting cycles are disallowed as a requirement.

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -30,151 +31,91 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.Monomorphization
...
         public static QsCompilation Apply(QsCompilation compilation)
         {
             var globals = compilation.Namespaces.GlobalCallableResolutions();
+            var concretizations = new List<QsCallable>();
+            var concreteNames = new Dictionary<ConcreteCallGraphNode, QsQualifiedName>();
 
-            var intrinsicCallableSet = globals
-                .Where(kvp => kvp.Value.Specializations.Any(spec => spec.Implementation.IsIntrinsic))
-                .Select(kvp => kvp.Key)
+            var nodes = new ConcreteCallGraph(compilation).Nodes
```
samarsha, line 40: This will cause a regression in monomorphization until #662 is fixed. #662 will probably be resolved by disallowing the cases that cause the bug, but the error message will not be as nice and won't show up at design time until then.

ScottCarda-MS: What regression is being caused? I'm not clear on what you are referring to.

ScottCarda-MS: Based on the issue linked, if the language verification were correctly forbidding the situations listed, as is the issue in question, than there would be no change as a result of this PR. The fact that the updated monomorphization more closely follows the language specifications you linked in the issue by throwing errors if the input doesn't follow it seems like it is more correct that before, as now the situations listed do not compile, which is intended. The linked issue seems to refer to a bug in the language verification where it is not invalidating programs that, by the language specifications, are supposed to be invalid.

samarsha: The regression is that those programs compile and run normally in the current version of the compiler, but cause the compiler to throw an exception in this PR. From the user's point of view, compiler exceptions look like a bug in the compiler. It would be better if there was a proper error message.

ScottCarda-MS: I agree that it should be handled better, but that is a job for the language verification and not specific to the call graph or monomorphization. The situations listed should be rejected, for example, even if monomorphization is not being run as a rewrite step, as not allowing these is part of the language definition.

samarsha: My request is only that we either hold off on merging this PR until #662 is resolved, or document that this is a known issue in the release notes or somewhere else.

src/QsCompiler/Tests.Compiler/TestCases/CycleValidation.qs
samarsha: OK, the finite cycles restriction makes sense, and I understand that this is a conservative approximation of that restriction. Thanks for the explanation.

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
```diff
@@ -0,0 +1,597 @@
...
+
+    let MakeNode name specKind (paramRes : _ list) =
+        let qualifiedName = { Namespace = NonNullable<_>.New Signatures.TypeParameterResolutionNS; Name = NonNullable<_>.New name }
+        let res = paramRes.ToImmutableDictionary((fun kvp -> (qualifiedName, NonNullable<_>.New (fst kvp))), (fun kvp -> ResolvedType.New (snd kvp)))
+        ConcreteCallGraphNode(qualifiedName, specKind, res)
+
+    let DecorateWithNamespace (ns : string) (input : string list list) =
+        List.map (List.map (fun name -> { Namespace = NonNullable<_>.New ns; Name = NonNullable<_>.New name })) input
+
+    let ResolutionFromParamName (res : (QsQualifiedName * NonNullable<string> * QsTypeKind<_,_,_,_>) list) =
```
bettinaheim, line 48: This one here seems to be unused.

src/QsCompiler/Tests.Compiler/TestCases/TypeParameterResolution.qs
```diff
@@ -0,0 +1,267 @@
+﻿// Copyright (c) Microsoft Corporation.
```
bettinaheim, line 1: nit: The majority of test cases in here don't seem to be related to type parameter resolution, so it may make sense to rename the file (and possibly the namespace).

src/QsCompiler/Transformations/CodeTransformations.cs
```diff
@@ -10,12 +10,12 @@
 using Microsoft.Quantum.QsCompiler.Transformations.FunctorGeneration;
 using Microsoft.Quantum.QsCompiler.Transformations.SearchAndReplace;
 
-namespace Microsoft.Quantum.QsCompiler
+namespace Microsoft.Quantum.QsCompiler.Transformations
```
bettinaheim, line 13: This renaming of the namespace is still a breaking change - wasn't that the reason that you needed to add additional usings in a couple of files?

bettinaheim (changes requested):

src/QsCompiler/Tests.Compiler/TestUtils/Signatures.fs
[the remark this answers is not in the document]
bettinaheim: That should probably be fine. We might need to change it in the future.

src/QsCompiler/Transformations/Monomorphization.cs
bettinaheim: Since these are edge cases that I am not sure have a real use case, I am fine with taking this in to unblock other work before they are addressed. I tentatively assigned the issue to November.

src/QsCompiler/Tests.Compiler/CallGraphTests.fs
ScottCarda-MS: removed

src/QsCompiler/Tests.Compiler/TestCases/TypeParameterResolution.qs
ScottCarda-MS: Renamed to PopulateCallGraph, since that is the test category all of these tests fall under.

src/QsCompiler/Transformations/CodeTransformations.cs
ScottCarda-MS: Reverted to old namespace. Removed added usings in ConjugationInlining.cs, FullPreEvaluation.cs, and FunctorGeneration.cs.

src/QsCompiler/CompilationManager/TypeChecking.cs
```diff
@@ -1850,5 +1871,16 @@ internal static void UpdateTypeChecking(this FileContentManager file, Compilatio
                 compilation.ExitWriteLock();
             }
         }
+
+        private static void CheckForGlobalCycleChange(FileContentManager file, List<Diagnostic> diagnostics)
+        {
+            var numCycleDiagnosticsChange = file.CurrentSemanticDiagnostics().Count(m => m.Code == ErrorCode.InvalidCyclicTypeParameterResolution.Code())
```
bettinaheim, line 1877:
```suggestion
            var numCycleDiagnosticsChange = file.CurrentSemanticDiagnostics().Count(DiagnosticTools.ErrorType(ErrorCode.InvalidCyclicTypeParameterResolution))
                - diagnostics.Count(DiagnosticTools.ErrorType(ErrorCode.InvalidCyclicTypeParameterResolution));
```

src/QsCompiler/CompilationManager/TypeChecking.cs
```diff
@@ -1850,5 +1871,16 @@ internal static void UpdateTypeChecking(this FileContentManager file, Compilatio
...
+                - diagnostics.Count(m => m.Code == ErrorCode.InvalidCyclicTypeParameterResolution.Code());
```
bettinaheim, line 1878:
```suggestion
```

bettinaheim (approved):

PR #656, opened 2020-09-29
 Enable C# nullable reference types in test projects

samarsha:
I ended up using the null-forgiving operator `!` in a few places - mainly after `Assert.IsNotNull`, because the compiler doesn't realize the variable isn't nullable after this statement, and in one place where a non-nullable field in a test class is initialized after the class is created. Normally I would try to avoid `!`, but for a test project it seems OK.


src/QsCompiler/Tests.LanguageServer/ProjectLoaderTests.cs
```diff
@@ -129,10 +129,10 @@ public void LoadNonQsharpProjects()
         public void LoadOutdatedQsharpProject()
         {
             var (projectFile, context) = this.Context("test9");
-            var projDir = Path.GetDirectoryName(projectFile);
+            var projDir = Path.GetDirectoryName(projectFile) ?? "";
```
cesarzc, line 132:
```suggestion
           var projDir = Path.GetDirectoryName(projectFile) ?? string.Empty;
```

src/QsCompiler/Tests.LanguageServer/ProjectLoaderTests.cs
```diff
@@ -129,10 +129,10 @@ public void LoadNonQsharpProjects()
...
             Assert.IsNotNull(context);
-            Assert.AreEqual("test9.dll", Path.GetFileName(context.Properties.OutputPath));
-            Assert.IsTrue(Path.GetDirectoryName(context.Properties.OutputPath).StartsWith(projDir));
+            Assert.AreEqual("test9.dll", Path.GetFileName(context!.Properties.OutputPath));
+            Assert.IsTrue((Path.GetDirectoryName(context.Properties.OutputPath) ?? "").StartsWith(projDir));
```
cesarzc, line 135:
```suggestion
            Assert.IsTrue((Path.GetDirectoryName(context.Properties.OutputPath) ?? string.Empty).StartsWith(projDir));
```

src/QsCompiler/Tests.LanguageServer/ProjectLoaderTests.cs
cesarzc: I think using string.Empty is slightly better since it expresses intent and the person that is looking at the code can be sure it was not a typo or something similar.

cesarzc: I thought it was a just these two lines but it seems like there are more like this one. Please let me know if you agree with this change before going ahead and making a bunch of suggestions that might end up not being integrated :P

src/QsCompiler/Tests.LanguageServer/TestSetup.cs
```diff
@@ -18,8 +18,8 @@ public sealed partial class BasicFunctionality : IDisposable
     {
         // basic setup
 
-        private Connection connection;
-        private JsonRpc rpc;
+        private Connection? connection;
+        private JsonRpc rpc = null!; // Initialized in SetupServerConnectionAsync.
```
cesarzc, line 22: I can't add a comment in a line that has not been edited but should the null-conditional operator in line 47 be a null-forgiving operator?
`this.rpc?.Dispose();` -> `this.rpc!.Dispose();`

samarsha: No, I don't think so - in that case, if the class is disposed before `this.rpc` is initialized, then `this.rpc.Dispose()` will throw a NullPointerException. I think the intended behavior is to do nothing.

src/QsCompiler/Tests.LanguageServer/ProjectLoaderTests.cs
samarsha: I prefer `""`, since it's shorter and consistent with non-empty strings.

cesarzc: Ok, I don't have a string preference for string.Empty so I'll close the suggestions regarding `""` vs `string.Empty`.

cesarzc (approved):
Looks good! Thanks again for making all these changes.

src/QsCompiler/Tests.LanguageServer/TestUtils.cs
```diff
@@ -107,16 +107,13 @@ internal static int GetRangeLength(VisualStudio.LanguageServer.Protocol.Range ra
...
-            {
-                throw new ArgumentNullException(nameof(content));
-            }
             if (!content.Any())
             {
                 throw new ArgumentException("the given content has to have at least on line");
             }
 
-            Assert.IsTrue(IsValidRange(change?.Range) && change.Text != null);
+            Assert.IsNotNull(change);
```
ScottCarda-MS, line 115: I think it is not possible for `change` to be null here because it is not a nullable `List<string>`, right? If that's the case, this Assert seems unnecessary.

src/QsCompiler/TestTargets/Simulation/Example/Example.csproj
```diff
@@ -8,6 +8,8 @@
     <IncludeQsharpCorePackages>false</IncludeQsharpCorePackages>
     <CsharpGeneration>false</CsharpGeneration> <!--we provide our own C# generation for the sake of also generating C# for all references-->
     <NoWarn>0219</NoWarn> <!-- suppress C# warning for unused variables -->
+    <Nullable>enable</Nullable>
```
ScottCarda-MS, line 11: Since this is a Q# project, does this then apply the nullable references feature to the generated C#?

ScottCarda-MS (approved):

src/QsCompiler/TestTargets/Simulation/Example/Example.csproj
samarsha: Hmm, I didn't think about that, but it looks like it doesn't apply to the generated code (it doesn't compile if I manually add `#nullable enable` to one of the generated files). I'm not sure why.

samarsha: I removed `<Nullable>enable</Nullable>` from the Q# projects since it seems to not be doing anything, and I don't think the generated code works with nullable enabled.

PR #657, opened 2020-09-30
Remove null argument checks and thrown ArgumentNullExceptions

samarsha:
With C#'s nullable reference types feature enabled in all projects, checking if every argument is null and throwing `ArgumentNullException` is not as important. Null values can still appear even with non-nullable types, especially with public members or when interacting with code that doesn't have nullable references types enabled, but it's much less likely now. We have decided that the benefit of catching nulls at runtime that should already be checked by the compiler is not worth the cost of decreased code readability, so they're being removed.

### DIFF
--- a/src/QsCompiler/CommandLineTool/Commands/Build.cs
+++ b/src/QsCompiler/CommandLineTool/Commands/Build.cs
@@ -152,14 +152,9 @@ namespace Microsoft.Quantum.QsCompiler.CommandLineCompiler
         /// <summary>
         /// Reads the content off all given response files and tries to parse their concatenated content as command line arguments.
         /// Logs a suitable exceptions and returns null if the parsing fails.
-        /// Throws an ArgumentNullException if the given sequence of responseFiles is null.
         /// </summary>
         private static BuildOptions? FromResponseFiles(IEnumerable<string> responseFiles)
         {
-            if (responseFiles == null)
-            {
-                throw new ArgumentNullException(nameof(responseFiles));
-            }
             var commandLine = string.Join(" ", responseFiles.Select(File.ReadAllText));
             var args = SplitCommandLineArguments(commandLine);
             var parsed = Parser.Default.ParseArguments<BuildOptions>(args);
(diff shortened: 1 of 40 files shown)


cgranade:
If the issue is one of code volume, would it be worth onboarding to .NET 5 and C# 9 at build-time, but using .NET Core 3.1 TFMs such as `netcoreapp31` to get access to [simplified null validation](https://github.com/dotnet/csharplang/pull/2144/files)? That would require .NET 5 to be installed when building the compiler, but would not pass that requirement along to users.

samarsha:
I haven't heard about that proposal. To be honest, I think that proposal is the wrong approach - I'm not thrilled about the idea of putting `!`s on every parameter. If it was integrated with NRTs such that every non-nullable parameter automatically generated a check, then I think it would be a lot more useful.

Even with all of the `ArgumentNullException`s, I think we still had null-unsound code, for example in F# which strongly discourages null but can't prevent non-F# code, or F# code using types defined in non-F# code, from adding nulls. So even with the minimal null-checking proposal, we couldn't catch all null pointers. It makes sense to prevent as much as possible at compile-time, but I think there are too many places where null can be hiding at runtime to make checking worth it.

cgranade:
> I haven't heard about that proposal. To be honest, I think that proposal is the wrong approach - I'm not thrilled about the idea of putting `!`s on every parameter. If it was integrated with NRTs such that every non-nullable parameter automatically generated a check, then I think it would be a lot more useful.

Fair enough, I just thought I'd ask in light of that feature being on track for C# 9.

> Even with all of the `ArgumentNullException`s, I think we still had null-unsound code, for example in F# which strongly discourages null but can't prevent non-F# code, or F# code using types defined in non-F# code, from adding nulls. 

If I recall, support for C# 8's nullability metadata is a major focus for F# 5, so as to address that precise issue (https://github.com/fsharp/fslang-design/blob/master/RFCs/FS-1060-nullable-reference-types.md). Perhaps with your recent improvements to use `<Nullable>enable</Nullable>` throughout the C# parts of the compiler, it may be worth investigating if F# 5 can make good use of that metadata as well?

samarsha:
I could be wrong, but my understanding is that F# 5 will not include integration with C# nullable reference types, since the PR for that proposal is still open: https://github.com/dotnet/fsharp/pull/6804. Maybe there is another PR that is in F# 5 that I'm not aware of though?

cgranade:
Ah, got it; I think I'd seen it in early planning docs, but didn't realize it was deferred, sorry.

src/QsCompiler/CommandLineTool/Commands/Build.cs
```diff
@@ -177,18 +172,9 @@ private static IEnumerable<string> SplitCommandLineArguments(string commandLine)
         /// <summary>
         /// Builds the compilation for the Q# code or Q# snippet and referenced assemblies defined by the given options.
         /// Returns a suitable error code if one of the compilation or generation steps fails.
-        /// Throws an ArgumentNullException if any of the given arguments is null.
         /// </summary>
         public static int Run(BuildOptions options, ConsoleLogger logger)
         {
-            if (options == null)
```
ScottCarda-MS, line 184: Don't we still need these checks for parameters on public facing methods? If I recall correctly, the fact that the method could be called from code that does not use nullable references meant that one could pass in a null as an argument to a public method. Has that been addressed?

samarsha: @bettinaheim and I believe that the checks are not worth it even for public methods. We recommend that users enable nullable reference types in their projects and avoid passing nulls to non-nullable parameters.

ScottCarda-MS (approved):
I've been wanting this for a while now. Thanks!

PR #663, opened 2020-10-03
Bond schema to compiler object translator

cesarzc:
This change implements translation from bond schema objects to compiler objects based on F# types.


swernli (approved):
LGTM. Looking forward to seeing these changes in action!

PR #664, opened 2020-10-05
Add vsce to devDependencies

samarsha:
This pins `vsce` to version `1.81.1` and installs it locally to `node_modules`, instead of using `npm install -g` to install it as a global tool. This prevents `vsce` updates from affecting builds of the QDK extension for VS Code without being tested first.


rmshaffer (approved):
Makes sense to me!

ricardo-espinoza (approved):

PR #668, opened 2020-10-06
Fix some additional minor nullability issues

samarsha:
These are a few minor nullability issues that don't seem to be caught by the C# compiler or Visual Studio, but Rider marks them as errors. They seem to be legitimate issues, so here are some fixes:

* ExternalRewriteSteps.cs: `T` could be a value type, in which case unboxing the potentially null result from `?.Invoke` could throw an exception. Using `default` handles both value and reference types without errors.
* ProjectLoader.cs: `p.Name` is known to be not null here, but `p.Name?.ToLowerInvariant()` implies that it could be null, and should convert the type from `string` to `string?`. But the value is used as a dictionary key which must be non-null.
* LanguageServer/Utils.cs: `TResult` may be null, but it is not known to be either a value type or a reference type, so it's not possible to use `TResult?`. A `[MaybeNullWhen(false)]` out parameter can't be easily used either, because local functions don't support attributes until C# 9. The simplest thing here seems to be to use the null-forgiving operator, because the return value is used safely in the code below.

### DIFF
--- a/src/QsCompiler/Compiler/ExternalRewriteSteps.cs
+++ b/src/QsCompiler/Compiler/ExternalRewriteSteps.cs
@@ -43,15 +43,21 @@ namespace Microsoft.Quantum.QsCompiler
                 this.InterfaceMethod($"get_{name}")?.Invoke(this.selfAsObject, null);
 
             [return: MaybeNull]
-            private T GetViaReflection<T>(string name) =>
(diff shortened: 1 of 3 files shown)


swernli (approved):
Looks good. I hadn't actually known about the [default literal](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/default#default-literal) before, very cool to see it used here!

Issue #610, opened 2020-09-08
Support for binary BigInt constants


msoeken:
**Describe the solution you'd like**
For consistency, I suggest to support binary `BigInt` constants, e.g., `0b1100101011111110L`.

**Describe alternatives you've considered**
Alternatives are not using binary constants, e.g., `0xCAFEL` or converting a binary `Int` constant (if the value is small enough), e.g., `IntAsBigInt(0b1100101011111110)`.

samarsha:
I noticed this a few days ago, too. I agree, I think it would make sense to support the same notations for `BigInt` that we do for `Int`. There is also octal notation `0o123` that does not work for `BigInt` with `0o123L`.

cesarzc:
Please take a look at the `numericLiteral` function in [QsExpressionParsing.fs](https://github.com/microsoft/qsharp-compiler/blob/main/src/QsCompiler/TextProcessor/QsExpressionParsing.fs#L173) which is the one that parses Int, BigInt and Double literals.

Also, remember adding new cases to the `Expression literal tests` [test](https://github.com/microsoft/qsharp-compiler/blob/b19fe1705020a253c4ad867f966648128c95c3be/src/QsCompiler/Tests.Compiler/SyntaxTests.fs#L207) to make sure the new supported BigInt literals are parsed as expected.

ziqi-ma:
Seems like C#'s NumberStyles Enum (argument of BigInteger.Parse) only supports hex? Does that mean we need to write our own logic of converting binary/octal strings to hex?

cesarzc:
Hello Ziqi Ma, thanks for your interest in fixing this issue!

Yes, logic needs to be added to convert binary and octal strings to hex such that BigInteger.Parse can be used.

cesarzc:
> 
> 
> Seems like C#'s NumberStyles Enum (argument of BigInteger.Parse) only supports hex? Does that mean we need to write our own logic of converting binary/octal strings to hex?
> Well, or slice it up and parse each part as int...
> And are we assuming the number is positive? (seems so from code?)

Hello Ziqi Ma, thanks for your interest in fixing this issue!

Yes, logic needs to be added to convert binary and octal strings to hex such that BigInteger.Parse can be used.

samarsha:
> And are we assuming the number is positive? (seems so from code?)

I think this is correct. My understanding is that negative numbers are parsed as `negation-operator numeric-literal`, rather than parsing the `-` as part of the literal itself.

PR #673, opened 2020-10-13
adding support for binary and octal BigInt

ziqi-ma:
Addressing #610 
But note this currently only adds binary BigInt, not yet octal. I was thinking if this looks alright then I will first rewrite octal into binary string and then parse binary string to hex, which is less efficient than writing octal directly to hex - but that requires enumerating too many cases...


samarsha:
Thanks for the PR! I'll make sure it's reviewed soon.

src/QsCompiler/TextProcessor/QsExpressionParsing.fs
```diff
@@ -183,9 +183,41 @@ let internal numericLiteral =
...
         let isDouble = not nl.IsInteger && format = 10                  && nl.SuffixLength = 0 
         let returnWithRange kind = preturn (kind, range)
+        let binaryStringToHex binaryString:string = 
+            // first pad 0's so that length is multiple of 4, so we can match from left rather than right
+            let nZeroPad = (4 - string(binaryString).Length % 4) % 4 // if str.Length is already multiple of 4 then we don't pad
+            let rec padZero(curString, numPad) =
+                match numPad with
+                | 0 -> (curString,0)
+                | _ -> padZero(curString |>sprintf("0%s"), numPad-1)
+            let (binaryString, _) = padZero(binaryString, nZeroPad)
```
samarsha, line 196:
```suggestion
            let binaryString = binaryString.PadLeft (nZeroPad + String.length binaryString, '0')
```

src/QsCompiler/TextProcessor/QsExpressionParsing.fs
```diff
@@ -183,9 +183,41 @@ let internal numericLiteral =
...
+                | '0'::'1'::'1'::'0'::rest -> '6'::binaryListToHex rest
+                | '0'::'1'::'1'::'1'::rest -> '7'::binaryListToHex rest
+                | '1'::'0'::'0'::'0'::rest -> '8'::binaryListToHex rest
+                | '1'::'0'::'0'::'1'::rest -> '9'::binaryListToHex rest
+                | '1'::'0'::'1'::'0'::rest -> 'A'::binaryListToHex rest
+                | '1'::'0'::'1'::'1'::rest -> 'B'::binaryListToHex rest
+                | '1'::'1'::'0'::'0'::rest -> 'C'::binaryListToHex rest
+                | '1'::'1'::'0'::'1'::rest -> 'D'::binaryListToHex rest
+                | '1'::'1'::'1'::'0'::rest -> 'E'::binaryListToHex rest
+                | '1'::'1'::'1'::'1'::rest -> 'F'::binaryListToHex rest
```
samarsha, line 217: `Seq.chunkBySize 4` might come in handy here - I think something like `binaryString |> Seq.chunkBySize 4 |> Seq.map System.String` will give you a sequence of strings of length 4, so you could map over them with

```fsharp
// ...
| "1001" -> "9"
| "1010" -> "A"
// ...
```

At that point, you might be able to avoid enumerating all the cases by round-tripping from string-to-int-to-string like:

```fsharp
> System.Convert.ToByte("1010", 2).ToString "X";;
val it : string = "A"
```

samarsha (commented):
I think this looks like the right approach! Just added some comments for simplifying parts of the code.

src/QsCompiler/TextProcessor/QsExpressionParsing.fs
```diff
@@ -183,9 +183,41 @@ let internal numericLiteral =
             let trimmed = nl.String.TrimStart '+'
             if format = 10 || format = 0 then trimmed else trimmed.Substring 2 |> sprintf "0%s" // leading 0 is required to keep numbers positive
         let isInt    = nl.IsInteger     && format <> 0                  && nl.SuffixLength = 0 // any format is fine here
-        let isBigInt = nl.IsInteger     && (format = 10 || format = 16) && nl.SuffixLength = 1 && System.Char.ToUpperInvariant(nl.SuffixChar1) = 'L'
+        let isBigInt = nl.IsInteger     && format <> 0  && nl.SuffixLength = 1 && System.Char.ToUpperInvariant(nl.SuffixChar1) = 'L'
```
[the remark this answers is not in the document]
ziqi-ma: added octal support in new PR

src/QsCompiler/TextProcessor/QsExpressionParsing.fs
```diff
@@ -183,9 +183,41 @@ let internal numericLiteral =
...
         let isDouble = not nl.IsInteger && format = 10                  && nl.SuffixLength = 0 
         let returnWithRange kind = preturn (kind, range)
+        let binaryStringToHex binaryString:string = 
+            // first pad 0's so that length is multiple of 4, so we can match from left rather than right
+            let nZeroPad = (4 - string(binaryString).Length % 4) % 4 // if str.Length is already multiple of 4 then we don't pad
```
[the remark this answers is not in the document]
ziqi-ma: addressed

src/QsCompiler/TextProcessor/QsExpressionParsing.fs
ziqi-ma: good to know:) addressed

src/QsCompiler/TextProcessor/QsExpressionParsing.fs
ziqi-ma: latest PR no longer enumerates - but still need to enumerate octal->binary since "X" is the only keyword that's recognized

samarsha: Hmm, it looks like `System.Convert.ToString (x, 2)` will convert a number to a binary string. I wonder if there is a way to go directly from octal to hex, but the digits don't seem to line up nicely, so it's probably easier to go through binary.

samarsha: Oh, I think you can convert 4 digits of octal to 3 digits of hex. Since you can use `System.Convert` for both ways instead of enumerating the cases, that might be better?

ziqi-ma: I tried that first:) -  it doesn't work, since Byte only allows for values up to 255, and 4 digits of octal is much bigger than that

samarsha: You can use `System.Convert.ToInt32` (or `ToInt16`, or `ToInt64`, also):

```fsharp
> System.Convert.ToInt32 ("7777", 8);;
val it : int = 4095
```

ziqi-ma: Duh... never used F# before haha. Updated to this in latest PR.

src/QsCompiler/Tests.Compiler/SyntaxTests.fs
```diff
@@ -244,7 +244,13 @@ let ``Expression literal tests`` () =
...
         ("-1L",                   true,    toExpr (NEG (toBigInt "1")),                                           []); 
-        ("10000000000000000L",    true,    toBigInt "10000000000000000",                                          []); 
+        ("10000000000000000L",    true,    toBigInt "10000000000000000",                                          []);
+        ("0b111L",                true,    toBigInt "7",                                                          []);
+        ("0b1101L",               true,    toBigInt "13",                                                         []);
+        ("0b1100101011111110L",   true,    toBigInt "51966",                                                      []);
+        ("0o1L",                  true,    toBigInt "1",                                                          []);
+        ("0o105L",                true,    toBigInt "69",                                                         []);
+        ("0o12345L",              true,    toBigInt "5349",                                                      []);
```
samarsha, line 253: Added one more space for alignment:

```suggestion
        ("0o12345L",              true,    toBigInt "5349",                                                       []);
```

src/QsCompiler/TextProcessor/QsExpressionParsing.fs
```diff
@@ -183,9 +183,22 @@ let internal numericLiteral =
...
+        let isBigInt = nl.IsInteger     && format <> 0  && nl.SuffixLength = 1 && System.Char.ToUpperInvariant(nl.SuffixChar1) = 'L'
         let isDouble = not nl.IsInteger && format = 10                  && nl.SuffixLength = 0 
         let returnWithRange kind = preturn (kind, range)
+        let binaryToHex binaryString = 
+            // first pad 0's so that length is multiple of 4, so we can match from left rather than right
+            let nZeroPad = (4 - String.length binaryString % 4) % 4 // if str.Length is already multiple of 4 then we don't pad
+            let binaryString = binaryString.PadLeft (nZeroPad + String.length binaryString, '0')
+            // now match from left
+            binaryString |> Seq.chunkBySize 4 |> Seq.map (fun x -> System.Convert.ToByte(System.String x, 2).ToString "X") |> System.String.Concat
+        // seems like toString only takes keyword "X", so still need to enumerate for octal
```
samarsha, line 195:
```suggestion
```

src/QsCompiler/TextProcessor/QsExpressionParsing.fs
```diff
@@ -183,9 +183,22 @@ let internal numericLiteral =
...
+            let binaryString = binaryString.PadLeft (nZeroPad + String.length binaryString, '0')
+            // now match from left
+            binaryString |> Seq.chunkBySize 4 |> Seq.map (fun x -> System.Convert.ToByte(System.String x, 2).ToString "X") |> System.String.Concat
+        // seems like toString only takes keyword "X", so still need to enumerate for octal 
+        let octalToHex octalString =
+            // first pad 0's so that length is multiple of 4, so we can match from left rather than right
+            let nZeroPad = (4 - String.length octalString % 4) % 4 // if str.Length is already multiple of 4 then we don't pad
+            let octalString = octalString.PadLeft (nZeroPad + String.length octalString, '0')
+            // now match from left
+            octalString |> Seq.chunkBySize 4 |> Seq.map (fun x -> System.Convert.ToInt32(System.String x, 8).ToString "X") |> System.String.Concat
```
samarsha, line 201: `binaryToHex` and `octalToHex` are very similar - the only difference seems to be the parsed base (2 or 8). Can these be refactored into `let baseToHex base str = ...`, where `base` is either 2 or 8? (For simplicity we can just use `Convert.ToInt32` in both cases instead of `ToByte` for the binary one.)

ziqi-ma: Addressed

samarsha (approved):
It looks good, thanks for your contribution!

PR #675, opened 2020-10-14
Merge 'main' into 'feature/serialization-deserialization-performance'

cesarzc:
Merges `main` into `feature/serialization-deserialization-performance` to keep up to date during development.


ScottCarda-MS (approved):

samarsha:
I don't think it's possible to use a PR directly from main, since merging requires the source branch to be up-to-date with respect to the base branch. But main is protected so it can't be made up-to-date here.

cesarzc:
> 
> 
> I don't think it's possible to use a PR directly from main, since merging requires the source branch to be up-to-date with respect to the base branch. But main is protected so it can't be made up-to-date here.

Thanks for the pointer Sarah! I will create a merge branch locally and publish another PR.

PR #677, opened 2020-10-14
Monomorphization Access Modifiers

ScottCarda-MS:
Fixes a bug where access modifiers for concretizations generated during monomorphization were simply assigned the access of their generic parent. This causes an issue if any of the concretizing types are more restrictive than the original generic callable. This PR fixes this by taking the most restrictive of all the access modifiers between the original callable and all of the concretizing types.

This also adds a unit test to make sure type arguments are handled correctly by monomorphization.


src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/Monomorphization.qs
```diff
@@ -34,6 +34,39 @@ namespace Microsoft.Quantum.Testing.Monomorphization {
...
+    
+    // The first (Internal|Public) describes the access of the operation.
+    // The second (Internal|Public) describes the type that will be passed in.
+    internal operation InternalInternalFoo<'A>(x : 'A) : Unit { }
+    internal operation InternalPublicFoo<'A>(x : 'A) : Unit { }
+    operation PublicInternalFoo<'A>(x : 'A) : Unit { }
+    operation PublicPublicFoo<'A>(x : 'A) : Unit { }
+
+    @ EntryPoint()
+    operation TestAccessModifiers() : Unit {
```
bettinaheim, line 60: It would be good to also add a test for when the internal type is 
- in the return type
- in the signature of an argument that is passed
- in the signature of a callable that is returned
- the array base type (argument and return)
- the type of a tuple item (argument and return)

ScottCarda-MS: I've added these tests.

bettinaheim (approved):

PR #594, opened 2020-08-28
Initial work on documentation generation tool as rewrite step

cgranade:
This PR beings work on a new documentation generation tool that directly outputs DocFX-Flavored Markdown (DFM) from the Q# compilation process, instead of the YAML → template process currently used. This enables:

- Direct usage by third-party libraries (i.e..: no additional template needed).
- Resolving long-standing documentation limitations (e.g.: https://github.com/microsoft/qsharp-compiler/issues/113), and fixing bugs in templates (https://github.com/MicrosoftDocs/quantum-docs-pr/issues/991).
- Adapting output generation for new language features.

The compilation rewrite step in this PR also embeds documentation metadata into new attributes (PR on qsharp-runtime forthcoming), so that downstream tools such as IQ# can consume documentation without re-parsing documentation comments.

By way of implementation, this PR uses the existing documentation output tool as a parser to power the new compilation step.

NB: This PR replaces and supercedes https://github.com/microsoft/qsharp-runtime/pull/253.

Breaking changes:
- UIDs generated for Q# are no longer lower-cased, following standards for UIDs in other languages and frameworks. This will require fixing existing `@""` and `xref:` links to use fully-qualified names as UIDs instead of transforming to lowercase.

TODO:
- [x] Open corresponding PRs on qsharp-runtime, iqsharp, and QuantumKatas with changes needed to support this PR.
- [x] Finish extension method for converting type signatures to Markdown links (operations and tuples are currently `TODO`).
- [ ] Adapt QuantumLibraries documentation for UDTs to include new `/// # Named Items` documentation section.
- [ ] Adapt QuantumLibraries projects to provide package name.
- [x] Adapt symbol encoder in IQ# to utilize new information.
- [x] Package new step as nuget package.
- [x] Adapt Microsoft.Quantum.Sdk to automatically import new package when outputting documentation.
- [x] Generate/update TOC and namespace pages
- [x] Add new MSBuild properties to documentation.

Currently in draft for early feedback, targeting new feature branch.

PR #678, opened 2020-10-15
Minor fixes to new documentation generation tool

cgranade:
This PR makes a couple fixes to the new documentation generation tool that were discovered using beta packages from the previous PR (#594):

-  Default package ID not being set correctly.
-  Output path not being passed through to assembly constants due to issue with MSBuild conditional attributes.
- ToMarkdownLink falling through to `"__invalid__"` for type parameters as well as for invalid types.


swernli (approved):
LGTM!

Issue #679, opened 2020-10-15
Assembly property values set from within response files cannot end with quote marks


cgranade:
**Describe the bug**

When using the `--assembly-properties` command line argument to `qsc.dll` from within a response file, assembly property values cannot end with a double-quote character (`"`").

**To Reproduce**

1. In a `qsc.rsp` file, add an argument to the `--assembly-properties` option that ends with a double-quote character. E.g.:
   
    ```
    --assembly-properties ProcessorArchitecture:Unspecified QsharpOutputType:QsharpLibrary ExecutionTarget:Any DocsPackageId:Microsoft.Quantum.Standard DocsOutputPath:"C:\Users\cgran\source\repos\microsoft\quantumlibraries\Standard\src\docs"
    ```

2. Invoke `qsc.dll` with the given response file.

**Expected behavior**

The `DocsOutputPath` assembly property to be set to `"C:\Users\cgran\source\repos\microsoft\quantumlibraries\Standard\src\docs"`. Due to https://github.com/microsoft/qsharp-compiler/blob/feature/docgen/src/QsCompiler/CommandLineTool/Commands/Build.cs#L136, the closing `"` is stripped by the call to `Trim`.

**System information**

- Beta package version: Microsoft.Quantum.Sdk/0.12.201016171-beta

cgranade:
As a side note, it also appears that assembly properties cannot contain a colon due to the call to `Split(":")` in Options.cs at L93.

PR #680, opened 2020-10-15
Fix issues with using absolute paths in docgen

cgranade:
The previous version of the new docgen tool works correctly when using relative paths, but fails when using absolute paths on Windows. Upon investigation, this comes down to three distinct issues:

- The docgen path was not properly quoted in MSBuild properties.
- `qsc.dll` was stripping one of the two quotation marks (#679)
- `qsc.dll` was truncating assembly properties at the first colon (#679)

This PR fixes all three issues, and adds more diagnostics to the documentation writer component to help understand future issues.


swernli (approved):
LGTM

PR #684, opened 2020-10-17
Fix reversal of nested adjoint expressions

### LINKED ISSUE #615: Order of evaluation in expressions can generate incorrect adjoints
**Describe the bug**

When auto-generating an adjoint specialization, Q# reverses the order of statements, but does not reverse the order of evaluation in expressions. Because evaluation can have side effects, this means auto-generated adjoint specializations may apply operations in an incorrect order.

**To Reproduce**

```qsharp
operation Evaluator(a : Unit, b : Unit) : Unit is Adj { }

operation DoubleCNOT(a : Qubit, b : Qubit) : Unit is Adj {
    Evaluator(CNOT(a, b), CNOT(b, a));
}

@EntryPoint()
operation Demo() : Unit {
    using ((a, b) = (Qubit(), Qubit())) {
        within {
            X(a);
            DoubleCNOT(a, b);
        } apply { }
        Message($"({M(a)}, {M(b)})");
    }
}
```

Outputs `(Zero, One)` because the `within` block is not uncomputed correctly.

**Expected behavior**

The code above should output `(Zero, Zero)`, or it should not be allowed at all.

**System information**

Q# compiler v0.12.20082513

swernli:
When generating an adjoint operation, first lift any nested expressions that are operation calls into their own lines, replacing the inline call with Unit, ie: `( )`, since all adjointable operations must return Unit anyway. Doing this lifting before the statements in the scope get reversed ensures that they will be properly reversed instead of left in-order in the in-line expression.

Fixes #615


src/QsCompiler/Transformations/FunctorGeneration.cs
```diff
@@ -148,6 +152,99 @@ public override QsStatementKind OnConjugation(QsConjugation stm)
...
+            public override ExpressionKind OnOperationCall(TypedExpression method, TypedExpression arg)
+            {
+                // An operation in an adjoint scope must return Unit, so lift the operation to be an
+                // an additional statement and then replace it with Unit.
+                this.SharedState.AdditionalStatements.Add(
+                    new QsStatement(
+                        QsStatementKind.NewQsExpressionStatement(
+                            new TypedExpression(
+                                base.OnOperationCall(method, arg),
+                                TypeArgsResolution.Empty, // TODO(swernli): How do I determine the right argument types here?
```
swernli, line 235: This works for the cases in the test, but doesn't seem right. How do I fill in the right arg types here? Do I need to, or is that optional? Looking at other code, it usually either fills in `Empty` like I'm doing now, or collects the types from the higher level expression, but I can't seem to get access to those data structures here.

src/QsCompiler/Transformations/FunctorGeneration.cs
```diff
@@ -148,6 +152,99 @@ public override QsStatementKind OnConjugation(QsConjugation stm)
...
     }
 
+    /// <summary>
+    /// Scope transformation that splits any nested operation calls into separate statements so
+    /// that they can be properly reversed. This is necessary to avoid out of order execution of the
+    /// automatically generated adjoint. It is safe to do because an adjointable operation must return
+    /// Unit, so any nested calls can  be replaced by Unit and those calls moved to separate,
+    /// ordered statements.
+    /// </summary>
+    public class LiftOperationCalls
```
ScottCarda-MS, line 162: nit: "Lift" is used elsewhere in the transformations to refer to when a piece of code is "lifted" into its own generated callable. Using that term here may cause confusion. Is there another term that could be used here? Maybe "Fan Out" or "Unnest"?

swernli: That’s a good point, I don’t want to conflate the meaning. There’s got to be a term for this, but so far “unnest” is my favorite.

swernli: At least one resource I found refers to this as “extracting” so maybe something like “ExtractNestedOperationCalls” would be nice and descriptive. 

https://refactoring.guru/extract-variable

src/QsCompiler/Transformations/FunctorGeneration.cs
ScottCarda-MS: You may need to store the current expression in the shared state as you traverse down to this point in the syntax tree (and maybe the current statement as well), so that you can populate these fields with whatever the original values were. I would do this populating with the type arg resolutions, the return type (even though it currently always has to be unit), the inferred information, and the location. The location is important for accurately reporting diagnostics (even though these are generated statements, they represent original code). I would also copy over the statement location for the generated statements as well for the same reason, but the comment and local declarations only need to be on the original, I think.

ScottCarda-MS: As a reference for something similar the concrete call graph walker has to keep track of if it is looking at expressions inside a controlled functor application: https://github.com/microsoft/qsharp-compiler/blob/79503a2a3ce31e9beeb7c65fc4d24fdc0c2cbdf4/src/QsCompiler/Transformations/CallGraph/ConcreteCallGraphWalker.cs#L276
This could be similar to how you would need to keep track of the current expression, due to expressions being possibly nested. Instead of `this.SharedState.HasControlledDependency` it would be something like `this.SharedState.CurrentExpression` and instead of being in `OnControlledApplication` it would be in an override for `OnExpression`.

swernli: Thanks for the tip! I think that worked, and I've got a new version that keeps track of the currently processing execution.

src/QsCompiler/Transformations/FunctorGeneration.cs
```diff
@@ -148,6 +152,125 @@ public override QsStatementKind OnConjugation(QsConjugation stm)
...
+
+        public ExtractNestedOperationCalls()
+        : base(new TransformationsState())
+        {
+            this.Statements = new ExtractNestedOperationCalls.StatementTransformation(this);
+            this.Expressions = new ExtractNestedOperationCalls.ExpressionTransformation(this);
+            this.ExpressionKinds = new ExtractNestedOperationCalls.ExpressionKindTransformation(this);
+        }
+
+        public class StatementTransformation
```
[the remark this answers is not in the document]
ScottCarda-MS: This goes for the expression kind and expression transformation classes as well.

src/QsCompiler/Transformations/FunctorGeneration.cs
```diff
@@ -148,6 +152,125 @@ public override QsStatementKind OnConjugation(QsConjugation stm)
...
+        : StatementTransformation<TransformationsState>
+        {
+            public StatementTransformation(SyntaxTreeTransformation<TransformationsState> parent)
```
ScottCarda-MS, line 186: Do you need this constructor? I suppose if the class is public, giving people the option is nice, but if you make it private as per my other suggestion, then I think this constructor may go unused.
This goes for the expression kind and expression transformations class constructors as well.

src/QsCompiler/Transformations/FunctorGeneration.cs
```diff
@@ -148,6 +152,125 @@ public override QsStatementKind OnConjugation(QsConjugation stm)
...
+                    new QsStatement(
+                        QsStatementKind.NewQsExpressionStatement(
+                            new TypedExpression(
+                                base.OnOperationCall(method, arg),
+                                curExpression.TypeParameterResolutions.Select(x => Tuple.Create(x.Key.Item1, x.Key.Item2, x.Value)).ToImmutableArray(),
+                                curExpression.ResolvedType,
+                                curExpression.InferredInformation,
+                                curExpression.Range)),
+                        LocalDeclarations.Empty,
+                        QsNullable<QsLocation>.Null,
```
ScottCarda-MS, line 266: nit: Could you keep track of and fill out the statement location the same way you did for the expression location? Keeping location information around for as long as possible can help transformations down the line that my want/need to use them.

src/QsCompiler/Transformations/FunctorGeneration.cs
```diff
@@ -148,6 +152,125 @@ public override QsStatementKind OnConjugation(QsConjugation stm)
...
     }
 
+    /// <summary>
+    /// Scope transformation that splits any nested operation calls into separate statements so
+    /// that they can be properly reversed. This is necessary to avoid out of order execution of the
+    /// automatically generated adjoint. It is safe to do because an adjointable operation must return
+    /// Unit, so any nested calls can be replaced by Unit and those calls moved to separate,
+    /// ordered statements.
+    /// </summary>
+    public class ExtractNestedOperationCalls
```
ScottCarda-MS, line 162: nit: In order to reduce additions to the public API, one trick that I've been doing is wrapping the transformation class in a public static class, making the transformation class private and adding a public static method to the wrapper class that instantiates and uses the (now private) transformation class in only the way that you intend it to be used. This removes the TransformationState nested class from the public API and ensures that the transformation is only used as intended, reducing the API addition to a single static method.
See Example: https://github.com/microsoft/qsharp-compiler/blob/79503a2a3ce31e9beeb7c65fc4d24fdc0c2cbdf4/src/QsCompiler/Transformations/ClassicallyControlled.cs#L31

ScottCarda-MS (commented):
Looks good! I've added some suggestions for reducing the additions to the public API.

src/QsCompiler/Transformations/FunctorGeneration.cs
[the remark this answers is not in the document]
swernli: I can make these private and also make the `ExtractNestedOperationCalls` that wraps them internal. In fact, all the top level transformations defined in this file can be changed to internal.

src/QsCompiler/Transformations/FunctorGeneration.cs
swernli: I think you're right... let me try taking those out.

swernli: Actually, for some reason it's the other constructor I can remove and this one needs to be kept. It will compile but fail the tests, even though I update the calls above to use the remaining constructor.

src/QsCompiler/Transformations/FunctorGeneration.cs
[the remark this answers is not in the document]
ScottCarda-MS: If the transformation as a whole is internal, than the TransformationState class can also be internal, but ideally there would only be one way to use the transformation and we could limit the use to just that one way with a static wrapper class. Please see my other comment for that.

src/QsCompiler/Transformations/FunctorGeneration.cs
swernli: In this case, these transformations are already only used in the static methods in CodeTransformations.cs, so I just updated the whole file to be internal and now we don't have to worry about them being part of the public API surface.

src/QsCompiler/Transformations/FunctorGeneration.cs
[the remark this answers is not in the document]
ScottCarda-MS: Also I'm not sure about changing the accessibility of the other existing transformations in this file. Ideally they would have been made with more restrictive access initially, but now that they have been released as public others may have taken dependencies on them. There may also be some other reasons why they are public that I just don't know about. At any rate, if you remove things from the existing public API, please make sure to get @bettinaheim 's approval first.

src/QsCompiler/Transformations/FunctorGeneration.cs
ScottCarda-MS: Please check with @bettinaheim about making the other existing transformations in this file internal before merging. I've mentioned that in the other comment, so I'll close this one out and we can consolidate the conversation there.

src/QsCompiler/Transformations/FunctorGeneration.cs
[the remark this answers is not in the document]
swernli: Having another static wrapper class would be redundant, since there already exists one in CodeTransformations.cs, and the `GenerateAdjoint` static method is the only place this transformation is used. Unfortunately I can't make it all the way private because they are in different namespaces. Once these classes are internal, they are not in the public API anymore, so it is really that much different to have them be private if it requires adding even more indirection on top of what already exists?

src/QsCompiler/Transformations/FunctorGeneration.cs
ScottCarda-MS: Oh whoops. You're right, I commented the wrong one of the two types of constructors. You have removed the one I meant.

src/QsCompiler/Transformations/FunctorGeneration.cs
swernli: Good idea! I'm working on that now...

src/QsCompiler/Transformations/FunctorGeneration.cs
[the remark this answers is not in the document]
ScottCarda-MS: The difference between internal vs private is less of an issue here I think because then it scopes any breaking changes to just breaking our own code, which we can manage. Having a single static method as an entry point prevents our own code from jumping into the middle of the transformation, which could be important here as this is a stateful transformation. For example we could do something like `new ExtractNestedOperationCalls().ExpressionKind.OnOperationCall(method, arg);` which would be confusing because we would not have any `CurrentExpression` to utilize. As an internal transformation though, that would only be on us to make sure not to misuse the transformation.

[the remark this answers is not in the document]
ScottCarda-MS: It might also be unnecessary overhead, as you suggest, just to police our own code in this case. It's up to you which way you want to go with it. I'll close this thread out and make a separate thread regarding the access modifiers for existing transformations.

src/QsCompiler/Transformations/FunctorGeneration.cs
```diff
@@ -15,13 +15,17 @@
...
+    using TypeArgsResolution = ImmutableArray<Tuple<QsQualifiedName, NonNullable<string>, ResolvedType>>;
+
     /// <summary>
     /// Scope transformation that replaces each operation call within a given scope
     /// with a call to the operation after application of the functor given on initialization.
     /// The default values used for auto-generation will be used for the additional functor arguments.
     /// Additional functor arguments will be added to the list of defined variables for each scope.
     /// </summary>
-    public class ApplyFunctorToOperationCalls
+    internal class ApplyFunctorToOperationCalls
```
ScottCarda-MS, line 28: Please check with @bettinaheim about changing the access modifiers for the existing transformations from public to internal.

ScottCarda-MS (approved):

src/QsCompiler/Transformations/FunctorGeneration.cs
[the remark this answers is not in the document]
swernli: @bettinaheim To weigh in on the design considerations here...

One option would be to move the static methods in CodeTransformations.cs like `GenerateAdjoint` into FunctorGeneration.cs so that those entrypoints and the code that implements the transformations can be part of the same static class. Then we'd be able to get closer to the design used in other places without needing an intermediate wrapper class. Would that work?

src/QsCompiler/Transformations/FunctorGeneration.cs
swernli: True... while we don't expect anything else to be using them, it is technically a change in the API.

bettinaheim: Let's properly deprecate things when we do that. In this case, while it seems like it is no longer in use by anything outside the compiler, my inclination would be to keep it as part of the public API. Or do you see any reason why that particular piece would likely change in the future? Same for IgnoreOuterBlockInConjugations<T>.
As far as AddVariableDeclarations<T> is concerned, it may make sense to properly deprecate it here, and add an internal class to the BasicTransformations namespace instead.

src/QsCompiler/Transformations/FunctorGeneration.cs
```diff
@@ -148,6 +152,123 @@ public override QsStatementKind OnConjugation(QsConjugation stm)
...
+            /// </summary>
+            public QsNullable<QsLocation> StatementLocation = QsNullable<QsLocation>.Null;
+        }
+
+        public ExtractNestedOperationCalls()
+        : base(new TransformationsState())
+        {
+            this.Statements = new StatementTransformation(this);
+            this.Expressions = new ExpressionTransformation(this);
+            this.ExpressionKinds = new ExpressionKindTransformation(this);
```
bettinaheim, line 191: Since you are not doing any type transformations, it would be nice to disable them so the transformation doesn't unnecessarily recur into and rebuild all types. 
```suggestion
            this.ExpressionKinds = new ExpressionKindTransformation(this);
            this.Types = new TypeTransformation<TransformationState>(this, TransformationOptions.Disabled);
```

src/QsCompiler/Transformations/FunctorGeneration.cs
swernli: I'll put `ApplyFunctorToOperationCalls` and `IgnoreOuterBlockInConjugations` back to public. I think I might need a little more clarification on how to properly deprecate `AddVariableDeclarations`; do I add a decorator on this one and copy the implementation to the internal replacement? I'm fine with just putting this one back to public and leaving the API surface as is until we want to do a more formal review of the compiler public API, if that would be more straightforward for now.

bettinaheim: The suggestion is to copy the implementation for AddVariableDeclarations to the BasicTransformations (it can be kept internal), have this one here just inherit from the one in BasicTransformations to keep it functional, and decorate it with the Obsolete attribute.

swernli: Unfortunately, I can't have the one in BasicTransformations be internal while having one here inherit and be public, because a public class cannot inherit from an internal class:
`FunctorGeneration.cs(116,18): error CS0060: Inconsistent accessibility: base class 'AddVariableDeclarations<T>' is less accessible than class 'AddVariableDeclarations<T>'`
So they either both have to be public or both have to be internal, and once they have the same accessibility it seems to defeat the point of duplicating it, since we did that just to deprecate the public one. Should I just have two copies of the same implementation?

bettinaheim: Ah, right, that's unfortunate. I guess duplicating will have to do if we want to remove it from the public API. I think that should be fine. I'll create a task to make sure we are not forgetting about removing it.

PR #686, opened 2020-10-17
Create folders and split out large files in Core

samarsha:
* Created a new `Transformations` folder to put all the base transformation files.
* The types in `SymbolTable.fs` were marked as mutually recursive using `and` and defined all in one file, but this was not necessary. I split them out into 3 separate files and moved them into the new `SymbolTable` folder.


ScottCarda-MS (approved):
Nice change! A little more organization can go a long ways.

bettinaheim (approved):

PR #692, opened 2020-10-20
Bond translators integration

cesarzc:
This change integrates Bond translators.


src/QsCompiler/CompilationManager/AssemblyLoader.cs
```diff
@@ -98,13 +106,18 @@ public static bool LoadReferencedAssembly(Uri asm, out References.Headers header
         public static bool LoadSyntaxTree(
             Stream stream,
             [NotNullWhen(true)] out QsCompilation? compilation,
+            SyntaxTreeSerializationProtocol serializationProtocol = SyntaxTreeSerializationProtocol.NewtonsoftBinary,
```
bettinaheim, line 109: There shouldn't be a default value here, since the right protocol depends on the stream, so whoever gets the stream needs to know the protocol.

src/QsCompiler/CompilationManager/AssemblyLoader.cs
```diff
@@ -25,6 +27,12 @@ namespace Microsoft.Quantum.QsCompiler
     /// </summary>
     public static class AssemblyLoader
     {
+        public enum SyntaxTreeSerializationProtocol
```
bettinaheim, line 30: Hm, I would love to eliminate this from the public API; while we will want to keep the Newtonsoft infrastructure for backwards compatibility, I also think we should deprecate it in the near future (though we should wait quite a while before we actually remove it).

src/QsCompiler/CompilationManager/AssemblyLoader.cs
bettinaheim: But looking more closely at the code, I don't thing the two ways of loading should be unified in the same method. I think LoadSyntaxTree should be the loading via Bond (i.e. it should always be the loading of what is currently generated by the compiler). If we for now want to support processing old dlls then I would add a separate method for loading the old version. It may make sense to consider if this method should be taking something else than a stream, in which case both methods could have the same name but a different first argument.

src/QsCompiler/CompilationManager/AssemblyLoader.cs
cesarzc: I removed this and now the backwards compatibility with Newtonsoft is just encapsulated in the `FromResource` private method.

src/QsCompiler/CompilationManager/AssemblyLoader.cs
cesarzc: I went with you recommendation and created another `LoadSyntaxTree` method that receives a byte array as first argument. Should we mark the one that receives a stream as deprecated?

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -977,12 +978,18 @@ private bool SerializeSyntaxTree(MemoryStream ms)
                 return false;
             }
 
-            using var writer = new BsonDataWriter(ms) { CloseOutput = false };
             var fromSources = this.CompilationOutput.Namespaces.Select(ns => FilterBySourceFile.Apply(ns, s => s.Value.EndsWith(".qs")));
             var compilation = new QsCompilation(fromSources.ToImmutableArray(), this.CompilationOutput.EntryPoints);
             try
             {
-                Json.Serializer.Serialize(writer, compilation);
```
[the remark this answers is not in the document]
cesarzc: The new centralized place for serialization/deserialization methods is `Protocols.cs` in the `BondSchemas` project.

src/QsCompiler/CompilationManager/AssemblyLoader.cs
bettinaheim: Yes, I think that makes sense.

src/QsCompiler/DataStructures/ReservedKeywords.fs
```diff
@@ -276,6 +276,7 @@ module CommandLineArguments =
 /// contains specific names used within Q# dlls
 module DotnetCoreDll = 
     let ResourceName = "__qsharp_data__.bson"
```
bettinaheim, line 278: Maybe we should deprecate this one, too, though with the same comment that it is still in use for older dlls.

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -965,32 +962,27 @@ private References LoadAssemblies(IEnumerable<string> refs, bool loadTestNames,
...
         /// </summary>
-        private bool SerializeSyntaxTree(MemoryStream ms)
+        private bool WriteSyntaxTreeSerialization(MemoryStream ms)
         {
             void LogError() => this.LogAndUpdate(
                 ref this.compilationStatus.Serialization, ErrorCode.SerializationFailed, Enumerable.Empty<string>());
 
-            this.compilationStatus.Serialization = 0;
-            if (this.CompilationOutput == null)
+            void LogExceptionAndError(Exception ex)
```
bettinaheim, line 970: This looks like it is still work in progress, so I'll take a look when it's done.

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -977,12 +978,18 @@ private bool SerializeSyntaxTree(MemoryStream ms)
...
             {
-                Json.Serializer.Serialize(writer, compilation);
+                var outputBuffer = new OutputBuffer();
+                var serializer = BondSchemas.Factory.GetFastBinarySerializer();
+                var fastBinaryWriter = new FastBinaryWriter<OutputBuffer>(outputBuffer);
+                var bondCompilation = BondSchemas.BondSchemaTranslator.CreateBondCompilation(compilation);
+                serializer.Serialize(bondCompilation, fastBinaryWriter);
+                ms.Write(outputBuffer.Data);
+                ms.Flush();
+                ms.Position = 0;
```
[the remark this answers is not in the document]
cesarzc: I created a `WriteBinary` API that is analogous to the `ReadBinary` one to make sure that for any given version, both APIs are compatible.

Tests can be found here: `src/QsCompiler/Tests.Compiler/CompilationLoaderTests.fs`

src/QsCompiler/CompilationManager/AssemblyLoader.cs
cesarzc: Deprecated the API with the appropriate message.

src/QsCompiler/Compiler/CompilationLoader.cs
cesarzc: Ready to review!

src/QsCompiler/DataStructures/ReservedKeywords.fs
cesarzc: Deprecated with an appropriate message.

src/QsCompiler/BondSchemas/BondSchemas.csproj
```diff
@@ -1,7 +1,11 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
+  <Import Project="..\..\Common\AssemblyCommon.props" />
+
   <PropertyGroup>
+    <OutputType>Library</OutputType>
     <TargetFramework>netstandard2.1</TargetFramework>
+    <AssemblyName>Microsoft.Quantum.BondSchemas</AssemblyName>
```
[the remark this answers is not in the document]
cesarzc: I will add this as part of the merge into `main` since the feature branch currently does not have the latest version of `manifest.ps1`.

src/QsCompiler/CompilationManager/AssemblyLoader.cs
```diff
@@ -91,10 +91,35 @@ public static bool LoadReferencedAssembly(Uri asm, out References.Headers header
         // tools for loading the compiled syntax tree from the dll resource (later setup for shipping Q# libraries)
 
         /// <summary>
-        /// Given a stream containing the binary representation of compiled Q# code, returns the corresponding Q# compilation.
-        /// Returns true if the compilation could be deserialized without throwing an exception, and false otherwise.
+        /// Given a binary representation of compiled Q# code, returns the corresponding Q# compilation.
+        /// Returns true if the compilation could be deserialized without throwing an exception, and it is not empty. False otherwise.
```
bettinaheim, line 95: I think the phrasing is a bit misleading; the method returns true even if the compilation has no namespaces but just an empty array. It only checks if that array can be accessed without throwing. 
```suggestion
        /// Returns true if the compilation could be deserialized without throwing an exception, and it is properly instantiated. False otherwise.
```

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -965,32 +962,27 @@ private References LoadAssemblies(IEnumerable<string> refs, bool loadTestNames,
...
+        private bool WriteSyntaxTreeSerialization(MemoryStream ms)
         {
             void LogError() => this.LogAndUpdate(
                 ref this.compilationStatus.Serialization, ErrorCode.SerializationFailed, Enumerable.Empty<string>());
 
-            this.compilationStatus.Serialization = 0;
-            if (this.CompilationOutput == null)
+            void LogExceptionAndError(Exception ex)
             {
+                this.LogAndUpdate(ref this.compilationStatus.TargetSpecificReplacements, ex);
```
bettinaheim, line 972:
```suggestion
                this.LogAndUpdate(ref this.compilationStatus.Serialization, ex);

```

bettinaheim (approved):
Aside from the two edits I suggested, looks good.

src/QsCompiler/CompilationManager/AssemblyLoader.cs
```diff
@@ -91,10 +91,35 @@ public static bool LoadReferencedAssembly(Uri asm, out References.Headers header
...
+                onDeserializationException?.Invoke(ex);
+                return false;
+            }
+
+            return compilation != null && IsValidCompilation(compilation);
+        }
+
+        /// <summary>
+        /// Given a stream containing a binary representation of compiled Q# code, returns the corresponding Q# compilation.
+        /// Returns true if the compilation could be deserialized without throwing an exception, and it is not empty. False otherwise.
```
cesarzc, line 119:
```suggestion
        /// Returns true if the compilation could be deserialized without throwing an exception, and it is properly instantiated. False otherwise.
```

PR #694, opened 2020-10-20
Concrete Call Graph Support Self-Adjoint Generator Directives

ScottCarda-MS:
The concrete call graph assumes that the functor generation has been applied to the given compilation and that all generator directives have been processed and removed. However the functor generation does not remove the self-inverse generator directives and leaves them in the compilation. The concrete call graph is updated in the PR to accommodate the self-inverse generator directives by adding dependencies from adjoint and controlled adjoint specializations marked as 'self' to the body and controlled specializations, respectively, of the same callable.

### DIFF
--- a/src/QsCompiler/Transformations/CallGraph/ConcreteCallGraphWalker.cs
+++ b/src/QsCompiler/Transformations/CallGraph/ConcreteCallGraphWalker.cs
@@ -170,6 +170,7 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.CallGraphWalker
                 public bool HasAdjointDependency = false;
                 public bool HasControlledDependency = false;
                 public Func<QsQualifiedName, IEnumerable<QsSpecializationKind>> GetSpecializationKinds = _ => Enumerable.Empty<QsSpecializationKind>();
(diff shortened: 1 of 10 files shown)


src/QsCompiler/Transformations/CallGraph/ConcreteCallGraphWalker.cs
```diff
@@ -245,7 +267,32 @@ public NamespaceWalker(SyntaxTreeTransformation<TransformationState> parent) : b
...
+                        else
+                        {
+                            throw new ArgumentException("\"self\" can only be used on adjoint and controlled adjoint specializations.");
+                        }
+
+                        return directive;
+                    }
+                    else
+                    {
+                        throw new ArgumentException("Encountered unresolved generated specialization while constructing concrete call graph.");
```
bettinaheim, line 294: There should be a doc comment somewhere that needs to be updated to reflect that a "self" generation directive is ok.

bettinaheim (approved):

src/QsCompiler/Transformations/CallGraph/ConcreteCallGraphWalker.cs
ScottCarda-MS: Updated the doc comment for the concrete call graph walker.

PR #695, opened 2020-10-21
Catch docgen feature up to main.

cgranade:


samarsha (approved):

cgranade:
The test failures look to be happening only in IQ# and don't seem to be related to this PR. @bettinaheim, would you be good with merging this to the feature branch so that we can start e2e builds there and get that ready for merging to `main`?

rmshaffer:
For more context, the IQ# failure appears to be related to the recent cross-repo change between `qsharp-compiler` and `iqsharp`, and this branch somehow leading to an inconsistent state across the two repos in the E2E build. I'd recommend we merge this as-is to get the feature branch up-to-date, and then rely on the E2E build when merging the feature branch back to `main`.

rmshaffer (approved):

cgranade:
Merging this PR with failing e2e builds as per @rmshaffer's diagnosis above and with @bettinaheim's signoff.

PR #696, opened 2020-10-21
Complete new documentation generation tool.

cgranade:
This PR merges the work started with #594 to main.

PR #697, opened 2020-10-21
Finish documentation generation feature.

cgranade:
This PR merges the work started with #594 to main. Note that this is the same PR as #696, but from an unprotected branch to allow updating from main more easily.


ricardo-espinoza (approved):

anpaz (approved):

src/QuantumSdk/Sdk/Sdk.props
```diff
@@ -23,26 +23,39 @@
...
+    <PackageReference Condition="'$(ResolvedProcessorArchitecture)' == 'QCIProcessor' And $(IncludeProviderPackages)" Include="Microsoft.Quantum.Providers.QCI" Version="0.12.20090427-beta" />
     <!-- TODO: UNCOMMENT THESE ONCE WE HAVE THE CORRESPONDING PACKAGES. -->
     <!--PackageReference Condition="'$(ResolvedQuantumIntrinsics)' == 'OpenQASM'" Include="Microsoft.Quantum.Intrinsics.OpenQASM" Version="0.11.2006.2118-beta" IsTargetPackage='true' />
     <PackageReference Condition="'$(ResolvedQuantumIntrinsics)' == 'ExtendedQASM'" Include="Microsoft.Quantum.Intrinsics.ExtendedQASM" Version="0.11.2006.2118-beta" IsTargetPackage='true' /-->
   </ItemGroup>
 
+  <!-- Include the documentation generator package only if generating docs using Markdown output. -->
+  <ItemGroup>
+    <PackageReference Condition="$(QsharpDocsGeneration)" 
+                      Include="Microsoft.Quantum.DocumentationGenerator"
```
samarsha, line 40: Has `Microsoft.Quantum.DocumentationGenerator` been added to the package list in the QDK build scripts? It doesn't look like the version number is getting replaced with the current build number [here](https://github.com/microsoft/qsharp-compiler/blob/pull/v0.12.2010.17080/src/QuantumSdk/Sdk/Sdk.props#L39) or [here](https://github.com/microsoft/qsharp-compiler/blob/pull/v0.12.2010.17072/src/QuantumSdk/Sdk/Sdk.props#L39). (In my experience, the build script regex also doesn't work unless the `Include` and `Version` attributes are on the same line.)

cgranade: There's an open PR on the build repo to do that, but it's blocked on completing this PR. Thanks for the note about Include and Version being on the same line, I'll fix that so that we can merge with them on the same line, then. I keep forgetting that it's a regex-based script and that it doesn't look at the semantic content of the XML...

PR #704, opened 2020-10-23
Call Graph Clears Type-Parameter Resolutions before Processing a Statement

ScottCarda-MS:
Fixes a bug in call graphs where type parameter resolutions from previous statements would be carried over to subsequent statements, causing incorrect type parameter resolutions to occur in the construction of the graph.


cesarzc (approved):
Looks good to me.

PR #710, opened 2020-10-27
Parentheses on left-hand side of curried call are optional

samarsha:
This PR makes the parentheses on the left-hand side of a curried call optional, so these are equivalent:

```qsharp
((Foo(x))(y))(z)
Foo(x)(y)(z)
```


cgranade:
💕!!!

src/QsCompiler/TextProcessor/QsExpressionParsing.fs
```diff
@@ -383,10 +383,11 @@ let private argumentTuple =
...
 /// meaning they process the left-most part of the call-like expression and thus need to be evaluated *after* the callLikeExpr parser. 
-let internal callLikeExpr = 
-    attempt ((itemAccessExpr <|> identifier <|> tupledItem expr) .>>. argumentTuple) // identifier needs to come *after* arrayItemExpr
-    |>> fun (callable, arg) -> applyBinary CallLikeExpression () callable arg
-
+let internal callLikeExpr =
+    // identifier needs to come *after* arrayItemExpr
+    itemAccessExpr <|> identifier <|> tupledItem expr .>>. many1 argumentTuple
+    |>> List.Cons
+    |>> List.reduce (applyBinary CallLikeExpression ())
```
samarsha, line 390: So I think the reason why this change was simple is because it doesn't support item access directly after a call, like `Foo()[0]` - it needs to be tupled like `(Foo())[0]`. @bettinaheim Do we want to support that without parentheses as well?

samarsha: I believe supporting that wouldn't be *too* difficult - we could do something like `many1 (argumentTuple <|> namedItem <|> arrayIndex)` and reduce that list into a tree.

bettinaheim: Fundamentally, there are three options: 
Option 1: 
Item access and call combinator have the same precedence and lower precedence than modifiers (functors and unwrap). 
This would allow `Adjoint Unitary()` but the currently supported `Adjoint opArray[i]` would no longer be possible. 
Option 2:
Item access and call combinator have the same precedence and higher precedence than modifiers. 
This is definitively undesirable since then `Adjoint Unitary()` would no longer work.
Option 3: 
Item access and call combinator retain the precedence that they currently have. 
This is the option that is implemented, and I believe the best option. While that disallows to freely combine item access and call combinators (e.g. GetUnitary()[i] is not supported as you point out above), the two scenarios  `Adjoint Unitary()` and  `Adjoint opArray[i]` that I believe to be more important are supported.

msoeken:
Will this also work for UDT member access? For example:

```qsharp
let numbers = [Complex(1.0, 2.0), Complex(2.0, 3.0)];
let firstReal = Head(numbers)::Real;
```

samarsha:
No, both named and array items must use parentheses:

```qsharp
(Head(numbers))::Real
(Tail(numbers))[0]
```

bettinaheim:
@msoeken The choice here is between supporting `Adjoint opArray[i]` or supporting `Head(numbers)::Real;` since I think that both named item access and numbered item access should have the same precedence (hence we can support either one but not both).

bettinaheim (approved):

samarsha:
@bettinaheim Would requiring parentheses in `Adjoint (opArray[i])` be that bad? Is that a very common case?

msoeken:
Thanks for the additional info @smarsha and @bettinaheim. From my own experience, I tend to add parentheses around `Adjoint op` and forget them around variables before member or array access.

samarsha:
Merging this for now since I expect any future changes to the parser related to functors/item access will preserve the repeated call syntax without parentheses.

PR #717, opened 2020-10-30
Include name of required runtime capability in diagnostic message

### LINKED ISSUE #716: More specific error messages for runtime capability verification
Error messages for calling operations that require runtime capabilities that aren't supported by the execution target look like:

> The callable Teleport requires runtime capabilities that are not supported by the target IonQProcessor.

It would be better if the message mentioned the specific capabilities that aren't supported, such as BasicMeasurementFeedback.

N/A

**Additional context**
N/A

samarsha:
Closes #716.


cesarzc (approved):
Great to have the specific capability that is not supported in the message.

PR #719, opened 2020-11-03
Fix package links in new documentation generator.

cgranade:
This PR fixes the "Package" link that appears in the output from the new Markdown-based documentation tool. In particular, this PR fixes a typo in Sdk.targets that caused the package ID to not be passed to the rewrite step correctly, and adds a check for `IsNullOrWhitespace` to ensure that empty package IDs are treated the same way as `null`.


samarsha (approved):

cgranade:
Looks like CI failed due to a temporary outage with a required NuGet feed. Will rerun.

cgranade:
@samarsha: Thanks for the review! Would you be willing to take a quick look again? @msoeken brought another minor issue to my awareness such that I modified the PR to address that as well. Thanks!

src/Documentation/DocumentationGenerator/Extensions.cs
```diff
@@ -179,11 +179,37 @@ TypedExpression input
...
-            SyntaxTreeToQsharp.DeclarationSignature(
+        internal static string ToSyntax(this QsFunctor functor) =>
+            functor.Tag switch
+            {
+                QsFunctor.Tags.Adjoint => "Adj",
+                QsFunctor.Tags.Controlled => "Ctl",
+                _ => "__invalid__"
+            };
+
+        internal static string ToSyntax(this QsCallable callable, Action<IRewriteStep.Diagnostic>? onDiagnostic = null)
```
samarsha, line 190: `onDiagnostic` is an unused parameter?

src/Documentation/DocumentationGenerator/Extensions.cs
```diff
@@ -179,11 +179,37 @@ TypedExpression input
...
+        {
+            var signature = SyntaxTreeToQsharp.DeclarationSignature(
                 callable, SyntaxTreeToQsharp.Default.ToCode
             );
```
samarsha, line 194:
```suggestion
                callable, SyntaxTreeToQsharp.Default.ToCode);
```

https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1111.md

Looks like there are a lot of StyleCop violations in this file, actually. :) Out of curiosity, are the squigglies showing up for you? It should look like:

image

src/Documentation/DocumentationGenerator/Extensions.cs
```diff
@@ -179,11 +179,37 @@ TypedExpression input
...
+            // the characteristics (e.g.: `is Adj + Ctl`) for the callable, so
+            // we add that here.
+            var characteristics = callable.Signature.Information.Characteristics.SupportedFunctors switch
+            {
+                var nullable when nullable.IsNull => "",
+                { Item: var item } => item switch
+                {
+                    { Count: 0 } => "",
+                    // Be sure to add the leading space before is!
+                    _ => $" is {string.Join("+", item.Select(functor => functor.ToSyntax()))}"
```
samarsha, line 206: Is there not supposed to be a space around `+` in the documentation?

src/Documentation/DocumentationGenerator/Extensions.cs
```diff
@@ -179,11 +179,37 @@ TypedExpression input
...
+                }
```
samarsha, line 207:
```suggestion
                { Item: { Count: 0 } } => "",
                // Be sure to add the leading space before is!
                { Item: var functors} => $" is {string.Join("+", functors.Select(functor => functor.ToSyntax()))}"
```

samarsha (approved):

src/Documentation/DocumentationGenerator/Extensions.cs
```diff
@@ -179,11 +179,37 @@ TypedExpression input
...
+            var signature = SyntaxTreeToQsharp.DeclarationSignature(
                 callable, SyntaxTreeToQsharp.Default.ToCode
             );
 
+            // The signature provided by SyntaxTreeToQsharp doesn't include
+            // the characteristics (e.g.: `is Adj + Ctl`) for the callable, so
+            // we add that here.
+            var characteristics = callable.Signature.Information.Characteristics.SupportedFunctors switch
+            {
+                var nullable when nullable.IsNull => "",
```
samarsha, line 201:
```suggestion
                { IsNull: true } => "",
```

...maybe. :)

src/Documentation/DocumentationGenerator/Extensions.cs
cgranade: Good catch, thank you!
```suggestion
                    _ => $" is {string.Join(" + ", item.Select(functor => functor.ToSyntax()))}"
```

src/Documentation/DocumentationGenerator/Extensions.cs
cgranade: For some reason I thought that nested pattern was a C# 9 feature... thanks for the suggestion!

src/Documentation/DocumentationGenerator/Extensions.cs
cgranade: I'll admit, I find code that follows SA1111 _much_ more difficult to read... maybe I've spent too long in PEP 8 codebases, which requires the exact opposite. In any case, thanks for raising that, will go on and fix.

src/Documentation/DocumentationGenerator/Extensions.cs
cgranade: Good catch, thank you; it used to be used, but that was stripped out in response to one of @bettinaheim's earlier points of feedback; must have missed a case.

src/Documentation/DocumentationGenerator/Extensions.cs
samarsha: Looks like this change was not applied - did nested patterns turn out to be a C# 9 feature? (It worked in my IDE but I didn't try building it.)

cgranade: Ah, sorry, the web UI wouldn't allow me to accept the suggestion so I did it locally then forgot to push. Fixed now.

cgranade:
It looks like the e2e test failure is in an unrelated component; are builds working at the moment?

```
       "/home/vsts/work/1/s/drop/Tests/tests.generated/tests-all/tests-all.csproj" (VSTest target) (1:7) ->
       "/home/vsts/work/1/s/drop/Tests/tests.generated/tests-all/tests-all.csproj" (default target) (1:8) ->
       (CoreCompile target) -> 
         CoreTests.cs(43,24): error CS0246: The type or namespace name 'AbstractFactory<>' could not be found (are you missing a using directive or an assembly reference?) [/home/vsts/work/1/s/drop/Tests/tests.generated/tests-all/tests-all.csproj]
```

rmshaffer:
> It looks like the e2e test failure is in an unrelated component; are builds working at the moment?

No, I believe the E2E build is broken at the moment until https://github.com/microsoft/qdk/pull/27 is merged.

PR #721, opened 2020-11-03
Integrate 'feature/serialization-deserialization-performance' to 'main'

cesarzc:
This PR integrates the changes in `feature/serialization-deserialization-performance` into `main`.


src/QsCompiler/DataStructures/SyntaxTokens.fs
```diff
@@ -56,6 +56,19 @@ type CharacteristicsKind<'S> =
 | Union of 'S * 'S
 | Intersection of 'S * 'S
 | InvalidSetExpr
+    with
+    member this.TryGetSimpleSet(simpleSet: OpProperty byref) =
```
bettinaheim, line 60: What is the reason for wanting to provide these TryGet* methods instead of using something like `value as SimpleSet set` (assuming they are primarily for use within C#)?

src/QsCompiler/BondSchemas/BondSchemaTranslator.cs
```diff
@@ -0,0 +1,1344 @@
...
+            };
+
+        private static Identifier ToBondSchema(this SyntaxTree.Identifier identifier)
+        {
+            string? bondLocalVariable = null;
+            QsQualifiedName? bondGlobalCallable = null;
+            NonNullable<string> compilerLocalVariable = default;
+            SyntaxTree.QsQualifiedName? compilerGlobalCallable = null;
+            IdentifierKind kind;
+            if (identifier.TryGetLocalVariable(ref compilerLocalVariable))
```
bettinaheim, line 50: You could instead express this as 
```suggestion
            if (identifier is SyntaxTree.Identifier.LocalVariable var)
            {
                kind = IdentifierKind.LocalVariable;
                bondLocalVariable = var.Item.Value;
            }            
```

src/QsCompiler/DataStructures/SyntaxTokens.fs
cesarzc: I think there is no strong reason to have the TryGet* methods. I will change it to use the `is` pattern instead.

src/QsCompiler/BondSchemas/BondSchemaTranslator.cs
cesarzc: Yes, that way we don't need all the `TryGet`* methods. I will change the pattern to remove these helper methods.

bettinaheim: Thanks!

cesarzc: I moved everything to follow this pattern in the latest changes :)

src/QsCompiler/DataStructures/SyntaxTokens.fs
cesarzc: Removed these methods since the pattern you suggested no longer needs them.

bettinaheim (approved):
I'm excited to have this format switch pulled in!

PR #723, opened 2020-11-04
Exclude Type Param Keys from Monomorphization Rewrite

ScottCarda-MS:
Exclude Type Parameter Resolution keys from being rewritten by replacement transformation in Monomorphization. This PR updates the monomorphization rewrite step so that it handles type parameter collection/calculation more like the way the call graph does, bringing greater alignment with each other and reducing the chance for errors.


src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -332,9 +348,10 @@ public static QsCallable Apply(QsCallable current, GetConcreteIdentifierFunc get
 
             public class TransformationState
             {
-                public readonly Concretization CurrentParamTypes = new Concretization();
+                public readonly Stack<TypeParameterResolutions> CurrentTypeParamResolutions = new Stack<TypeParameterResolutions>();
```
cesarzc, line 351: It is not completely clear to me why there was a change from `Dictionary` to `Stack` here, can you please expand a bit more on the reason for this change?

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -15,10 +15,9 @@
 
 namespace Microsoft.Quantum.QsCompiler.Transformations.Monomorphization
```
cesarzc, line 16: Can you please add a documentation comment on the `Apply` public static method?

src/QsCompiler/Transformations/Monomorphization.cs
ScottCarda-MS: Individual type parameter resolutions are dictionaries, and in exploring an expression the transformation can come across many of these dictionaries. These dictionaries need to be combined to create a more complete picture of how the type parameters are resolved in that expression. Previously this file was trying to do that combining itself as it went along, only holding keeping track of the current calculated dictionary. However there is a utility that performs this combining now, which is used by the call graph. Since we are calculating these resolutions in order to compare them against the resolutions found by the call graph, it is safer if these resolutions were calculated the same way, using the same tool, which is the TypeParameterCombination utility.

ScottCarda-MS: In short, it was changed to a stack to match how the call graph handles type parameter resolutions.

src/QsCompiler/Transformations/Monomorphization.cs
ScottCarda-MS: Sure, most of the explanation is on the class doc, so the doc for the Apply method will be minimal.

src/QsCompiler/Transformations/Monomorphization.cs
cesarzc: Makes sense, thanks for the explanation Scott!

cesarzc (approved):

PR #725, opened 2020-11-05
Add automerge GitHub action

### LINKED ISSUE #718: Support automerge for pull requests
Pull requests require automatic builds to complete before merging, but these builds take a long time.

It would be nice to have something like the "automerge" label that https://github.com/microsoft/qsharp-runtime has, which will automatically merge a PR once all the required status checks have passed.

N/A

**Additional context**
N/A

samarsha:
I configured it a little differently from qsharp-runtime. Instead of the `automerge` label always squashing the PR, I set it up to recognize two labels: `automerge` (create merge commit) and `autosquash` (create squash commit). This is to allow automerge to be used with PRs to feature branches that are allowed to create merge commits.

Closes #718.


cgranade:
Would it make sense to copy the autosquash configuration to other repos as well? That seems like a good idea now that we're using both merge and squash commits on different PRs.

samarsha:
Yes, I think it'd make sense to use the same label configuration on all the repos.

samarsha:
After talking with @anpaz, it looks like the e2e build will fail for this PR due to some GitHub permissions issues caused by this PR adding a new GitHub workflow file. So I will override the build checks if this is approved.

bettinaheim (approved):

PR #727, opened 2020-11-07
Initialize Bond serializers and deserializers asynchronously

cesarzc:
This change asynchronously triggers the creation of Bond serializers and deserializers at the beginning of the compilation process in order to speed up syntax tree serialization and deserialization.

### DIFF
--- a/src/QsCompiler/BondSchemas/Protocols.cs
+++ b/src/QsCompiler/BondSchemas/Protocols.cs
@@ -1,20 +1,30 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System.Diagnostics;
 using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
 using Bond;
 using Bond.IO.Unsafe;
 using Bond.Protocols;
 
 namespace Microsoft.Quantum.QsCompiler.BondSchemas
 {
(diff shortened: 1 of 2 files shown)


cesarzc:
Speeding up the compiler 💨

src/QsCompiler/BondSchemas/Protocols.cs
```diff
@@ -39,34 +66,52 @@ public static class Protocols
...
+
             stream.Flush();
             stream.Position = 0;
         }
 
-        private static Deserializer<SimpleBinaryReader<InputBuffer>> GetSimpleBinaryDeserializer()
+        private static SimpleBinaryDeserializer GetSimpleBinaryDeserializer()
         {
-            if (simpleBinaryDeserializer == null)
+            Debug.Assert(Monitor.IsEntered(BondSharedDataStructuresLock));
```
bettinaheim, line 85: Debug.Asserts are not properly registering when running the extensions in debug and there were at some point also certain cross-platform issues (not sure if there still are). As awkward as it is, could we instead use 
     #if DEBUG
     // a check that throws if it is not satisfied
     #endif
Same for the other Debug.Asserts.

cesarzc: Replaced Debug.Assert with the `VerifyLockAcquired` that uses `#if DEBUG`.

bettinaheim (approved):

PR #729, opened 2020-11-09
Remove NonNullable

### LINKED ISSUE #703: Remove NonNullable<'T>
`NonNullable<'T>` adds a lot of noise to the code, and does not actually provide much null safety, because it is easily bypassed using `default(NonNullable<T>)` in C# or `Unchecked.defaultof<NonNullable<'T>>` in F#. We should remove it and use the convention that all types are non-nullable by default.

samarsha:
Closes #703.

Required PRs in other repos:

- [ ] microsoft/qsharp-runtime#431
- [ ] microsoft/iqsharp#363
- [ ] microsoft/QuantumKatas#559
- [x] microsoft/qdk#29


src/QsCompiler/DataStructures/Serialization.fs
```diff
@@ -15,18 +15,6 @@ open Newtonsoft.Json
 open Newtonsoft.Json.Linq
 open Newtonsoft.Json.Serialization
 
-
-type NonNullableConverter<'T when 'T: equality and 'T: null>()  =
```
bettinaheim, line 19: Did you double check whether the extension including this change works fine when opening an older project? I.e. the references serialized based on the old data structures are properly loaded?

bettinaheim (approved):
Thanks for doing this!

src/QsCompiler/BondSchemas/BondSchemaTranslator.cs
```diff
@@ -88,7 +88,7 @@ private static Identifier ToBondSchema(this SyntaxTree.Identifier identifier)
         private static LocalDeclarations ToBondSchema(this SyntaxTree.LocalDeclarations localDeclarations) =>
             new LocalDeclarations
             {
-                Variables = localDeclarations.Variables.Select(v => v.ToBondSchemaGeneric(typeTranslator: ToBondSchema)).ToList()
+                Variables = localDeclarations.Variables.Select(v => v.ToBondSchemaGeneric(typeTranslator: s => s)).ToList()
```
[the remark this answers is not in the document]
samarsha: It's intentional. The overload of `ToBondSchema` this line was using was:

```csharp
private static string ToBondSchema(this NonNullable<string> s) => s.Value;
```

which simply becomes the identity function since `NonNullable` is removed.

src/QsCompiler/DataStructures/Serialization.fs
samarsha: I opened a project using version 0.12.20100504 of the SDK with a language server built from this branch, and it worked.

bettinaheim: Thanks for confirming!

Issue #699, opened 2020-10-22
Use consistent tags in documentation comments


samarsha:
Most documentation comments in the compiler only use the `<summary>` tag. For short comments, this is fine, but in many cases the summary is several paragraphs long and would be easier to read if it used standard documentation tags to structure the information. This is especially important for exceptions - exceptions mentioned in the `<summary>` tag are easy to miss, whereas the special `<exception>` tag makes them stand out more.

I think we need to use some rules for our doc comments. My proposal is:

1. Limit [`<summary>`](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/summary) to one paragraph. Anything longer than that should go in specific tags as appropriate, or the [`<remarks>`](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/remarks) tag if none of the other tags apply.
2. Always put exceptions in an [`<exception>`](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/exception) tag.

cgranade:
Agreed entirely. Can I also request splitting parameter descriptions in to `<param>` elements so that they show up in intelliSense correctly?

kevinhartman:
I'd be interested in working on this issue. Assuming I wouldn't be toe-stepping, my plan would be to start with exceptions, working out of [this branch](https://github.com/kevinhartman/qsharp-compiler/tree/keha/except-tag-docs).

samarsha:
Hi @kevinhartman, that would be great! Let us know if you have any questions.

kevinhartman:
Thanks @samarsha! 

I just submitted a PR to add the exception tag for C# doc comments. I didn't do this for the F# code, since most of it uses the implicit summary tag (omits the `<summary>` tag). Thoughts on using explicit XML tags for that portion of the codebase?

I was also wondering how strictly the `<paramref>` tag should be used in exception text. In most cases, I was worried that it'd make the documentation less readable when looking at source files. But perhaps if the XML is being used to generate API reference docs, it'd be pretty helpful!

samarsha:
Thanks @kevinhartman for the PR! I'll take a look soon.

For the F# comments, I think switching from an implicit summary tag to an explicit one when an exception tag (or other tag)  is needed makes sense.

[...]

PR #732, opened 2020-11-10
Use the XML exception tag in doc comments (C#)

kevinhartman:
Adds proper XML tags for any exceptions previously mentioned within the summary tag.

Currently, this is only done for C# projects, since most of the F# code uses [implicit summary tags](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/xml-documentation#comments-without-xml-tags).

Ref #699


src/QsCompiler/CompilationManager/TypeChecking.cs
```diff
@@ -1161,10 +1167,10 @@ internal static void ImportGlobalSymbols(this FileContentManager file, Compilati
         /// Given a sequence of tree nodes, builds the corrsponding array of Q# statements (ignoring invalid fragments)
         /// using and updating the given scope context and adding the generated diagnostics to the given list of diagnostics,
         /// provided each statement consists of a suitable statement header followed by the required continuation(s), if any.
```
kevinhartman, line 1169: Any recommendations for a concise way to include some of this text in place of "Thrown if this is not the case" in the exception tag?

samarsha: If I'm interpreting the doc comment correctly, maybe something like "Thrown if any statement is missing a header or a required continuation, or ..."?

src/QsCompiler/CompilationManager/DataStructures.cs
```diff
@@ -479,9 +479,9 @@ internal struct HeaderEntry<T>
         /// If the symbol of the extracted declaration is not an unqualified symbol,
         /// verifies that it corresponds instead to an invalid symbol and returns null unless the keepInvalid parameter has been set to a string value.
         /// If the keepInvalid parameter has been set to a (non-null) string, uses that string as the SymbolName for the returned HeaderEntry instance.
-        /// Throws an ArgumentException if this verification fails as well.
-        /// Throws an ArgumentException if the extracted declaration is Null.
         /// </summary>
+        /// <exception cref="ArgumentException">Thrown if the symbol of the extracted declaration is not an unqualified symbol and doesn't instead correspond to an invalid symbol.</exception>
+        /// <exception cref="ArgumentException">Thrown if the extracted declaration is Null.</exception>
```
samarsha, line 484: What do you think about combining two exception tags into one tag if they have the same exception type? I think this would be more consistent with some of the other doc comments that say "thrown if X, or Y, or Z."

samarsha (commented):
Thanks for doing this! It's much easier to see which exceptions a function throws now. :)

src/QsCompiler/CompilationManager/DataStructures.cs
kevinhartman: You know, I hadn't thought about that. I just now spent a little time digging.

The Sandcastle Help File Builder documentation seems to indicate that [only a single tag should be present per exception type](https://github.com/EWSoftware/SHFB/blob/ea37113a5aa5e6d998213d37f43cc4f49acac2a7/Documentation/XMLCommentsGuide/Content/SectionElements/exception.aml#L23), though this documentation hasn't been updated since 2012.

On the flip side, I found an MS example where [multiple entries are used](https://github.com/dotnet/corefx/blob/a10890f4ffe0fadf090c922578ba0e606ebdd16c/src/Common/src/CoreLib/System/Threading/Tasks/TaskCompletionSource.cs#L178). In this case, whatever tool that generated the API docs seems to have [automatically joined these entries](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskcompletionsource-1.trysetexception?view=netcore-2.1#System_Threading_Tasks_TaskCompletionSource_1_TrySetException_System_Collections_Generic_IEnumerable_System_Exception__) with `<para>- or -</para>` between each.

Personally, I think I'm in favor of combining, but perhaps with some structure for different causes. Maybe something like:

```xml
<exception cref="ArgumentException">
  Causes:
  <list type="bullet">
    <item>
      <description>The symbol of the extracted declaration is not an unqualified symbol and doesn't instead correspond to an invalid symbol.</description>
    </item>
    <item>
      <description>The extracted declaration is Null.</description>
    </item>
  </list>
</exception>
```

Which when generated with DocFX looks like this:
Screen Shot 2020-11-15 at 2 42 21 AM

I think for nearly identical causes (e.g. an exception is thrown if any of a few parameters are null), a single sentence like "thrown if X, or Y, or Z" is plenty readable and fine as is.

So I would propose:
- For any exception tags with the same `cref`, combine into a single tag using a list format.
- Leave any single sentences like "thrown if X, or Y, or Z" alone, unless they're obviously overloaded with logically separate causes.

I think that's sort of what's done at least for [this example](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.appendalllines?view=net-5.0), where `ArgumentNullException` uses a sentence and `UnauthorizedAccessException` uses a list (kinda...).

kevinhartman:
A few TODOs prior to merge:

- [ ] Remove "Thrown if" prefix from exception tag descriptions. Otherwise, generated documentation might be redundant (e.g. DocFX will put these descriptions inside column entitled "Conditions").
- [ ] Add `paramref` tags for all parameter references in exception descriptions.
- [ ] Combine exception tags with the same `cref` (approach TBD).
- [ ] Add `see` tags for code references to exception tag descriptions.

src/QsCompiler/CompilationManager/DataStructures.cs
samarsha: Oh, interesting, I didn't realize the .NET documentation used multiple `exception` tags like that. I think as long as we're consistent in how we apply the distinction, like you mentioned (closely related conditions vs. unrelated conditions), do you see any downside to using separate tags to represent an "or" of unrelated conditions? It looks less verbose in the source code compared to the list, and IMO the list is a little ambiguous about whether the list items are connected by "or" or "and," whereas with separate tags it makes more sense to me that the conditions would be a list of separate possibilities (so "or" instead of "and").

kevinhartman: > do you see any downside to using separate tags to represent an "or" of unrelated conditions?

I think my only concern is how tools will handle it when rendering. Unfortunately, it looks like DocFX creates a new table row for every exception tag, even if they share the same `cref` (I'm not sure what .NET docs use). This would be hard to read in API docs IMO. I wonder if you might be able to test this out in Rider?

> It looks less verbose in the source code compared to the list

Completely agree. I rather like the way the separate tags approach looks in the source.

> the list is a little ambiguous about whether the list items are connected by "or" or "and,"

Darn. I thought maybe I'd found a more elegant solution than separating conditions with `"- or -"` by heading the list with `Causes:`.

We could always just go with the "- or -" thing. It's actually fairly readable in the source IMO.

```c#
///     <para>The symbol of the extracted declaration is not an unqualified symbol and doesn't instead correspond to an invalid symbol.</para>
///     <para>- or -</para>
///     <para>The extracted declaration is Null.</para>
/// </exception>
```

Here's a render in DocFX next to the list approach:

Screen Shot 2020-11-17 at 12 47 26 AM

samarsha: I think that I would prefer to pick a source representation that has the most flexibility in how it's rendered, since the documentation generator is easier to change than re-doing all of the doc comments.

The multiple `exception` tags seem to encode more information in a machine-readable format than the list or paragraph versions. I think it'd be easier to have the docs generator transform the multiple `exception` tags into the format that we want, than it is to transform the list or paragraph versions into a different format if we ever decide to change the way the rendered docs look like. It also has the benefit of being simpler to write in the source code.

We don't currently generate docs for the Q# compiler, but if/when we do I think we can address the multiple exceptions of the same type problem as being a problem with the generator/template we use, instead of a problem with the source. Does that seem reasonable?

Also, I tried it in Rider and it shows each exception tag separately (like DocFX) - that's probably harder to change than the docs generator template, but personally I think it's not that much of a readability issue to work around it in the source code.

kevinhartman: > The multiple exception tags seem to encode more information in a machine-readable format

+1

> we can address the multiple exceptions of the same type problem as being a problem with the generator/template we use

That's a great idea! It'd be pretty easy to add a post-processing step for the full XML files that the C# compiler spits out. This also means less work for me to get this PR ready :D

src/QsCompiler/CompilationManager/TypeChecking.cs
kevinhartman: I'll just leave this one as is, since it's probably at least clear enough for now in the context of the summary text.

kevinhartman:
This one is ready for another look @samarsha 

Thanks for all of the great discussion so far!

src/QsCompiler/CompilationManager/DataStructures.cs
```diff
@@ -479,9 +479,9 @@ internal struct HeaderEntry<T>
         /// If the symbol of the extracted declaration is not an unqualified symbol,
         /// verifies that it corresponds instead to an invalid symbol and returns null unless the keepInvalid parameter has been set to a string value.
         /// If the keepInvalid parameter has been set to a (non-null) string, uses that string as the SymbolName for the returned HeaderEntry instance.
-        /// Throws an ArgumentException if this verification fails as well.
-        /// Throws an ArgumentException if the extracted declaration is Null.
         /// </summary>
+        /// <exception cref="ArgumentException">The symbol of the extracted declaration is not an unqualified symbol and does not instead correspond to an invalid symbol.</exception>
```
samarsha, line 483: I think this is a simpler way to say it:

```suggestion
        /// <exception cref="ArgumentException">The symbol of the extracted declaration is not an unqualified or invalid symbol.</exception>
```

src/QsCompiler/CompilationManager/DataStructures.cs
```diff
@@ -534,8 +534,10 @@ public FileHeader(ReaderWriterLockSlim syncRoot)
         /// <summary>
         /// Invalidates (i.e. removes) all elements in the range [start, start + count), and
         /// updates all elements that are larger than or equal to start + count with the given lineNrChange.
-        /// Throws an ArgumentOutOfRange exception if start or count are negative, or if lineNrChange is smaller than -count.
         /// </summary>
+        /// <exception cref="ArgumentOutOfRangeException">
+        /// <paramref name="start"/> or <paramref name="count"/> are negative, or if <paramref name="lineNrChange"/> is smaller than -<paramref name="count"/>.
```
samarsha, line 539:
```suggestion
        /// <paramref name="start"/> or <paramref name="count"/> are negative, or <paramref name="lineNrChange"/> is smaller than -<paramref name="count"/>.
```

src/QsCompiler/CompilationManager/TypeChecking.cs
samarsha: I think it would be nice if the exception tags could be read separately without referring back to the summary too much, do you think it would be OK to actually replace the "this is not the case" part here?

kevinhartman: No problem! Done.

samarsha (approved):
Looks great! Thanks again for working on this!

PR #735, opened 2020-11-11
Interpolated String Delimiters

ScottCarda-MS:
Added context information to CodeLine class to keep track of whether the code line leaves open a string, an interpolated string, and interpolated argument, or a string in an interpolated argument. Throws an error if an interpolated argument contains a nested interpolated string.

Future work for this feature branch will be to create proper error handling for nested interpolated strings, as they will be disallowed moving forward.


ScottCarda-MS:
A merge in from the main branch has polluted this PR. Please wait to review until it can be resolved.

src/QsCompiler/CompilationManager/DataStructures.cs
```diff
@@ -41,69 +50,357 @@ internal class CodeLine
...
+            {
+                delimiterBuilder.Add(stringLength);
+            }
+
+            errorDelimiters = errorDelimiterBuilder.ToImmutable();
+            return delimiterBuilder.ToImmutable();
         }
 
-        internal CodeLine SetCommentIndex(int newCommentIndex)
+        private static (StringContext, bool) MoveToNextState(StringContext curr, string input)
```
DmitryVasilevsky, line 316: Would be helpful to add a comment about what bool means in return tuple. Alternatively create a named type with the named field.

src/QsCompiler/CompilationManager/DataStructures.cs
```diff
@@ -41,69 +50,357 @@ internal class CodeLine
...
+            this.ErrorDelimiterPositions = errorDelimiterPositions.ToImmutableArray();
+            this.StringDelimiters = stringDelimiters.ToImmutableArray();
+            this.BeginningStringContext = beginningStringContext;
+            this.EndingStringContext = endingStringContext;
+
+            ScopeTracking.VerifyStringDelimiters(text, stringDelimiters);
+            ScopeTracking.VerifyExcessBracketPositions(this, excessBracketPositions);
+        }
+
+        private CodeLine(StringContext beginningStringContext)
```
DmitryVasilevsky, line 89: As there're multiple constructors with different arguments it would be nice to have comments on how they are different.

src/QsCompiler/CompilationManager/DataStructures.cs
ScottCarda-MS: Thanks! I've given names to the tuple fields to make it clearer what the fields are for.

src/QsCompiler/CompilationManager/DataStructures.cs
ScottCarda-MS: That's a good idea. I realize I made a lot more constructors in this PR so adding comments to help sort out which ones are for what is good. I've added descriptions for these constructors to the code to help make things a bit cleaner to understand.

src/QsCompiler/CompilationManager/DataStructures.cs
```diff
@@ -23,6 +23,15 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder.DataStructures
     /// </summary>
     internal class CodeLine
     {
+        internal enum StringContext
+        {
+            NoOpenString,
+            OpenInterpolatedArgument,
+            OpenString,
```
DmitryVasilevsky, line 30: The order is important here. Suggest making a comment about this.

src/QsCompiler/CompilationManager/DataStructures.cs
```diff
@@ -41,69 +50,374 @@ internal class CodeLine
...
+            {
+                delimiterBuilder.Add(-1);
+            }
+
+            commentIndex = -1;
+            var stringLength = text.Length;
+            while (text != string.Empty)
+            {
+                // Check for a comment start if we are outside a string
+                commentIndex = stringContext >= StringContext.OpenString ? -1 : text.IndexOf("//");
```
DmitryVasilevsky, line 248: Suggest making this check a separate function as it has a meaning.
stringContext >= StringContext.OpenString

src/QsCompiler/CompilationManager/DataStructures.cs
```diff
@@ -23,6 +23,15 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder.DataStructures
     /// </summary>
     internal class CodeLine
     {
+        internal enum StringContext
```
DmitryVasilevsky, line 26: Suggest to make a comment about how these states relate to context in the text:
NOS " OS
NOS $" OIS { OIA " OSIOIA

src/QsCompiler/CompilationManager/DataStructures.cs
```diff
@@ -41,69 +50,374 @@ internal class CodeLine
 
         internal readonly int Indentation; // Note: This denotes the initial indentation at the beginning of the line
         internal readonly ImmutableArray<int> ExcessBracketPositions;
+        internal readonly ImmutableArray<int> ErrorDelimiterPositions;
 
         // convention: first one opens a string, and we have the usual inclusions: {0,5} means the chars 0..4 are the content of a string
         // -> i.e. -1 means the string starts on a previous line, and an end delimiter that is equal to Text.Length means that the string continues on the next line
-        internal readonly ImmutableArray<int> StringDelimiters; // note that this property is only properly initialized *after* knowing the sourrounding lines
+        internal readonly ImmutableArray<int> StringDelimiters; // note that this property is only properly initialized *after* knowing the surrounding lines
```
DmitryVasilevsky, line 57: Why do you need to keep these positions in the object?

src/QsCompiler/CompilationManager/DataStructures.cs
```diff
@@ -41,69 +50,374 @@ internal class CodeLine
...
+                    case StringContext.OpenInterpolatedArgument:
+                        // Find the next " or }
+                        index = text.IndexOfAny(new[] { '"', '}' });
+                        break;
+                    case StringContext.OpenInterpolatedString:
+                        // Find the next " or {, neither or which being preceded by \
+                        index = text.IndexOfAny(new[] { '"', '{' });
+                        while (index > 0 && text[index - 1] == '\\')
+                        {
+                            var next = text.Substring(index + 1).IndexOfAny(new[] { '"', '{' });
```
DmitryVasilevsky, line 266: Instead of allocating these arrays dynamically, you can allocated it once (static modifier) and use it in several places.

src/QsCompiler/CompilationManager/DataStructures.cs
```diff
@@ -41,69 +50,374 @@ internal class CodeLine
...
+            this.ExcessBracketPositions = excessBrackets.ToImmutableArray();
+            ScopeTracking.VerifyStringDelimiters(text, this.StringDelimiters);
+            ScopeTracking.VerifyExcessBracketPositions(this, excessBrackets);
         }
 
-        internal CodeLine SetStringDelimiters(IEnumerable<int> newStringDelimiters)
+        /// <summary>
+        /// Computes the location of the string delimiters within a given text.
+        /// </summary>
+        private static IEnumerable<int> ComputeStringDelimiters(string text, ref StringContext stringContext, out int commentIndex, out IEnumerable<int> errorDelimiters)
```
DmitryVasilevsky, line 233: This function is a seems to play a role of tokenizer in analyzing the input. As written, it does a lot of scanning over the line (i.e. repeatable 'indexof' with different arguments. A typical DFA approach would consume less CPU time.

src/QsCompiler/CompilationManager/DataStructures.cs
```diff
@@ -41,69 +50,374 @@ internal class CodeLine
...
+                    else
+                    {
+                        errorDelimiterBuilder.Add(pos);
+                    }
+                }
+                else
+                {
+                    delimiterBuilder.Add(pos);
+                }
+                text = text.Substring(index + 1);
```
DmitryVasilevsky, line 314: Repeatedly reallocating string consumes excessive memory (strings are immutable). All indexof functions take starting posistion for search, so you don't need to reallocate.

src/QsCompiler/CompilationManager/DataStructures.cs
ScottCarda-MS: These are used in the parser code.

src/QsCompiler/CompilationManager/DataStructures.cs
ScottCarda-MS: I've added that to the summary for the enum.

src/QsCompiler/CompilationManager/DataStructures.cs
ScottCarda-MS: This is now encapsulated in the IsInAString function.

src/QsCompiler/CompilationManager/DataStructures.cs
ScottCarda-MS: I've added an example to the doc comment for the enum.

src/QsCompiler/CompilationManager/DataStructures.cs
ScottCarda-MS: Good idea, done

src/QsCompiler/CompilationManager/DataStructures.cs
ScottCarda-MS: I've reworked the function so that it should only perform a single pass through the text now.

src/QsCompiler/CompilationManager/DataStructures.cs
ScottCarda-MS: I've reworked the function so that it will keep the same text but update the position in that text, instead of allocating new strings for each pass as you pointed out.

src/QsCompiler/CompilationManager/DataStructures.cs
ScottCarda-MS: They are used later on in the parser logic. Since the CodeLines were previously storing these delimiters, I don't want to change that here and have to refactor a larger portion of the parser code than I strictly need to.

src/QsCompiler/CompilationManager/DataStructures.cs
```diff
@@ -41,69 +61,416 @@ internal class CodeLine
...
+            var delimiterBuilder = ImmutableArray.CreateBuilder<int>();
+            var errorDelimiterBuilder = ImmutableArray.CreateBuilder<int>();
+
+            if (IsInAString(stringContext))
+            {
+                delimiterBuilder.Add(-1);
+            }
+
+            commentIndex = -1;
+            var noOpenStringDelimiters = new[] { '"', '/' };
```
DmitryVasilevsky, line 285: These are still allocated on every invocation of ComputeStringDelimiters. I suggest moving them out of the function and making them static readonly. I.e.
`        static readonly char[] noOpenStringDelimiters = new[] { '"', '/' };`

DmitryVasilevsky (approved):

src/QsCompiler/CompilationManager/DataStructures.cs
ScottCarda-MS: Oh you are right, I've now made them static fields, as you suggest.

PR #739, opened 2020-11-13
Fix some nullable reference type errors in Visual Studio 16.8

samarsha:
It looks like there were some changes to nullable reference type analysis in Visual Studio 16.8, causing some errors when building the Q# compiler. (This is despite having a global.json that sets the SDK version to 3.1 - `dotnet build` would succeed, but building in VS itself would fail.) The changes needed to fix the errors continue to work in 3.1, and mostly seem reasonable.

There's one (small) breaking change. I removed the default parameter from the `TypedExpressionWalker` constructor:

```csharp
public TypedExpressionWalker(Action<TypedExpression> onExpression, T internalState = default)
```

This could be worked around with nullable attributes or the null-forgiving operator, but having a null transformation state does not seem like a great idea, anyway.

### DIFF
--- a/src/QsCompiler/Tests.Compiler/LinkingTests.fs
+++ b/src/QsCompiler/Tests.Compiler/LinkingTests.fs
@@ -307,7 +307,7 @@ type LinkingTests (output:ITestOutputHelper) =
                     Assert.Equal(0, ex.TypeParameterResolutions.Count)
             | _ -> ()
 
-        let walker = new TypedExpressionWalker<unit>(new Action<_>(onExpr));
+        let walker = TypedExpressionWalker(Action<_> onExpr, ())
         walker.Transformation.OnCompilation compilation |> ignore
         Assert.True(gotLength)
         Assert.True(gotIndexRange)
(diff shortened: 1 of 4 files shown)


filipw:
possibly because it was not set to 3.1.403?

samarsha:
No, changing the version in global.json to 3.1.403 or 3.1.404 doesn't seem to work. I think the global.json is working, because `dotnet build` succeeds, but the analyzer/compiler built into Visual Studio seems to have changed behavior slightly.

src/QsCompiler/Tests.Compiler/LinkingTests.fs
```diff
@@ -307,7 +307,7 @@ type LinkingTests (output:ITestOutputHelper) =
                     Assert.Equal(0, ex.TypeParameterResolutions.Count)
             | _ -> ()
 
-        let walker = new TypedExpressionWalker<unit>(new Action<_>(onExpr));
+        let walker = TypedExpressionWalker(Action<_> onExpr, ())
```
bettinaheim, line 310: Hu, I am curious: Do you know what the change is related to new?

swernli (approved):

bettinaheim (approved):

src/QsCompiler/Tests.Compiler/LinkingTests.fs
samarsha: In F#, `new` is optional unless the type implements `IDisposable`. I always prefer to omit `new`.

bettinaheim: That makes sense - not a needed change then.

PR #752, opened 2020-11-24
Add Bond generated C# code to the VS solution instead of performing Bond code generation during build

cesarzc:
This change adds Bond generated C# code to the VS solution instead of performing Bond code generation during build.

This is meant to enable building Q# compiler on linux.


src/QsCompiler/BondSchemas/BondSchemas.csproj
```diff
@@ -8,6 +8,10 @@
     <AssemblyName>Microsoft.Quantum.BondSchemas</AssemblyName>
   </PropertyGroup>
 
+  <ItemGroup>
+    <BondCodegen Remove="CompilerDataStructures.bond" />
```
anpaz, line 12: I would do "*.bond"

anpaz (approved):

src/QsCompiler/BondSchemas/BondSchemas.csproj
cesarzc: Integrated this suggestion.

Issue #737, opened 2020-11-12
Language server does not work after installing .NET 5


samarsha:
**Describe the bug**

After installing the .NET 5 SDK, the language server can't load Q# projects:

> [Info  - 12:15:44] workspace folder: c:\Users\samarsha\Desktop\QSharpSandbox
[Error - 12:15:45] Error on loading project 'c:\Users\samarsha\Desktop\QSharpSandbox\Library\Library.csproj': Der SDK-Resolvertyp "WorkloadSdkResolver" konnte nicht geladen werden. Could not load file or assembly 'System.Runtime, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. Das System kann die angegebene Datei nicht finden.  c:\Users\samarsha\Desktop\QSharpSandbox\Library\Library.csproj.
Ignoring non-Q# project 'c:\Users\samarsha\Desktop\QSharpSandbox\Library\Library.csproj'.
[Error - 12:15:45] Error on loading project 'c:\Users\samarsha\Desktop\QSharpSandbox\App\App.csproj': Der SDK-Resolvertyp "WorkloadSdkResolver" konnte nicht geladen werden. Could not load file or assembly 'System.Runtime, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. Das System kann die angegebene Datei nicht finden.  c:\Users\samarsha\Desktop\QSharpSandbox\App\App.csproj.
Ignoring non-Q# project 'c:\Users\samarsha\Desktop\QSharpSandbox\App\App.csproj'.
[Info  - 12:15:45] The file c:\Users\samarsha\Desktop\QSharpSandbox\App\App.qs is not associated with a compilation unit. Only syntactic diagnostics are generated.

**To Reproduce**

1. Open a Q# project in VS Code

**System information**

- QDK extension v0.13.20111004 for VS Code
- Windows 10, .NET 5.0.100
- VS Code 1.51.1

**Additional context**

The Q# project file is:

```xml
<Project Sdk="Microsoft.Quantum.Sdk/0.13.20111004">
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp3.1</TargetFramework>
    <QscVerbosity>detailed</QscVerbosity>
  </PropertyGroup>
</Project>
```

The project builds and runs successfully from the command line.

samarsha:
[...]

PR #753, opened 2020-11-24
Detect absence of .NET SDK 3.1 and display error message

vxfield:
This PR is a mitigation of issue #737.

### DIFF
--- a/src/QsCompiler/LanguageServer/LanguageServer.cs
+++ b/src/QsCompiler/LanguageServer/LanguageServer.cs
@@ -132,21 +132,21 @@ namespace Microsoft.Quantum.QsLanguageServer
 
         internal async void CheckDotNetSdkVersion()
         {
-            var sdkVersions = DotNetSdkHelper.GetSdkVersions();
-            if (sdkVersions == null)
+            var isDotNet31Installed = DotNetSdkHelper.IsDotNet31Installed();
+            if (isDotNet31Installed == null)
             {
                 this.LogToWindow("Unable to detect .NET SDK versions", MessageType.Error);
             }
             else
             {
-                if (!sdkVersions.HasDotNet31)
+                if (isDotNet31Installed != true)
                 {
                     const string dotnet31Url = "https://dotnet.microsoft.com/download/dotnet-core/3.1";
-                    this.LogToWindow($".NET SDK 3.1 not found. Quantum Development Kit Extension requires .NET Core SDK 3.1 to work properly ({dotnet31Url}).", MessageType.Error);
+                    this.LogToWindow($".NET Core SDK 3.1 not found. Quantum Development Kit Extension requires .NET Core SDK 3.1 to work properly ({dotnet31Url}).", MessageType.Error);
                     var downloadAction = new MessageActionItem { Title = "Download" };
                     var cancelAction = new MessageActionItem { Title = "No, thanks" };
                     var selectedAction = await this.ShowDialogInWindowAsync(
-                        "Quantum Development Kit Extension requires .NET Core SDK 3.1 to work properly. Please install .NET SDK 3.1.",
+                        "Quantum Development Kit Extension requires .NET Core SDK 3.1 to work properly. Please install .NET Core SDK 3.1.",
                         MessageType.Error,
                         new[] { downloadAction, cancelAction });
                     if (selectedAction != null
--- a/src/QsCompiler/LanguageServer/LanguageServer.cs
+++ b/src/QsCompiler/LanguageServer/LanguageServer.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Reactive.Linq;
@@ -123,12 +124,15 @@ namespace Microsoft.Quantum.QsLanguageServer
         internal Task NotifyClientAsync(string method, object args) =>
             this.rpc.NotifyWithParameterObjectAsync(method, args);  // no need to wait for completion
 
+        internal Task<T> InvokeAsync<T>(string method, object args) =>
+            this.rpc.InvokeWithParameterObjectAsync<T>(method, args);
+
         internal Task PublishDiagnosticsAsync(PublishDiagnosticParams diagnostics) =>
             this.NotifyClientAsync(Methods.TextDocumentPublishDiagnosticsName, diagnostics);
 
-        internal void CheckDotNetSDKVersion()
+        internal async void CheckDotNetSdkVersion()
         {
-            var sdkVersions = DotNetSDKHelper.GetSDKVersions();
+            var sdkVersions = DotNetSdkHelper.GetSdkVersions();
             if (sdkVersions == null)
             {
                 this.LogToWindow($"Unable to detect .NET SDK versions", MessageType.Error);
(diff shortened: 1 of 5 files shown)


src/QsCompiler/LanguageServer/LanguageServer.cs
```diff
@@ -126,6 +126,25 @@ public void Dispose()
...
+            {
+                this.LogToWindow($"Unable to detect .NET SDK versions", MessageType.Error);
+            }
+            else
+            {
+                if (!sdkVersions.HasDotNet31)
+                {
+                    this.LogToWindow($".NET SDK 3.1 not found. QDK requires .NET SDK 3.1 to work properly.", MessageType.Error);
+                    this.ShowInWindow(
+                        $"QDK requires .NET SDK 3.1 to work properly. Please install .NET SDK 3.1 from https://dotnet.microsoft.com/download/dotnet-core/3.1",
```
samarsha, line 142: I think the rest of the QDK actually works when only .NET 5 is installed, it's only the language server that's broken.

src/QsCompiler/LanguageServer/LanguageServer.cs
```diff
@@ -126,6 +126,25 @@ public void Dispose()
         internal Task PublishDiagnosticsAsync(PublishDiagnosticParams diagnostics) =>
             this.NotifyClientAsync(Methods.TextDocumentPublishDiagnosticsName, diagnostics);
 
+        internal void CheckDotNetSDKVersion()
```
samarsha, line 129: Nitpick, but the naming convention is for acronyms longer than 2 letters to be cased as if they were a normal word, so `SDK` should be `Sdk`.

src/QsCompiler/LanguageServer/LanguageServer.cs
vxfield: updated to "Quantum Development Kit Extension requires .NET Core SDK 3.1 to work properly".
- Spelled out QDK since users might not immediately recognize QDK.
- Added Extension, to clarify that the rest of QDK works fine without .NET Core 3.1

src/QsCompiler/LanguageServer/LanguageServer.cs
vxfield: Updated to "Sdk"

samarsha:
We should bring this change into `main` as well - maybe we could retarget this PR to `main` and then cherry-pick into the release branch?

src/QsCompiler/LanguageServer/Utils.cs
```diff
@@ -123,4 +143,42 @@ public override void Initialize(IEventSource eventSource)
...
+    internal static class DotNetSdkHelper
+    {
+        private static readonly Regex DotNet31Regex = new Regex(@"^3\.1\.\d+", RegexOptions.Multiline | RegexOptions.Compiled);
+        private static readonly Regex DotNet50Regex = new Regex(@"^5\.0\.\d+", RegexOptions.Multiline | RegexOptions.Compiled);
+
+        public class DotNetSDKVersions
+        {
+            public bool HasDotNet50 { get; set; }
+
+            public bool HasDotNet31 { get; set; }
```
samarsha, line 156: It looks like we are only using `HasDotNet31`. Can this class simply be a function that returns true if .NET Core SDK 3.1 is installed?

src/QsCompiler/LanguageServer/Utils.cs
```diff
@@ -33,6 +36,23 @@ internal static void ShowInWindow(this QsLanguageServer server, string text, Mes
...
+        internal static async Task<MessageActionItem> ShowDialogInWindowAsync(this QsLanguageServer server, string text, MessageType severity, MessageActionItem[] actionItems)
+        {
+            var message =
+                new ShowMessageRequestParams()
+                {
+                    Message = text,
+                    MessageType = severity,
+                    Actions = actionItems
+                };
+            QsCompilerError.Verify(server != null && message != null, "cannot show message - given server or text was null");
```
samarsha, line 51: This can't happen unless someone overrides the nullable type checks, and in that case the line below will throw NRE anyway.

```suggestion
```

src/QsCompiler/LanguageServer/Utils.cs
```diff
@@ -33,6 +36,23 @@ internal static void ShowInWindow(this QsLanguageServer server, string text, Mes
...
+            var action = await server.InvokeAsync<MessageActionItem>(Methods.WindowShowMessageRequestName, message);
+            return action;
```
samarsha, line 53:
```suggestion
            return await server.InvokeAsync<MessageActionItem>(Methods.WindowShowMessageRequestName, message);
```

src/QsCompiler/LanguageServer/LanguageServer.cs
```diff
@@ -123,9 +124,44 @@ public void Dispose()
...
+            else
+            {
+                if (!sdkVersions.HasDotNet31)
+                {
+                    const string dotnet31Url = "https://dotnet.microsoft.com/download/dotnet-core/3.1";
+                    this.LogToWindow($".NET SDK 3.1 not found. Quantum Development Kit Extension requires .NET Core SDK 3.1 to work properly ({dotnet31Url}).", MessageType.Error);
+                    var downloadAction = new MessageActionItem { Title = "Download" };
+                    var cancelAction = new MessageActionItem { Title = "No, thanks" };
+                    var selectedAction = await this.ShowDialogInWindowAsync(
+                        $"Quantum Development Kit Extension requires .NET Core SDK 3.1 to work properly. Please install .NET SDK 3.1.",
```
samarsha, line 149: I think we should consistently write ".NET Core SDK 3.1" instead of ".NET SDK 3.1".

src/QsCompiler/LanguageServer/LanguageServer.cs
```diff
@@ -123,9 +124,44 @@ public void Dispose()
...
+
         internal Task PublishDiagnosticsAsync(PublishDiagnosticParams diagnostics) =>
             this.NotifyClientAsync(Methods.TextDocumentPublishDiagnosticsName, diagnostics);
 
+        internal async void CheckDotNetSdkVersion()
+        {
+            var sdkVersions = DotNetSdkHelper.GetSdkVersions();
+            if (sdkVersions == null)
+            {
+                this.LogToWindow($"Unable to detect .NET SDK versions", MessageType.Error);
```
samarsha, line 138:
```suggestion
                this.LogToWindow("Unable to detect .NET SDK versions", MessageType.Error);
```

src/QsCompiler/LanguageServer/LanguageServer.cs
samarsha, line 149:
```suggestion
                        "Quantum Development Kit Extension requires .NET Core SDK 3.1 to work properly. Please install .NET SDK 3.1.",
```

src/QsCompiler/LanguageServer/Utils.cs
vxfield: Done

src/QsCompiler/LanguageServer/LanguageServer.cs
vxfield: Done

ricardo-espinoza (approved):

bettinaheim (commented):
I think that will work fine together with packaging a suitable global.json with the server, and setting the working directory to the server path.

src/QsCompiler/LanguageServer/LanguageServer.cs
```diff
@@ -123,9 +124,44 @@ public void Dispose()
...
+                        $"Quantum Development Kit Extension requires .NET Core SDK 3.1 to work properly. Please install .NET SDK 3.1.",
+                        MessageType.Error,
+                        new[] { downloadAction, cancelAction });
+                    if (selectedAction != null
+                        && selectedAction.Title == downloadAction.Title)
+                    {
+                        Process.Start(new ProcessStartInfo
+                        {
+                            FileName = dotnet31Url,
+                            UseShellExecute = true
```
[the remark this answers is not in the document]
bettinaheim: Actually, looking more closely, I believe this simply opens a webpage; in that case I suggest to not have the console pop up (unless there is a reason to that I am missing?).

src/QsCompiler/LanguageServer/LanguageServer.cs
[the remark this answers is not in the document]
vxfield: done

src/QsCompiler/LanguageServer/LanguageServer.cs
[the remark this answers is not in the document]
vxfield: the ShellExecute should not create a window, but I added  CreateNoWindow = true, just in case

src/QsCompiler/LanguageServer/Program.cs
```diff
@@ -121,7 +126,6 @@ private static int Run(Options options)
             catch (Exception ex)
             {
                 Log("[ERROR] MsBuildLocator could not register defaults.", options.LogFile);
```
bettinaheim, line 128:
```suggestion
                Log("[ERROR] MsBuildLocator could not register defaults.", options.LogFile);
                // Don't exit here, since exiting without establishing a connection will result in a cryptic failure of the extension. 
                // Instead, we proceed to create a server instance and establish the connection. 
                // Any errors can then be properly processed via the standard server-client communication as needed. 

```

bettinaheim (approved):

PR #755, opened 2020-11-25
Detect absence of .NET SDK 3.1 and display error message

vxfield:
This PR mitigates issue #737 by:
1. Adding a global.json in the LanguageServer folder
1. Forcing the Working Directory of the LanguageServer to be the same as the. LanguageServer is (and. therefore using the global.json from step above)
1. Not quitting the LanguageServer if there was an issue with the MSBuildLocator when .NET Core SDK 3.1 is not found
1. Check if .NET Core SDK 3.1 is not installed and display a message to the user with an option to download it (opens the download page) and asking the user to restart VS/Code.

### DIFF
--- a/src/QsCompiler/LanguageServer/global.json
+++ b/src/QsCompiler/LanguageServer/global.json
@@ -1,0 +1,6 @@
+{
+  "sdk": {
+    "version": "3.1.100",
+    "rollForward": "latestFeature"
+  }
+}
--- a/build/init.yml
+++ b/build/init.yml
@@ -16,7 +16,7 @@ steps:
     versionSpec: '5.6.0'
 
 - task: UseDotNet@2
-  displayName: 'Use .NET Core SDK 3.1.300'
+  displayName: 'Use .NET SDK 5.0.101'
   inputs:
     packageType: sdk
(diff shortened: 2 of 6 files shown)


vxfield:
@samarsha, @bettinaheim: this is exactly the same approved PR #753 but now targeting the main branch.

src/QsCompiler/LanguageServer/global.json
```diff
@@ -0,0 +1,6 @@
+{
+  "sdk": {
+    "version": "3.1.100",
+    "rollForward": "latestFeature"
+  }
+}
```
samarsha, line 6: We should have only one global.json that applies to the compiler repository. Currently, the global.json at the root of the repository also sets the SDK version to 3.1, but it may be updated to support 5 soon. In that case, running `dotnet` from the LanguageServer folder will require 3.1, but may use 5 when run from any other folder.

Since this global.json is intended to only be used by the language server after it's built and running, is it possible for it to have a different name in the source folder that `dotnet` won't recognize, but have it be renamed to global.json in the output directory?

samarsha (changes requested):

Issue #757, opened 2020-11-28
Dependencies for unused specializations are pruned, causing compilation failures


swernli:
The call graph design includes the functionality to trim to a concrete call graph that includes only those callables that are part of the entrypoint dependency tree. However, this means that if a given callable has a specialization that is unused, any callables only referred to in that specialization will have no edges and will not be included in the concrete graph, even though the specialization itself is not pruned. The end result is that the specialization will use types corresponding to callabes that are not generated, resulting in a compilation failure.

To see this in action, create a stand-along executable project using the latest release (0.14.2011120240), and the following code:
```qsharp
    @EntryPoint()
    operation Test() : Result {
        using (q = Qubit()) {
            ApplyPauli([PauliX], [q]);
            return M(q);
        }
    }
```
If you add an execution target to the project file that triggers classical control lifting and monomorphization, such as `<ExecutionTarget>honeywell.test</ExecutionTarget>`, the compilation of the generated C# will fail due to missing type definitions:

obj\qsharp\src\Microsoft.Quantum.Standard.g.cs(282,19): error CS0246: The type or namespace name '_dc509b171dd743d58d632169c29726fe_ApplyPauli' could not be found (are you missing a using directive or an assembly reference?)

This is because `ApplyPauli` has if-statements that turn into multiple callables via classical control lifting, and then because the controlled, adjoint, and controlled-adjoint specializations are unused, those generated callabes are removed. This leaves the call locations in place (the specializations themselves are not pruned, by design), while the dependent types for the lifted callables are missing.

The code that handles adding edges for a call (found in [ConcreteCallGraphWalker.cs](https://github.com/microsoft/qsharp-compiler/blob/main/src/QsCompiler/Transformations/CallGraph/ConcreteCallGraphWalker.cs#L201)) will only add an edge to the specialization actually used in a call. If instead it is a non-call usage, like setting the callable to a variable, then all specializations get an edge. So if the above code is changed to:
```qsharp
    @EntryPoint()
    operation Test() : Result {
        using (q = Qubit()) {
            let op = ApplyPauli;
            op([PauliX], [q]);
            return M(q);
        }
    }
```
the compilation will succeed.

ScottCarda-MS:
[...]

PR #758, opened 2020-11-28
Fix pruning of specialization dependencies

swernli:
This comments out some code so that any dependency on an operation results in an edge to each specialization of that operation instead of just the specific specialization used. This is a patch to eliminate the issue, but we should consider either eliminating the commented out code and making this design change permanent (never prune dependencies of specializations even if they are unused), or allow for the pruning of both unused dependencies and the unused specialization itself.

Note that this also skips tests that fail due to the given work around.

Works around #757

### DIFF
--- a/src/QsCompiler/Tests.Compiler/CallGraphTests.fs
+++ b/src/QsCompiler/Tests.Compiler/CallGraphTests.fs
@@ -334,7 +334,7 @@ type CallGraphTests (output:ITestOutputHelper) =
         AssertNotInConcreteGraph graph FooEmpty
         AssertNotInConcreteGraph graph BarEmpty
 
-    [<Fact>]
+    [<Fact(Skip="Workaround for Issue #757 means specialization pruning is disabled.")>]
     [<Trait("Category","Populate Call Graph")>]
     member this.``Concrete Graph Trims Specializations`` () =
         let graph = PopulateCallGraphWithExe 10 |> ConcreteCallGraph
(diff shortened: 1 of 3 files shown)


swernli:
@bettinaheim, I've been able to confirm that this workaround fixes the issue I was seeing using the generated packages and building code that uses `AppyPauli` and targets a provider.

swernli:
Annoyingly, this change means that the C# files produced will have more callables in them, which seems to be just enough to hit the timeout on the Linux perf tests. I tried a clean e2e build of main, and the Linux perf tests completed in 59 minutes, 4 seconds (see [here](https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=19559&view=logs&j=0375038c-6c3c-5d51-0b30-0aa0e0fe9d6b&t=35d59a6e-6924-5f7c-5d3f-53442006109d&l=640)). The limit is 60 minutes, so this PR is cancelled at 59 minutes and 59 seconds. From the logs, it looks like it gets cancelled at about 4/5ths of the way through, so making the test just slightly slower has resulted in hitting the timeout.

bettinaheim:
Given that this PR here https://github.com/microsoft/qsharp-runtime/pull/450 also had the runtime perf cancelled, the timeout may not actually be related to this change.

bettinaheim:
@swernli Closing this one, since I #764 deals with the same issue. I think we should have the feature branch there ready to merge to main soon, and otherwise we should cherry-pick that fix.

PR #759, opened 2020-11-29
Update example to 0.14.2011120240 and addressed the removal of NonNullable

filipw:
Bumped the example from the old SDK to the latest one and removed the usage of `NonNullable` from it, following https://github.com/microsoft/qsharp-compiler/pull/729


bettinaheim (approved):
Thanks for this update!

PR #764, opened 2020-12-04
Compiler integration and clean up for QIR generation

bettinaheim:
This PR 
- splits out the QIR data structures that are not specific to Q#, 
- cleans up the API of the QIR Generator
- properly integrates the QIR generation pass into the compiler
- adds an additional pass to arrange the target intrinsics in the format that the QIR generator expects them (one specialization per intrinsic callable, tagged with the appropriate attributes)
- enables QIR emission as part of the Sdk (i.e. via project property)
- contains some minor clean-up to the docs generator extension 
- ensures that the concrete call graph always contains all specializations for a used callable


src/QsCompiler/Transformations/CallGraph/ConcreteCallGraphWalker.cs
```diff
@@ -256,9 +258,18 @@ private void AddEdge(QsQualifiedName identifier, QsSpecializationKind kind, Type
                     var called = new ConcreteCallGraphNode(identifier, kind, typeRes);
                     var edge = new ConcreteCallGraphEdge(referenceRange);
                     this.Graph.AddDependency(this.CurrentNode, called, edge);
-                    if (!this.RequestStack.Contains(called) && !this.ResolvedNodeSet.Contains(called))
+
+                    // make sure we always build the dependencies for all specializations
```
ScottCarda-MS, line 262: Is this for issue #757? Does this make PR #758 unnecessary?

ScottCarda-MS (commented):
Would it be possible to split this PR into several different PRs into a feature branch so that the content can be more easily reviewed? It seems like some of these changes could be reviewed independently.

alan-geller (approved):

src/QsCompiler/Transformations/CallGraph/ConcreteCallGraphWalker.cs
bettinaheim: I think it should, but that needs to be confirmed.

PR #765, opened 2020-12-04
Response Files Reformat

ScottCarda-MS:
Use Absolute Paths for the proj, input and output arguments of the generated response file.
Formats the response file to make it easier to read and edit.


cesarzc:
Thanks for doing this Scott! It would be great if you could include an example of a generated response file as a comment.

src/QuantumSdk/Sdk/Sdk.targets
```diff
@@ -71,25 +71,33 @@
...
-      <_QscPackageLoadFallbackFoldersFlag Condition="@(ResolvedPackageLoadFallbackFolders->Count()) &gt; 0">--package-load-fallback-folders "@(ResolvedPackageLoadFallbackFolders,'" "')"</_QscPackageLoadFallbackFoldersFlag>
-      <_QscCommandArgs>--proj "$(PathCompatibleAssemblyName)" $(_QscCommandIsExecutableFlag) $(_QscCommandInputFlag) $(_QscCommandOutputFlag) $(_QscCommandReferencesFlag) $(_QscCommandLoadFlag) $(_QscCommandRuntimeFlag) $(_QscCommandTargetDecompositionsFlag) $(_QscPackageLoadFallbackFoldersFlag) $(_QscCommandTestNamesFlag) $(_QscCommandAssemblyPropertiesFlag) $(AdditionalQscArguments)</_QscCommandArgs>
+      <_QscCommandIsExecutableFlag Condition="'$(ResolvedQsharpOutputType)' == 'QsharpExe'">$(_NewLine)--build-exe</_QscCommandIsExecutableFlag>
+      <_QscCommandInputFlag Condition="@(QsharpCompile->Count()) &gt; 0">$(_NewLine)--input$(_NewLineIndent)"@(QsharpCompile->'%(FullPath)', '"$(_NewLineIndent)"')"</_QscCommandInputFlag>
+      <_QscCommandOutputPath>$([System.IO.Path]::GetFullPath('$(BaseIntermediateOutputPath)\qsharp\'))</_QscCommandOutputPath>
```
cesarzc, line 81: In the previous version of `Sdk.targets`, the output flag used `$(GeneratedFilesOutputPath)`. Would it make sense to keep using it but getting its full path instead of what could possibly be a relative path?

Since `$(GeneratedFilesOutputPath)` is guaranteed to be created and "clean", it seems safer to keep using it.

src/QuantumSdk/Sdk/Sdk.targets
```diff
@@ -71,25 +71,33 @@
...
+      <_QscCommandOutputPath>$([MSBuild]::Unescape('$(_QscCommandOutputPath)').Replace('\','/'))</_QscCommandOutputPath>
+      <_QscCommandOutputPath Condition="!HasTrailingSlash('$(_QscCommandOutputPath)')">$(_QscCommandOutputPath)/</_QscCommandOutputPath>
+      <_QscCommandOutputFlag>$(_NewLine)--output$(_NewLineIndent)"$(_QscCommandOutputPath)"</_QscCommandOutputFlag>
+      <_QscCommandReferencesFlag Condition="@(ResolvedQsharpReferences->Count()) &gt; 0">$(_NewLine)--references$(_NewLineIndent)"@(ResolvedQsharpReferences,'"$(_NewLineIndent)"')"</_QscCommandReferencesFlag>
+      <_QscCommandLoadFlag Condition="@(_PrioritizedResolvedQscReferences->Count()) &gt; 0">$(_NewLine)--load$(_NewLineIndent)"@(_PrioritizedResolvedQscReferences,'"$(_NewLineIndent)"')"</_QscCommandLoadFlag>
+      <_QscCommandRuntimeFlag>$(_NewLine)--runtime$(_NewLineIndent)$(ResolvedRuntimeCapabilities)</_QscCommandRuntimeFlag> 
+      <_QscCommandTargetDecompositionsFlag Condition="@(ResolvedTargetSpecificDecompositions->Count()) &gt; 0">$(_NewLine)--target-specific-decompositions$(_NewLineIndent)"@(ResolvedTargetSpecificDecompositions,'"$(_NewLineIndent)"')"</_QscCommandTargetDecompositionsFlag>
```
cesarzc, line 88: This might be a bit subjective, but I would find it more readable if the formatted values for the flags were encapsulated in another variable.

E.g.:
```
<_QscCommandTargetDecompositionsFlagFormattedValue>$(_NewLineIndent)"@(ResolvedTargetSpecificDecompositions,'"$(_NewLineIndent)"')"</_QscCommandTargetDecompositionsFlagFormattedValue>
```

My only concern is with the conditions, but that might not be a problem at all.

src/QuantumSdk/Sdk/Sdk.targets
```diff
@@ -71,25 +71,33 @@
...
+      <_QscCommandPredefinedAssemblyProperties Condition="'$(QsharpDocsGeneration)'">$(_QscCommandPredefinedAssemblyProperties)$(_NewLineIndent)DocsOutputPath:"$(QsharpDocsOutputPath)"</_QscCommandPredefinedAssemblyProperties>
+      <_QscCommandAssemblyPropertiesFlag>$(_NewLine)--assembly-properties$(_QscCommandPredefinedAssemblyProperties)</_QscCommandAssemblyPropertiesFlag>
+      <_QscCommandAssemblyPropertiesFlag Condition="'$(QscCommandAssemblyProperties)' != ''">$(_QscCommandAssemblyPropertiesFlag)$(_NewLineIndent)$(QscCommandAssemblyProperties)</_QscCommandAssemblyPropertiesFlag>
+      <AdditionalQscArguments Condition="'$(AdditionalQscArguments)' != ''">$(_NewLine)$(AdditionalQscArguments)</AdditionalQscArguments>
+      <_QscCommandArgs>$(_QscCommandProjFlag)$(_QscCommandIsExecutableFlag)$(_QscCommandInputFlag)$(_QscCommandOutputFlag)$(_QscCommandReferencesFlag)$(_QscCommandLoadFlag)$(_QscCommandRuntimeFlag)$(_QscCommandTargetDecompositionsFlag)$(_QscPackageLoadFallbackFoldersFlag)$(_QscCommandTestNamesFlag)$(_QscCommandAssemblyPropertiesFlag)$(AdditionalQscArguments)</_QscCommandArgs>
```
cesarzc, line 100: This might also be subjective, but I think that including `<_NewLine>` and `<_NewLineIndent>` here instead of in the flag commands themselves makes everything more readable.

src/QuantumSdk/Sdk/Sdk.targets
ScottCarda-MS: If the condition is met, then we want the value and the formatting. But if the condition is not met, we don't want to throw in the extra newlines. So if we went with a regular value and a formatted value, they would both have to have the condition on them.

src/QuantumSdk/Sdk/Sdk.targets
ScottCarda-MS: We don't want to add the newlines and indentation unless there is a value to print out, so these need to be encapsulated into a value that has the condition attached. The formatting tags can't be included here or we would end up printing out a bunch on extra newlines when conditions are met.

src/QuantumSdk/Sdk/Sdk.targets
ScottCarda-MS: I was able to clean this up a little bit and use the GeneratedFilesOutputPath as a starting point. But I still have to redo the unescaping and backslash replacement.

src/QuantumSdk/Sdk/Sdk.targets
ScottCarda-MS: Hmm, after trying to do something like this, I'm not sure this can be done in a consistent way for all the flags and still keep a reasonable amount of tags.

cesarzc: That's true, it would add one extra tag per flag, and it would require duplicated condition handling. I am ok with leaving it as it currently is.

cesarzc (approved):
Looks good to me, thanks Scott!

PR #766, opened 2020-12-06
Use the XML exception tag in doc comments (F#)

kevinhartman:
Changes:
- Adds proper XML tags for any exceptions previously mentioned within the summary tag.
- Enables warning of malformed F# doc comments (FS3390) as an error.

A few things to note:
- At this time, only F# Tools 11.0.0 supports XML doc comment validation ([release notes](https://github.com/dotnet/fsharp/blob/687d2118488e84317c9965ac93ea454069a51618/release-notes.md#fsharp-tools-1100)).
- Code references don't seem to be validated yet, even by 11.0.0 (i.e. an incorrect `cref` won't be caught).
- Parameter name reference validation actually works :)

Ref #699


samarsha:
Hi @kevinhartman, sorry for the delay in reviewing this! I'll take a look on Monday. Also, #769 created some merge conflicts here - although that PR didn't actually change any of the doc comments, so the conflicts shouldn't be too bad to resolve. Since #769 was my fault, I'll resolve the merge conflicts for you unless you want to handle them. :)

kevinhartman:
No worries, @samarsha! I'm happy to resolve them :D I should be able to get to in the next day or two.

src/QsCompiler/SyntaxProcessor/ScopeTools.fs
```diff
@@ -52,11 +52,14 @@ type private TrackedScope =
             requiredSupport [] functors
 
 
-/// The SymbolTracker class does *not* make a copy of the given NamespaceManager,
-/// but instead will throw an InvalidOperationException upon accessing its content if that content has been modified
-/// (i.e. if the version number of the NamespaceManager has changed).
-/// The constructor throws an ArgumentException if the given NamespaceManager does not contain all resolutions,
-/// or if a callable with the given parent name does not exist in the given NamespaceManager.
+/// <summary>
+/// The <see cref="SymbolTracker"/> class does *not* make a copy of the given <see cref="NamespaceManager"/>,
```
kevinhartman, line 56: Adding code refs inside summary tags isn't in the original scope of this PR, but I could resist for this one :)

src/QsCompiler/Core/SyntaxGenerator.fs
```diff
@@ -283,10 +299,12 @@ module SyntaxGenerator =
         | _ -> ArgumentException "expecting the given argument tuple to be a QsTuple" |> raise
         |> QsTuple
 
+    /// <summary>
     /// Given a typed expression that is used as argument to an operation and a typed expression for the control qubits,
     /// combines them to a suitable argument for the controlled version of the originally called operation under the assumption that the argument was correct.
     /// The range information for the built expression is set to Null.
-    /// Throws an ArgumentException if the given expression for the control qubits is a valid type but not of type Qubit[].
+    /// </summary>
+    /// <exception cref="ArgumentException"><paramref name="ctlQs"/> is a valid type but not of type <see cref="T:Qubit[]"/>.</exception>
```
samarsha, line 307: I think `Qubit[]` here is actually referring to the Q# type, so `ArrayType Qubit` in F# (but I'm not sure if it's worth trying to create a `<see>` tag for that).

src/QsCompiler/SyntaxProcessor/ScopeTools.fs
samarsha: Looks good. :)

src/QsCompiler/SyntaxProcessor/ScopeTools.fs
```diff
@@ -226,9 +233,11 @@ type SymbolTracker(globals: NamespaceManager, sourceFile, parent: QsQualifiedNam
             LocalVariableDeclaration<_>.New mut ((decl.Position, decl.Range), name, decl.Type, qDep)
             |> this.TryAddVariableDeclartion
 
+    /// <summary>
     /// Updates the quantum dependency of the local variable with the given name.
-    /// Throws an ArgumentException if no local variable with the given name is visible on the current scope.
-    /// Throws an InvalidOperationException if no scope is currently open, or the variable is immutable.
+    /// </summary>
+    /// <exception cref="ArgumentException">No local variable with the given name is visible on the current scope.</exception>
+    /// <exception cref="InvalidOperationException">No scope is currently open, or the variable is immutable.</exception>
```
samarsha, line 240:
```suggestion
    /// <exception cref="ArgumentException"><paramref name="varName"/> is not visible in the current scope.</exception>
    /// <exception cref="InvalidOperationException">No scope is currently open, or <paramref name="varName"/> is immutable.</exception>
```

src/QsCompiler/Core/SymbolResolution.fs
```diff
@@ -553,8 +562,9 @@ module SymbolResolution =
     /// and generates suitable diagnostics if they are not, replacing them by the Q# type denoting an invalid type.
     /// Returns the resolved type as well as an array with diagnostics.
     /// IMPORTANT: for performance reasons does *not* verify if the given the given parent and/or source file is inconsistent with the defined callables.
-    /// May throw an ArgumentException if no namespace with the given name exists, or the given source file is not listed as source of that namespace.
-    /// Throws a NotSupportedException if the QsType to resolve contains a MissingType.
+    /// </summary>
+    /// <exception cref="NotSupportedException"><paramref name="qsType"/> contains a <see cref="MissingType"/>.</exception>
+    /// <exception cref="ArgumentException">No namespace with the given name exists, or the given source file is not listed as source of that namespace.</exception>
```
samarsha, line 567: I can't remember where exactly our discussion on this landed, but did we want to split these kinds of unrelated conditions into two tags?

```suggestion
    /// <exception cref="ArgumentException">No namespace with the given name exists</exception>
    /// <exception cref="ArgumentException">The given source file is not listed as source of that namespace.</exception>
```

samarsha (commented):
Thanks for working on this and for resolving the merge conflicts! The compiler checks for invalid doc comments sound great. :)

src/QsCompiler/Core/SymbolResolution.fs
kevinhartman: I think we decided to split them up for completely unrelated causes (like the case we have here), since it's then easy to generate documentation with a formatting that makes the separate causes easy to distinguish. LGTM!

src/QsCompiler/Core/SyntaxGenerator.fs
kevinhartman: Just pushed a separate commit which adds `<see>` tags for both `ArrayType` and `Qubit`.

I don't believe there's a better way to compose them in this case, unless we were to refer directly to `QubitArray`, but this is likely not possible (or at least not hip) since it's a private member.

If you feel strongly that this is not readable, we can easily drop the commit before merging.

kevinhartman:
> Thanks for working on this and for resolving the merge conflicts! The compiler checks for invalid doc comments sound great. :)

Thanks! :) Sorry about all of the param refs I'd missed. I'm used to Scala's multiple parameter lists for currying, but F#'s curried syntax without parenthesis threw me off!

src/QsCompiler/SyntaxProcessor/ScopeTools.fs
```diff
@@ -226,9 +233,11 @@ type SymbolTracker(globals: NamespaceManager, sourceFile, parent: QsQualifiedNam
             LocalVariableDeclaration<_>.New mut ((decl.Position, decl.Range), name, decl.Type, qDep)
             |> this.TryAddVariableDeclartion
 
+    /// <summary>
     /// Updates the quantum dependency of the local variable with the given name.
-    /// Throws an ArgumentException if no local variable with the given name is visible on the current scope.
-    /// Throws an InvalidOperationException if no scope is currently open, or the variable is immutable.
+    /// </summary>
+    /// <exception cref="ArgumentException"><paramref name="varName"/> is not visible in the current scope.</exception>
+    /// <exception cref="InvalidOperationException">No scope is currently open, or <paramref name="varName"/> is immutable.</exception>
```
samarsha, line 240: I think this might also work better as two separate tags:

```suggestion
    /// <exception cref="InvalidOperationException">No scope is currently open.
    /// <exception cref="InvalidOperationException"><paramref name="varName"/> is immutable.</exception>
```

samarsha (approved):
> Sorry about all of the param refs I'd missed. I'm used to Scala's multiple parameter lists for currying, but F#'s curried syntax without parenthesis threw me off!

No worries! Looks good, thanks again!

src/QsCompiler/Core/SyntaxGenerator.fs
samarsha: I think the current version looks fine, thanks!

src/QsCompiler/SyntaxProcessor/ScopeTools.fs
```diff
@@ -226,9 +233,12 @@ type SymbolTracker(globals: NamespaceManager, sourceFile, parent: QsQualifiedNam
             LocalVariableDeclaration<_>.New mut ((decl.Position, decl.Range), name, decl.Type, qDep)
             |> this.TryAddVariableDeclartion
 
+    /// <summary>
     /// Updates the quantum dependency of the local variable with the given name.
-    /// Throws an ArgumentException if no local variable with the given name is visible on the current scope.
-    /// Throws an InvalidOperationException if no scope is currently open, or the variable is immutable.
+    /// </summary>
+    /// <exception cref="ArgumentException"><paramref name="varName"/> is not visible in the current scope.</exception>
+    /// <exception cref="InvalidOperationException">No scope is currently open.
```
samarsha, line 240:
```suggestion
    /// <exception cref="InvalidOperationException">No scope is currently open.</exception>
```

Sorry, it looks like I missed the closing `</exception>` tag in my last suggestion. I'm surprised the doc comment verification didn't cause the build to fail, though. It looks like it actually is working in Visual Studio (this comment has a red squiggly), but running `dotnet build` from the command line doesn't catch the error.

Do you know why this might be happening - maybe the `<OtherFlags>--warnon:3390</OtherFlags>` tag isn't working with `dotnet build` for some reason?

kevinhartman: Hmm that's a great question. Does passing the warning flag on the command line work? If not, is it possible that a version of F# Tools older than 11.0.0 is being used under the hood? I also wonder what version is used in CI.

I should be able to poke around at some point over the next day or two. Thanks for your patience on working through this PR!

samarsha: Oh, I forgot that only F# from .NET 5 can do the validation. :) I think that would explain it - I'm using Visual Studio 16.8 which comes built-in with .NET 5, but `dotnet` on the command line follows the global.json file which pins the version to .NET Core 3.1. The CI is also using 3.1. So it sounds like XML comment validation will probably start working when we update those to .NET 5.

kevinhartman: Ah, that's right! I forgot that I updated the config for .NET 5 when I was testing. Yes, it should start working as soon as that upgrade is complete :)

PR #767, opened 2020-12-07
Fix ClassicallyControlled Rewrite bugs

### LINKED ISSUE #763: Classical Control Rewrite step does not process ExpressionKind.NOT
When the classical control rewrite step executes, the expectation is that any conditionals that can be refactored and lifted are translated into calls to the appropriate ApplyIf* callables. However, the code only handles restructuring ExpressionKind.AND and ExpressionKind.OR, but does not process ExpressionKind.NOT, so the inclusion of a `not` in the expression will cause processing to stop and leave behind an untransformed conditional. For targets that expect classical controls to be either refactored or fully evaluated, this can cause failures later during compilation or execution.  For example, the if-condition `if (M(q) != One)` will correctly be handled (since it is ExpressionKind.NEQ), but the condition `if (not (M(q) == One))` will not be transformed and then will fail target translation at execution time.

The fix should be to extend [ReshapeConditional](https://github.com/microsoft/qsharp-compiler/blob/main/src/QsCompiler/Transformations/ClassicallyControlled.cs#L167) to handle NOT in addition to AND and OR.

swernli:
This change includes 3 fixes/updates to the ClassicallyControlled rewrite step:
1. It adds handling for ExpressionKind.NOT so that those conditions get properly restructured and processed for lifting (fixes #763)
2. If a conditional statement has an empty scope, it gets replaced with a call to `NoOp` instead of being lifted into an empty callable.
3. <s>Updates the propagation of `IsValidScope` through the recursive calls to avoid lifting outer conditional scopes when inner scopes are not valid for lifting.</s> The fix here was incorrect, so it was removed and a separate issue https://github.com/microsoft/qsharp-compiler/issues/768 was filed.

Things left to do before check-in:

- [x] Add test cases for conditional lifting with NOT (top level and nested condition)
- [x] Add test case for lifting an empty scope (becomes `NoOp`)
- [x] <s>Add test case for nested conditionals that cannot be lifted</s>

### DIFF
--- a/src/QsCompiler/Transformations/ClassicallyControlled.cs
+++ b/src/QsCompiler/Transformations/ClassicallyControlled.cs
@@ -160,6 +160,41 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.ClassicallyControlled
                     }
                 }
 
+                private (bool, QsConditionalStatement) ProcessNOT(QsConditionalStatement conditionStatement)
+                {
+                    // This method expects elif blocks to have been abstracted out
+                    if (conditionStatement.ConditionalBlocks.Length != 1)
+                    {
+                        return (false, conditionStatement);
+                    }
+
+                    var (condition, block) = conditionStatement.ConditionalBlocks[0];
+
+                    if (condition.Expression is ExpressionKind.NOT notCondition)
+                    {
+                        var newConditionalBlock = conditionStatement.Default;
+                        if (newConditionalBlock.IsNull)
+                        {
+                            var emptyScope = new QsScope(
+                                ImmutableArray<QsStatement>.Empty,
+                                LocalDeclarations.Empty);
+                            newConditionalBlock = QsNullable<QsPositionedBlock>.NewValue(
+                                new QsPositionedBlock(
+                                    emptyScope,
+                                    QsNullable<QsLocation>.Null,
+                                    QsComments.Empty));
+                        }
+
+                        return (true, new QsConditionalStatement(
+                            ImmutableArray.Create(Tuple.Create(notCondition.Item, newConditionalBlock.Item)),
+                            QsNullable<QsPositionedBlock>.NewValue(block)));
+                    }
+                    else
+                    {
+                        return (false, conditionStatement);
+                    }
+                }
+
                 /// <summary>
                 /// Converts conditional statements to nested structures so they do not
                 /// have elif blocks or top-most OR or AND conditions.
@@ -170,13 +205,14 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.ClassicallyControlled
                     {
                         var stm = condition.Item;
                         (_, stm) = this.ProcessElif(stm);
-                        bool wasOrProcessed, wasAndProcessed;
+                        bool wasOrProcessed, wasAndProcessed, wasNotProcessed;
                         do
                         {
                             (wasOrProcessed, stm) = this.ProcessOR(stm);
                             (wasAndProcessed, stm) = this.ProcessAND(stm);
+                            (wasNotProcessed, stm) = this.ProcessNOT(stm);
                         }
-                        while (wasOrProcessed || wasAndProcessed);
+                        while (wasOrProcessed || wasAndProcessed || wasNotProcessed);
 
                         return new QsStatement(
                             QsStatementKind.NewQsConditionalStatement(stm),
@@ -184,6 +220,7 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.ClassicallyControlled
                             statement.Location,
                             statement.Comments);
                     }
+
                     return statement;
                 }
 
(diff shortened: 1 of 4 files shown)


src/QsCompiler/Transformations/ClassicallyControlled.cs
```diff
@@ -160,6 +160,41 @@ public StatementTransformation(SyntaxTreeTransformation parent) : base(parent)
...
+
+                    if (condition.Expression is ExpressionKind.NOT notCondition)
+                    {
+                        var newConditionalBlock = conditionStatement.Default;
+                        if (newConditionalBlock.IsNull)
+                        {
+                            var emptyScope = new QsScope(
+                                ImmutableArray<QsStatement>.Empty,
+                                LocalDeclarations.Empty);
+                            newConditionalBlock = QsNullable<QsPositionedBlock>.NewValue(
```
ScottCarda-MS, line 181: There doesn't seem to be a need to wrap this in a QsNullable as it has an item and that item is what is used in line 189.

src/QsCompiler/Transformations/ClassicallyControlled.cs
```diff
@@ -870,7 +929,7 @@ public override QsStatementKind OnConditionalStatement(QsConditionalStatement st
                         }
 
                         this.SharedState.GeneratedOpParams = contextParams;
-                        this.SharedState.IsValidScope = contextValidScope;
+                        this.SharedState.IsValidScope = this.SharedState.IsValidScope && contextValidScope;
```
ScottCarda-MS, line 932: I am thinking that it may be the case where an inner scope is not valid for lifting, but an outer scope may be.
For example if there is a mutable variable defined in the outer scope that is updated in the inner scope. The inner scope would not be valid because it is setting a mutable variable defined outside of itself, but the outer scope would be fine for lifting because that mutable is defined inside itself.

src/QsCompiler/Transformations/ClassicallyControlled.cs
swernli: Yup, you are right, I can restructure the shape of the if-statement and get rid of the extra layer here. I'll make that update!

src/QsCompiler/Transformations/ClassicallyControlled.cs
swernli: Oh, that's a good point. I had been so focused on the scenario where the mutable is defined outside of the outer scope and then modified inside the inner scope. This may not be as easy of a fix then, because an inner scope that is not valid for lifting would need to return which variables made it invalid for lifting so that the outer scope could then differentiate between those mutable updates that were made to something declaring in the outer scope vs not. I'll follow up with you to get your thoughts on how to handle this, but it may be that this is better saved for the larger review and fix to the use of `IsValidScope` that is pending.

ScottCarda-MS: It should work the following way:
```
operation Foo() {
    if (...) {
        mutable x = 3;
        if (...) {
            set x <- 4;
        }
    }
}
```
The inner scope is evaluated first and determined to be invalid because of the mutable set. The state after this is:
```
operation Foo() {
    if (...) {
        mutable x = 3;
        if (...) {
            set x <- 4;
        }
    }
}
```
Then the outer scope is evaluated and looks at its entire contents at that point. It is determined to be valid for lifting because the mutable variable that is set is defined in its scope. The state after this is:
```
LiftedOp (...) {
    mutable x = 3;
    if (...) {
        set x <- 4;
    }
}
...
operation Foo() {
    LiftedOp(...);
}
```

ScottCarda-MS: If instead we have an invalidation due to return:
```
operation Foo() {
    if (...) {
        if (...) {
            return;
        }
    }
}
```
Then the inner scope is determined to be invalid from having the return.
Then the outer scope is independently evaluated and seen to be invalid from the same return.
This results in no lifting and no change.

ScottCarda-MS: If we have:
```
operation Foo() {
    if (...) {
        if (...) {
            // Valid code for lifting
        }
        return;
    }
}
```
Then the inner scope is valid, resulting in:
```
LiftedOp (...) {
    // Valid code for lifting
}
...
operation Foo() {
    if (...) {
        LiftedOp(...);
        return;
    }
}
```
Then the outer scope is evaluated and found to be invalid.

ScottCarda-MS: Finally, a fully valid case:
```
operation Foo() {
    if (...) {
        if (...) {
            // Valid for lifting
        }
        // Also valid for lifting
    }
}
```
Then the inner scope is evaluated, found valid, and lifted:
```
operation LiftOp1(...) {
    // Valid for lifting
}
...
operation Foo() {
    if (...) {
        LiftOp1(...);
        // Also valid for lifting
    }
}
```
Then the outer scope is evaluated, found valid, and lifted:
```
operation LiftOp1(...) {
    // Valid for lifting
}
...
operation LiftOp2(...) {
    LiftOp1(...);
    // Also valid for lifting
}
...
operation Foo() {
    LiftOp2(...);
}
```

swernli: Yeah, that makes sense.  The case I'm worried about is this:
```qsharp
operation Foo() {
    mutable x = 3;
    if (...) {
        if (...) {
            set x = 4;
        }
    }
}
```
In that case, the inner scope correctly doesn't get lifted, but then the outer scope DOES get lifted because it didn't have an explicit set inside of it (the nested set isn't seen). My approach of just propagating up the validity of the inner scope fixes this bug but breaks the example you posted. We need some way to differentiate inner scope invalidity that should propagate to the parent and those that shouldn't, which would depend on where the initial definition of the mutable variable came from.

ScottCarda-MS: Yah after testing this I can see that it doesn't do what is expected. The whole scope needs to be evaluated again when it comes to the outer scope, but maybe that is not efficient? Either way this seems like an open issue.

swernli: Yeah, agreed. Since it just trades one issue for another, I'll remove the change I made for this here and file it as a separate issue.

swernli: Filed as https://github.com/microsoft/qsharp-compiler/issues/768

ScottCarda-MS: Thanks for filing that issue: #768

ScottCarda-MS (approved):

PR #771, opened 2020-12-09
Show detailed diagnostic messages for callables with an unsupported required capability

### LINKED ISSUE #736: Better targeting-related error messages for multi-project Q# applications
### Is your feature request related to a problem? Please describe.

When building a Q# application for Azure Quantum with multiple local projects, targeting-related error messages point to the source code in the executable project. This is not very informative, as the actual error may be in a referenced local library project, and the error doesn't tell me specifically what I'm doing that is unsupported.

For example, if I have a library project with the following code:

```
namespace AzureTestLibrary {
    open Microsoft.Quantum.Canon;
    open Microsoft.Quantum.Convert;
    open Microsoft.Quantum.Measurement;

    operation MeasureEqualSuperposition() : Result[] {
        using (qubits = Qubit[3])
        {
            ApplyToEach(H, qubits);
            let results = ResultArrayAsBoolArray(MultiM(qubits));
            return BoolArrayAsResultArray(results);
        }
    }
}
```

and an executable project targeting `ionq.simulator`, referencing the above project, with the following code:

```
namespace AzureTestExecutable {
    open Microsoft.Quantum.Canon;

    @EntryPoint()
    operation GenerateRandomNumber() : Result[] {
        return AzureTestLibrary.MeasureEqualSuperposition();
    }
}
```

then I get the following compilation error:

```
C:\QDK\Test\AzureTestExecutable\Program.qs(7,16): error QS5027: The callable MeasureEqualSuperposition requires runtime capabilities that are not supported by the target IonQProcessor. [C:\QDK\Test\AzureTestExecutable\AzureTestExecutable.csproj]
```

### Describe the solution you'd like
It would be much better if these error messages pointed to the source code in the local library that actually needs to be changed. For example, I'd hope to get something like:

```
C:\QDK\Test\AzureTestLibrary\Library.qs(12,27): error QS5027: The callable ResultArrayAsBoolArray requires runtime capabilities that are not supported by the target IonQProcessor. [C:\QDK\Test\AzureTestLibrary\AzureTestLibrary.csproj]
```

### Describe alternatives you've considered

The only alternative I am aware of would be to put all of the code in a single executable project. But for code-sharing purposes, this is not a practical solution.

### Additional context

This is also necessary in order to merge https://github.com/microsoft/iqsharp/pull/362, which will improve the experience for Azure Quantum users in Python and Jupyter. IQ# uses a similar project structure as described above - the user code is compiled into one or more library assemblies, and then a targeted executable references those library assemblies.

samarsha:
Adds secondary diagnostics for unsupported capabilities that give the reasons why a callable from a library isn't supported. For example:

```qsharp
namespace Library {
    function MyResultAsBool(result : Result) : Bool {
        return result == One ? true | false;
    }
}
```

```qsharp
namespace App {
    open Library;

    operation Foo() : Bool {
        return MyResultAsBool(Zero);
    }
}
```

gives the diagnostics when compiled for Honeywell:

> Error QS5027:
C:\Users\samarsha\Desktop\QSharpSandbox\App\App.qs
Position: [ln 5, cn 16]
The callable MyResultAsBool requires the FullComputation runtime capability, which is not supported by the target HoneywellProcessor.

> Warning QS5024:
C:\Users\samarsha\Desktop\QSharpSandbox\App\App.qs
Position: [ln 5, cn 16]
MyResultAsBool: Measurement results cannot be compared here. The target HoneywellProcessor only supports comparing measurement results as part of the condition of an if- or elif-statement in an operation. [C:\Users\samarsha\Desktop\QSharpSandbox\Library\Library.qs:3:16]

Resolves #736.


src/QsCompiler/BondSchemas/BondSchemaTranslator.cs
```diff
@@ -374,7 +374,7 @@ qsResult.Tag switch
                 Kind = qsSpecialization.Kind.ToBondSchema(),
                 Parent = qsSpecialization.Parent.ToBondSchema(),
                 Attributes = qsSpecialization.Attributes.Select(a => a.ToBondSchema()).ToList(),
-                SourceFile = qsSpecialization.SourceFile,
+                SourceFile = qsSpecialization.Source.CodeFile,
```
cesarzc, line 377: We are losing a bit of data from the `Source` object by just including the value of `CodeFile`. This is consistent with what we had before but we would not have the additional information from loaded references. Is this by design?

samarsha: Yes, because `AssemblyFile` is always set to the DLL file path after deserializing a reference, I can't see any situation in which a serialized `AssemblyFile` property would be used.

src/QsCompiler/BondSchemas/CompilerObjectTranslator.cs
```diff
@@ -201,7 +202,7 @@ bondQsBindingKind switch
                 fullName: bondQsCallable.FullName.ToCompilerObject(),
                 attributes: bondQsCallable.Attributes.Select(a => a.ToCompilerObject()).ToImmutableArray(),
                 modifiers: bondQsCallable.Modifiers.ToCompilerObject(),
-                sourceFile: bondQsCallable.SourceFile,
+                source: new Source(bondQsCallable.SourceFile, QsNullable<string>.Null),
```
cesarzc, line 205: I think it would be good to encapsulate the creation of a `Source` object into a `ToCompilerObject()` method. I know it is only one line of code but it would be consistent with the pattern used in this file, and it would be less error prone in case the `Source` type changes or what we pass to the constructor changes.

src/QsCompiler/CompilationManager/CompilationUnit.cs
```diff
@@ -51,9 +50,9 @@ public class Headers
             public Headers(string source, IEnumerable<QsNamespace> syntaxTree)
                 : this(
                     source,
-                    syntaxTree.Callables().Where(c => c.SourceFile.EndsWith(".qs")).Select(CallableDeclarationHeader.New),
-                    syntaxTree.Specializations().Where(c => c.SourceFile.EndsWith(".qs")).Select(s => (SpecializationDeclarationHeader.New(s), (SpecializationImplementation?)s.Implementation)),
-                    syntaxTree.Types().Where(c => c.SourceFile.EndsWith(".qs")).Select(TypeDeclarationHeader.New))
+                    syntaxTree.Callables().Where(c => c.Source.AssemblyFile.IsNull).Select(CallableDeclarationHeader.New),
```
cesarzc, line 53: Does this mean that any `Source` with a null `AssemblyFile` is a Q# source?

Filtering based on this condition makes this code a bit harder to understand. Would it be possible to add a property to `Source` whose name is more explicit (e.g. `IsQSharp`)?

src/QsCompiler/CompilationManager/EditorSupport/SymbolInformation.cs
```diff
@@ -244,7 +243,7 @@ bool OverlapsWithPosition(Range symRange)
                     return false;
                 }
                 referenceLocations = parent.Specializations
-                    .Where(spec => spec.SourceFile == file.FileName)
+                    .Where(spec => spec.Source.AssemblyOrCode == file.FileName)
```
cesarzc, line 246: Is `spec.Source.AssemblyOrCode == file.FileName` correct? Should this be `spec.Source.CodeFile == file.FileName` instead?

src/QsCompiler/Core/DeclarationHeaders.fs
```diff
@@ -108,24 +108,55 @@ module DeclarationHeader =
...
         Type: ResolvedType
         TypeItems: QsTuple<QsTypeItem>
         Documentation: ImmutableArray<string>
     }
+
     [<JsonIgnore>]
     member this.Location = DeclarationHeader.CreateLocation(this.Position, this.SymbolRange)
 
-    member this.FromSource source = { this with SourceFile = source }
+    // TODO: RELEASE 2021-07: Remove TypeDeclarationHeader.SourceFile.
```
cesarzc, line 155: Just curious about why is this being removed in 2021-07 release. Is this because it is six months after the release of this change?

samarsha: Yes, I believe the compiler API tries to follow a 6-month deprecation period when possible.

src/QsCompiler/Core/Utils.fs
```diff
@@ -1,7 +1,8 @@
-﻿module internal Microsoft.Quantum.QsCompiler.Utils
+﻿module Microsoft.Quantum.QsCompiler.Utils
 
 /// Converts a C# Try-style return value into an F# option.
-let tryToOption =
+[<CompiledName "TryOption">]
```
cesarzc, line 4: Since this is now public, should the comment above be enclosed in `<summary></summary>` tags?

src/QsCompiler/CompilationManager/EditorSupport/SymbolInformation.cs
cesarzc: Looking at the code a bit more it seems like `spec.Source.AssemblyOrCode == file.FileName` is correct. I'll just wait for confirmation in case I have misread the code again :)

src/QsCompiler/DataStructures/SyntaxTree.fs
```diff
@@ -671,6 +673,37 @@ and QsStatement =
...
+                       [<Optional; DefaultParameterValue null>] ?assemblyFile) =
+        { source with
+            CodeFile = codeFile |> Option.defaultValue source.CodeFile
+            AssemblyFile = assemblyFile |> QsNullable<_>.FromOption |> QsNullable.orElse source.AssemblyFile
+        }
+
+/// Operations for source files.
+module Source =
+    /// The assembly file path for this source if one exists, otherwise the code file path.
+    [<CompiledName "AssemblyOrCode">]
```
cesarzc, line 703: While reading the code that uses `AssemblyOrCode`, I originally interpreted it as getting the path to the assembly or getting the code itself. I think it would help readability to be more explicit in the name, e.g. `AssemblyFileOrCodeFile`.

src/QsCompiler/DataStructures/SyntaxTree.fs
```diff
@@ -671,6 +673,37 @@ and QsStatement =
...
+    /// </summary>
+    member source.With([<Optional; DefaultParameterValue null>] ?codeFile,
+                       [<Optional; DefaultParameterValue null>] ?assemblyFile) =
+        { source with
+            CodeFile = codeFile |> Option.defaultValue source.CodeFile
+            AssemblyFile = assemblyFile |> QsNullable<_>.FromOption |> QsNullable.orElse source.AssemblyFile
+        }
+
+/// Operations for source files.
+module Source =
```
cesarzc, line 701: I don't have a ton of experience with F# so maybe this question is a bit naive. Why is a `Source` module needed? When an object of `Source` type is present, wouldn't be enough just to call the `AssemblyOrCode` member of the object?

src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -173,7 +173,9 @@ type ErrorCode =
     | ResultComparisonNotInOperationIf = 5024
     | ReturnInResultConditionedBlock = 5025
     | SetInResultConditionedBlock = 5026
-    | UnsupportedCapability = 5027
+    | UnsupportedCallableCapability = 5027
+    // TODO: RELEASE 2021-07: Remove ErrorCode.UnsupportedCapability.
+    | [<Obsolete "Renamed to UnsupportedCallableCapability.">] UnsupportedCapability = 5027
```
bettinaheim, line 178: What is the reason for renaming it? In general, there is always a temptation to rename items, but since it is a burden to the user (even with a nice deprecation), we should try to avoid unnecessarily deprecating/renaming things.

src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -897,6 +905,27 @@ type DiagnosticItem =
                 "This expression may be short-circuited, and operation calls may not be executed."
             | WarningCode.DeprecationWithRedirect -> "{0} has been deprecated. Please use {1} instead."
             | WarningCode.DeprecationWithoutRedirect -> "{0} has been deprecated."
+            | WarningCode.UnsupportedResultComparison ->
```
bettinaheim, line 908: What are the 4 additional arguments to this message?

src/QsCompiler/DataStructures/SyntaxTree.fs
```diff
@@ -671,6 +673,37 @@ and QsStatement =
...
+    }
+
+    /// The assembly file path for this source if one exists, otherwise the code file path.
+    member source.AssemblyOrCode = source.AssemblyFile.ValueOr source.CodeFile
+
+    /// <summary>
+    /// Returns a copy of this source with the given <paramref name="codeFile"/> or <paramref name="assemblyFile"/> if
+    /// provided.
+    /// </summary>
+    member source.With([<Optional; DefaultParameterValue null>] ?codeFile,
```
bettinaheim, line 693: I like this for compatibility!

src/QsCompiler/DataStructures/SyntaxTree.fs
```diff
@@ -720,6 +753,34 @@ type SpecializationImplementation =
     /// indicates that the specialization is to be generated by the compiler according to the given functor generator directive
     | Generated of QsGeneratorDirective // Invert and Distribute will be replaced by Provided before sending to code gen
 
+/// <summary>
+/// The schema for <see cref="QsSpecialization"/> that is used with JSON serialization.
+/// </summary>
+[<CLIMutable>]
+[<DataContract>]
+type internal QsSpecializationSchema =
```
bettinaheim, line 761: I think it would make more sense to have these in the Serialization file - are they needed here?

src/QsCompiler/DataStructures/Diagnostics.fs
bettinaheim: Reading more, I believe the arguments are callable name, source file, start line, start column. The format with [source : line : column] seems a bit cryptic to me. Would it make sense to instead just make it another sentence that also includes the callable name? That would be more in line with how other diagnostics look I think.

src/QsCompiler/Core/Utils.fs
samarsha: No, `<summary>` is implicit in F#. It's only needed when other documentation tags are used.

src/QsCompiler/DataStructures/SyntaxTree.fs
samarsha: How about `AssemblyOrCodeFile`?

src/QsCompiler/DataStructures/SyntaxTree.fs
samarsha: It's not strictly necessary, but it's more convenient and idiomatic to use module functions instead of properties in F#. Functions support type inference:

```fsharp
let foo source =
    "File: " + Source.assemblyOrCode source

let foo (source: Source) = // Type annotation is needed.
    "File: " + source.AssemblyOrCode
```

Functions are also easier to pass as a value than a property getter:

```fsharp
let paths = sources |> List.map Source.assemblyOrCode

// Lambda is needed.
let paths = sources |> List.map (fun source -> source.AssemblyOrCode)
```

src/QsCompiler/CompilationManager/CompilationUnit.cs
samarsha: `IsQSharp` seems weird to me - all sources are Q# sources. Maybe naming the opposite is easier, `IsReference`?

src/QsCompiler/DataStructures/Diagnostics.fs
samarsha: The previous name is misleading, and it confused me several times while working on this PR. `UnsupportedCapability` makes it sound like it's generated for every capability error, but it's only used when an unsupported callable is referenced.

src/QsCompiler/DataStructures/Diagnostics.fs
samarsha: I don't think there are any other diagnostics that refer to a different file. However, it looks like the format for file locations displayed by the compiler is `file(line,column)`.

src/QsCompiler/CompilationManager/EditorSupport/SymbolInformation.cs
samarsha: Yes, `Source.AssemblyOrCode` was made to be backwards compatible with the original `SourceFile` property. It should always be OK to replace usages of `SourceFile` with it. I think if we didn't use it here, it would be something like `spec.Source.AssemblyFile.IsNull && spec.Source.CodeFile == file.FileName`. Code file paths in references could match a local file path by coincidence.

src/QsCompiler/CompilationManager/CompilationUnit.cs
cesarzc: `IsReference` sounds good to me.

src/QsCompiler/Core/Utils.fs
cesarzc: Thanks for the explanation!

src/QsCompiler/DataStructures/SyntaxTree.fs
cesarzc: `AssemblyOrCodeFile` sounds good :)

src/QsCompiler/DataStructures/Diagnostics.fs
samarsha: Oh, it seems like `file(line,column)` is only from `dotnet build`. The compiler itself shows

> File: filename.qs
> Position: [ln X, cn Y]

cesarzc (approved):

samarsha:
@bettinaheim Could you take another look at your remaining open comments?

src/QsCompiler/DataStructures/Diagnostics.fs
bettinaheim: Ok.

src/QsCompiler/DataStructures/Diagnostics.fs
bettinaheim: The compiler itself has two modes to display diagnostic: one that is intended for best human readability, and then one mode that formats the diagnostics in a way that is picked up by the IDE (that only happens if the emitted diagnostics follow a specific format). 
Whatever format you choose is fine by me; that's just my two cents.

bettinaheim (approved):

src/QsCompiler/DataStructures/Diagnostics.fs
samarsha: I can't really think of a good way to refer to another file and location from inside a diagnostic message. Ideally, we would just set the diagnostic location itself instead of inserting it into the message, but the compiler doesn't support that at the moment. I think copying the compiler's human-readable output format is OK for now.

PR #772, opened 2020-12-09
Only initialize Bond protocols when needed

cesarzc:
This change aims to remove impact on Katas' performance by not initializing Bond serializer/deserializer when is not needed (even if it is done asynchronously).

### DIFF
--- a/src/QsCompiler/Compiler/CompilationLoader.cs
+++ b/src/QsCompiler/Compiler/CompilationLoader.cs
@@ -477,10 +477,16 @@ namespace Microsoft.Quantum.QsCompiler
             this.RaiseCompilationTaskStart(null, "OverallCompilation");
 
             // loading the content to compiler
-            BondSchemas.Protocols.Initialize();
             this.logger = logger;
             this.LoadDiagnostics = ImmutableArray<Diagnostic>.Empty;
             this.config = options ?? default;
+
+            // When loading references is done through the generated C# and the syntax tree is not serialized,
+            // Bond protocols are not needed.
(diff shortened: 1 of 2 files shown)


src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -477,11 +477,17 @@ public CompilationLoader(SourceLoader loadSources, ReferenceLoader loadReference
...
 
             // loading the content to compiler
-            BondSchemas.Protocols.Initialize();
             this.logger = logger;
             this.LoadDiagnostics = ImmutableArray<Diagnostic>.Empty;
             this.config = options ?? default;
 
+            // When loading references is done through the generated C# and the syntax tree is not serialized,
+            // Bond protocols are not needed.
+            if (!this.config.LoadReferencesBasedOnGeneratedCsharp && !this.config.SerializeSyntaxTree)
```
rmshaffer, line 486: Based on the comment, is the logic supposed to be like this?
```suggestion
            if (!this.config.LoadReferencesBasedOnGeneratedCsharp || this.config.SerializeSyntaxTree)
```

cesarzc: Since `LoadReferencesBasedOnGeneratedCsharp` is directly related to the deserializer and `SerializeSyntaxTree` is directly related to the serializer, I added and used new Bond protocols APIs to be able to initialize each one independently. This is an improvement over the previous setup since the compiler only initializes what is strictly needed.

rmshaffer (approved):

Issue #641, opened 2020-09-23
QIR generator code clean up


alan-geller:
**Is your feature request related to a problem? Please describe.**
There's some ugly code in there...

**Describe the solution you'd like**
Some specific items:
- [obsolete/solved differently] Create a RuntimeFunctionBuilder that generates calls to runtime functions, including all prep code.
- [x] Track known information about values, such as lengths of array literals, and use it to simplify code.
- [x] In particular, simplify for loop condition evaluation when the loop direction is known.
- [x] Avoid materializing tuple literals that are immediately deconstructed, such as `let (a, b) = (1, 2);`.
- [obsolete/will be solved differently] Avoid generating callable values for partial applications that are only used locally.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

bettinaheim:
@alan-geller I believe the subsequent work I did largely took care of those.
As for the partial applications, as part of implementing support for lambda expressions in Q#, I would like to just move the responsibility to properly lift lambdas and partial applications into a separate compiler pass, and remove the partial application handling from the QIR emission entirely. I'll hence go ahead and close this isse.

PR #774, opened 2020-12-09
QIR generation bug fixes

bettinaheim:
This PR fixes
- an issue for partial applications involving nested tuples
- an issue in the monomorphization that caused QIR generation to fail when using type parameterized callables
- an issue with unwrap application (didn't push to value stack in some cases)
- an issue when there is a return statement that returns Unit
- some issues around using Unit in signatures and returns 
- the lack of implementation for supporting passing callables that return tuples as arguments

The PR also 
- adds additional expression tests
- adds doc comments and edits the method names to not call three different things all "wrappers"
- cleans up the used naming for functions (specializations)


src/QsCompiler/QirGeneration/Context.cs
```diff
@@ -1131,33 +1131,49 @@ Value BuildLoadForArg(QsArgumentTuple arg, Value value)
...
+                    for (int i = 1; i <= tupleType.Item.Length; i++)
                     {
-                        // TODO: complete implementation
+                        var resItem = this.CurrentBuilder.GetElementPtr(
+                             resultTupleType,
+                             resultValue,
+                             new[] { this.Context.CreateConstant(0L), this.Context.CreateConstant(i) });
+                        var itemOutputPointer = GetOutputItem(i);
+
+                        // if the subitem is a tuple we are good
```
alan-geller, line 1155: This comment doesn't match the code. I think the code is correct.

src/QsCompiler/QirGeneration/Context.cs
```diff
@@ -1131,33 +1131,49 @@ Value BuildLoadForArg(QsArgumentTuple arg, Value value)
...
+                        // if it is not then we need to get an output tuple to populate for the recursion to work... -> change recursion?
+                        if (tupleType.Item[i].Resolution.IsTupleType)
+                        {
+                            PopulateResultTuple(tupleType.Item[i], resItem, itemOutputPointer);
+                        }
+                        else
+                        {
+                            this.CurrentBuilder.Store(resItem, itemOutputPointer);
```
alan-geller, line 1163: You need to Load from resItem and store the result at itemOutputPointer; resItem is a pointer, not a final value.

src/QsCompiler/QirGeneration/Context.cs
```diff
@@ -1131,33 +1131,49 @@ Value BuildLoadForArg(QsArgumentTuple arg, Value value)
...
+                             resultTupleType,
+                             resultValue,
+                             new[] { this.Context.CreateConstant(0L), this.Context.CreateConstant(i) });
+                        var itemOutputPointer = GetOutputItem(i);
+
+                        // if the subitem is a tuple we are good
+                        // if it is not then we need to get an output tuple to populate for the recursion to work... -> change recursion?
+                        if (tupleType.Item[i].Resolution.IsTupleType)
+                        {
+                            PopulateResultTuple(tupleType.Item[i], resItem, itemOutputPointer);
```
alan-geller, line 1159: Don't you need to create a new tuple before this, based on the item type, and store the pointer to the new tuple in itemOutputPointer?
And resItem is a pointer to the pointer to the source tuple, so you need to Load from it to get a pointer to the source tuple.
When the item is a tuple, which is to say a pointer to a struct, itemOutputPointer will be a pointer to a pointer to the new copy, not a pointer to the new copy.

alan-geller (approved):
I think there might be some Loads missing. It's hard to tell from the diff, though, so sorry for the spam if I got it wrong!

ScottCarda-MS (approved):
Thanks for addressing the specialization naming bug.

ScottCarda-MS:
> Thanks for addressing the specialization naming bug.

What is the plan for getting this and other bug fixes from these PRs into main? Are we going to have to wait until the QIR feature is complete to get them?

src/QsCompiler/QirGeneration/Context.cs
bettinaheim: Thanks. How come LLVM/LLVM.NET doesn't complain here? There is a type distinction between the value and the pointer, and I have seen that in other places, the emission (and specifically Store) throws when there is a type mismatch. So how come it didn't throw here?

src/QsCompiler/QirGeneration/Context.cs
bettinaheim: Ah, yes, I did not read the code sufficiently to realize that there is no space allocated for the output tuple yet.

src/QsCompiler/QirGeneration/Context.cs
alan-geller: Hmm... I don't know. Maybe I'm misreading and you don't need to Load, but I'm not sure.
Does the generated code look correct?

bettinaheim: This was me not thinking right about when this code would get invoked; the code only gets executed if a callable that has a tuple-valued output is passed as argument, and it indeed complains as it should.

src/QsCompiler/QirGeneration/Context.cs
bettinaheim: Actually, since the tuples already exists I think it is sufficient to fill in the pointer to the existing tuple (the on in resultValue) into the right location of the output tuple, much like non-tuple items. I hence think the case distinction or any kind of recursion is not actually necessary, unless I am somehow missing that the tuple constructed as part of building the resultValue gets deallocated somewhere? @alan-geller Could you take another look?

alan-geller: Yes, that should work, you can just copy the tuple pointer. The only subtlety is that you need to make sure you add a new reference to the tuple; you can use the scope manager to make sure this happens.

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestExpressions.ll
```diff
@@ -0,0 +1,72 @@
...
+  %42 = call i64 @Microsoft__Quantum__Testing__QIR__ReturnBXOr__body()
+  %43 = call i64 @Microsoft__Quantum__Testing__QIR__ReturnBOr__body()
+  %44 = call %BigInt* @Microsoft__Quantum__Testing__QIR__ReturnBAnd__body()
+  %45 = call i1 @Microsoft__Quantum__Testing__QIR__ReturnNot__body()
+  %46 = call i64 @Microsoft__Quantum__Testing__QIR__ReturnBNot__body()
+  %47 = call double @Microsoft__Quantum__Testing__QIR__ReturnNegative__body(double 3.000000e+00)
+  call void @__quantum__rt__callable_unreference(%Callable* %0)
+  call void @__quantum__rt__callable_unreference(%Callable* %1)
+  call void @__quantum__rt__callable_unreference(%Callable* %4)
+  %48 = bitcast { %TupleHeader, { %TupleHeader, i2, i64 }*, double }* %5 to %TupleHeader*
```
irinayat-MS, line 54: when is the inner TupleHeader released?

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestExpressions.ll
```diff
@@ -0,0 +1,72 @@
...
+  %7 = call %Callable* @Microsoft__Quantum__Testing__QIR__ReturnControlledApplication__body()
+  %8 = call { %TupleHeader, i64, { %TupleHeader, %Callable*, %String* }* }* @Microsoft__Quantum__Testing__QIR__ReturnTuple__body()
+  %9 = call %String* @Microsoft__Quantum__Testing__QIR__ReturnArrayItem__body()
+  %10 = call i2 @Microsoft__Quantum__Testing__QIR__ReturnNamedItem__body()
+  %11 = call %Array* @Microsoft__Quantum__Testing__QIR__ReturnArray__body()
+  %12 = call %Array* @Microsoft__Quantum__Testing__QIR__ReturnNewArray__body()
+  %13 = call %String* @Microsoft__Quantum__Testing__QIR__ReturnString__body()
+  %14 = call %Range @Microsoft__Quantum__Testing__QIR__ReturnRange__body()
+  %15 = call %Array* @Microsoft__Quantum__Testing__QIR__ReturnCopyAndUpdateArray__body()
+  %16 = call { %TupleHeader, { %TupleHeader, i2, i64 }*, double }* @Microsoft__Quantum__Testing__QIR__ReturnCopyAndUpdateUdt__body()
```
irinayat-MS, line 18: why these TupleHeader's don't need to be unreferenced (unlike %8)?

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestOpArgument.ll
```diff
@@ -19,8 +19,37 @@ entry:
...
+  %12 = call %TupleHeader* @__quantum__rt__tuple_create(i64 ptrtoint ({ %TupleHeader, %Callable*, %Qubit*, %Qubit* }* getelementptr ({ %TupleHeader, %Callable*, %Qubit*, %Qubit* }, { %TupleHeader, %Callable*, %Qubit*, %Qubit* }* null, i32 1) to i64))
+  %13 = bitcast %TupleHeader* %12 to { %TupleHeader, %Callable*, %Qubit*, %Qubit* }*
+  %14 = getelementptr { %TupleHeader, %Callable*, %Qubit*, %Qubit* }, { %TupleHeader, %Callable*, %Qubit*, %Qubit* }* %13, i64 0, i32 1
+  %15 = call %Callable* @__quantum__rt__callable_create([4 x void (%TupleHeader*, %TupleHeader*, %TupleHeader*)*]* @Microsoft__Quantum__Testing__QIR___Choose, %TupleHeader* null)
+  store %Callable* %15, %Callable** %14
+  %16 = getelementptr { %TupleHeader, %Callable*, %Qubit*, %Qubit* }, { %TupleHeader, %Callable*, %Qubit*, %Qubit* }* %13, i64 0, i32 2
+  store %Qubit* %q, %Qubit** %16
+  %17 = getelementptr { %TupleHeader, %Callable*, %Qubit*, %Qubit* }, { %TupleHeader, %Callable*, %Qubit*, %Qubit* }* %13, i64 0, i32 3
+  store %Qubit* %q, %Qubit** %17
+  %18 = bitcast %TupleHeader* %8 to { %TupleHeader, %Qubit*, %Qubit* }*
```
irinayat-MS, line 31: why? %18 doesn't seem to be consumed by anybody below

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestOpArgument.ll
```diff
@@ -19,8 +19,37 @@ entry:
...
+  %25 = getelementptr { %TupleHeader, %Callable*, %Qubit*, %Qubit* }, { %TupleHeader, %Callable*, %Qubit*, %Qubit* }* %21, i64 0, i32 3
+  store %Qubit* %q, %Qubit** %25
+  %26 = bitcast %TupleHeader* %8 to { %TupleHeader, %Qubit*, %Qubit* }*
+  %27 = call %Callable* @__quantum__rt__callable_create([4 x void (%TupleHeader*, %TupleHeader*, %TupleHeader*)*]* @PartialApplication__4, %TupleHeader* %20)
+  call void @Microsoft__Quantum__Testing__QIR__Apply__body(%Callable* %27)
   call void @__quantum__rt__qubit_release(%Qubit* %q)
   call void @__quantum__rt__callable_unreference(%Callable* %5)
   call void @__quantum__rt__callable_unreference(%Callable* %11)
+  call void @__quantum__rt__callable_unreference(%Callable* %19)
+  call void @__quantum__rt__callable_unreference(%Callable* %27)
```
irinayat-MS, line 50: where are %20 and %23 are released?

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestOpArgument.ll
bettinaheim: Hu, this seems wrong in any case; %8 contains the pointer to the constant array containing the functions for SWAP. It seems like there is something wrong here. I'll look into it - thanks for pointing it out!

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestOpArgument.ll
bettinaheim: There are a bunch or unreferences missing it seems. I'll look into it.

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestOpArgument.ll
bettinaheim: This is fixed now and this bitcast no longer exists (it is now part of the lifted partial application).

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestExpressions.ll
bettinaheim: Looking more closely, I believe the idea was that a call to tuple_unreference, callable_unreference and similar would recur and also call the unreference for inner tuples and capture tuples respectively. However, I don't think the runtime actually can do that, since I believe it may not have sufficient type information to do so. Hence, I think we may need to actually handle that in the QIR emission. @alan-geller Am I missing something?

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestOpArgument.ll
bettinaheim: closing this to continue the conversation on the thread about unreferencing inner tuples above.

src/QsCompiler/QirGeneration/Context.cs
bettinaheim: It is not currently doing that for other references like callables or arrays either. I think we need to take a closer look at the reference counting in any case.

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestExpressions.ll
bettinaheim: I am working on some changes that will add an unreference to %16. However, I would prefer to make this a separate PR.

src/QsCompiler/QirGeneration/Context.cs
bettinaheim: After taking a closer look: we do not unreference the return value, and I hence believe there is no need to add an additional reference here.

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestOpArgument.ll
bettinaheim: To give a bit more context: The idea was that when a callable is constructed such that it can be used as part of a partial application, rather than adding a reference to the constructed value when it is used in that way, the corresponding unreference was removed instead. This, however, I believe does not always lead to the right behavior, since it is possible that an additional unreference is introduced later on that would then incorrectly cause the value to be released (this would be the case e.g. when the partial application is stored in a named item that is then updated). I will hence change this with the PR that changes the reference handling in general.

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestExpressions.ll
irinayat-MS: For the tuples, the runtime doesn't know the types and cannot unreference inner elements. For the callables, the runtime could unreference the capture tuple, but it should be called out in the QIR spec: 1. whether the runtime should increment the reference of the given tuple, when creating a callable, and 2. that it should decrement the reference when releasing the callable.

alan-geller: Note that some of this is covered in issue #641 .

bettinaheim: Array items also need to be unreferenced and can't be by the runtime since it won't know the their type.

bettinaheim:
I'll merge this PR. The PR for reference handling can be found here: https://github.com/microsoft/qsharp-compiler/pull/784 (it still needs more work).

PR #777, opened 2020-12-09
Add remarks to constructors and methods that must not be invoked through a Task

cesarzc:
This change adds remarks to constructors and methods that must not be invoked through a Task because it could cause a deadlock if the thread pool is already full when they are invoked.


samarsha:
Could you explain a bit more what the deadlock condition is for these methods?

cesarzc:
> 
> 
> Could you explain a bit more what the deadlock condition is for these methods?

These methods queue (or might queue) Tasks to initialize Bond serializers/deserializers, and at some point, the thread that queued the Task waits for it to complete.

By default, a Task is executed in a threadpool thread. If a Task is queued from a threadpool thread and the threadpool does not have any available threads for the Task to be executed on, the thread from which the Task was queued would be blocked waiting for the Task to complete but the Task would not complete until there is an available thread in the threadpool. An extreme example would be calling one of these methods from a threadpool thread when the size of the threadpool is fixed at one.

The scenario in which this happens is not a common one but it is a possible one under high workload conditions.

samarsha:
Thanks for the explanation! Is this something that's specific to the way these methods use tasks, or does it apply to any method that uses tasks internally?

cesarzc:
> 
> 
> Thanks for the explanation! Is this something that's specific to the way these methods use tasks, or does it apply to any method that uses tasks internally?

It applies to any method that use Tasks internally with the default `TaskScheduler` and that do an explicit or implicit `Wait` on the Task.

src/QsCompiler/BondSchemas/Protocols.cs
```diff
@@ -30,6 +30,7 @@ public static class Protocols
         /// Deserializes a Q# compilation object from its Bond simple binary representation.
         /// </summary>
         /// <param name="byteArray">Bond simple binary representation of a Q# compilation object.</param>
+        /// <remarks>This must not be invoked through a <see cref="Task"/>.</remarks>
```
samarsha, line 33: Maybe a comment like this would be clearer:

```suggestion
        /// <remarks>This method waits for <see cref="Task"/>s to complete and may deadlock if invoked through a <see cref="Task"/>.</remarks>
```

cesarzc: Yes, that is much better! Thanks for the suggestion :) I'll change it in the other comments too.

samarsha (approved):

PR #783, opened 2020-12-14
Adds ConcreteIntrinsicGen Property

ScottCarda-MS:
Adds ConcreteIntrinsicGen Property as an assembly constant and Q# project property. This property will be used by the C# generation rewrite step to determine if an intrinsic operation is generated in C# as an abstract class (original and default behavior) or a concrete class (new behavior from using property). The logic that will utilize this property is found in the runtime repo under the branch 'swernli/decomp-interfaces', specifically here: https://github.com/microsoft/qsharp-runtime/blob/swernli/decomp-interfaces/src/Simulation/CsharpGeneration/SimulationCode.fs

### DIFF
--- a/src/QsCompiler/DataStructures/ReservedKeywords.fs
+++ b/src/QsCompiler/DataStructures/ReservedKeywords.fs
@@ -246,6 +246,7 @@ module AssemblyConstants =
     let ToffoliSimulator = "ToffoliSimulator"
     let ResourcesEstimator = "ResourcesEstimator"
     let ExposeReferencesViaTestNames = "ExposeReferencesViaTestNames"
+    let ConcreteIntrinsicGen = "ConcreteIntrinsicGen"
 
     /// The runtime capabilities supported by an execution target. The names of the capabilities here match the ones
     /// defined by the SDK.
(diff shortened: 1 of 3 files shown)


src/QsCompiler/DataStructures/ReservedKeywords.fs
```diff
@@ -246,6 +246,7 @@ module AssemblyConstants =
     let ToffoliSimulator = "ToffoliSimulator"
     let ResourcesEstimator = "ResourcesEstimator"
     let ExposeReferencesViaTestNames = "ExposeReferencesViaTestNames"
+    let ConcreteIntrinsicGen = "ConcreteIntrinsicGen"
```
swernli, line 249: I have a slight preference for using the active verb at the front, similar to `ExposeReferencesViaTestName`, so I think I'd like `GenerateConcreteIntrinsic` instead of `ConcreteIntrinsicGen`. I feel like that more clearly underlines the behavior that will change when the property is set to true.

ScottCarda-MS: Changed to use GenerateConcreteIntrinsic

swernli (approved):

PR #786, opened 2020-12-16
More detailed breakdown of reference loading performance

cesarzc:
This change adds details to the reference loading performance accounting.

### DIFF
--- a/src/QsCompiler/CompilationManager/AssemblyLoader.cs
+++ b/src/QsCompiler/CompilationManager/AssemblyLoader.cs
@@ -48,11 +48,17 @@ namespace Microsoft.Quantum.QsCompiler
             using var assemblyFile = new PEReader(stream);
             if (ignoreDllResources || !FromResource(assemblyFile, out var compilation, onDeserializationException))
             {
+                PerformanceTracking.TaskStart(PerformanceTracking.Task.HeaderAttributesLoading);
                 var attributes = LoadHeaderAttributes(assemblyFile);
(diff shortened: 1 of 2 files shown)


src/QsCompiler/CompilationManager/AssemblyLoader.cs
```diff
@@ -48,11 +48,17 @@ public static bool LoadReferencedAssembly(Uri asm, out References.Headers header
...
             {
+                PerformanceTracking.TaskStart(PerformanceTracking.Task.HeaderAttributesLoading);
                 var attributes = LoadHeaderAttributes(assemblyFile);
+                PerformanceTracking.TaskEnd(PerformanceTracking.Task.HeaderAttributesLoading);
+                PerformanceTracking.TaskStart(PerformanceTracking.Task.ReferenceHeadersCreation);
                 headers = new References.Headers(id, attributes);
+                PerformanceTracking.TaskEnd(PerformanceTracking.Task.ReferenceHeadersCreation);
                 return ignoreDllResources || !attributes.Any(); // just means we have no references
             }
+            PerformanceTracking.TaskStart(PerformanceTracking.Task.ReferenceHeadersCreation);
```
ScottCarda-MS, line 59: This task has the same name as an earlier one, in line 54. If they both get executed, does this task start-end block override the other one?

cesarzc: No, they add up. The total accounted time for the task would be the sum of the time spent in both sections.

ScottCarda-MS (approved):

PR #787, opened 2020-12-17
Eliminating tuple headers in QIR

bettinaheim:
With this PR, QIR would no longer use tuple headers at all. Instead, concrete tuples represented as normal structs are used locally, and an opaque pointer %Tuple is passed around. The bitcasts remain the same. 

The current target branch is just to show a proper diff. I will retarget it once the reference handling is done and merged.


irinayat-MS:
Could the QIR spec possibly be updated before merging this PR?

bettinaheim:
@irinayat-MS Since this is so far just a merge to my own branch, I'll merge this and then put up a spec update PR before this goes to main.

PR #791, opened 2020-12-22
Make optional and deprecate parentheses in block statements

samarsha:
This is part of the implementation for the [implicitly-scoped qubit allocation proposal](https://github.com/microsoft/qsharp-language/blob/main/Approved/implicitly-scoped-qubit-allocation.md), which specifies that parentheses for statement headers should be removed.

In the case of `if`, `elif`, `while`, and `until`, the parentheses surrounded an expression, so no deprecation will occur - the parentheses will simply be made optional, and will be parsed as a tuple expression.

In the case of `for`, `using`, and `borrowing`, the parentheses are deprecated and will be not be supported in the future.

Depends on microsoft/iqsharp#385.


src/QsCompiler/Tests.Compiler/TestCases/FunctorGeneration.qs
```diff
@@ -680,21 +680,21 @@ namespace Microsoft.Quantum.Testing.FunctorGeneration {
...
-            for (i in 1..10) { set _ = 1; }
+            for i in 1..10 { set _ = 1; }
         }
         adjoint auto;    
     }
 
     operation InvalidAutoInversion10(q : Qubit) : Unit {
         body (...) {
-            using (c = Qubit()) { return (); }
+            using c = Qubit() { return (); }
```
bettinaheim, line 690: It would be good to coordinate when this PR is shipped with the PR that deprecates using and introduces non-block qubit allocation statements. I don't want to force people to update usings twice, and that also makes it clearer why we are doing this change. Could you please retarget this PR to a feature branch?

bettinaheim (approved):
Approving for the merge to a feature branch.

PR #796, opened 2020-12-25
Allow Language Server to work with .NET 5.0. Detect absence of .NET SDK 3.1 or greater and display error message

vxfield:
This PR mitigates issue #737 by using the [latest MS Build Locator package](https://www.nuget.org/packages/Microsoft.Build.Locator/1.3.2) ([fix](https://github.com/microsoft/MSBuildLocator/commit/ecf817d34558011e08f2c3755413cb45180b04be)).

It picks some changes from the November QDK release fix #755 with:
- Forcing the Working Directory of the LanguageServer to be the same as the LanguageServer is (in case we might need to add a global.json to select the .NET SDK version)
- Not quitting the LanguageServer if there was an issue with the MSBuildLocator when .NET Core SDK 3.1 or greater is not found
- Check if .NET Core SDK 3.1 or greater is not installed and display a message to the user with an option to download it (opens the download page) and asking the user to restart VS/Code.


src/QsCompiler/LanguageServer/LanguageServer.cs
```diff
@@ -123,9 +124,46 @@ public void Dispose()
...
+            {
+                if (!installedVersionsOfDotNetSdk.DotNetSdk31Installed
+                    && !installedVersionsOfDotNetSdk.DotNetSdk5Installed)
+                {
+                    const string dotnetSdkUrl = "https://dotnet.microsoft.com/download/dotnet-core/";
+                    this.LogToWindow($".NET Core SDK 3.1 or greater not found. Quantum Development Kit Extension requires .NET Core SDK version 3.1 or greater ({dotnetSdkUrl}).", MessageType.Error);
+                    var downloadAction = new MessageActionItem { Title = "Download" };
+                    var cancelAction = new MessageActionItem { Title = "No, thanks" };
+                    var selectedAction = await this.ShowDialogInWindowAsync(
+                        "Quantum Development Kit Extension requires either .NET Core SDK version 3.1 or greater. Please install the latest .NET Core SDK and restart Visual Studio.",
```
samarsha, line 150: Based on @rainersigwald's comment in #737, I believe we need to target the language server for `net5` in order to support both 3.1 and 5 SDKs. Since the language server targets `netcoreapp3.1`, we would still require specifically that .NET Core SDK 3.1 is installed, even with the Microsoft.Build.Locator update. This dialog message shown to users implies that having only .NET 5 SDK installed would work, though.

@vxfield Does this match your testing when using the language server on a computer that only has .NET 5 SDK installed - does it fail to load the project with this PR applied?

samarsha (changes requested):

build/init.yml
```diff
@@ -16,10 +16,16 @@ steps:
     versionSpec: '5.6.0'
 
 - task: UseDotNet@2
-  displayName: 'Use .NET Core SDK 3.1.300'
+  displayName: 'Use .NET 5 SDK'
   inputs:
     packageType: sdk
-    version: '3.1.300'
+    version: '5.0.101'
```
samarsha, line 22:
```suggestion
    version: '5.0.102'
```

global.json
```diff
@@ -1,6 +1,6 @@
 {
-    "sdk": {
-        "version": "3.1.300",
-        "rollForward": "latestFeature"
-    }
-}
\ No newline at end of file
+  "sdk": {
+    "version": "5.0.101",
```
samarsha, line 3: For a lower bound, I think we can use the first .NET 5 release.

```suggestion
    "version": "5.0.100",
```

src/QsCompiler/LanguageServer/LanguageServer.csproj
```diff
@@ -19,16 +19,16 @@
...
     <PackageReference Include="CommandLineParser" Version="2.6.0" />
-    <PackageReference Include="Microsoft.Build.Locator" Version="1.2.6" />
-    <PackageReference Include="Microsoft.Build" Version="16.3.0" ExcludeAssets="runtime" />
-    <PackageReference Include="Microsoft.Build.Framework" Version="16.3.0" ExcludeAssets="runtime" />
-    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="16.3.0" ExcludeAssets="runtime" />
-    <PackageReference Include="Microsoft.VisualStudio.LanguageServer.Protocol" Version="16.3.57" />
+    <PackageReference Include="Microsoft.Build.Locator" Version="1.3.2" />
+    <PackageReference Include="Microsoft.Build" Version="16.8.0" ExcludeAssets="runtime" />
+    <PackageReference Include="Microsoft.Build.Framework" Version="16.8.0" ExcludeAssets="runtime" />
+    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="16.8.0" ExcludeAssets="runtime" />    <PackageReference Include="Microsoft.VisualStudio.LanguageServer.Protocol" Version="16.3.57" />
```
samarsha, line 25:
```suggestion
    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="16.8.0" ExcludeAssets="runtime" />
    <PackageReference Include="Microsoft.VisualStudio.LanguageServer.Protocol" Version="16.3.57" />
```

src/QsCompiler/LanguageServer/LanguageServer.csproj
```diff
@@ -19,16 +19,16 @@
...
     <PackageReference Include="Newtonsoft.Json" Version="12.0.2" />
     <PackageReference Include="StreamJsonRpc" Version="2.2.34" />
     <PackageReference Include="System.Reactive" Version="4.2.0" />
+    <PackageReference Include="NuGet.Frameworks" Version="5.8.0" />
```
samarsha, line 29: It worries me a little that it's necessary to include this dependency, since we had issues with depending on a specific version `NuGet.ProjectModel` before. @bettinaheim @ricardo-espinoza

src/QsCompiler/LanguageServer/Program.cs
```diff
@@ -83,6 +83,11 @@ private static int LogAndExit(ReturnCode code, string? logFile = null, string? m
 
         public static int Main(string[] args)
         {
+            // We need to set the current directory to the same directory of
+            // the LanguageServer executable so that it will pick the global.json file
+            // and force the MSBuildLocator to use a particular version of .NET Core SDK
+            Directory.SetCurrentDirectory(AppDomain.CurrentDomain.BaseDirectory);
+
```
samarsha, line 90: Since the language server doesn't have a `global.json` anymore, I think this can be removed?

```suggestion
```

samarsha:
Updating the language server to .NET 5 is blocked until the QDK build pipeline can be updated. I don't know when that will be, so I'll close this for now.

PR #798, opened 2020-12-29
Further unification of tuple and array handling

bettinaheim:
This PR
- introduces a TupleValue class to more conveniently pass around both the typed and opaque version of a tuple (allows to avoid some unnecessary casts)
- introduces an ArrayValue in analogy to abstract away the array creation similar to the tuple handling
- uses the standard CreateTuple function to instantiate the return value for type constructors (results in changes to reference handling)
- uses CreateTuple for constructing the inner argument tuple in partial applications (adds some missing calls related to reference handling)
- uses CreateTuple to populate the capture tuple in partial applications (adds some missing calls related to reference handling)
- fixes an issue where the bitcasts for partial applications that don't result in the creation of an inner tuple are incorrect
- largely eliminates explicit usage of tuple related bitcasts in favor of using the defined tuple helpers

The PR introduces the following API to eliminate all tuple and array related bitcasts from the rest of the code:
- GetArrayElementPointer
- GetArrayElement
- GetTupleElementPointer
- GetTupleElement
- GetTupleElementPointers
- GetTupleElements


bettinaheim:
@alan-geller, @irinayat-MS I am periodically making PRs as I am working to get the reference handling in a decent state to make it a little bit easier to follow what changes I am making in the process. I don't expect reviews, but certainly appreciate any inputs & feedback if you want to take a look. Unfortunately, I underestimated the amount of code it makes sense to restructure in order to facilitate the reference management. My apologies for the amount coming together here.

Issue #795, opened 2020-12-24
QEP 1: Implicitly-Scoped Qubit Allocation


samarsha:
This issue tracks the implementation of [QEP 1: Implicitly-Scoped Qubit Allocation](https://github.com/microsoft/qsharp-language/blob/main/Approved/1-implicitly-scoped-qubit-allocation.md). The feature branch is [feature/inline-qubit-scope](https://github.com/microsoft/qsharp-compiler/tree/feature/inline-qubit-scope).

- [x] Deprecate `using` and `borrowing` keywords and replace with `use` and `borrow` (#804)
- [x] Allow `use` and `borrow` statements to be followed by either a block `{ ... }` or a semicolon `;` (#803)
- [x] Deprecate parentheses around block statement headers (#791)

cgranade:
Thinking about `using` → `use` and `borrowing` → `borrow` a bit more, it feels a bit awkward to use [present participles](https://www.merriam-webster.com/dictionary/present%20participle) to represent an action taken by an operation, given that we elsewhere use finite verb phrases to directly indicate the tense and subject of that verb. Concretely, the gerund "using" on its own doesn't have a tense to make it clear _when_ something is used. Moving to "use" makes that tense explicit, which strikes me as a very good thing for readability.

tcNickolas:
When this change becomes breaking, will we offer the users a tool to mass-migrate to the new syntax automatically, like we did in the past (I think it was around 0.3 version)? I've seen a mention of a code action in the proposal, but migrating a sizeable code base would require a lot of effort without such tool.

samarsha:
We've been discussing using [QsFmt](https://github.com/samarsha/qsfmt) to automatically update deprecated syntax across an entire project. It needs some more work to be able to do that, though.

PR #803, opened 2021-01-05
Allow using and borrowing statements to use the current scope

samarsha:
This PR allows `using` and `borrowing` statements to be followed by either a block `{ ... }` or a semicolon `;`. If they're followed by a semicolon, the qubits are bound for the remainder of the current scope. See #795.


src/QsCompiler/CompilationManager/TypeChecking.cs
```diff
@@ -531,18 +531,37 @@ internal static void ImportGlobalSymbols(this FileContentManager file, Compilati
         /// If the set of required functors is unspecified or null, then the functors to support are determined by the parent scope.
         /// </summary>
         private static QsScope BuildScope(
-            IReadOnlyList<FragmentTree.TreeNode> nodeContent,
+            IEnumerator<FragmentTree.TreeNode> nodes,
```
cesarzc, line 534: What was the motivation to change this from `IReadOnlyList` to `IEnumerator`?

samarsha: When building an implicit scope, the statement nodes for the scope body re-use the same enumerator from the parent scope. Using an `IReadOnlyList` would require consuming the rest of the enumerator all at once and building a new list, which is unnecessary because `BuildScope` simply enumerates that list from beginning to end.

cesarzc (approved):
Looks good to me.

PR #806, opened 2021-01-06
Update npm packages for VS Code extension with npm audit

samarsha:
Component governance is blocking builds because of a vulnerability in axios 0.18.1. This PR used `npm audit` to fix the package versions.

The package update changed the VS Code test API; since we don't actually have any tests (they're just placeholders), I removed them.


src/VSCodeExtension/package.json.v.template
```diff
@@ -114,26 +113,25 @@
         "vscode-extension-telemetry": "0.0.18",
         "vscode-languageclient": "5.1.0",
         "which": "1.3.1",
-        "yeoman-environment": "^2.10.3",
+        "yeoman-environment": "^2.9.6",
```
ricardo-espinoza, line 116: Hi @samarsha , last time when we had dependency issues with Yeoman environment, we pinned the version to 2.10.3.
Why do we want to roll back to 2.9.6 in particular?

samarsha: We get `axios` by the dependency chain: `yeoman-environment -> npm-api -> paged-request -> axios ^0.18.0`. It seems like 2.9.6 is the last version of `yeoman-environment` that does not depend on `npm-api`, so `npm audit` picked that version to avoid the dependency on `axios ^0.18.0`.

ricardo-espinoza (approved):

PR #807, opened 2021-01-07
Fix remaining StyleCop warnings and update StyleCop version

samarsha:
* Fix all of the remaining StyleCop warnings
* Add StyleCop to AssemblyCommon.props so it automatically applies to all C# projects
* Update StyleCop version to latest
* Configure StyleCop to use the standard copyright header
* Restore some custom rule settings to their defaults (the resulting changes were fairly minor)


src/Common/AssemblyCommon.props
```diff
@@ -18,6 +18,11 @@
     <WarningsAsErrors>3390</WarningsAsErrors>
   </PropertyGroup>
 
+  <ItemGroup Condition="'$(MSBuildProjectExtension)' == '.csproj'">
```
bettinaheim, line 21: Ah yes, I had done that in the qir feature branch only - that's definitively helpful!

bettinaheim (approved):

Issue #773, opened 2020-12-09
Monomorphization is not setting the correct parent for specializations


bettinaheim:
The name of the original callable is listed as parent for monomorphized specializations.

bettinaheim:
Fixed in https://github.com/microsoft/qsharp-compiler/pull/774.

ScottCarda-MS:
@bettinaheim Since #744 does not target main, should we wait until this fix is in main to close this issue? Should the fix in #744 be separated out into a different PR targeting main?

ScottCarda-MS:
Merged into Main with #809.

PR #809, opened 2021-01-07
Monomorphization Bug Fixes

ScottCarda-MS:
Addresses #773 and #757.
Adds support for optional parameter to Monomorphization rewrite step that allows intrinsic operations to be kept or removed if unused as specified in #764.


src/QsCompiler/Compiler/RewriteSteps/Monomorphization.cs
```diff
@@ -28,16 +30,21 @@ internal class Monomorphization : IRewriteStep
 
         public bool ImplementsPostconditionVerification => true;
 
-        public Monomorphization()
+        /// <summary>
+        /// Constructor for the Monomorphization Rewrite Step. If keepAllIntrinsics is
+        /// set to true, intrinsics will not be removed as part of the rewrite step.
```
cesarzc, line 35: It would be good to move the description of the `keepAllIntrinsics` parameter to a `<param>` tag.

ScottCarda-MS: done

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -138,7 +144,8 @@ private bool NamespaceElementFilter(QsNamespaceElement elem)
                 {
                     if (elem is QsNamespaceElement.QsCallable call)
                     {
-                        return BuiltIn.RewriteStepDependencies.Contains(call.Item.FullName) || this.SharedState.IntrinsicCallableSet.Contains(call.Item.FullName);
+                        return BuiltIn.RewriteStepDependencies.Contains(call.Item.FullName) ||
```
cesarzc, line 147: Does this method return `true` when an element should be filtered out?

ScottCarda-MS: This method returns `true` if an element needs to be retained, even if Monomorphization would otherwise remove it.

cesarzc: Looking at the description of #757, it seems like some specializations are not included in the call graph in some instances when they should. Given this description, I haven't yet understood why the `KeepAllIntrinsics` is included as an AND condition rather than an OR, since the AND condition restricts even more the elements that are retained.

ScottCarda-MS: The `KeepAllIntrinsics` option allows the `IntrinsicsCallableSet` to be retained here. If `KeepAllIntrinsics` is false, we no longer want to special-case the intrinsics to keep them around. This is the change migrated from #764. The issue from #757 is that when a specialization for a callable is entered into the graph, we want to make sure that all the specialization from that same callable make it into the graph as well. The changes in ConcreteCallGraphWalker.cs reflect this.

cesarzc: Thanks for the explanation Scott!

cesarzc (approved):

PR #814, opened 2021-01-12
Support for multi-thread performance tracking

cesarzc:
This change implements support for multi-thread performance tracking.

This is needed to keep performance tracking when loading references in parallel is implemented.


ScottCarda-MS (approved):
lgtm

PR #815, opened 2021-01-13
QEP 1: Implicitly-Scoped Qubit Allocation

### LINKED ISSUE #795: QEP 1: Implicitly-Scoped Qubit Allocation

samarsha:
This PR merges the `feature/inline-qubit-scope` branch to `main`. Closes #795.

Blocked by:
- [x] microsoft/iqsharp#385
- [x] microsoft/QuantumKatas#587


cgranade:
While waiting for https://github.com/microsoft/QuantumKatas/pull/587, can we spawn an e2e build with katas disabled, so that I can get a beta to use in adapting the samples repo? Thanks!

samarsha:
I started an e2e build that uses the branch from microsoft/QuantumKatas#587, so katas validation should succeed.

samarsha:
@cgranade QDK version 0.14.210121247-alpha should be ready to use now.

cgranade:
Awesome, thank you, @samarsha!

cgranade:
Started working on adapting samples at https://github.com/microsoft/Quantum/pull/454, will start adapting libraries shortly.

src/QsCompiler/Tests.Compiler/CompletionParsingTests.fs
```diff
@@ -500,7 +507,34 @@ let ``Operation statement parser tests`` () =
         [
             ("repeat ", [])
             ("within ", [])
-            ("using ", [])
+            ("use ", [ Declaration ])
```
bettinaheim, line 510: Are using and borrowing removed from the list of suggested keywords we give for code completion?

src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -896,6 +900,9 @@ type DiagnosticItem =
             | WarningCode.UseOfFutureReservedKeyword -> "The symbol will be reserved for internal use in the future."
             | WarningCode.UseOfUnderscorePattern ->
                 "Double underscores as well as underscores before a dot or at the end of a namespace name will be reserved for internal use in the future."
+            | WarningCode.DeprecatedTupleBrackets ->
+                "Deprecated syntax. Parentheses here are no longer required and will not be supported in the future."
```
bettinaheim, line 904: Hm, I think this would be a good place to also let people know that a non-block version of the statement exists. Maybe add something like "Consider using a simply use-statement" (C# has a similar message - I'd look at their phrasing there and use that).

bettinaheim (approved):
Looks good!

src/QsCompiler/Tests.Compiler/CompletionParsingTests.fs
samarsha: No, it suggests both `use` and `using`, and `borrow` and `borrowing`.

src/QsCompiler/DataStructures/Diagnostics.fs
samarsha: There's no warning specifically for the block form; the extra parentheses and deprecated keyword warnings are more general than that. At the moment, the only deprecated keywords are `using` and `borrowing`, so I could repurpose that warning and make it specifically about the `using` or `borrowing` blocks.

But, it also feels like this is really a different kind of warning, unrelated to the keyword change. Like a "unnecessary qubit allocation block" warning that could detect `use`/`borrow` blocks that are the last statement of a scope...

samarsha: I updated the warning message to mention the non-block version.

src/QsCompiler/Tests.Compiler/CompletionParsingTests.fs
samarsha: I removed `using` and `borrowing` from code completion.

bettinaheim: Thanks!

PR #819, opened 2021-01-15
Update language server target to .NET 5

samarsha:


filipw:
very nice

filipw:
did you try upgrading MSBuild packages to 16.8?

samarsha:
Same error after updating the Microsoft.Build.* packages to 16.8.0. :(

filipw:
does it fail locally too or only in CI?

I can see that there is a workaround ("hack") in place to load Nuget assemblies from the MSBuild locator https://github.com/microsoft/qsharp-compiler/blob/main/src/QsCompiler/LanguageServer/Program.cs#L106-L119

I guess the reason is to avoid taking an explicit dependency on Nuget packages? For example in OmniSharp we still directly reference NuGet packages and actually old ones (https://github.com/OmniSharp/omnisharp-roslyn/blob/master/build/Packages.props#L7) as they provide the lowest supported boundary. 

On a separate note, I think in the future the workaround won't be needed once this ships https://github.com/microsoft/MSBuildLocator/pull/115

samarsha:
> does it fail locally too or only in CI?

It was failing locally with the same error, too.

> I guess the reason is to avoid taking an explicit dependency on Nuget packages? For example in OmniSharp we still directly reference NuGet packages and actually old ones (https://github.com/OmniSharp/omnisharp-roslyn/blob/master/build/Packages.props#L7) as they provide the lowest supported boundary.

Yeah, I think it was to avoid the explicit dependency, since the language server was breaking every time there was a .NET Core update until we updated our dependencies... It looks like adding an explicit dependency on NuGet.Frameworks 5.8.0 fixes the tests, though. Without the dependency, an older version (5.0.0 or 5.1.0 I think) was getting pulled in from somewhere (not sure why).

> On a separate note, I think in the future the workaround won't be needed once this ships microsoft/MSBuildLocator#115

:+1:

samarsha:
This is blocked until the QDK build pipeline has .NET 5. I don't know when that will be, so I'll close this for now.

PR #821, opened 2021-01-19
Language server asks user to install .NET Core 3.1 SDK if missing

samarsha:
This is #755 to main, with a couple changes:

* Rename language server's `global.json` to `global.language-server.json` to avoid confusing `dotnet` when run from the language server folder; rename it back to `global.json` in the output folder.
* Replace `LogAndExit` with `Log` instead of removing the log call entirely to avoid swallowing exceptions at startup.

Co-authored-by: Victor XField <vxfield@outlook.com>


src/QsCompiler/LanguageServer/Utils.cs
```diff
@@ -124,4 +142,30 @@ public override void Initialize(IEventSource eventSource)
...
+    {
+        private static readonly Regex DotNet31Regex = new Regex(@"^3\.1\.\d+", RegexOptions.Multiline | RegexOptions.Compiled);
+
+        public static bool? IsDotNet31Installed()
+        {
+            var listSdkProcess = Process.Start(
+                new ProcessStartInfo()
+                {
+                    FileName = "dotnet",
+                    Arguments = "--list-sdks",
```
bettinaheim, line 156: Given that the command lists one the sdks, and I don't see that list ever getting to long this is fine here. Just as an FYI, ReadToEnd may deadlock if the output is too long.

bettinaheim (approved):

src/QsCompiler/LanguageServer/Utils.cs
samarsha: Doesn't that only apply if the stream never ends, not if the stream is simply too long? (It's possible a broken or malicious version of `dotnet` could output forever, though, since it's an external command.)

bettinaheim: No, it can be an issue even for finite streams; but for this case it should be irrelevant.

PR #822, opened 2021-01-20
Treat C# and F# warnings as errors

samarsha:
Updated the CI to build with `/property:TreatWarningsAsErrors=true`. This means that warnings are allowed when developing locally, but all warnings must be fixed or suppressed before a PR can be merged.

This applies to C# and F# warnings, but not MSBuild and Q# warnings.


swernli:
How will this affect things like classes being marked as deprecated? I know we generally try to not to use deprecated structures, but sometimes during multi-part work it's convenient to add the new infrastructure, mark the old one as deprecated, and then go around update all the consumers.

samarsha:
Yeah, you can use `#pragma warning disable` (C#) or `#nowarn` (F#) to suppress warnings that are false positives or warnings that will be fixed later. So this PR just requires that every warning is looked at and is either fixed or marked as OK.

src/Documentation/Tests.DocGenerator/DocParsingTests.cs
```diff
@@ -13,6 +13,8 @@
 using static Microsoft.Quantum.QsCompiler.Documentation.Testing.Utils;
 using Range = Microsoft.Quantum.QsCompiler.DataTypes.Range;
 
+#pragma warning disable 618
```
cesarzc, line 16: I think it would be valuable to add a brief comment on what the disabled warning is about. E.g.:
```
// A class member was marked with the Obsolete attribute is referenced.
```

samarsha: Added!

cesarzc (approved):
Thanks for doing this Sarah!

PR #831, opened 2021-01-22
Caching Llvm.NET binaries directly into compiler project

swernli:


src/QsCompiler/QirGeneration/Context.cs
```diff
@@ -1391,20 +1391,20 @@ internal void CreateForLoop(Value startValue, Func<Value, Value> evaluateConditi
             // Contains the loop header that creates the phi-node, evaluates the condition,
             // and then branches into the body or exits the loop depending on whether the condition evaluates to true.
             var headerName = this.GenerateUniqueName("header");
-            var headerBlock = this.CurrentFunction.AppendBasicBlock(headerName);
+            BasicBlock headerBlock = this.CurrentFunction.AppendBasicBlock(headerName);
```
cesarzc, line 1394: I am curious why this was changed to explicitly declare `headerBlock` as `BasicBlock` instead of keep using `var`.

swernli: Great question. In dotnet 3.1 the `var` usage should be fine, but for some reason on my machine it was being treated as a nullable (the inferred type became `BasicBlock?`) which caused a number of warnings-treated-as-errors later in the code. @guenp saw something similar when working with this branch, but not Bettina. Because I couldn't figure out what in my environment was triggering this stricter behavior, and since this level of validation will become standard in dotnet 5.0, it seemed reasonable to go ahead and update it now such that the nullable issues went away.

cesarzc (approved):

bettinaheim:
@swernli Before I forget: Could you please update the Notice.txt at the repo root to reflect the llvm.net dependency?

bettinaheim:
> @swernli Before I forget: Could you please update the Notice.txt at the repo root to reflect the llvm.net dependency? Usually we should generate it as part of the release process, but if we include dlls directly in the compiler, I believe it makes sense to manually edit it. Hopefully the work to get this all contributed back and integrated into the llvm.net package should not take too long, but better to update the notice now in case it takes longer.

@swernli
I took the liberty to copy over the licence into the notice and merge this PR.

PR #835, opened 2021-01-25
Resolve the default access modifier to Public or Internal in the syntax tree

### LINKED ISSUE #667: Access modifiers are not resolved in the syntax tree
**Describe the bug**

The `AccessModifier` type is defined as:

```fsharp
type AccessModifier =
    | DefaultAccess
    | Internal
```

This is ambiguous, because `DefaultAccess` can represent either public or internal, depending on if it applies to a specialization whose parent callable is internal. This ambiguity exists in the syntax, because specializations don't require access modifier keywords and automatically inherit the parent callable's accessibility, but there's no reason to carry it through to the resolved syntax tree.

**Expected behavior**

A type like

```fsharp
type AccessModifier =
    | Internal
    | Public
```

would be more appropriate for the resolved syntax tree. The missing keyword in the syntax can be represented by `QsNullable<AccessModifier>` in the unresolved syntax tree.

**Additional context**

The current type definition means there is no well-defined ordering of access modifiers, so for example it is not easy to take the minimum of a sequence of modifiers. This could be useful, though, when determining the accessibility of a callable based on the accessibility of the types it uses.

samarsha:
Updates access modifiers to be `Internal | Public` instead of `DefaultAccess | Internal`. In the unresolved syntax tree, a default access modifier is `QsNullable.Null` instead of `DefaultAccess`. In the resolved syntax tree, it's set to `Public`, which is the default. This way, AST transformations don't need to know what the default access is, and access modifiers can be compared using standard comparison operators like `<` or `Min`.

Requires microsoft/qsharp-runtime#481. Closes #667.


cesarzc:
I am a bit hesitant on the rename from `AccessModifier` to `Visibility`. C# refers to them as [access modifiers](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/access-modifiers), F# as [access control](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/access-control) and other languages like C++ also use the term access modifiers so I think it makes sense to use the same term that other languages use.

Even though this is the compiler implementation and not documentation, I think it would be easier for both contributors and users of the compiler public APIs if we used a more familiar term.

I am not completely opposed to the rename but I would like to know what other people think. @ScottCarda-MS, @swernli, @bettinaheim do you have an opinion about renaming `AccessModifier` to `Visibility`?

ScottCarda-MS:
> I am a bit hesitant on the rename from `AccessModifier` to `Visibility`. C# refers to them as [access modifiers](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/access-modifiers), F# as [access control](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/access-control) and other languages like C++ also use the term access modifiers so I think it makes sense to use the same term that other languages use.
> 
> Even though this is the compiler implementation and not documentation, I think it would be easier for both contributors and users of the compiler public APIs if we used a more familiar term.
> 
> I am not completely opposed to the rename but I would like to know what other people think. @ScottCarda-MS, @swernli, @bettinaheim do you have an opinion about renaming `AccessModifier` to `Visibility`?

I agree with the sentiment @cesarzc made. It is better to use terms that are already out there and familiar to programmers. Additionally, changing the name is a breaking change that I don't think is particularly necessary here, although the behavior of the modifiers is changing in a breaking change anyway. Not changing the name would also make for a much smaller PR with much less code changes, which is always appreciated :) .

samarsha:
> changing the name is a breaking change that I don't think is particularly necessary here

It's not strictly necessary, but (1) the removal of `DefaultAccess` and addition of `Public`, and (2) the removal of the `Modifiers` type are already breaking changes to the access modifier API, so the renaming is pretty minor comparatively.

I could call it `Access` instead. I just think `AccessModifier` is a particularly long and cumbersome name, even though I named it originally. :)

swernli:
I don't have strong feelings about changing the name vs not, but I think that the name `Visibility` might be relying too much on colloquial terminology. I usually talk about a class' visibility but from the C# docs it seems they use the term "accessibility" which makes sense given that the accessibility is modified by the Access Modifiers (https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/access-modifiers). That said, "accessibility" as a noun always makes me think of designing products for use by the differently abled (https://en.wikipedia.org/wiki/Accessibility), so that brings me right back to `AccessModifier` as both accurate and descriptive without being confused with other concepts.

cesarzc:
I like Sarah's suggestion of calling it `Access`. It is shorter than `AccessModifiers` and uses a term that people are already familiar with.

samarsha:
Renamed `Visibility` to `Access`.

src/QsCompiler/CompilationManager/TypeChecking.cs
```diff
@@ -262,7 +272,15 @@ private static T ContainingParent<T>(Position pos, IReadOnlyCollection<(Position
...
                     file.GetCallableDeclarationHeaderItems(),
-                    (pos, name, decl, att, doc) => ContainingParent(pos, namespaces).TryAddCallableDeclaration(file.FileName, new QsLocation(pos, name.Item2), name, Tuple.Create(decl.Item1, decl.Item3), att, decl.Item2, doc),
+                    (pos, name, decl, att, doc) =>
+                        ContainingParent(pos, namespaces).TryAddCallableDeclaration(
+                            file.FileName,
+                            new QsLocation(pos, name.Item2),
+                            name,
+                            Tuple.Create(decl.Item1, decl.Item2.Signature),
+                            att,
+                            decl.Item2.Access.ValueOr(Access.Public),
```
cesarzc, line 282: If I understand correctly, before this change if the callable did not specify an access modifier keyword, the value of `decl.Item2.AccessModifier` would be `DefaultAccess` which is what would be passed to this method. After this change, `decl.Item2.Access` can be `null` and its is `null` when an access modifier keyword is not specified, correct?

samarsha: Yes, that's right.

cesarzc: Thanks!

cesarzc (approved):
Looks good to me.

src/QsCompiler/BondSchemas/BondSchemaTranslator.cs
```diff
@@ -126,7 +120,7 @@ qsBindingKind.Tag switch
                 Kind = qsCallable.Kind.ToBondSchema(),
                 FullName = qsCallable.FullName.ToBondSchema(),
                 Attributes = qsCallable.Attributes.Select(a => a.ToBondSchema()).ToList(),
-                Modifiers = qsCallable.Modifiers.ToBondSchema(),
+                Modifiers = new Modifiers { Access = qsCallable.Access.ToBondSchema() },
```
bettinaheim, line 123: Why doesn't this generate a warning; I thought Modifiers was obsolete? Does this mean we will need to update the schema version when we remove the deprecated pieces?

src/QsCompiler/Core/SymbolTable/Namespace.fs
```diff
@@ -3,6 +3,8 @@
 
 namespace Microsoft.Quantum.QsCompiler.SymbolManagement
 
+#nowarn "44" // AccessModifier is deprecated.
```
bettinaheim, line 6: Why is this using the deprecated AccessModifier, and what will be needed to update it?

samarsha: It's only used by `IsDeclarationAccessible`, which is itself deprecated and not used anywhere (but is public). We can just remove `IsDeclarationAccessible` when `AccessModifier` is removed.

src/QsCompiler/BondSchemas/BondSchemaTranslator.cs
samarsha: This is `Microsoft.Quantum.QsCompiler.BondSchemas.Modifiers`, which is generated by the Bond schema. I don't think a new schema version is necessary, since we can keep the Bond versions of `Modifiers` and `AccessModifier` indefinitely, even after the other deprecated types are removed.

src/QsCompiler/Core/SymbolTable/Namespace.fs
bettinaheim: Ah, I see, thanks for the explanation.

bettinaheim (approved):

PR #836, opened 2021-01-25
Improve doc comments in CompilationManager

kevinhartman:
As per #699, limits summary text to a single paragraph and adds parameter names and code references in documentation comments, where possible.


kevinhartman:
@samarsha 

Sorry for the delay on this. I'm realizing this is much more time-consuming than I was expecting, even with [CSharpDocRewriter](https://github.com/kevinhartman/CSharpDocRewriter) 😄.

I'm creating this as a draft for now, since I'd like to do one more pass to make sure I was consistent before asking others to review. I think I got a bit better at editing (or at least started to stick to a few patterns in some cases) as I went.

Please let me know what you think. Happy to change things up if you think any sort of different patterns should be followed.

src/QsCompiler/CompilationManager/AssemblyLoader.cs
```diff
@@ -27,15 +27,20 @@ namespace Microsoft.Quantum.QsCompiler
...
+        /// <param name="asm">The uri of the referenced assembly.</param>
+        /// <returns>
+        /// False if some of the content could not be loaded successfully,
         /// possibly because the referenced assembly has been compiled with an older compiler version.
-        /// If onDeserializationException is specified, invokes the given action on any exception thrown during deserialization.
+        /// </returns>
+        /// <exception cref="FileNotFoundException"><paramref name="asm" /> does not exist.</exception>
+        /// <exception cref="ArgumentException"><paramref name="asm" /> is not an absolute file URI.</exception>
+        /// <remarks>
+        /// If <paramref name="onDeserializationException" /> is specified, it will be invoked on any exception thrown during deserialization.
```
samarsha, line 41: Maybe this sentence could be the param tag for `onDeserializationException`? I saw you mentioned that you didn't add param tags for all methods, I'm not sure if that applies to methods that would need new documentation to be written for the params, or also ones with existing text that could be repurposed. Either way is fine. :)

src/QsCompiler/CompilationManager/AssemblyLoader.cs
```diff
@@ -27,15 +27,20 @@ namespace Microsoft.Quantum.QsCompiler
     public static class AssemblyLoader
     {
         /// <summary>
-        /// Loads the Q# data structures in a referenced assembly given the Uri to that assembly,
-        /// and returns the loaded content as out parameter.
-        /// Returns false if some of the content could not be loaded successfully,
+        /// Loads the Q# data structures in a referenced assembly given <paramref name="asm" />,
```
samarsha, line 30: Nitpicking, I've mostly seen self-closing tags written without a space at the end, like `<paramref name="asm"/>`. (Like shown [here](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/paramref) or [here](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/inheritdoc).)

samarsha (commented):
Thank you for doing this, it looks like a lot of work!

I remember you mentioned something about enabling C# doc generation in order to validate the doc comments. Would it make sense to do that in this PR to catch any mistakes with all of the tag changes? (Or a separate PR that we could merge before this one?)

src/QsCompiler/CompilationManager/AssemblyLoader.cs
kevinhartman: This is definitely one that I missed :) In general, I opted to favor repurposing any text into parameter tags where possible, but I wasn't as strict about doing that in the beginning. I'll add a comment below (hopefully tomorrow) with some of the decisions I found myself making :D

src/QsCompiler/CompilationManager/AssemblyLoader.cs
kevinhartman: Sounds good! I can easily change all of these automatically with the CSharp doc comment rewriter tool. This is just what C#'s built in XML writer seems to do out of the box for whatever reason.

kevinhartman:
Sorry for the delay.

Here's my (rough) list of decisions. @samarsha, let me know if you think we should reconsider any of these, or if you want more context for any of them.

- If method description starts with "Returns X..." and has no other content, keep this as summary (i.e. don't add a returns tag, since the summary would be empty).
- If method description like "Given X, Y, Z, returns S...", document X, Y, Z as <param> tags, and convert to "Returns S..."
- True and false as english words, not types.
- Private methods can use <see> tags for private fields in scope.
- Document return failure cases in <remarks> instead of <returns> for brevity.
- When referring to idiomatic data structures (e.g. list, enumerable, set, map), use english word rather than <see> tag.
- Replace text like "given an X" with <paramref> tag, unless:
	- we're documenting the behavior of a function returned from the current function/member (because <paramref> cannot refer to the returned function's parameters).
	- we're describing the input abstractly (e.g. it's the combination of multiple parameters, or multiple fields of a struct parameter).
	- we're describing a specific field/member of a specific parameter. In this case, use <paramref> and then <see> tag to describe which field is read from the parameter. (???)
- Add a TODO comment at bottom of comments I could use careful review help on :)

samarsha:
Those all seem reasonable to me!

kevinhartman:
Quick update: I have not abandoned this PR. I haven't had time to do a second pass yet, but I'll try to do so this week!

In the meantime, I was also wondering if you have any opinion on adding indentation for each tag body.

E.g.
```xml
<summary>
    This is a multi-line summary. It continues onto the next line, and when it does so, it
    looks like this!
</summary>
<returns>
    Return text here.
</returns>
```

This is something that Rider can do automatically for doc comments, and I think it improves readability. If you're on the fence, perhaps I can run the relevant Rider code cleanup task as a final commit to this PR, and then we can just drop it if it's decided against?

samarsha:
> Quick update: I have not abandoned this PR. I haven't had time to do a second pass yet, but I'll try to do so this week!

Sounds good!

> In the meantime, I was also wondering if you have any opinion on adding indentation for each tag body.

I think indenting does make some sense, but I think I've most commonly seen it not intended. For example, all of the examples here don't indent the tag bodies: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/summary
I'm not sure if it's explicitly called out the Microsoft or StyleCop style guides though.

(Side note, I actually would prefer to use the Rider code formatter on the entire codebase, not just for doc comments, but that would be a pretty huge change.)

kevinhartman:
@samarsha, moving out of draft, so (hopefully) ready for another look!

> I've most commonly seen it not intended.

I'd actually never seen it indented, until I ran the Rider code cleanup task. I think it does look pretty good, but I'm not sure how well VS would behave when editing indented comments, so forgoing it for now is probably best anyhow.

> I remember you mentioned something about enabling C# doc generation in order to validate the doc comments. 

Here's a branch off of this PR branch that includes an additional commit to enable doc generation and warn as error as discussed [here](https://github.com/microsoft/qsharp-compiler/issues/699#issuecomment-757623941), which I've used to validate this XML.

https://github.com/kevinhartman/qsharp-compiler/tree/keha/enable-docgen

Should I also create a separate PR enabling doc generation for the entire solution (as is done there)? This will cause all projects to generate an XML file in their build output dir.

samarsha (commented):
It looks pretty good so far. I added some comments for minor changes.

Something I noticed a few times is that documentation for the return value is sometimes split between `<summary>` and `<remarks>`. For example,

```csharp
/// <summary>
/// Returns names that match an alternative casing of the identifier at <paramref name="pos"/> in <paramref name="file"/>.
/// </summary>
/// <remarks>
/// Returns the empty enumerable if there is no valid identifier at <paramref name="pos"/>.
/// </remarks>
private static IEnumerable<string> IdCaseSuggestions(
    this FileContentManager file, Position pos, CompilationUnit compilation)
```

I sort of see `<remarks>` as being for extra, non-essential information about the method. For documentation about whether the return value is null or empty if some condition is true, it seems better to group it together with the other documentation for the return value (whether that is in `<summary>` or a dedicated `<returns>` tag). What do you think @kevinhartman?

kevinhartman:
@samarsha, thanks for the review! I'm sorry for the delay on addressing it.

> I sort of see `<remarks>` as being for extra, non-essential information about the method. For documentation about whether the return value is null or empty if some condition is true, it seems better to group it together with the other documentation for the return value (whether that is in `<summary>` or a dedicated `<returns>` tag).

I definitely agree with that thinking, and originally I did strictly place information about the return value in `<returns>`. However, I found that `<returns>` text written in the canonical way tends to become very complex when you start describing the behavior of the function, rather than the meaning of the value. As a mitigation, I decided to move text about the return value that seems more to describe function behavior than the result value to `<remarks>`.

```c#
/// <summary>
/// Returns names that match an alternative casing of the identifier at <paramref name="pos"/> in <paramref name="file"/>.
/// </summary>
/// <remarks>
/// Returns the empty enumerable if there is no valid identifier at <paramref name="pos"/>.
/// </remarks>
private static IEnumerable<string> IdCaseSuggestions(
    this FileContentManager file, Position pos, CompilationUnit compilation)
```

From the example, I thought of the empty enumerable result as more of a description of the function's behavior in edge case. A `<returns>` tag for this function could be introduced (I didn't add one for this one), and would probably read something like `"An enumerable of the matching names."`.

bettinaheim (approved):
Thanks for doing this!

PR #837, opened 2021-01-26
Updates from main

bettinaheim:
This PR just updates the feature branch.


swernli (approved):

src/QsCompiler/QirGeneration/Context.cs
```diff
@@ -28,8 +28,10 @@ namespace Microsoft.Quantum.QsCompiler.QIR
     /// This class holds the shared state used across a QIR generation pass.
     /// It also holds a large number of shared utility routines.
     /// </summary>
+#pragma warning disable SA1404 // Code analysis suppression should have justification
     [SuppressMessage("StyleCop.CSharp.ReadabilityRules", "SA1124:Do not use regions")]
```
samarsha, line 32: FYI, the intended way to avoid this warning is by adding a justification:

```suggestion
    [SuppressMessage("StyleCop.CSharp.ReadabilityRules", "SA1124:Do not use regions", Justification = "Existing code already used regions.")]
```

bettinaheim:
It looks like the build is failing because the llvm.net package is not signed.

bettinaheim:
Since this issue is unrelated to the updates from main and will need to be addressed by replacing the package in question, I'll go ahead and merge these updates to unblock other work based on qir emission.

PR #838, opened 2021-01-26
Continue with CI even if style check has errors

samarsha:
This allows the style check to run independently of the build and tests, so that the rest of the CI can continue even if Fantomas reports style errors.

### DIFF
--- a/build/init.yml
+++ b/build/init.yml
@@ -16,10 +16,7 @@ steps:
     versionSpec: '5.6.0'
 
 - task: UseDotNet@2
-  displayName: 'Use .NET Core SDK 3.1.300'
+  displayName: 'Use .NET Core SDK 3.1.405'
   inputs:
     packageType: sdk
-    version: '3.1.300'
-
-- script: dotnet tool restore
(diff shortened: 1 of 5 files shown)


build/init.yml
```diff
@@ -16,10 +16,7 @@ steps:
...
 - task: UseDotNet@2
-  displayName: 'Use .NET Core SDK 3.1.300'
+  displayName: 'Use .NET Core SDK 3.1.405'
   inputs:
     packageType: sdk
-    version: '3.1.300'
-
-- script: dotnet tool restore
-  displayName: 'Restore .NET tools'
+    version: '3.1.405'
```
cesarzc, line 22: Is there a reason to update the .NET Core version? I looked at the history of the file and it looks like the version is updated when people happen to be changing something related.

samarsha: No, there was no specific reason to update it - just felt like it would be nice to use the latest version, especially since there have been many security patches since 3.1.300.

cesarzc (approved):

PR #839, opened 2021-01-27
[Docgen] Hide internal, ```Q# → ```qsharp, and fix ms.topic.

cgranade:


guenp (approved):

src/Documentation/DocumentationGenerator/ProcessDocComments.cs
```diff
@@ -141,7 +149,10 @@ public override QsCustomType OnTypeDeclaration(QsCustomType type)
                     range: null, // TODO: provide more exact locations once supported by DocParser.
                     source: type.Source.AssemblyOrCodeFile);
 
-                this.writer?.WriteOutput(type, docComment)?.Wait();
+                if (!type.Modifiers.Access.Equals(QsCompiler.SyntaxTokens.AccessModifier.Internal))
```
samarsha, line 152: Could you add a `using` for that?

```suggestion
                if (!type.Modifiers.Access.Equals(AccessModifier.Internal))
```

src/Documentation/DocumentationGenerator/ProcessDocComments.cs
```diff
@@ -195,7 +206,10 @@ public override QsCallable OnCallableDeclaration(QsCallable callable)
                     range: null, // TODO: provide more exact locations once supported by DocParser.
                     source: callable.Source.AssemblyOrCodeFile);
 
-                this.writer?.WriteOutput(callable, docComment)?.Wait();
+                if (!callable.Modifiers.Access.Equals(QsCompiler.SyntaxTokens.AccessModifier.Internal))
```
samarsha, line 209:
```suggestion
                if (!callable.Modifiers.Access.Equals(AccessModifier.Internal))
```

samarsha (approved):

samarsha:
By the way, I think the pipelines should automatically run on every push without needing `/azp run` - is that not working for you?

cgranade:
@samarsha: I've had enough cases where  that hasn't worked that I've gotten a bit paranoid, I guess...

PR #840, opened 2021-01-27
Adding project templates for targeted executables

bettinaheim:


src/ProjectTemplates/Quantum.Honeywell.App1/Program.qs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Quantum.App1 {
+
+    open Microsoft.Quantum.Canon;
+    open Microsoft.Quantum.Intrinsic;
+    
+    @EntryPoint()
+    operation HelloQ() : Result {
```
cgranade, line 7: Minor, but should we have this be `SayHello` so that it's a verb?

cgranade (approved):
Looks good to me; would it be reasonable to also add a yeoman generator for VS Code users?

src/ProjectTemplates/Quantum.Honeywell.App1/Program.qs
ScottCarda-MS: `HelloQ` is the initial operation in all of the console templates so far, so if we change it here, we would need to change it in all the other ones as well to keep them consistent.

ScottCarda-MS: Also, I think it is a play on `Hello World` which is extremely common and well known in the programming world as an initial program in various settings.

cgranade: Believe me, I do understand the "hello, world" reference. My suggestion does stand, though, that we should ideally keep to our own naming conventions in the templates that we offer users!

bettinaheim: I updated all names in the templates, though since Message is very suboptimal for AQ project templates, I went with preparing a |+> and returning the random measurement result. Calling it SayHello hence just didn't seem accurate, so I (somewhat boringly) named it GetRandomResult.

ScottCarda-MS: I think this name change will require an update on the qdk sanity tests as well. The names will need to be changed here: https://github.com/microsoft/qdk/blob/main/Tests/Templates/xunit/tests-all/OOTBTests.qs

bettinaheim: @cgranade Scott is right, and I only got back to check on the build now. I undid the name change for the existing templates; it's getting late and I still need two builds to complete. It makes more sense doing that on the side during the day ahead of time for the next release instead of watching builds all night.

PR #841, opened 2021-01-27
Allow .NET 5 in global.json and build with .NET 5 in CI

samarsha:
With #821 merged, we no longer need a `global.json` in the root of the repository to keep the language server tests passing when .NET 5 is installed.

I'd also like to update the CI to use .NET 5, even though we aren't using any .NET 5 features yet:

* It makes sure that the compiler can always build with .NET 5, so that when we upgrade later there won't be any surprises. Compatibility with .NET Core 3.1 is still checked by the e2e build.
* The build supports XML validation for F# doc comments (#766), but only with .NET 5. This will allow CI to enforce that F# doc comments stay valid.
* Newer versions of Fantomas (for F# formatting) have some bug fixes, but require .NET 5. (Fantomas doesn't run on the e2e build so we can still use the .NET 5 versions without updating e2e.)


bettinaheim:
@anjbur @vxfield for awareness and input.

vxfield (approved):
Makes sense to me

PR #843, opened 2021-01-28
Added Template ReadMe File

ScottCarda-MS:
Added a README file to the custom templates folder to provide some links to documentation for custom templates and the full schema description.

### DIFF
--- a/src/ProjectTemplates/README.md
+++ b/src/ProjectTemplates/README.md
@@ -1,0 +1,5 @@
+This folder contains the definitions for the custom dotnet templates provided by the Quantum Development Kit.
+
+Microsoft documentation for creating custom templates: https://docs.microsoft.com/en-us/dotnet/core/tools/custom-templates
+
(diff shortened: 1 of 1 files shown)


src/ProjectTemplates/README.md
```diff
@@ -0,0 +1,5 @@
+This folder contains the definitions for the custom dotnet templates provided by the Quantum Development Kit.
```
bettinaheim, line 1: There are actually three sets of templates, so these are unfortunately not the only ones and e.g. have nothing to do with what is in the new project dialogue for VS, or the corresponding command in VS Code. These templates here are merely a NuGet package that can be installed separately to work with the command line. They are not included automatically by any of the other packages or extensions that we ship. 

```suggestion
This folder contains the definitions for the NuGet package containing templates for Q# projects. These templates are provided for working with the command line; there is a separate set of templates each that is included with the VS and the VS Code extension. 
```

bettinaheim (approved):

PR #845, opened 2021-01-30
Add Q# source code formatter

samarsha:
This adds an experimental Q# source code formatter. See the [README](https://github.com/microsoft/qsharp-compiler/blob/samarsha/formatter/src/Formatter/README.md) for current limitations, instructions on how to use it, and an overview of the design.


src/Formatter/App/App.fsproj
```diff
@@ -0,0 +1,16 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <AssemblyName>qsfmt</AssemblyName>
+    <RootNamespace>QsFmt.App</RootNamespace>
```
samarsha, line 6: Does anyone have any preferences for what to call this? We could keep QsFmt and use the namespace `Microsoft.Quantum.QsFmt`. Or something like QsFormat or QsFormatter? Or a longer option is "Q# Formatter" with namespace `Microsoft.Quantum.QSharp.Formatter`...

Then there is also samarsha/qsfmt#2. :)

Co0olboi (approved):

src/Formatter/Formatter.Tests/Examples.fs
```diff
@@ -0,0 +1,78 @@
...
+} else {
+return x + 1;
+}
+}
+}""",
+
+    """namespace Foo {
+    function Bar() : Int {
+        let x = 5;
+        if (x == 5) {
```
cgranade, line 41: Minor question, but do we want the formatter to keep parens here? I thought that parens after `if` were generally discouraged now that they're optional?

samarsha: Yes, that would be a good formatting rule to add. The formatter hasn't been taught to remove parentheses yet, and I think this test case was written before QEP 1 was implemented. I would like to delay new formatting rule additions to later PRs though.

swernli (approved):
Excited to see progress on this!

src/Formatter/App/App.fsproj
samarsha: We decided to keep QsFmt.

Issue #776, opened 2020-12-09
Unnecessary code in generated QIR


irinayat-MS:
Am using feature/qir branch at commit e5ecd6f7.
Generated attached ll file using attached qs and QirTarget/QirCore qs files from qsharp-compiler.

The generated QIR contains unnecessary defines and declarations such as:
Microsoft__Quantum__Targeting__TargetInstruction__body
Microsoft__Quantum__Core__RangeReverse__body
etc.

The extra code makes native build longer and also makes it harder to understand the actual dependencies of the generated QIR.

[qir-test-arrays.zip](https://github.com/microsoft/qsharp-compiler/files/5668192/qir-test-arrays.zip)

bettinaheim:
I removed the unnecessary declaration with PR https://github.com/microsoft/qsharp-compiler/pull/774.

irinayat-MS:
I'm still seeing the following in the generated QIR from https://github.com/microsoft/qsharp-compiler/commit/d78fef5c0a9f4bccb72256ad9e346cda32a670cd:

```
define %Range @Microsoft__Quantum__Core__RangeReverse__body(%Range %range) {
entry:
  %0 = call %Range @__quantum__qis__rangereverse__body(%Range %range)
  ret %Range %0
}

declare %Range @__quantum__qis__rangereverse__body(%Range)

define i64 @Microsoft__Quantum__Core___c5f751d561914d34ba7ac9335ce7ca34_Length__body(%Array* %array) {
entry:
  call void @__quantum__rt__array_add_access(%Array* %array)
  %0 = call i64 @__quantum__qis___c5f751d561914d34ba7ac9335ce7ca34_length__body(%Array* %array)
  call void @__quantum__rt__array_remove_access(%Array* %array)
  ret i64 %0
}

declare i64 @__quantum__qis___c5f751d561914d34ba7ac9335ce7ca34_length__body(%Array*)
```
It's in the way of the linker because we don't provide __quantum__qis__rangereverse__body or __quantum__qis___c5f751d561914d34ba7ac9335ce7ca34_length__body as part of the runtime.

bettinaheim:
Sorry, I overlooked the two that are kept since they are listed as dependencies for rewrite steps. Trimming these as well makes sense.

PR #846, opened 2021-01-31
Built-in functions for QIR emission

bettinaheim:
This PR splits out a class for handling functions for which QIR code is generated as part of QIR emission and thus don't need to be provided by the runtime. This fixes https://github.com/microsoft/qsharp-compiler/issues/776.


swernli (approved):

samarsha:
> CSC : error CS8002: Referenced assembly 'Ubiquity.NET.Llvm, Version=2.55194.55264.21762, Culture=neutral, PublicKeyToken=null' does not have a strong name. [D:\a\1\s\submodules\qsharp-compiler\src\QsCompiler\QirGeneration\QirGeneration.csproj]
CSC : error CS8002: Referenced assembly 'Ubiquity.NET.Llvm.Interop, Version=2.55194.55264.21762, Culture=neutral, PublicKeyToken=null' does not have a strong name. [D:\a\1\s\submodules\qsharp-compiler\src\QsCompiler\QirGeneration\QirGeneration.csproj]

I suspect this is only a warning that's being treated as an error in CI. You could try `<NoWarn>8002</NoWarn>`.

bettinaheim:
@samarsha Thanks for the hint! @swernli has been looking into signing the corresponding dlls, and we should be able to resolve this shortly.

PR #851, opened 2021-02-02
Fixing a naming issue

bettinaheim:
This PR fixes an issue that incorrectly named the functions used for memory management of capture tuples when memory management tables were generated by more than one callable.


src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestOpCall2.ll
```diff
@@ -9,12 +9,12 @@ entry:
...
-  call void @__quantum__rt__array_update_alias_count(%Array* %__controlQubits____inline__1, i64 1)
-  call void @__quantum__qis__x__ctl(%Array* %__controlQubits____inline__1, %Qubit* %target)
-  call void @__quantum__rt__array_update_alias_count(%Array* %__controlQubits____inline__1, i64 -1)
+  %__controlQubits__1 = call %Array* @__quantum__rt__array_concatenate(%Array* %__controlQubits__, %Array* %3)
+  call void @__quantum__rt__array_update_alias_count(%Array* %__controlQubits__1, i64 1)
+  call void @__quantum__qis__x__ctl(%Array* %__controlQubits__1, %Qubit* %target)
+  call void @__quantum__rt__array_update_alias_count(%Array* %__controlQubits__1, i64 -1)
   call void @__quantum__rt__array_update_reference_count(%Array* %3, i64 -1)
-  call void @__quantum__rt__array_update_reference_count(%Array* %__controlQubits____inline__1, i64 -1)
+  call void @__quantum__rt__array_update_reference_count(%Array* %__controlQubits__1, i64 -1)
```
swernli, line 17: I'm not sure how easy this is to do, but is it worth having a test that verifies the increment occurs when a second instance of the named inline is needed, ie: `%__controlQubits__2`?

swernli (approved):

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestOpCall2.ll
bettinaheim: It's harder than you would think to come up with a test case due to all the caching. I added it to the inlining test.

PR #859, opened 2021-02-11
Fixing incorrect ref counts

bettinaheim:
This PR fixes an issue where the generated QIR resulted in accessing released memory when arrays of pointers are updated in place and the updated item value is constructed in an inner scope.

Longer explanation looking at the test case:

```qsharp
function TestRefCounts(cond : Bool) : Unit {
    mutable ops = new Int[][5];
    if (cond) {
        set ops w/= 0 <- new Int[3];
    }
}
```
The first line gets translated into first creating and then populating the Int[][], After creation and populating it, the array and all its items have ref count 1. Then that array is assigned to the mutable variable. This assignment leads to the array and all its items having ref count 2, and an alias count 1. 
Elaborating more on why the ref count needs to be 2 and can't remain 1, think of the case where the array assigned to the mutable variable has been passed in as an argument. Suppose that argument initially has ref count 1 and is "owned" by the calling function. Then if we kept the ref count at 1, and updated an item in ops, the old item's ref count would drop to 0, releasing it. Hence (assuming we can't know which items will be updated), we increase both the alias and the ref count when assigning to mutable variables for the array and all its item. 
 
Continuing into the if-branch, we create a new array that has ref count 1. Upon assigning it to item 0 in the array, its alias count and ref count are increased. At the same time, the alias count and ref count of the old item are decreased. The ops array and all its items now have an alias count 1 and a ref count 2. The alias count of the old item is now 0, and its ref count is 1. 
 
Now we exit the scope. When we exit the scope, the array value created inside the if-branch goes out of scope. Its ref count is hence decreased by 1. This now causes a problem when we exit TestRefCounts; upon exiting, we decrease the alias count and the ref count of ops and all its items by 1. The alias count of ops and all its items is then 0 (ok), and the ref count of all items except item 0 is 1, while the ref count of item 0 is 0. 
And here is where things go wrong; we also release the initially created array (variable %0), which is correct, since the value goes out of scope. However, the original item at index 0 is no longer accessible by getting the 0-element pointer of %0 - instead of getting the old item that still has a ref count 1 that needs to be set to 0, we get the new item that has a ref count 0 already. 
 
Now why is it relevant whether the ops array has been copied inside the if-branch or not? Suppose after the first line there is another variable defined that is bound to ops, such that the if-branch actually does create the copy. In that case when we exit TestRefCounts, accessing the item at index 0 of %0 indeed still accesses the old item, and everything works fine. 

Whatever scheme we pursue to resolve the issue, it ultimately boils down to that we need to ensure that when %0 goes out of scope, we have a way to access all its items at the time of declaration. Thinking of queuing the unreferencing for each item separately when we increase the ref count for the array and all its items upon the initial assignment to ops, the problem is that iterating over the array means that the items are loaded inside a loop body, and I no longer have access to them when we exit the function. Hence, the simplest and I think cleanest solution I can come up with is that I inject an additional ref count increase for the new item and ref count decrease for the old item when an array item is modified in place. 

I should say that the reason why I think the additional count change has to be exactly 1 is that I don't see a way for the array to be unreferenced more than once at the end of the scope while also not being copied; I believe the only way it would be unreferenced multiple times is if there is another alias for it, in which case the copy would get executed and everything should work.

### DIFF
--- a/src/QsCompiler/QirGeneration/Context.cs
+++ b/src/QsCompiler/QirGeneration/Context.cs
@@ -378,17 +378,16 @@ namespace Microsoft.Quantum.QsCompiler.QIR
             this.runtimeLibrary.AddFunction(RuntimeLibrary.TupleCreate, this.Types.Tuple, this.Context.Int64Type);
             this.runtimeLibrary.AddFunction(RuntimeLibrary.TupleUpdateAliasCount, this.Context.VoidType, this.Types.Tuple, this.Types.Int);
             this.runtimeLibrary.AddFunction(RuntimeLibrary.TupleUpdateReferenceCount, this.Context.VoidType, this.Types.Tuple, this.Types.Int);
-            this.runtimeLibrary.AddFunction(RuntimeLibrary.TupleCopy, this.Types.Tuple, this.Types.Tuple, this.Context.BoolType);
+            this.runtimeLibrary.AddFunction(RuntimeLibrary.TupleCopy, this.Types.TypedTuple(this.Types.Bool, this.Types.Tuple).CreatePointerType(), this.Types.Tuple, this.Context.BoolType);
 
             // array library functions
             this.runtimeLibrary.AddVarArgsFunction(RuntimeLibrary.ArrayCreate, this.Types.Array, this.Context.Int32Type, this.Context.Int32Type);
             this.runtimeLibrary.AddVarArgsFunction(RuntimeLibrary.ArrayGetElementPtr, this.Context.Int8Type.CreatePointerType(), this.Types.Array);
-            // TODO: figure out how to call a varargs function and get rid of these two functions
             this.runtimeLibrary.AddFunction(RuntimeLibrary.ArrayCreate1d, this.Types.Array, this.Context.Int32Type, this.Context.Int64Type);
             this.runtimeLibrary.AddFunction(RuntimeLibrary.ArrayGetElementPtr1d, this.Context.Int8Type.CreatePointerType(), this.Types.Array, this.Context.Int64Type);
             this.runtimeLibrary.AddFunction(RuntimeLibrary.ArrayUpdateAliasCount, this.Context.VoidType, this.Types.Array, this.Types.Int);
             this.runtimeLibrary.AddFunction(RuntimeLibrary.ArrayUpdateReferenceCount, this.Context.VoidType, this.Types.Array, this.Types.Int);
-            this.runtimeLibrary.AddFunction(RuntimeLibrary.ArrayCopy, this.Types.Array, this.Types.Array, this.Context.BoolType);
+            this.runtimeLibrary.AddFunction(RuntimeLibrary.ArrayCopy, this.Types.TypedTuple(this.Types.Bool, this.Types.Array).CreatePointerType(), this.Types.Array, this.Context.BoolType);
             this.runtimeLibrary.AddFunction(RuntimeLibrary.ArrayConcatenate, this.Types.Array, this.Types.Array, this.Types.Array);
             this.runtimeLibrary.AddFunction(RuntimeLibrary.ArraySlice, this.Types.Array, this.Context.Int32Type, this.Types.Array, this.Types.Range);
             this.runtimeLibrary.AddFunction(RuntimeLibrary.ArraySlice1d, this.Types.Array, this.Types.Array, this.Types.Range, this.Context.BoolType);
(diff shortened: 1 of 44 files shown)


irinayat-MS:
I'd suggest, rather than thinking in terms of scopes, we try to outline the semantics of ownership when copy is involved. Following is the set of statements that I believe would define a correct model:

1. Arrays (and tuples) own their content with shared semantics, thus:
  - when arrays' ref count goes to 0 there must be code that would decrease ref-counts of its items
  - when item in an array is replaced there must be code that decreases ref-count on the items being removed and increases ref-count on the item being added

2. It might be desirable to elide the copy sometimes, thus have two different behaviours:
  - when array is actually copied, the new copied array has ref-count of one and owns the content, so there must be code that increases ref-counts of all items
  - when the copy is elided, the original array's ref-count is incremented but the content's ref-count shouldn't because there is no new owner of the content

Ideally, the runtime (which does the copy elision) would handle the different in behaviour re the ref-counts of the content, but because the runtime doesn't have type information, the task falls onto the Q# compiler while generating QIR.

I can see two possible solutions to this:
1. The compiler checks the returned pointer from the copy call and if the pointer matches the original array, the compiler infers that the copy was elided and doesn't increment ref-counts of the content.
2. The compiler checks the condition for copy elision (i.e. the alias count) and either does non-optimized copy (and increments the ref-counts of the content) or proceeds with updating the original array in place.

Personally, I like option 2 more because it makes the split of responsibilities between the compiler and runtime cleaner, but it does mean this extra handling will have to be done for all arrays/tuples whether or not they contain reference types.

kuzminrobin:
Looks like I still need to know too much to understand the logic behind the original @bettinaheim 's explanation. At least looks like the currently existing logic of ref-counting (implemented in Q# compiler/runtime) significantly differs from what I would come up with based on my C/C++ way of thinking. 
What @irinayat-MS is [writing above](https://github.com/microsoft/qsharp-compiler/pull/859#issuecomment-777691283) seems to me very close to how I imagine the ref-counting.

Here is my understanding/vision in more details:
The original code can be written in any language (Q#, C/C++, whatever) but the same `.ll` can be generated from the original code regardless of the language. Now I will consider the original code step-by step (to be more precise - I'll consider what I would expect from the `.ll` generated from that original code), based on the `QirArray` data type in [qsharp-runtime/src/QirRuntime/lib/QIR/qirTypes.hpp](https://github.com/microsoft/qsharp-runtime/blob/main/src/QirRuntime/lib/QIR/qirTypes.hpp). Original:
```qsharp
function TestRefCounts(cond : Bool) : Unit {
    mutable ops = new Int[][5];
    if (cond) {
        set ops w/= 0 <- new Int[3];
    }
}
```
The fragment `new Int[][5]` heap-allocates an instance of `QirArray` and passes to its constructor 
`5` as a number of items, 
and `sizeof(void *)` as an item size in bytes. 
The constructor heap-allocates one more block - a block for 5 pointers - and saves the address of that block in the `buffer` member variable of the `QirArray` instance. Those 5 pointers are initialized to null.
The address of the just created instance of `QirArray` is saved in the `.ll`'s internal variable `%0`, thus we get one reference to the `QirArray` instance (that is why the ref-count of this instance is set to 1 by the constructor). 
The internal variable `%0` is the `.ll`'s abstraction. In real code the internal variable `%0` can be implemented as a CPU register (in the particular case) or as a compiler-generated variable allocated on the stack (in the common case or if the value does not fit in the CPU register). Since the stack-allocated entities typically survive until the end of the scope only (until the `}`), I believe, that `.ll` assumes, that _the internal variable `%0` will only be valid until the end of the scope_.
Then, during the fragment `mutable ops = `, the address is copied from the internal variable `%0` to the variable `ops`, and we get 2 references to the same `QirArray` instance. Thus the ref-count for the `QirArray` instance is incremented from 1 to 2.

Then the execution enters the `if` and reaches the fragment `new Int[3]`. This fragment heap-allocates a second instance of `QirArray` and passes to its constructor 
`3` as a number of items, 
and `sizeof(int64_t)` as an item size in bytes. 
The constructor heap-allocates one more block - a block for 3 64-bit values - and saves the address of that block in the `buffer` member variable of the second `QirArray` instance. Those 3 64-bit values can be left *un*initialized (or can be initialized to 0).
The address of the second `QirArray` instance is saved in the internal variable `%1` (and we get one reference to the second `QirArray` instance), and the second `QirArray` instance's ref-count is set to 1 (by the constructor).
Then the fragment `set ops w/= 0 <-` (which in C++ I interpret like `ops[0] =`) copies the address from internal variable `%1` to the `ops[0]` (to the first-most item of the heap-allocated block for 5 pointers). Thus we get a second reference to the second `QirArray` instance, and we increment the ref-count for this instance from 1 to 2.
Then the execution reaches the `if`'s end of scope `}`, where the internal variable `%1` goes out of scope and stop being valid (as described earlier for the internal variable `%0`). That is why the ref-count of the second `QirArray` instance is decremented from 2 to 1 (and only the `ops[0]` refers to the second `QirArray` instance).

Now we get to the second `}` - the end of the scope of the `function TestRefCounts()`. All the variables in of this scope - `%0` and 
`ops[]` go out of the scope - stop being valid. Since they all might have been allocated on the stack (in the real code in the common case), they go out of the scope (popped from the stack) in the reverse order of allocation. That is why `ops[]` goes out of the scope first. And the ref-count for the first `QirArray` instance is decremented from 2 to 1 (now only the `%0` refers to the first `QirArray` instance). 
Then the `%0` goes out of the scope and the ref-count for the first `QirArray` instance is decremented from 1 to 0.
To be continues... temporarily saving the result.

kuzminrobin:
For the future:
If the `.ll` files are written manually then would be nice to see the comments in high-level programming language or pseudocode :-)

kuzminrobin (approved):

bettinaheim:
> For the future:
> If the `.ll` files are written manually then would be nice to see the comments in high-level programming language or pseudocode telling what is going on in the code :-)

Unfortunately the tests currently compare what is in those ll files with what the compiler emits. It would be nice to insert comments into the emitted code, but I don't think I currently have a way to do that with the wrappers as is.

bettinaheim:
@irinayat-MS 
The statements 1 and 2 you phrased are indeed consistent with how the ref counting works, though for the second bullet the implementation while equivalent in behavior currently deviates. Having the ref count change for the full array (meaning all items) inside a conditional block would prevent that these changes can be optimized out, but I would need to think more about whether there is a way to have the code be optimal when we conditionally update the counts. What the code is doing currently is indeed equivalent to that, it just reads slightly differently. I'd prefer to potentially doing further changes in that regard in a separate PR and preferable after we have a full set of integration tests. 

@kuzminrobin I believe our conversation last week largely addressed your comment. Let me know if there was something we did not cover. Let me just very briefly post the examples we had talked about here: 
```qsharp
function TestRefCounts(cond : Bool, arr : Int[][]) : Unit {
    mutable ops = arr;
    if (cond) {
        set ops w/= 0 <- new Int[3];
    }
    // do something 
}
```
We had considered the case of a caller invoking TestRefCounts when a) the arr argument is assigned to a variable in the caller, b) when the array is created as part of the call, c) what one could do after the call invocation with arr, and what if ops were returned from TestRefCounts.

kuzminrobin:
> @kuzminrobin I believe our conversation last week largely addressed
@bettinaheim, my long comment was written _before_ the discussion, sorry, I forgot to mention that.

PR #868, opened 2021-02-18
Simplify resolution of file edit in code action to avoid unnecessary execute command

ricardo-espinoza:
In the current implementation, handling the code action involves creating an execute command to edit the file. However, this leads to a deadlock in which the apply edit is waiting for the code action handling to complete and this can't complete until the edit action is applied.

It's unclear when this regression was introduced, but it's possible that recent changes in the synchronization model of the protocol are no longer allowing this approach.

This change is shortcutting this by returning an edit Microsoft.VisualStudio.LanguageServer.Protocol.CodeAction directly instead.

### DIFF
--- a/src/QsCompiler/LanguageServer/LanguageServer.cs
+++ b/src/QsCompiler/LanguageServer/LanguageServer.cs
@@ -581,10 +581,13 @@ namespace Microsoft.Quantum.QsLanguageServer
         [JsonRpcMethod(Methods.TextDocumentCodeActionName)]
         public object OnCodeAction(JToken arg)
         {
-            Command BuildCommand(string title, WorkspaceEdit edit)
+            CodeAction CreateAction(string title, WorkspaceEdit edit)
             {
-                var commandArgs = new ApplyWorkspaceEditParams { Label = $"code action \"{title}\"", Edit = edit };
-                return new Command { Title = title, CommandIdentifier = CommandIds.ApplyEdit, Arguments = new object[] { commandArgs } };
+                return new CodeAction
+                {
+                    Title = title,
+                    Edit = edit
+                };
             }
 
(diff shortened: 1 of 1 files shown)


milopezc (approved):

src/QsCompiler/LanguageServer/LanguageServer.cs
```diff
@@ -581,10 +581,13 @@ public object OnTextDocumentReferences(JToken arg)
         [JsonRpcMethod(Methods.TextDocumentCodeActionName)]
         public object OnCodeAction(JToken arg)
         {
-            Command BuildCommand(string title, WorkspaceEdit edit)
+            CodeAction CreateAction(string title, WorkspaceEdit edit)
```
samarsha, line 584: You could turn this into an expression function:

```csharp
CodeAction CreateAction(string title, WorkspaceEdit edit) =>
    new CodeAction { Title = title, Edit = edit };
```

samarsha (approved):

src/QsCompiler/LanguageServer/LanguageServer.cs
samarsha: or move it inline into the LINQ expression, since it's pretty small.

ricardo-espinoza: Sounds good. I'll leave as is for now to avoid resetting the approval (including the one from the VS SDK team), but if another change is required I'll take the suggestion as part of it.

samarsha: On the compiler repo, approvals aren't reset if new commits are pushed.

ricardo-espinoza:
Overriding because of transient break in CI due to Lodash audit failure.

PR #873, opened 2021-02-19
Update scope of quantumDevKit.languageServerPath config variable.

ricardo-espinoza:


ricardo-espinoza:
Abandoning pull request to avoid source branch re-use.  Will use a separate branch for fix port.

PR #880, opened 2021-02-24
Fixing duplicate intrinsics

swernli:
This fixes an issue where `body intrinsic` callables would appear twice in monomorphized results by pruning the list of concretized callables based on intrinsic settings.

### DIFF
--- a/src/QsCompiler/Transformations/Monomorphization.cs
+++ b/src/QsCompiler/Transformations/Monomorphization.cs
@@ -101,7 +101,12 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.Monomorphization
         {
             public static QsCompilation Apply(QsCompilation compilation, List<QsCallable> callables, ImmutableHashSet<QsQualifiedName> intrinsicCallableSet, bool keepAllIntrinsics)
             {
-                var filter = new ResolveGenerics(callables.ToLookup(res => res.FullName.Namespace), intrinsicCallableSet, keepAllIntrinsics);
+                var filter = new ResolveGenerics(
+                    callables
+                        .Where(call => !keepAllIntrinsics || !intrinsicCallableSet.Contains(call.FullName))
+                        .ToLookup(res => res.FullName.Namespace),
+                    intrinsicCallableSet,
+                    keepAllIntrinsics);
 
                 return filter.OnCompilation(compilation);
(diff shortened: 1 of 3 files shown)


swernli:
@ScottCarda-MS I could use some tips on testing this change. I was able to manually verify from a built package that it eliminates the issue, but perhaps there are tests I can add in the runtime repo once this is merged?

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -101,7 +101,12 @@ private class ResolveGenerics : SyntaxTreeTransformation<ResolveGenerics.Transfo
         {
             public static QsCompilation Apply(QsCompilation compilation, List<QsCallable> callables, ImmutableHashSet<QsQualifiedName> intrinsicCallableSet, bool keepAllIntrinsics)
             {
-                var filter = new ResolveGenerics(callables.ToLookup(res => res.FullName.Namespace), intrinsicCallableSet, keepAllIntrinsics);
+                var filter = new ResolveGenerics(
+                    callables
+                        .Where(call => !keepAllIntrinsics || !intrinsicCallableSet.Contains(call.FullName))
```
bettinaheim, line 106: You could pull this out of the iteration.

bettinaheim (approved):
Ah, thanks for finding this!

ScottCarda-MS (approved):

ScottCarda-MS:
@swernli For testing, there are some Monomorphization test cases found here: https://github.com/microsoft/qsharp-compiler/blob/main/src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/Monomorphization.qs
and the associated tests for these are found here: https://github.com/microsoft/qsharp-compiler/blob/main/src/QsCompiler/Tests.Compiler/LinkingTests.fs#L244

You could add a test case that defines a custom intrinsic, and then uses this intrinsic in the entry point. The testing logic would then verify that there is exactly the expected operations in the syntax tree after Monomorphization is run (it is run for all such test cases). I believe even just running a Signatures test on this would have that effect. We can discuss that more offline if you want to know more about how to do that.

swernli:
@ScottCarda-MS Thanks, that's perfect! Don't know how I missed that in my searches...

PR #881, opened 2021-02-25
Allow Multiple Entry Points

ScottCarda-MS:
This PR removes the logic that throws an error when multiple entry points are present in a compilation.

Major Changes:
 - Removed logic that throws error when multiple entry points are present.
 - Made associated error codes obsolete.
 - Added a public-facing `WriteFile` method to `CompilationLoader` that allows files to be written even when they are not related to a source file. This is used by the runtime PR to write the new generated C# Entry Point file.
 - Updated the SDK targets file to include all C# files in the compilation, in-line with this guide (https://github.com/microsoft/qsharp-compiler/blob/main/src/QuantumSdk/README.md#injected-c-code).

This is should be merged in tandem with this PR on the runtime PR: https://github.com/microsoft/qsharp-runtime/pull/528


swernli (approved):
LGTM

src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -248,8 +248,6 @@ type ErrorCode =
     | UserDefinedTypeInEntryPointSignature = 6233
     | InnerTupleInEntryPointArgument = 6234
     | ArrayOfArrayInEntryPointArgument = 6235
-    | OtherEntryPointExists = 6236
-    | MultipleEntryPoints = 6237
```
samarsha, line 252: It's unlikely that anyone is using these directly, but technically they should be marked `Obsolete` first before being removed.

src/QsCompiler/Core/SymbolTable/NamespaceManager.fs
```diff
@@ -408,16 +408,7 @@ type NamespaceManager(syncRoot: IReaderWriterLock,
...
-                match GetEntryPoints() |> Seq.tryHead with
-                | None -> isExecutable, errs
-                | Some (epName, epSource) ->
-                    let msgArgs = [ sprintf "%s.%s" epName.Namespace epName.Name; epSource ]
-
-                    errs.Add
-                        (offset,
-                         range |> orDefault |> QsCompilerDiagnostic.Error(ErrorCode.OtherEntryPointExists, msgArgs))
-
-                    false, errs
```
samarsha, line 420: We should coordinate this with microsoft/qsharp-runtime#528, since allowing multiple entry points will break C# generation without that PR. Since that PR can't be fully tested without also using this PR, maybe using a feature branch would be better?

samarsha (changes requested):

src/QsCompiler/DataStructures/Diagnostics.fs
ScottCarda-MS: done

src/QsCompiler/Core/SymbolTable/NamespaceManager.fs
ScottCarda-MS: This PR is dependent on the runtime PR https://github.com/microsoft/qsharp-runtime/pull/528, but that runtime PR is not dependent on this PR. That PR has tests to ensure continuity of existing behavior and just makes room for this PR to work. That PR will need to be merged in first, but I don't think a feature branch is needed here.

cgranade: That would also help make sure we can test the alphas from that feature branch against IQ#'s entry point generator.

samarsha: The runtime PR's purpose is to add support for multiple entry points, but it currently has no tests against multiple entry points. I'm concerned that we'll merge both of these PRs, then add driver tests for multiple entry points and realize that there are bugs/design issues that only came up after using it in practice. That could cause more changes/revisions to the entry point driver code. Using a feature branch to make sure it's thoroughly tested would avoid the potential churn on main.

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -1206,5 +1206,26 @@ public static string GeneratedFile(string fileId, string outputFolder, string fi
...
+            var outputUri = new Uri(FullDirectoryName(outputFolder));
+            var fileDir = Path.GetDirectoryName(outputUri.LocalPath);
+            var targetFile = Path.GetFullPath(Path.Combine(fileDir, fileName));
+
+            if (!Directory.Exists(fileDir))
+            {
+                Directory.CreateDirectory(fileDir);
+            }
+            File.WriteAllText(targetFile, content);
+        }
```
samarsha, line 1229: This has some code in common with `GeneratedFile`. Could that be factored out?

src/QuantumSdk/Sdk/Sdk.targets
```diff
@@ -156,7 +156,7 @@
     <Exec Condition="$(DefaultEntryPoint) != ''" Command="$(DefaultEntryPointGeneration) $(DefaultEntryPoint)" IgnoreExitCode="false" />
     <ItemGroup>
       <EmbeddedResource Include="$(GeneratedFilesOutputPath)$(PathCompatibleAssemblyName).bson" LogicalName="__qsharp_data_bond_v1__.bson" Visible="false" />
-      <Compile Condition="$(CSharpGeneration)" Include="$(GeneratedFilesOutputPath)**/*.g.cs" Exclude="@(Compile)" AutoGen="true" />
+      <Compile Condition="$(CSharpGeneration)" Include="$(GeneratedFilesOutputPath)**/*.cs" Exclude="@(Compile)" AutoGen="true" />
```
samarsha, line 159: There are other places in this file that use the `*.g.cs` pattern that should be updated.

samarsha (changes requested):

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -1206,5 +1206,26 @@ public static string GeneratedFile(string fileId, string outputFolder, string fi
             File.WriteAllText(targetFile, content);
             return targetFile;
         }
+
+        /// <summary>
+        /// Writes the given contents to the given file name at the given output folder.
+        /// Throws the corresponding exception if any of the path operations fails or if the writing fails.
+        /// </summary>
+        public static void WriteFile(string fileName, string outputFolder, string content)
```
bettinaheim, line 1214: Why is this separate method needed opposed to using GeneratedFile?

ScottCarda-MS: GeneratedFile only works on files that have a matching Q# source file. We need something more, like this, to generate a separate file for the C# Main.

bettinaheim: Do you mean because of the use of TryGetUri? No Q# file needs to exist - the only requirement is that the format of the path matches what the compiler expects (take a look at TryGetUri - I believe it was a full absolute path).

ScottCarda-MS: The `TryGetUri` will not succeed (it will return false with the out var as null) if the given file path does not point to an existing folder, i.e. an existing Q# file. This is probably a good check when generating files that are based on existing Q# files, like what is done in `GeneratedFile`, but is not a valid check when generating the Main file, which is not tied to any one specific Q# file.

ScottCarda-MS: There is also no reason to get a Uri to any existing file when generating the Main file. At least as far as I can tell.

ScottCarda-MS: Hmm, maybe I'm not right about `TryGetUri` failing if the file doesn't exist. But then what would I provide for arguments to the `GeneratedFile` method to produce the `EntryPoint.g.Main.cs` file?

bettinaheim: You just need to full path for the file (opposed to just a local path). the fileId is then the full path (e.g. C:/SomePath/EntryPoint.g.Mail.cs), the output folder would be e.g. C:/SomePath/ (don't count on my slashes being right), the file ending would be .cs (or you could also give a fileId C:/SomePath/EntryPoint.qs and then a file ending of .g.Main.cs). 

To make sure you have the correct slashes, the easiest is to create an Uri and then use CompilationUnitManager.TryGetFileId.

ScottCarda-MS: I see, thanks! So the end result on the runtime repo will look like:
```
let mainSourceFile = "./EntryPoint" |> Path.GetFullPath |> Uri |> CompilationUnitManager.GetFileId
CompilationLoader.GeneratedFile(mainSourceFile, dir, ".g.Main.cs", content) |> ignore
```
This will remove the need for the `WriteFile` method on the `CompilationLoader`. Nice!

bettinaheim: Sounds good. I think you are missing the output folder that is configurable for the C# generation here somewhere (instead of "./EntryPoint" is should probabably be something like Path.Combine(outputfolder, "EntryPoint")).

src/QsCompiler/Compiler/CompilationLoader.cs
ScottCarda-MS: This method will be removed.

src/QsCompiler/Compiler/CompilationLoader.cs
ScottCarda-MS: The file id is just used to get the file name, not the path. In the `GeneratedFile` method it will use the output folder (passed in here as `dir`) and append the file name from the file id, and append the given file extension. So it doesn't matter what path we use in the file id, so long as it ends with the name we want for the resulting file, in this case "EntryPoint". It does have to be a valid, absolute path though.

src/QsCompiler/Compiler/CompilationLoader.cs
ScottCarda-MS: removed

src/QsCompiler/Compiler/CompilationLoader.cs
bettinaheim: The GeneratedFile method preserves the folder structure in the output folder. I checked and the structure is preserved relative to the current directory, which means what you are saying will work, though it feels a little bit dangerous to me to give a path that you know is not "correct". Just giving this as input to consider.

ScottCarda-MS: Can you give me an example of what you mean when you say 'The GeneratedFile method preserves the folder structure in the output folder.'? If what I think you are saying is correct, then https://github.com/microsoft/qsharp-runtime/issues/357 wouldn't be an issue.

ScottCarda-MS: @bettinaheim Oh I see, that issue is about files not under the current Q# project's folder structure. Okay, I now really think that the task of fixing the file name collisions has to be separated out into some separate work. There just keeps being more and more nuances and complexity added onto this, and it really doesn't have anything to do with Multiple Entry Points. I'm going to remove the enumeration in the name logic and try to get the Multiple Entry Points feature in without it.

ScottCarda-MS: Okay so given that there will be a folder structure under the output folder, will feeding the "./EntryPoint" into the GeneratedFile method always put the EntryPoint.g.Main.cs directly in the output folder (not a subfolder), or is it unsafe enough to were there is a possibility where it could go somewhere else? And would saying `Path.Combine(outputfolder, "EntryPoint")` fix that issue to make sure it is always directly in the output folder?

src/QuantumSdk/Sdk/Sdk.targets
ScottCarda-MS: These have been addressed now.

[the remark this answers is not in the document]
ScottCarda-MS: These have been addressed now.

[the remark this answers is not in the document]
ScottCarda-MS: The SDK targets file should be where I mean it to be, so it would be fine for review again.

src/QsCompiler/Compiler/CompilationLoader.cs
ScottCarda-MS: I'm moving the conversation on the `"./EntryPoint"` vs `Path.Combine(outputFolder, "EntryPoint")` to the runtime PR since it is more relevant there. Here is the link: https://github.com/microsoft/qsharp-runtime/pull/528#discussion_r598423034

bettinaheim: Yes, that definitively applied more to the runtime PR than this one, sorry.

src/QuantumSdk/Sdk/Sdk.targets
```diff
@@ -35,9 +35,9 @@
   <Target Name="QSharpClean" BeforeTargets="Clean">
     <ItemGroup>
       <_FilesToClean Include="$(GeneratedFilesOutputPath)**" />
-      <Compile Remove="$(_FilesToClean)**/*.g.cs" />
+      <Compile Remove="$(_FilesToClean)**/*.cs" />
     </ItemGroup>
-    <Delete Files="@(_FilesToClean)" />
+    <Delete Files="$(GeneratedFilesOutputPath)**/*.*" />
```
bettinaheim, line 40: Was there a reason for changing that? I think removing _FilesToClean should be fine.
```suggestion
    <Delete Files="@(_FilesToClean)" />
```

bettinaheim (approved):

src/QsCompiler/Core/SymbolTable/NamespaceManager.fs
bettinaheim: While this PR needs to be coordinated with the runtime PR, I don't think we need to require having a feature branch every time that is the case. Given that both PRs are ready and tested, let's move forward with this.

samarsha: @bettinaheim, I did not mean that a feature branch is needed every time a PR needs to be coordinated across multiple repos. This comment was made when the runtime PR did not have tests for multiple entry points, so I requested a feature branch until the tests were added. Those tests have now been added directly to the runtime PR, so I don't think a feature branch is needed anymore.

src/QuantumSdk/Sdk/Sdk.targets
ScottCarda-MS: @anpaz had suggested that the `<path>/**` format stored in _FilesToClean was unsafe and recommended that it be changed to `<path>/**/*.*` Since I'm not clear on what all this is, if you still think I made a mistake and it should be _FilesToClean instead, just respond to this thread again and let me know.

ScottCarda-MS: Actually now that I take a closer look at this, it _does_ seem like I would want the `<path>/**` syntax here, because we would want to remove any directories as well. I will make the change.

ScottCarda-MS: done

PR #887, opened 2021-03-05
NuGet compliance changes

anjbur:
Remove publishing qsc NuGet package and add <clear /> tag to NuGet.Config for compliance.

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -11,6 +11,7 @@
 -->
 <configuration>
   <packageSources>
+    <clear />
     <add key="qdk-alpha" value="https://pkgs.dev.azure.com/ms-quantum-public/Microsoft Quantum (public)/_packaging/alpha/nuget/v3/index.json" protocolVersion="3" />
   </packageSources>
 </configuration>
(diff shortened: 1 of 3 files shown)


NuGet.config
```diff
@@ -11,6 +11,8 @@
 -->
 <configuration>
   <packageSources>
+    <clear />
```
cgranade, line 14: 💕

build/pack.ps1
```diff
@@ -89,7 +89,6 @@ Publish-One '../src/QsCompiler/CommandLineTool/CommandLineTool.csproj'
 Publish-One '../src/QuantumSdk/Tools/BuildConfiguration/BuildConfiguration.csproj'
 
 Pack-One '../src/QsCompiler/Compiler/Compiler.csproj' '-IncludeReferencedProjects'
-Pack-One '../src/QsCompiler/CommandLineTool/CommandLineTool.csproj' '-IncludeReferencedProjects'
```
cgranade, line 92: I agree with this change, but it may be worth reconsidering in the future whether we want to publish the command-line tool as a tool in its own right.

cgranade (approved):

PR #888, opened 2021-03-06
undo changes to NuGet.config from PR#887

anjbur:
Revert NuGet.config change as it introduced a break into end to end builds that wasn't caught by PR validations.


cgranade (approved):

anjbur:
Manually merging this, because this is broken across repos and the end-to-end checks will be blocking each other.

Issue #801, opened 2021-01-04
QIR: Functions that return Range by value produce IR that isn't compatible with C++


irinayat-MS:
Related to https://github.com/microsoft/qsharp-language/issues/31

Define a C-type that has the same layout as %Range in QIR:
```
0:000> dt qir_static_tests!QirRange
   +0x000 start            : Int8B
   +0x008 step             : Int8B
   +0x010 end              : Int8B
```

This C-function invokes a Q# function that returns a range by value and the assembly, generated for it by Clang:
```
int64_t TestBody()
{
    QirRange range = Microsoft__Quantum__Testing__QIR__CreateRange__body(9, 17, 3);
    return range.end * 5;
}
0:000> uf qir_static_tests!TestBody
qir_static_tests!TestBody [D:\repos\qsharp-runtime\src\QirRuntime\test\QIR-static\qir-driver.cpp @ 319]:
  319 00007ff7`3772f3d0 4883ec38        sub     rsp,38h
  320 00007ff7`3772f3d4 488d4c2420      lea     rcx,[rsp+20h] // memory for 'range' is allocated on the stack, rcx contains the address now
  320 00007ff7`3772f3d9 ba09000000      mov     edx,9
  320 00007ff7`3772f3de 41b811000000    mov     r8d,11h
  320 00007ff7`3772f3e4 41b903000000    mov     r9d,3
  320 00007ff7`3772f3ea e8e1d90a00      call    qir_static_tests!Microsoft__Quantum__Testing__QIR__CreateRange__body (00007ff7`377dcdd0)
  321 00007ff7`3772f3ef 486944243005000000 imul  rax,qword ptr [rsp+30h],5 // expects to find range.end at rsp+30h, that is, rsp+20h+10h
  321 00007ff7`3772f3f8 4883c438        add     rsp,38h
  321 00007ff7`3772f3fc c3              ret
```

The Q# function to create a range, the QIR generated for it by Q# compiler and the assembly code, generated by Clang from the IR:
```
function CreateRange(start: Int, end: Int, step: Int) : Range { return start..step..end; }

define %Range @Microsoft__Quantum__Testing__QIR__CreateRange__body(i64 %start, i64 %end, i64 %step) {
entry:
  %0 = load %Range, %Range* @EmptyRange
  %1 = insertvalue %Range %0, i64 %start, 0
  %2 = insertvalue %Range %1, i64 %step, 1
  %3 = insertvalue %Range %2, i64 %end, 2
[...]

PR #889, opened 2021-03-08
Entry point mapping

bettinaheim:
This PR implements implements generating an entry point wrapper during QIR emission that has a signature that does not rely on any QIR specific/opaque types. Generating this wrapper allows to decouple e.g. the command line processing or other driver code from the runtime implementation. The generated wrapper itself relies on the existence of a malloc function (`__quantum_rt__memory_allocate`) in the runtime.

More tests will follow shortly.
If the e2e build fails, that would be a good sign: The e2e should currently fail if it is properly running the runtime tests against the compiler edits, since I finally adjusted the string creation signature to match the specs.

This PR also contains 
- a bug fix that the declarations for monomorphized functions that are implemented either directly in llvm or via a call to a runtime function were not properly stripped (see OriginalNameFromMonomorphized)
- fixes an issue for when discarding the loop variable in for-loops over ranges (in StatementTransformation.cs)
- replaces the call to __quantum__rt__int_power with suitable casts and an invokation to llvm.powi instead
Sorry for not properly splitting out those changes - we have I think at least 3 more PRs queued up and it is getting a bit cumbersome to coordinate. We need to unify all of that and then working against main should facilitate keeping a proper structure for PRs...

Additional infos to facilitate review:
For this PR, the changes that contain the core of the functionality outlined in the description are in Interop.cs. The changes in particular are in the generation of the wrapper function for the entry point. The method GenerateWrapper generates that wrapper. MapToInteropType defines how QIR types in entry point signatures are replaced in the entry point wrapper. ProcessArguments does that replacement for argument; i.e. it maps "generic types" -> QIR types. ProcessReturnValue does that replacement for the return value; i.e. it maps QIR types -> "generic types". Having one general purpose function doing both, i.e. one function to map types A <-> types B, with the types A and B fully configurable would be nice, but makes the code really tedious both to write, read and maintain, so I decided to favor the two separate functions in this case. 
 
Configuration.cs has been removed. While it might in general be valuable to allow to configure the QIR emission, the only configuration options that were available so far was the names of the types that the QIR runtime uses to represent the opaque QIR types and other QIR pointer types. Having thought a bit about how to package everything, I believe it is a better approach not to use those in the wrapper to decouple the dependency on the runtime from the entry point driver (C++ or other native code calling into QIR) and command line handling. I believe this should save us some packaging grievances. This comes at the cost of needing the runtime to provide something like a malloc function (__quantum__rt__memory_allocate).


src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestOpArgument.ll
```diff
@@ -1,4 +1,4 @@
-define %String* @Microsoft__Quantum__Testing__QIR__TestOpArgument__body() #0 {
+define %String* @Microsoft__Quantum__Testing__QIR__TestOpArgument__body() {
```
swernli, line 1: This IR seems like it went from being an entrypoint to no longer being an entrypoint, but I don't see a corresponding change in any qs files to remove that attribute. Is there something I'm misunderstanding here?

bettinaheim: There is still an entry point in the file, just a separate one. The QIR generation emits the "normal" QIR compilant definition of the entry point function. Additionally, it generates a function that calls into that entry point and has a signature that does not rely on any of the runtime types, or contain any of the opaque QIR types and uses 8-bit multiples for intergers and similar things. If the signatures of both functions match, then the wrapper is not generated but the QIR function is just tagged as entry point (which was previously the case for TestOpArgument, but is no longer the case since we started to properly map strings).

swernli: Thanks, that clarifies it!

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestEntryPoint.ll
```diff
@@ -1,17 +1,36 @@
...
+define double @Microsoft__Quantum__Testing__QIR__TestEntryPoint({ i64, i8* }* %a, i8 %b) #0 {
 entry:
-  %0 = call %Array* @__quantum__rt__array_create_1d(i32 8, i64 %a__count)
-  %1 = icmp sgt i64 %a__count, 0
-  br i1 %1, label %copy, label %next
+  %0 = getelementptr { i64, i8* }, { i64, i8* }* %a, i64 0, i32 0
+  %1 = getelementptr { i64, i8* }, { i64, i8* }* %a, i64 0, i32 1
+  %2 = load i64, i64* %0
+  %3 = load i8*, i8** %1
+  %4 = call %Array* @__quantum__rt__array_create_1d(i32 8, i64 %2)
```
cesarzc, line 7: It seems like all arrays received by entry-points are the same size (64-bits). For use cases in which the data-type of the array is not 64 bits (e.g. Pauli), should the inter-op array encode Pauli values in 64-bits?

kuzminrobin (approved):
I was mostly looking at `.ll`s and  `.qs`s. I don't understand most of the other (`.cs`, `.fs`) files.
What I looked at seems good to me.

src/QsCompiler/QirGeneration/Interop.cs
```diff
@@ -0,0 +1,436 @@
...
+            }
+            else if (typeName == TypeNames.Callable || typeName == TypeNames.Qubit)
+            {
+                return bytePtrType;
+            }
+            else if (typeName == TypeNames.Result)
+            {
+                return this.sharedState.Context.Int8Type;
+            }
+            else if (t is IStructType st)
```
cesarzc, line 110: Does `Range` fall into the `IStructType` category?

Currently the way an entry point `Range` argument is generated as `%Range`. However, since the size of `%Range` is 24 bytes, interop with C++ is a bit strange because QIR expects an argument of that size but from C++ a 24 bytes single argument cannot be passed so three 8 bytes arguments have to be used.

For interop purposes, I think a `Range` argument in the entry-point operation should be received as a pointer `%Range*`.

kuzminrobin: I'll provide a bit more details to Cesar's observation.
During [that discussion](https://github.com/microsoft/qsharp-runtime/pull/520#discussion_r589733387), I asked Cesar offline to stop in the debugger at the caller and see in the disassembly how the caller passes the instance of the `struct Range` by value.
From Cesar's analysis I got an impression that the caller creates a copy (of the `struct Range` instance) on the stack (that's what I expected), but then passes to the callee the _address_ of that copy. As a result the callee, instead of getting the three 64-bit values in certain part of the stack, gets (in that part of the stack or in a CPU register) the _address_ to those values.

bettinaheim: Indeed there is an issue tracked on this already: https://github.com/microsoft/qsharp-compiler/issues/801
I believe the issue only occurs for interop, and indeed Range would fall under this IStructType here and should be passed as a pointer. However, I think I'll have to update the PR to make sure the actual memory storing the struct is allocated on the heap such that it won't be deallocated upon exiting QIR (I forgot).

bettinaheim: The ranges should now be handled properly (confirmation pending for whether everything executes as it should).

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestEntryPoint.ll
bettinaheim: Coming back to this, I can no longer reproduce; passing a Pauli[] as entry point argument instead of a Double[] changes this line to   
```
%4 = call %Array* @__quantum__rt__array_create_1d(i32 1, i64 %2)
```
It seems that one of the other changes I did since you reviewed fixed the issue.

cesarzc (approved):
Looks good to me.

PR #891, opened 2021-03-10
Create Bond schema to represent a Q# entry-point operation

cesarzc:
This change creates a Bond schema to represent a Q# entry-point operation.

This representation will be the input for the entry-point driver generator.


src/QsCompiler/BondSchemas/EntryPoint/EntryPoint.bond
```diff
@@ -0,0 +1,114 @@
...
+    45: nullable<ArrayValue> Array;
+}
+
+struct Argument
+{
+    5: DataType Type = BoolType;
+
+    10: string Name;
+
+    15: int32 Position;
```
swernli, line 99: I'm curious why position gets encoded here? Is there ever a time where an argument's position and its index in the `Arguments` vector inside of `EntryPointOperation` would be different? This seems like it could be redundant information  and might be eliminated.

src/QsCompiler/BondSchemas/EntryPoint/EntryPoint.bond
```diff
@@ -0,0 +1,114 @@
...
+    // Not null when Type is StringType.
+    40: nullable<string> String;
+
+    // Not null when Type is ArrayType.
+    45: nullable<ArrayValue> Array;
+}
+
+struct Argument
+{
+    5: DataType Type = BoolType;
```
swernli, line 95: `Argument` has a type but so does `ArgumentValue`. Since tuples are not supported on entrypoints and arrays must always be of the same type, it seems like only the type and array type at `Argument` should be necessary.

cesarzc: I was thinking that having each structure standalone but in this case it does not really make sense to have a `Type` member in `ArgumentValue`. I'll remove it.

src/QsCompiler/BondSchemas/EntryPoint/EntryPoint.bond
cesarzc: In theory, the position member should never differ from its position in the vector of arguments. However, since this might be transferred to other components and its critical for the correct functioning of the c++ entry-point driver generator, I think it is good to have this piece of redundant information.

src/QsCompiler/BondSchemas/EntryPoint/EntryPoint.bond
cesarzc: Removed the redundant members.

swernli (approved):

PR #895, opened 2021-03-11
Trim Rewrite Step

ScottCarda-MS:
Adds the Syntax Tree Trimming rewrite step. Removes syntax tree trimming logic from the Monomorphization rewrite step. Monomorphization will now only remove all generics that are not intrinsic. The new rewrite step will remove all unused callables from the syntax tree except those callables needed by later rewrite steps.


src/QsCompiler/Tests.Compiler/Tests.Compiler.fsproj
```diff
@@ -1,532 +1,535 @@
...
+    <None Include="TestCases\LinkingTests\Core.qs">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </None>
+    <None Include="TestCases\LinkingTests\Generics.qs">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </None>
+    <None Include="TestCases\LinkingTests\ReferenceLinking.qs">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </None>
+    <None Include="TestCases\LinkingTests\SyntaxTreeTrim.qs">
```
ScottCarda-MS, line 340: Here is the major change to the Tests.Compiler.fsproj file. At some point this file got the CRLF line endings, which is wrong, and my git is configured to only push files as having LF line endings, which is consistent with all the other files in the repo. With that in mind, I will keep the line ending changes to this file so that it is made consistent with the rest of the repo.

src/QsCompiler/Compiler/PluginInterface.cs
```diff
@@ -17,6 +17,12 @@ namespace Microsoft.Quantum.QsCompiler
     /// </summary>
     public static class RewriteStepPriorities
     {
+        /// <summary>
+        /// Priority of the built-in transformation that removes all
+        /// unused callables from the syntax tree.
+        /// </summary>
+        public const int SyntaxTreeTrimming = 1200;
```
kuzminrobin, line 24: Minor:
Are there any limitations for this value? Must this value be above or below (or both) certain threshold?

src/QsCompiler/Compiler/RewriteSteps/SyntaxTreeTrimming.cs
```diff
@@ -0,0 +1,54 @@
...
+        public IDictionary<string, string?> AssemblyConstants { get; } = new Dictionary<string, string?>();
+
+        public IEnumerable<IRewriteStep.Diagnostic> GeneratedDiagnostics => Enumerable.Empty<IRewriteStep.Diagnostic>();
+
+        public bool ImplementsPreconditionVerification => true;
+
+        public bool ImplementsTransformation => true;
+
+        public bool ImplementsPostconditionVerification => false;
+
```
kuzminrobin, line 31: 1. (My subjective opinion, feel free to ignore)
I prefer more dense code, code that occupies less space on the screen. I would not use empty lines here.
But if that is the requirement of the coding style then I'm totally fine with that.
2. (Minor) I suspect some of these are reasonable to be `const` (speaking in C++), or `readonly`? (speaking in C#).

src/QsCompiler/Compiler/RewriteSteps/SyntaxTreeTrimming.cs
```diff
@@ -0,0 +1,54 @@
...
+            this.keepAllIntrinsics = keepAllIntrinsics;
+        }
+
+        public bool PreconditionVerification(QsCompilation compilation) => compilation.EntryPoints.Any();
+
+        public bool Transformation(QsCompilation compilation, out QsCompilation transformed)
+        {
+            transformed = TrimSyntaxTree.Apply(compilation, this.keepAllIntrinsics);
+            return true;
+        }
```
kuzminrobin, line 47: (Minor) Looks like non-virtual function  always returning the same value. Why not return `void` then? (if C# has `void`, I don't know C# :-) )

src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/SyntaxTreeTrim.qs
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+// Trimmer Removes Unused Callables
+namespace Microsoft.Quantum.Testing.SyntaxTreeTrimming {
+    
+    @ EntryPoint()
+    operation Main() : Unit {
+        UsedOp();
+    }
```
kuzminrobin, line 10: (Minor)
In other files you removed the space after `@`. Here the space is present.
Is there a any specific meaning of that space?

src/QsCompiler/Compiler/RewriteSteps/SyntaxTreeTrimming.cs
cgranade: The signature here is required by the [`IRewriteStep` interface](https://github.com/microsoft/qsharp-compiler/blob/30880f669fe508e9ea777b7478b28a7b844aa4e1/src/QsCompiler/Compiler/PluginInterface.cs#L190).

kuzminrobin (approved):
I understand very little in C#/F#. What I understand, looks good to me. Approving.
I recommend at least one more approval.

src/QsCompiler/Compiler/RewriteSteps/SyntaxTreeTrimming.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public bool PostconditionVerification(QsCompilation compilation)
```
cesarzc, line 49: If there is not a post-condition verification, should this method always return `true` rather than raise an exception?

src/QsCompiler/Compiler/PluginInterface.cs
ScottCarda-MS: There are no limitations that I know about. I think that the exact numbers are somewhat arbitrary except the ordering they establish (which is not arbitrary).

src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/SyntaxTreeTrim.qs
ScottCarda-MS: 🤔Yah there does seem to be some inconsistency with this. I had just copied the style I had seen in other LinkingTests test cases, but elsewhere in the project I'm seeing attributes without the space. @cgranade Which of these two is the official style for Q#, and is it worth it to fix all the Q# attributes in the compiler to have the correct style in this PR?

src/QsCompiler/Compiler/RewriteSteps/SyntaxTreeTrimming.cs
ScottCarda-MS: All the rewrite tests we have written so far that don't implement one of the verifications raises the exception. I think that is just the way they are meant to be used. The method shouldn't get called so long as the corresponding `ImplementsPreconditionVerification` or `ImplementsProstconditionVerification` is set to `false`, so I suppose this exception raising acts as a failsafe incase that flag ever accidently gets set to `true` when we don't really implement verification for the step.

src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/SyntaxTreeTrim.qs
ScottCarda-MS: Looking at the style guide, it looks like the recommended style is to not have the space after the `@`. I've updated all the references in the compiler repo that I could find to not have the space.

src/QsCompiler/Transformations/SyntaxTreeTrimming.cs
```diff
@@ -0,0 +1,109 @@
...
+                    .Concat(compilation.EntryPoints)
+                    .Distinct()
+                    .ToImmutableArray();
+                var compilationWithBuiltIns = new QsCompilation(compilation.Namespaces, augmentedEntryPoints);
+                var callablesToKeep = new CallGraph(compilationWithBuiltIns, true).Nodes.Select(node => node.CallableName).ToImmutableHashSet();
+
+                // ToDo: convert to using ternary operator, when target-type
+                // conditional expressions are supported in C#
+                Func<QsNamespaceElement, bool> filter = elem => Filter(elem, callablesToKeep);
+                if (keepAllIntrinsics)
```
cesarzc, line 40: It would be great to add a comment on why `keepAllIntrinsics` is needed. Maybe in a documentation comment to the `Apply` method.

Is this used when executing this rewrite-step on libraries such that it doesn't remove all callables because they are not called anywhere that starts at an entry-point?

ScottCarda-MS: Hmm, this behavior was migrated from the Monomorphization rewrite step. Now that Monomorphization doesn't do trimming, it didn't make sense to keep in there. It was originally added by @bettinaheim I think, I'm not entirely sure what it is used for, I had thought maybe it was related to QIR emmission. @bettinaheim could you provide more context for this flag?

src/QsCompiler/Compiler/PluginInterface.cs
kuzminrobin: OK

src/QsCompiler/Compiler/RewriteSteps/SyntaxTreeTrimming.cs
kuzminrobin: OK. (I don't know C#, looks like these are not member _vars_ but are member _functions_... OK)

cgranade: As a side note, C# does have the `const` keyword as well as `readonly`. In particular, `const` denotes a compile-time constant, while a `readonly` field can only be set during a constructor; after that constructor returns, the field can no longer be modified.

src/QsCompiler/Transformations/SyntaxTreeTrimming.cs
bettinaheim: Yes, I can. The behavior before introducing the QIR setup was to always keep intrinsics as is in the compilation, since the C# runtime relies on abstract classes being generated for all of the intrinsics, independent on whether they are used or not. The new interface based approach for C# emission (which has been implemented but we haven't fully enabled it yet), will remove that requirement I believe (@swernli for confirmation). 
Conversely, the C++ runtime does not rely on any declarations in the emitted llvm code, and it hence makes much more sense to only generate the declarations that are actually needed, such that it is possible to run code using a runtime with limited capabilities (e.g. our bigint support is still missing), as long as the program doesn't actually require anything that is not implemented. Hence, for qir emission, we monomorphize and trim the intrinsics as well.

src/QsCompiler/Compiler/PluginInterface.cs
bettinaheim: There are indeed no limitations aside from data type limitations; the idea is just to leave enough space in between for other transformations that are added (by us or by compiler extensions).

src/QsCompiler/Compiler/RewriteSteps/SyntaxTreeTrimming.cs
bettinaheim: Also, the returned boolean is used and indicates whether the step concluded successfully (see the doc comments on the IRewriteStep interface).

src/QsCompiler/Core/Dependencies.fs
```diff
@@ -41,6 +41,54 @@ type BuiltIn =
     static member RewriteStepDependencies =
         ImmutableHashSet.Create(BuiltIn.RangeReverse.FullName, BuiltIn.Length.FullName)
 
+    /// The set of all built in callables and attributes
```
bettinaheim, line 44: We had chatted offline about the desire for a less ad-hoc setup for rewrite step dependencies at some point.
I am just mentioning it here for the sake of anybody else who wasn't in the conversation. For now, this is ok as is.

src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/SyntaxTreeTrim.qs
```diff
@@ -0,0 +1,59 @@
...
+    operation UsedOp() : Unit { UsedFunc(); }
+    operation UnusedOp() : Unit { UnusedFunc(); }
+
+    function UsedFunc() : Unit { }
+    function UnusedFunc() : Unit { }
+}
+
+// =================================
+
+// Trimmer Keeps UDTs
```
bettinaheim, line 21: Does the trimmer trim types as well, or does it only trim the constructors? If it is the latter I'd suggest to rename the test to "trimmer keeps udt constructors" to avoid confusion. Since constructors are intrinsic, I assume they are only trimmed if the flag to trim intrinsics is set to true? It would be good to have a test for both cases, i.e. to add a test for the case when keepAllIntrinsics is set to false.

src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/SyntaxTreeTrim.qs
```diff
@@ -0,0 +1,59 @@
...
+        UsedIntrinsic();
+    }
+
+    operation UsedIntrinsic() : Unit { body intrinsic; }
+    operation UnusedIntrinsic() : Unit { body intrinsic; }
+}
+
+// =================================
+
+// Trimmer Removes Intrinsics When Told
```
bettinaheim, line 49: I'd suggest to just use the same code snippet above, and basically for all code snippets test them with both keepAllIntrinsics=true and keepAllIntrinsics=false.

bettinaheim (approved):
I was hoping for one more test, but other than that looks good to me.

src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/SyntaxTreeTrim.qs
ScottCarda-MS: Thanks for pointing this out. I had thought that the constructors for the types would get bundled tightly with the UDTs themselves so that when I filtered for only the callables in the transformation, I would filter out the UDTs along with their constructors. The test worked as intended so I thought this was correct, but you've now pointed out that the tests was really only working by accident, with the constructors being body intrinsic and filtered out that way. The intent is to not ever filter out UDTs or their constructors, so I will need to update the transformation to do that, and update the tests to correctly test for that.

src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/SyntaxTreeTrim.qs
ScottCarda-MS: I think the keepAllIntrinsics sub-feature is orthogonal to the other sub-features of this feature, and I believe unit tests should test orthogonal pieces separately.

As far as duplicating the Q# test case code, it is either duplicate the test case code, or duplicate the testing logic in the unit test. My preference here is to have a separate Q# test case for the two tests, even though the content is identical, because that is the way the testing framework is setup: each unit test consumes its associated test case, and having two unit tests consume and operate off of the same test case is doable, but because it deviates from what the setup expects, it requires more code and work on the unit test side. Long story short, I think it is simpler to have a test case with duplicate content here than to complicate the unit test logic to work two different angles off of the same test case.

src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/SyntaxTreeTrim.qs
ScottCarda-MS: Okay I've fixed the issue with the transformation, and updated the existing tests so that they won't accidently pass from this behavior. The keepAllIntrinsics sub-feature was turned off for all the tests that don't test for that, making for cleaner and stronger tests. I do believe that the keepAllIntrinsic sub-feature should be tests separately and independently, and the issue before was caused by having that feature intermixed with the other tests.
I've also updated the name here to reflect that both the UDTs and their constructors should always be kept.

src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/SyntaxTreeTrim.qs
bettinaheim: Ok.

src/QsCompiler/Transformations/SyntaxTreeTrimming.cs
cesarzc: Thanks for the explanation!

cesarzc (approved):

src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/SyntaxTreeTrim.qs
bettinaheim: Ok, I am not sure I entirely follow what you are saying here, but looking at the tests, I see that there is a test ``Trimmer Keeps Intrinsics When Told`` that checks the functionality of the keepAllIntrinsics flag. So I think you are just saying that you prefer a separate test case. Sounds good.

Issue #897, opened 2021-03-13
Update QIR emission to be compliant with QIR spec update 1


bettinaheim:
The current implementation is partially compliant with the update defined in https://github.com/microsoft/qsharp-language/pull/83. Still missing are

- [x] adjust code to reflect that qubit release also unreferences array
- [x] all count updates take and i32 as argument rather than an i64
- [x] string creation now only takes an i8* as argument
- [x] the constants Zero and One need to be accessed via call rather than via a global constant
- [x] change runtime function that is called to update the counts for captures

bettinaheim:
Implemented by PR https://github.com/microsoft/qsharp-compiler/pull/901.

PR #901, opened 2021-03-15
Updating QIR emission to comply with QIR spec update 1

bettinaheim:
The current emission is partially compliant with the new spec. This PR implement the remaining changes. 
It addresses the points listed in https://github.com/microsoft/qsharp-compiler/issues/897, and the merge needs to be coordinated with the corresponding change on the runtime repo (i.e. the build will fail until the runtime changes are implemented as well).


src/QsCompiler/QirGeneration/ScopeManager.cs
```diff
@@ -158,18 +158,21 @@ public void RegisterVariable(string varName, IValue value)
             /// and makes sure the release function is invoked before unreferending the value.
```
swernli, line 158: Now that release registration has been separated into its own function, this comment should be updated.

swernli (approved):
Looks good!

src/QsCompiler/QirGeneration/ScopeManager.cs
bettinaheim: Thanks! I missed that.

kuzminrobin (approved):

PR #902, opened 2021-03-15
Enable C# doc generation

kevinhartman:
This allows the C# compiler to validate XML documentation comments for proper syntax and code references, and also produces an XML doc file per assembly, which is packaged with Nuget automatically.

Specific warnings are always promoted to error in cases where there's no reason not to fix them the moment they're introduced. Other doc warnings will be raised as errors (like any other warning) when running `build.ps1`.

Warnings [CS1591](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs1591) and [CS1573](https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs1573) are always suppressed, since they require that documentation be fully complete.

This changeset also includes required doc comment fixes.


src/QsCompiler/CompilationManager/DataStructures.cs
```diff
@@ -29,10 +29,10 @@ internal class CodeLine
         /// lesser are considered "non-string" states.
         ///
         /// Here is an example of how these states relate to Q# text:
-        ///     let str = <NoOpenString> " <OpenString> Hello World" <NoOpenString>;
-        ///     let str2 = <NoOpenString> $" <OpenInterpolatedString> Hello
-        ///         { <OpenInterpolatedArgument> " <OpenStringInOpenInterpolatedArgument> World" <OpenInterpolatedArgument> }
-        ///         <OpenInterpolatedString> " <NoOpenString>;
+        ///     let str = [NoOpenString] " [OpenString] Hello World" [NoOpenString];
```
kevinhartman, line 32: @ScottCarda-MS, any preference on this change?

`<` and `>` cannot be used in XML comments, but they can be encoded as `&lt;` and `&gt;`, respectively. I opted for square brackets for in-source readability.

src/QsCompiler/QirGeneration/Context.cs
```diff
@@ -51,7 +51,7 @@ static GenerationContext()
 
         /// <summary>
         /// The module used for QIR generation.
-        /// Generated functions to facilitate interoperability are created in a separate <see cref="InteropModule"/>.
+        /// Generated functions to facilitate interoperability are created via <see cref="Interop"/>.
```
kevinhartman, line 54: @bettinaheim, is this the best correction? I see that `Context` itself also appears to expose methods to facilitate interop.

src/QsCompiler/CompilationManager/DataStructures.cs
ScottCarda-MS: Works fine for me. Thanks!

samarsha (approved):
Looks great, thanks!

src/QsCompiler/QirGeneration/Context.cs
bettinaheim: Yes, that sounds good. Thanks for finding it.

Issue #876, opened 2021-02-22
QIR generation fails for callables that require conditional rewrite


irinayat-MS:
The following Q# code compiles without errors if either `<QirGeneration>True</QirGeneration>` or `<ExecutionTarget>honeywell.qpu</ExecutionTarget>` is not set.
```
    operation NoArgs() : Unit
    is Adj+Ctl {
        body (...) {
            use q = Qubit();
            X(q);
            if M(q) != One { fail("Unexpected qubit state"); }
        }
        adjoint self;
        controlled (ctrls, ...) {
            use q = Qubit();
            Controlled X(ctrls, q);
            if M(q) != One { fail("Unexpected qubit state"); }
        }
    }

    operation TestFunctorsNoArgs() : Unit {
        NoArgs();
    }
```
And fails with error: "microsoft.quantum.sdk\0.15.2102128704-alpha\Sdk\Sdk.targets(148,5): error : Q# compiler failed with exit code: -13." otherwise.

bettinaheim:
I've look at issue and the root of it is that the Classical Control transformation creates TupleValues that have only one item while the expression validation upon creating TypedExpressions ensures that the expression kind matches the type (and a tuple with a single item does not exist at as a type). We need to either enforce that ValueTuples cannot have a single item in the typed checked data structures (which is a bit painful since we are using the same data structure for the unchecked syntax noted generated by parsing), or we need to adjust the generation to strip the tuple wrapping for tuples that have a single item.

bettinaheim:
Fixed with #906.

samarsha:
> We need to either enforce that ValueTuples cannot have a single item in the typed checked data structures (which is a bit painful since we are using the same data structure for the unchecked syntax nodes generated by parsing)

What about normalizing tuples at parse time? Then we could change `QsExpressionKind.ValueTuple` to have a type that required at least two elements.

samarsha:
> > We need to either enforce that ValueTuples cannot have a single item in the typed checked data structures (which is a bit painful since we are using the same data structure for the unchecked syntax nodes generated by parsing)
> 
> What about normalizing tuples at parse time? Then we could change `QsExpressionKind.ValueTuple` to have a type that required at least two elements.

@bettinaheim thoughts?

PR #906, opened 2021-03-15
Handling ValueTuples that contain a single item in QIR emission

bettinaheim:
This fixes https://github.com/microsoft/qsharp-compiler/issues/876.
I went with the path of making the QIR emission handle ValueTuples that contain a single item, rather than preventing the construction of expressions of that kind in the first place. Please note that the type of such tuples would not be a TupleType, but indeed is the item type (as it should, see [singleton tuple equivalence](https://github.com/microsoft/qsharp-language/blob/main/Specifications/Language/4_TypeSystem/SingletonTupleEquivalence.md#singleton-tuple-equivalence)). For the type, this behavior is enforced, while for the expression kind it isn't.

### DIFF
--- a/src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestTargeting.ll
+++ b/src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestTargeting.ll
@@ -1,0 +1,19 @@
+define void @Microsoft__Quantum__Testing__QIR__NoArgs__body() {
+entry:
+  %q = call %Qubit* @__quantum__rt__qubit_allocate()
+  %0 = call %Result* @__quantum__qis__mz(%Qubit* %q)
+  %1 = call %Tuple* @__quantum__rt__tuple_create(i64 mul nuw (i64 ptrtoint (i1** getelementptr (i1*, i1** null, i32 1) to i64), i64 2))
+  %2 = bitcast %Tuple* %1 to { %Callable*, %Qubit* }*
+  %3 = getelementptr inbounds { %Callable*, %Qubit* }, { %Callable*, %Qubit* }* %2, i32 0, i32 0
+  %4 = getelementptr inbounds { %Callable*, %Qubit* }, { %Callable*, %Qubit* }* %2, i32 0, i32 1
+  %5 = call %Callable* @__quantum__rt__callable_create([4 x void (%Tuple*, %Tuple*, %Tuple*)*]* @Microsoft__Quantum__Testing__QIR_____GUID___NoArgs, [2 x void (%Tuple*, i64)*]* null, %Tuple* null)
+  store %Callable* %5, %Callable** %3, align 8
+  store %Qubit* %q, %Qubit** %4, align 8
+  call void @Microsoft__Quantum__Simulation__QuantumProcessor__Extensions_____GUID___ApplyIfOne__body(%Result* %0, { %Callable*, %Qubit* }* %2)
+  call void @__quantum__rt__qubit_release(%Qubit* %q)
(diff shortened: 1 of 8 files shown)


bettinaheim:
@ScottCarda-MS When I was looking for the dependencies to copy over, I was surprised to see that one was missing in the ClassicalControl tests. It turns out that even after removing all of the ApplyIf* etc declarations, the tests still pass. That worries me a bit; is this expected? Why where they there in the first place?

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestTargeting.ll
```diff
@@ -0,0 +1,19 @@
...
+  %q = call %Qubit* @__quantum__rt__qubit_allocate()
+  %0 = call %Result* @__quantum__qis__mz(%Qubit* %q)
+  %1 = call %Tuple* @__quantum__rt__tuple_create(i64 mul nuw (i64 ptrtoint (i1** getelementptr (i1*, i1** null, i32 1) to i64), i64 2))
+  %2 = bitcast %Tuple* %1 to { %Callable*, %Qubit* }*
+  %3 = getelementptr inbounds { %Callable*, %Qubit* }, { %Callable*, %Qubit* }* %2, i32 0, i32 0
+  %4 = getelementptr inbounds { %Callable*, %Qubit* }, { %Callable*, %Qubit* }* %2, i32 0, i32 1
+  %5 = call %Callable* @__quantum__rt__callable_create([4 x void (%Tuple*, %Tuple*, %Tuple*)*]* @Microsoft__Quantum__Testing__QIR_____GUID___NoArgs, [2 x void (%Tuple*, i64)*]* null, %Tuple* null)
+  store %Callable* %5, %Callable** %3, align 8
+  store %Qubit* %q, %Qubit** %4, align 8
+  call void @Microsoft__Quantum__Simulation__QuantumProcessor__Extensions_____GUID___ApplyIfOne__body(%Result* %0, { %Callable*, %Qubit* }* %2)
```
samarsha, line 12: Where did the error occur before this change? Was it the `ApplyIfOne` call?

samarsha (approved):

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestTargeting.ll
bettinaheim: Yes, it was the argument to the call. The issue was that the argument to ApplyIfOne was a tuple that the ClassicalControl transformation generated the AST for. While the data structures of the compiler enforce that there cannot be any tuple *type* that has less than two arguments, it does not enforce the same for the expression. What was hence injected by ClassicalControll was a typed expression that was of type Qubit and of expression kind ValueTuple [Qubit]. The QIR emission then build a tuple but tried to pass it as Qubit*.

ScottCarda-MS:
> @ScottCarda-MS When I was looking for the dependencies to copy over, I was surprised to see that one was missing in the ClassicalControl tests. It turns out that even after removing all of the ApplyIf* etc declarations, the tests still pass. That worries me a bit; is this expected? Why where they there in the first place?

These were useful when I was manually running the test cases through the compiler, but are not needed now, although it is a bit awkward. The unit tests just test that the transformation did what it was supposed to do, which is to insert references to those operations. The names of the referenced operations are hardcoded based on the details of the conditional statement they are converting, so the transformation doesn't really need the declarations. Although in this case the transformed code will be erroneous without those declarations as there will be references to global callables that don't exist, that's not really what the tests are for.

If anything this just underlines the necessity of the preconditions for the rewrite steps. The classical control rewrite step's precondition tests for the existence of these declarations, to make sure the transformation is not injecting calls to things that aren't declared.

PR #907, opened 2021-03-16
Adding a command line option to force the execution of rewrite steps

bettinaheim:
The current behavior is that if the precondition verification fails for rewrite steps, then the step is skipped and the compilation succeeds with a warning. While overall this seems reasonable behavior given that rewrite steps may come from a 3rd party, I also found this rather suboptimal for testing: When I set up the test for qir emission for a targeted executable, the test could technically succeed when/because the targeting step is skipped (since we test for a specific .ll it will fail regardless, but still seems risky since it is easy to overlook that the transformation in question is not executed when e.g. updating the test). 
I hence went ahead and added a command line flag to force execution of rewrite steps and generate an error when the precondition is not met (see the doc comments for more detail).


ScottCarda-MS:
This should just be a matter of elevating the warning thrown by precondition failure to be treated as an error. That's a .NET thing right? I would be surprised if this issue needs changes to the code, but maybe @bettinaheim, you have already looked into that?

samarsha:
No, diagnostic severities are handled by the Q# compiler, not by .NET or MSBuild.

bettinaheim:
Incorporated into https://github.com/microsoft/qsharp-compiler/pull/915.

PR #909, opened 2021-03-17
dispose Utf8JsonWriter

filipw:
I noticed today that `Utf8JsonWriter` instance was never disposed, which actually generated CA2000 warnings


bettinaheim (approved):
Lgtm, thanks!

Issue #910, opened 2021-03-17
Unclear how to build qsharp-compiler from sources.


kuzminrobin:
From the page https://github.com/microsoft/qsharp-compiler/ it is not clear what to do after the `bootstrap.ps1` step in order to build from sources the QsCompiler (not the extension for VS or VSCode).
The `README.md` likely needs to mention 
`QsCompiler.sln` 
and the scripts that are used by the CI builds.

Also what are the instructions to build for Linux and Mac?

bettinaheim:
Started a draft here: https://github.com/microsoft/qsharp-compiler/pull/911.

PR #911, opened 2021-03-18
Updating readme

bettinaheim:
I took an initial stab at updating and hopefully improving the readme. I could use your input for how to best make it clearer how to work with the compiler repo.


swernli (approved):
These changes look good. Reading through the whole readme, the only addition I can think of that is missing is a section on executing tests. Maybe just a mention of `dotnet test QsCompiler.sln` and brief description of what successful test output looks like?

kuzminrobin:
You might want to add to the description of this PR "Resolves #910".
Looks good. Approving.

ScottCarda-MS (approved):

cesarzc (approved):

bettinaheim:
> You might want to add to the description of this PR "Resolves #910".
> Looks good. Approving.

Hm, I wasn't sure if it was fully addressing it; the one thing I would have added still is how to package the Sdk. 
Rather than pointing to the build scripts (which at this time I wouldn't want to suggest as the mail way to build things, since they are somewhat specific to our build pipeline) I would have given the instruction to use `nuget pack`, but it looks like some deprecated fields are still used such that we should update them to work with newer nuget versions.

README.md
```diff
@@ -7,13 +7,17 @@ as well as the Q# language server included in our [Visual Studio extension](http
...
+The Q# [compiler](./src/QsCompiler/Compiler) is distributed as a [NuGet package](https://www.nuget.org/packages/Microsoft.Quantum.Compiler), and the [CompilationLoader class](https://github.com/microsoft/qsharp-compiler/blob/main/src/QsCompiler/Compiler/CompilationLoader.cs) exposes the different configuration options for building a compilation.
+The Q# [command line compiler](./src/QsCompiler/CommandLineTool) is included as a tool in the [Microsoft.Quantum.Sdk](./src/QuantumSdk) and provides an [extensibility mechanism](https://devblogs.microsoft.com/qsharp/extending-the-q-compiler/) for compilation steps. See the list of [project properties](./src/QuantumSdk#defined-project-properties) for more information about possible configuration options for Q# projects.
 
 - **[QsCompiler](./src/QsCompiler/)**: Q# compiler including the command line tool
 - **[QsCompiler/LanguageServer](./src/QsCompiler/LanguageServer/)**: Q# language server
+- **[Microsoft.Quantum.Sdk](./src/QuantumSdk)**: Sdk for building Q# projects and support for compiler extensions
```
filipw, line 15: very helpful update. I would also suggest that since you mention "compiler extensions" here, might be good to link to the [example](https://github.com/microsoft/qsharp-compiler/tree/main/examples/CompilerExtensions) too?

bettinaheim:
Now with the latest changes, I think it actually fixes https://github.com/microsoft/qsharp-compiler/issues/910.

PR #913, opened 2021-03-19
Create the base for the C++ entry-point QIR driver generator

cesarzc:
This change creates the base for the C++ entry-point QIR driver generator, including tests.

It uses [T4 Text Templates](https://docs.microsoft.com/en-us/visualstudio/modeling/code-generation-and-t4-text-templates?view=vs-2019) to generate C++ source code.


cesarzc:
This [document](https://docs.microsoft.com/en-us/visualstudio/modeling/code-generation-in-a-build-process?view=vs-2019) describes that T4 build tasks need Visual Studio installed. Since build agents do not have VS installed, to enable the use of T4 templates we integrate the generated C# file (_src/QsCompiler/Compiler/Templates/QirDriverCpp.cs_) to version control. This is the same approach we use with Bond schemas since the Bond compiler is not multi-platform.

src/QsCompiler/Tests.Compiler/TestCases/QirEntryPointTests/TestUnitReceivesBool.cpp
```diff
@@ -0,0 +1,11 @@
...
+// This code was generated by a tool.
+// <auto-generated />
+//------------------------------------------------------------------------------
+
+extern "C" void UseBoolArg(); // NOLINT
+
+int main(int argc, char* argv[])
+{
+    return 0;
+}
```
kuzminrobin, line 11: 1. Would be nice to have some more comments (in the code or in the PR) about why we need an auto-generated file in the repo.
2. Rather curious that a function is declared `UseBoolArg()` that is not called by anyone.

src/QsCompiler/Tests.Compiler/TestCases/QirEntryPointTests/TestUnitReceivesBool.json
```diff
@@ -0,0 +1,6 @@
+﻿{
+  "Name": "UseBoolArg",
+  "Arguments": [
+    { "Name": "BoolArg" }
+  ]
```
kuzminrobin, line 5: Looks like the "Arguments" are not use by anybody. Only the "Name".
Is the "Arguments" section a plan for the future?

kuzminrobin (approved):

src/QsCompiler/Tests.Compiler/TestCases/QirEntryPointTests/TestUnitReceivesBool.cpp
cesarzc: Regarding the first comment. This is a test case file that represents the expected generated C++ code. It is read by the test at _src/QsCompiler/Tests.Compiler/EntryPointGenerationTests.fs_ which compares it to the actual generated C++ code using _src/QsCompiler/Tests.Compiler/TestUnitReceivesBool.json_ as input.

Regarding the second comment, this PR is just intends to setup a base to start the implementation of the C++ entry-point driver generator. Next week, Sam will be picking this up to complete the implementation. I'll add a comment to the generated code to make this clear for now.

src/QsCompiler/Tests.Compiler/TestCases/QirEntryPointTests/TestUnitReceivesBool.json
cesarzc: The goal of this PR is just to setup a base to start the implementation of the C++ entry-point driver generator. It currently just uses the name of the entry-point operation to show an example but Arguments will be used to generate the rest of the C++ code. Sam will be working on this starting next week.

src/QsCompiler/Tests.Compiler/TestCases/QirEntryPointTests/TestUnitReceivesBool.cpp
cesarzc: Added comment clarifying the current state of C++ code generation.

PR #914, opened 2021-03-21
Separate entry point that prints the results

bettinaheim:
In addition to the interop wrapper that is created for entry points, this PR also adds another IrFunction that has the same argument list as the interop wrapper, calls into the entry point, but instead of returning the result, uses message to print it instead, and returns void.
This PR adds the missing support to process tuples, udts, and arrays in interpolated strings.
It also fixes an issue with diagnostics from rewrite steps that seems to be a result of a recent package update of the language server package (and some inconsistent nullable information in that package).

The build on this PR is currently failing due to a missing API function in the runtime.


src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
```diff
@@ -386,6 +386,252 @@ IValue CopyAndUpdateUdt(TupleValue originalValue)
             return ResolvedExpressionKind.InvalidExpr;
         }
 
+        /// <summary>
+        /// Creates a string literal that contains the given values as interpolation arguments.
+        /// The given string is expected to consist of text interspaced with integer constants between curly brackets.
+        /// The content between non-escaped curly brackets is parsed as integer,
+        /// and replaced with the string representation of the interpolation argument at that index.
+        /// Registers the built string with the scope manager.
+        /// </summary>
```
kuzminrobin, line 395: Consider adding an example (and maybe even starting with an example, such that the reader understands the most of the idea in a few seconds).
By looking just at the summary it is somewhat hard to understand, without an example, for a new/unaware reader.

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
```diff
@@ -386,6 +386,252 @@ IValue CopyAndUpdateUdt(TupleValue originalValue)
...
+                        if (j < 0)
+                        {
+                            throw new FormatException("Missing } in interpolated string");
+                        }
+                        var n = int.Parse(s[(i + 1)..j]);
+                        return (i, j + 1, n);
+                    }
+                    start = i + 1;
+                }
+            }
```
kuzminrobin, line 419: (I don't know C# and I can raise a false suspicion)
If the string `s` ends with `{` then after the call `var i = s.IndexOf('{', start)` the var `i` will be indexing the last char of `s`.
The subsequent call `var j = s.IndexOf('}', i + 1)` will try to access the `s[i + 1]` (which can be a character outside of the string `s`), and the topic [`IndexOf(Char, Int32)`](https://docs.microsoft.com/en-us/dotnet/api/system.string.indexof?view=net-5.0#System_String_IndexOf_System_Char_System_Int32_) makes me think that the `ArgumentOutOfRangeException` will be thrown. 
If that's the case then I just wanted to make sure that your code handles that case.

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
```diff
@@ -386,6 +386,252 @@ IValue CopyAndUpdateUdt(TupleValue originalValue)
...
+                    var i = s.IndexOf('{', start);
+                    if (i < 0)
+                    {
+                        return (-1, s.Length, -1);
+                    }
+                    else if ((i == start) || (s[i - 1] != '\\'))
+                    {
+                        var j = s.IndexOf('}', i + 1);
+                        if (j < 0)
+                        {
```
kuzminrobin, line 411: For the future, my own subjective opinion, feel free to ignore. I can also be looking at the wrong links/topics.

The topic [`IndexOf(Char, Int32)`](https://docs.microsoft.com/en-us/dotnet/api/system.string.indexof?view=net-5.0#System_String_IndexOf_System_Char_System_Int32_) tells me that the call **returns -1** if the char is not found.
The code here checks the returned values like this: `if (i < 0)`, `if (j < 0)`.

If in the future the .NET API changes, and the function `IndexOf(Char, Int32)` starts returning (for certain cases) -2, -3, -4, and so on, then the checks `if (i < 0)`, `if (j < 0)` will start working in the unexpected way. The refactoring will be needed, that refactoring may be huge.

For the future code I recommend checking exactly against the value described/promised in the documentation:
`if (i == -1)`, `if (j == -1)`.
The .NET documentation writers are _free to make such a change_ (to start returning -2, -3), because the only thing that they promised is `-1` (not `< 0`).

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
```diff
@@ -386,6 +386,252 @@ IValue CopyAndUpdateUdt(TupleValue originalValue)
...
+                            throw new FormatException("Missing } in interpolated string");
+                        }
+                        var n = int.Parse(s[(i + 1)..j]);
+                        return (i, j + 1, n);
```
kuzminrobin, line 415: Minor:
Just in case I want to make sure that your code, upon the call `int.Parse(..)`, correctly handles the cases when
`s` contains "{}" (char `}` immediately follows `{`), i.e. `i + 1` is equal to `j` in the fragment `s[(i + 1)..j]`;
`s` contains "{\<non Int>}", i.e. there is some identifier, or double value, etc. between the `{` and `}`.

I expect the call [`Parse(String)`](https://docs.microsoft.com/en-us/dotnet/api/system.int32.parse?view=net-5.0#System_Int32_Parse_System_String_) to throw certain exceptions in that case. So I wanted to make sure that those exceptions are handled properly.

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
```diff
@@ -386,6 +386,252 @@ IValue CopyAndUpdateUdt(TupleValue originalValue)
...
+            {
+                while (true)
+                {
+                    var i = s.IndexOf('{', start);
+                    if (i < 0)
+                    {
+                        return (-1, s.Length, -1);
+                    }
+                    else if ((i == start) || (s[i - 1] != '\\'))
+                    {
```
kuzminrobin, line 408: Just in case I want to make sure that the code correctly handles the case when `s` contains `"..\\{.."`, i.e. the backslash `\` is escaped, but the brace `{` is not escaped, and should be handled/parsed.

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
```diff
@@ -386,6 +386,252 @@ IValue CopyAndUpdateUdt(TupleValue originalValue)
...
+                    : sharedState.CurrentBuilder.BitCast(
+                        constantArray,
+                        sharedState.Types.DataArrayPointer);
+
+                var createString = sharedState.GetOrCreateRuntimeFunction(RuntimeLibrary.StringCreate);
+                return sharedState.CurrentBuilder.Call(createString, zeroLengthString);
+            }
+
+            // Creates a new string with reference count 1 that needs to be queued for unreferencing
+            // and contains the concatenation of both values. Both both arguments are unreferenced.
```
kuzminrobin, line 452: > **Both both** arguments

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
```diff
@@ -386,6 +386,252 @@ IValue CopyAndUpdateUdt(TupleValue originalValue)
...
+            Value DoAppend(Value? curr, Value next, bool unreferenceNext = true)
+            {
+                if (curr == null)
+                {
+                    return next;
+                }
```
kuzminrobin, line 458: Just in case I want to make sure that nothing else is needed here depending on the value of `unreferenceNext`.

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
```diff
@@ -386,6 +386,252 @@ IValue CopyAndUpdateUdt(TupleValue originalValue)
...
+                    var tupleElements = tuple.GetTupleElements();
+                    Value? comma = null;
+
+                    for (var idx = 0; idx < tupleElements.Length; ++idx)
+                    {
+                        str = DoAppend(str, ExpressionToString(tupleElements[idx]));
+                        var isLast = idx == tupleElements.Length - 1;
+                        var next = isLast
+                            ? CreateConstantString(")")
+                            : comma ?? CreateConstantString(",");
```
kuzminrobin, line 502: Minor:
(I don't know C#, I can be wrong)
I have an impression that `comma` is always `null` here.

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
```diff
@@ -386,6 +386,252 @@ IValue CopyAndUpdateUdt(TupleValue originalValue)
...
+                    var arr = (ArrayValue)evaluated;
+                    var str = CreateConstantString("[");
+                    Value comma = CreateConstantString(",");
+                    Value closeParens = CreateConstantString("]");
+
+                    sharedState.IterateThroughArray(arr, item =>
+                    {
+                        str = DoAppend(str, ExpressionToString(item));
+                        str = DoAppend(str, comma, unreferenceNext: false);
+                    });
```
kuzminrobin, line 574: I have an impression that this fragment inserts comma `,` after the last element and before `]`.
If that happens, then just in case I wanted to make sure that such an insertion is OK.

kuzminrobin:
C# is like Chinese for me. I don't understand most of it.
I looked at `ExpressionKindTransformation.cs` mostly (it contained a continuous block of code that seemed similar to C/C++).
All my notes can be false alarms. Feel free to ignore.
Approving.

kuzminrobin (approved):
C# is like Chinese for me. I don't understand most of it.
I looked at `ExpressionKindTransformation.cs` mostly (it contained a continuous block of code that seemed similar to C/C++).
All my notes can be false alarms. Feel free to ignore.
Approving.

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
bettinaheim: The idea in general is that rewrite steps (such as the entire QIR emission) can rely on the correctness of the AST; if we didn't do that, then each rewrite step would have to do a huge amount of validation and error handling. I believe it is hence better to assign the responsibility of doing that to the compiler, and each rewrite step in turn has the responsibility to generate valid AST if it does an AST transformation. One tool that would be nice to add (which we haven't had the time to build yet) is a separate tool for AST validation. While such a tool is not (yet) available, I believe it is still fine for rewrite steps to rely on correctness, since the compiler would at least do some proper error handling when a rewrite step throws, such that issues like that can be traced to which step is faulty with a little bit of work.

bettinaheim: And to make it explicit: The format of an interpolated string is expected to be valid in that each open { is closed, and the content between {} is expected to be a 32-bit integer.

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
bettinaheim: Throwing an exception in this case is expected, and acceptable, see my comment above regarding relying on the correctness of the AST.

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
bettinaheim: Thanks for the catch! I switched it to enter the block if there is an even number of backslashes before the {.

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
bettinaheim: No, nothing else is needed, since we return next; basically next takes the place of what would otherwise be a new string and we hence don't want to release it.

bettinaheim: Actually, while the code is correct for how it is used, the correct behavior for the function is to increase the reference count for next if unreferenceNext is false.

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
bettinaheim: Eh, yes; clearly I've been very distracted writing this code...! Thanks for the proper review; it was sorely needed.

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
bettinaheim: It does. I was on the fence about this. I could generate a conditional branch to not do that, but thinking of automatic parsing tools and that it is not uncommon for larger sequences to list them across multiple lines with each item being followed by a comma including the last one (making it easier to append), I opted for going with the more readable version of just leaving it there.

bettinaheim:
> C# is like Chinese for me. I don't understand most of it.
> I looked at `ExpressionKindTransformation.cs` mostly (it contained a continuous block of code that seemed similar to C/C++).
> All my notes can be false alarms. Feel free to ignore.
> Approving.

Thanks for the review; it was very useful and very much needed.

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
bettinaheim: I added an example string, not sure if that's what you had in mind.

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
bettinaheim: I changed it to -1. In general I agree that it is good to rely only on documented behavior and to not be "permissive" when writing code (trying to continue when dealing with an unexpected state/receiving unexpected input). In this specific case, anything negative will now result in an exception, since we are accessing an item at that index.

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
kuzminrobin: ok. 
(I assume that "content between {} is expected to be _unsigned_ 32-bit integer.")

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
kuzminrobin: I typically prefer to handle the first element separately.
And then, if there are more elements, add comma and one more element, comma and one more element, .. on each new iteration.
(if there is a chance that comma after the last element can cause a problem)

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
bettinaheim: Yes.

PR #915, opened 2021-03-22
Attaching QIR bitcode as resource in dll

bettinaheim:
This PR modifies the qir emission setup such that whenever qir is generated, we generate both the human readable version as well as the bitcode for it, and attach the bitcode as resource to the dll instead of the bson.
Incorporates the changes in https://github.com/microsoft/qsharp-compiler/pull/907.


bettinaheim:
abandoned in favor to #923

PR #884, opened 2021-03-02
Miscellaneous changes for HM type inference and QEP 2

samarsha:
* Add parsing for empty array literals.
* Support sized array literals in attribute arguments.
* Replace warning when a type parameter doesn't occur in the argument with a warning when the type parameter is unused (in both the argument and the return types).
* Add tests for integral, iterable, numeric, semigroup, and unwrap constraints.

PR #917, opened 2021-03-22
Add Hindley-Milner type inference module

samarsha:
This PR splits out the HM type inference module from #884, as well as additional changes that the module requires. The new module is not used by the compiler, so there's no change to type inference behavior yet (there will be another PR for that).

Additional changes:

* Added a range field to `ResolvedType` so that the HM module can generate diagnostics with accurate locations.
* Split ScopeTools.fs into two files so that the HM module can reference SymbolTracker, and ScopeContext can reference the HM module.
* Removed `WithinIfCondition` field from `ScopeContext` that has been deprecated for 6 months.

### DIFF
--- a/src/QsCompiler/BondSchemas/BondSchemaTranslator.cs
+++ b/src/QsCompiler/BondSchemas/BondSchemaTranslator.cs
@@ -536,9 +536,9 @@ namespace Microsoft.Quantum.QsCompiler.BondSchemas
             {
                 Origin = qsTypeParameter.Origin.ToBondSchema(),
                 TypeName = qsTypeParameter.TypeName,
-                Range = qsTypeParameter.Range.IsNull ?
-                    null :
-                    qsTypeParameter.Range.Item.ToBondSchema()
+#pragma warning disable 618 // QsTypeParameter.Range is obsolete.
+                Range = qsTypeParameter.Range.IsNull ? null : qsTypeParameter.Range.Item.ToBondSchema()
+#pragma warning restore 618
             };
 
         private static QsValueUpdate ToBondSchema(this SyntaxTree.QsValueUpdate valueUpdate) =>
@@ -675,9 +675,9 @@ namespace Microsoft.Quantum.QsCompiler.BondSchemas
             {
                 Namespace = userDefinedType.Namespace,
                 Name = userDefinedType.Name,
-                Range = userDefinedType.Range.IsNull ?
-                    null :
-                    userDefinedType.Range.Item.ToBondSchema()
+#pragma warning disable 618 // UserDefinedType.Range is obsolete.
+                Range = userDefinedType.Range.IsNull ? null : userDefinedType.Range.Item.ToBondSchema()
+#pragma warning restore 618
             };
 
         private static CharacteristicsKindComposition<TBond> ToBondSchemaGeneric<TBond, TCompiler>(
--- a/src/QsCompiler/BondSchemas/BondSchemaTranslator.cs
+++ b/src/QsCompiler/BondSchemas/BondSchemaTranslator.cs
@@ -536,9 +536,7 @@ namespace Microsoft.Quantum.QsCompiler.BondSchemas
             {
                 Origin = qsTypeParameter.Origin.ToBondSchema(),
                 TypeName = qsTypeParameter.TypeName,
-#pragma warning disable 618 // QsTypeParameter.Range is obsolete.
-                Range = qsTypeParameter.Range.IsNull ? null : qsTypeParameter.Range.Item.ToBondSchema()
-#pragma warning restore 618
+                Range = null
             };
 
         private static QsValueUpdate ToBondSchema(this SyntaxTree.QsValueUpdate valueUpdate) =>
@@ -675,9 +673,7 @@ namespace Microsoft.Quantum.QsCompiler.BondSchemas
             {
                 Namespace = userDefinedType.Namespace,
                 Name = userDefinedType.Name,
-#pragma warning disable 618 // UserDefinedType.Range is obsolete.
-                Range = userDefinedType.Range.IsNull ? null : userDefinedType.Range.Item.ToBondSchema()
-#pragma warning restore 618
+                Range = null
             };
 
         private static CharacteristicsKindComposition<TBond> ToBondSchemaGeneric<TBond, TCompiler>(
--- a/src/QsCompiler/Core/Transformations/TypeTransformation.fs
+++ b/src/QsCompiler/Core/Transformations/TypeTransformation.fs
@@ -144,5 +144,5 @@ type TypeTransformationBase(options: TransformationOptions) =
                 | ExpressionType.Pauli -> this.OnPauli()
                 | ExpressionType.Range -> this.OnRange()
 
-            let ResolvedType t = ResolvedType.New(true, t)
-            ResolvedType |> Node.BuildOr t transformed
+            let range = this.OnRangeInformation t.Range
+            Node.BuildOr t transformed (ResolvedType.create range)
--- a/src/QsCompiler/Core/Transformations/TypeTransformation.fs
+++ b/src/QsCompiler/Core/Transformations/TypeTransformation.fs
@@ -5,7 +5,9 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.Core
 
 #nowarn "44" // TypeParameter.Range and UserDefinedType.Range are deprecated.
 
+open System
 open System.Collections.Immutable
+
 open Microsoft.Quantum.QsCompiler.DataTypes
 open Microsoft.Quantum.QsCompiler.SyntaxExtensions
 open Microsoft.Quantum.QsCompiler.SyntaxTokens
@@ -16,15 +18,19 @@ type private ExpressionType = QsTypeKind<ResolvedType, UserDefinedType, QsTypePa
 
 
 type TypeTransformationBase(options: TransformationOptions) =
-
     let Node = if options.Rebuild then Fold else Walk
-    new() = new TypeTransformationBase(TransformationOptions.Default)
+
+    new() = TypeTransformationBase TransformationOptions.Default
 
 
     // supplementary type information
 
+    [<Obsolete "Use OnRangeInformation(TypeRange) instead.">]
     abstract OnRangeInformation: QsNullable<Range> -> QsNullable<Range>
     default this.OnRangeInformation range = range
+
+    abstract OnTypeRange: TypeRange -> TypeRange
+    default this.OnTypeRange range = range
 
     abstract OnCharacteristicsExpression: ResolvedCharacteristics -> ResolvedCharacteristics
     default this.OnCharacteristicsExpression fs = fs
@@ -123,12 +129,7 @@ type TypeTransformationBase(options: TransformationOptions) =
         if not options.Enable then
             t
         else
-            let range =
-                match t.Range with
-                | Annotated range -> Value range |> this.OnRangeInformation |> QsNullable<_>.Map Annotated
-                | Inferred range -> Value range |> this.OnRangeInformation |> QsNullable<_>.Map Inferred
-                | TypeRange.Generated -> Null
-                |> QsNullable.defaultValue TypeRange.Generated
+            let range = this.OnTypeRange t.Range
 
             let transformed =
                 match t.Resolution with
--- a/src/QsCompiler/Core/Transformations/TypeTransformation.fs
+++ b/src/QsCompiler/Core/Transformations/TypeTransformation.fs
@@ -25,6 +25,7 @@ type TypeTransformationBase(options: TransformationOptions) =
 
     // supplementary type information
 
+    // TODO: RELEASE 2021-10: Remove obsolete method.
     [<Obsolete "Use OnRangeInformation(TypeRange) instead.">]
     abstract OnRangeInformation: QsNullable<Range> -> QsNullable<Range>
     default this.OnRangeInformation range = range
--- a/src/QsCompiler/Core/Transformations/TypeTransformation.fs
+++ b/src/QsCompiler/Core/Transformations/TypeTransformation.fs
@@ -123,6 +123,13 @@ type TypeTransformationBase(options: TransformationOptions) =
         if not options.Enable then
             t
         else
+            let range =
+                match t.Range with
+                | Annotated range -> Value range |> this.OnRangeInformation |> QsNullable<_>.Map Annotated
+                | Inferred range -> Value range |> this.OnRangeInformation |> QsNullable<_>.Map Inferred
+                | TypeRange.Generated -> Null
+                |> QsNullable.defaultValue TypeRange.Generated
+
             let transformed =
                 match t.Resolution with
                 | ExpressionType.UnitType -> this.OnUnitType()
@@ -143,12 +150,5 @@ type TypeTransformationBase(options: TransformationOptions) =
                 | ExpressionType.Result -> this.OnResult()
                 | ExpressionType.Pauli -> this.OnPauli()
                 | ExpressionType.Range -> this.OnRange()
-
-            let range =
-                match t.Range with
-                | Annotated range -> Value range |> this.OnRangeInformation |> QsNullable<_>.Map Annotated
-                | Inferred range -> Value range |> this.OnRangeInformation |> QsNullable<_>.Map Inferred
-                | TypeRange.Generated -> Null
-                |> QsNullable.defaultValue TypeRange.Generated
 
             ResolvedType.create range |> Node.BuildOr t transformed
--- a/src/QsCompiler/Core/Transformations/TypeTransformation.fs
+++ b/src/QsCompiler/Core/Transformations/TypeTransformation.fs
@@ -144,5 +144,11 @@ type TypeTransformationBase(options: TransformationOptions) =
                 | ExpressionType.Pauli -> this.OnPauli()
                 | ExpressionType.Range -> this.OnRange()
 
-            let range = this.OnRangeInformation t.Range
-            Node.BuildOr t transformed (ResolvedType.create range)
+            let range =
+                match t.Range with
+                | Annotated range -> Value range |> this.OnRangeInformation |> QsNullable<_>.Map Annotated
+                | Inferred range -> Value range |> this.OnRangeInformation |> QsNullable<_>.Map Inferred
+                | TypeRange.Generated -> Null
+                |> QsNullable.defaultValue TypeRange.Generated
+
+            ResolvedType.create range |> Node.BuildOr t transformed
--- a/src/QsCompiler/Core/Transformations/TypeTransformation.fs
+++ b/src/QsCompiler/Core/Transformations/TypeTransformation.fs
@@ -3,6 +3,8 @@
 
 namespace Microsoft.Quantum.QsCompiler.Transformations.Core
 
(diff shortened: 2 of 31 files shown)


src/QsCompiler/SyntaxProcessor/ScopeContext.fs
```diff
@@ -0,0 +1,66 @@
...
+        IsInOperation: bool
+
+        /// The return type of the parent callable for the current scope.
+        ReturnType: ResolvedType
+
+        /// The runtime capability of the compilation unit.
+        Capability: RuntimeCapability
+
+        /// The name of the processor architecture for the compilation unit.
+        ProcessorArchitecture: string
```
swernli, line 34: I know this was there in the previous version and you are just preserving it, but why do we need the processor architecture for syntax processing?

swernli (approved):
Changes look good! The InferenceContext.fs file has a lot of logic that is hard to internalize when coming at the code fresh, so I'm wondering if it might be worth having a readme in the TypeInference folder with some brief descriptions of the high level concepts/strategy and a link to external info on Hindley-Milner type inference.

src/QsCompiler/SyntaxProcessor/ScopeContext.fs
samarsha: It's used only for diagnostic messages, so they can include the name of the quantum processor in capability errors.

swernli: Ah, ok. That does make sense. I have been so caught up in working on clang, LLVM, and CMakeLists.txt setup that I misinterpreted this as the classical processor architecture that the code was compiled on, which for Q# didn't seem relevant at all. Thanks for explaining!

samarsha:
@swernli I added a readme. Can you take a look at it?

src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -7,6 +7,9 @@ open System
 open System.Collections.Generic
 
 type ErrorCode =
+    | TypeMismatch = 1
```
cesarzc, line 10: I wonder if errors starting at 1001 was a conscious decision to show all Q# compilation errors look similar (e.g. _error QS2003_). Would it make sense to start type errors at 4001?

samarsha: Yeah, the error codes I picked are unusual compared to the existing ones. I think there are two separate questions:

1. Should error codes start at 0001 or 1001?
2. Are there groups (first digit, first two digits, etc.) of error codes for different kinds of errors?

For question 1, I would rather format error codes with leading zeroes than require that all error codes start at 1001, to avoid error code inflation. For comparison, [FS0001](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/compiler-messages/fs0001) is the F# error code for type mismatch.

For question 2, I'm not sure that there is a consistent grouping of error codes by leading digits. For example, ExpectingIntExpr and ArgumentTypeMismatch are both type errors, but ExpectingIntExpr is QS5012 and ArgumentTypeMismatch is QS6210. ExpectingSemicolon and ExpectingIdentifier are both syntax errors, but ExpectingSemicolon is QS2003 and ExpectingIdentifier is QS6103.

swernli:
@samarsha The readme is fantastic! Concise and still very informative for interpreting the accompanying code.

src/QsCompiler/DataStructures/Diagnostics.fs
cesarzc: I agree with you on both answers. Just before closing this conversation, with this change if there is a TypeMismatch error would the user see QS0001? When the error is formatted to be shown to the user is it adding the leading zeroes?

samarsha: If there is a pattern to error code number groups, then I think it would help to have comments that describe each group, since I don't think that they are very clear from context. I think @bettinaheim might know what the groups are?

samarsha: > Just before closing this conversation, with this change if there is a TypeMismatch error would the user see QS0001? When the error is formatted to be shown to the user is it adding the leading zeroes?

Currently, it's formatted as QS1, but I can change it now to add leading zeroes.

src/QsCompiler/SyntaxProcessor/TypeInference/Constraint.fs
```diff
@@ -0,0 +1,37 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.QsCompiler.SyntaxProcessing.TypeInference
+
+open Microsoft.Quantum.QsCompiler.SyntaxTokens
+open Microsoft.Quantum.QsCompiler.SyntaxTree
+
+type internal Constraint =
```
cesarzc, line 9: I am a curious about why the `Constraint` type is defined in two separate source files?

samarsha: Constraint.fsi is the signature file for Constraint.fs. In general, the signature file contains only the types of things in the implementation file (so Constraint.fsi has `val types: Constraint -> ResolvedType list`, while Constraint.fs has `let types = ...`).

For discriminated unions and records with public or internal cases/fields, the signature and implementation are actually the same, but it's still required to provide them separately, so unfortunately they are duplicated across both files.

src/QsCompiler/DataStructures/Diagnostics.fs
samarsha: It's formatted with leading zeroes now (QS0001).

src/QsCompiler/SyntaxProcessor/TypeInference/InferenceContext.fsi
```diff
@@ -0,0 +1,56 @@
...
+
+    /// <summary>
+    /// Creates a fresh type parameter originating from the given <paramref name="source"/> range.
+    /// </summary>
+    member internal Fresh: source:Range -> ResolvedType
+
+    /// <summary>
+    /// Unifies the <paramref name="expected"/> type with the <paramref name="actual"/> type.
+    /// </summary>
+    member internal Unify: expected:ResolvedType * actual:ResolvedType -> QsCompilerDiagnostic list
```
bettinaheim, line 33: Hm, what confuses me a bit is why you don't have a method for unification that computes the more general type of the two, and one that computes the less general type of the two ("covariant vs contravariant unification"). How does that work currently?

src/QsCompiler/Core/Transformations/TypeTransformation.fs
```diff
@@ -144,5 +144,5 @@ type TypeTransformationBase(options: TransformationOptions) =
                 | ExpressionType.Pauli -> this.OnPauli()
                 | ExpressionType.Range -> this.OnRange()
 
-            let ResolvedType t = ResolvedType.New(true, t)
-            ResolvedType |> Node.BuildOr t transformed
+            let range = this.OnRangeInformation t.Range
```
bettinaheim, line 147: Please make sure the order in which individual items in a node are processed matches other transformations; For the other nodes, ranges are processed before the kinds and the types.

bettinaheim (changes requested):
I am blocking the PR for now since I believe in its current state it would break some of the IDE commands due to the change in assigning ranges to types even when there is no matching piece of source code.

src/QsCompiler/SyntaxProcessor/TypeInference/InferenceContext.fsi
samarsha: That is `UnifyByOrdering`.

https://github.com/microsoft/qsharp-compiler/blob/70f506f9ad6b14554572bcdccba4485a4daf821c/src/QsCompiler/SyntaxProcessor/TypeInference/InferenceContext.fs#L362-L366

src/QsCompiler/SyntaxProcessor/SymbolTracker.fs
```diff
@@ -299,7 +297,7 @@ type SymbolTracker(globals: NamespaceManager, sourceFile, parent: QsQualifiedNam
             let argType, returnType =
                 decl.ArgumentType |> StripPositionInfo.Apply, decl.ReturnType |> StripPositionInfo.Apply
 
-            let idType = kind ((argType, returnType), decl.Information) |> ResolvedType.New
+            let idType = kind ((argType, returnType), decl.Information) |> ResolvedType.create qsSym.Range
```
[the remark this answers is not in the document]
samarsha: To clarify based on our other conversation, the PR didn't cause issues with IDE commands since it left alone the ranges in QsTypeParameter and UserDefinedType. However, having a range for types in multiple places was strange, so I updated the range in ResolvedType to distinguish between an inferred type and an annotated type, and marked the ranges for TPs and UDTs as obsolete.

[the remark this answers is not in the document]
samarsha: Currently, the IDE commands are still using the obsolete ranges instead of the new ResolvedType.Range, but I'd like to update that in another PR to avoid making this one too big.

src/QsCompiler/DataStructures/SyntaxTree.fs
```diff
@@ -342,66 +377,158 @@ type CallableInformation =
...
+    /// <summary>
+    /// Creates an <see cref="Inferred"/> type range if a range is given, or the <see cref="Generated"/> type range
+    /// otherwise.
+    /// </summary>
+    [<CompiledName "Inferred">]
+    let inferred = QsNullable<_>.Map Inferred >> QsNullable.defaultValue Generated
+
+    /// Returns the range if it exists.
+    [<CompiledName "TryRange">]
+    let tryRange =
```
bettinaheim, line 407: nit: TryGetRange to me is a bit more comprehensive than TryRange.

src/QsCompiler/Core/Transformations/TypeTransformation.fs
```diff
@@ -123,6 +123,13 @@ type TypeTransformationBase(options: TransformationOptions) =
         if not options.Enable then
             t
         else
+            let range =
```
bettinaheim, line 126: Hm, to make the change in behavior clearer, I would deprecate OnRangeInformation, and instead make separate methods to overload for annotated and inferred ranges. I think that will make it clearer that and how transformations need to be updated. Do you see any downsides to that?

src/QsCompiler/SyntaxProcessor/TreeVerification.fs
```diff
@@ -3,6 +3,8 @@
 
 module Microsoft.Quantum.QsCompiler.SyntaxProcessing.SyntaxTree
 
+#nowarn "44" // UserDefinedType.Range is deprecated.
```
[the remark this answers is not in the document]
bettinaheim: Finally my NuGet packages have restored and I can build things again locally; It looks like this can be removed.

bettinaheim (commented):
I think this setup make things a lot cleaner, thanks. I am still waiting for a response regarding SearchAndReplace; we need to update the transformation to not use the deprecated handles. The syntax tree transformation currently walks both the old and the new handle, which I believe would cause issues.

src/QsCompiler/BondSchemas/BondSchemaTranslator.cs
```diff
@@ -536,9 +536,9 @@ private static QsTypeItem ToBondSchema(this SyntaxTree.QsTypeItem qsTypeItem)
             {
                 Origin = qsTypeParameter.Origin.ToBondSchema(),
                 TypeName = qsTypeParameter.TypeName,
-                Range = qsTypeParameter.Range.IsNull ?
-                    null :
-                    qsTypeParameter.Range.Item.ToBondSchema()
+#pragma warning disable 618 // QsTypeParameter.Range is obsolete.
```
bettinaheim, line 539: Actually, why  do we need to keep this? Isn't this what gets serialized; i.e. can't we drop this and just make sure that the corresponding item in already compiled old dlls is ignored when deserializing it?

src/QsCompiler/DataStructures/SyntaxTree.fs
```diff
@@ -119,35 +119,70 @@ type QsLocation =
...
+
+        /// The name of the type parameter.
         TypeName: string
-        /// the range at which the type parameter occurs relative to the statement (or partial statement) root
-        /// -> is Null for auto-generated type information, i.e. in particular for inferred type information
+
+        /// The source code range where the type parameter occurs. Null for auto-generated or inferred types. Range is
+        /// ignored when comparing type parameters.
+        // TODO: RELEASE 2021-10: Remove obsolete property.
+        [<Obsolete "Use ResolvedType.Range instead.">]
```
bettinaheim, line 136: I saw that in the NamespaceManager there are a lot of usages of this now deprecated handle. They are not showing up here because the file already contains a nowarn due to the access modifier changes. Please take a look at what updates are required there - I believe it should be possible to update it entire to use only the new handles.

src/QsCompiler/Tests.Compiler/SerializationTests.fs
```diff
@@ -4,6 +4,8 @@
 // mock-up for the purpose of testing
 namespace Microsoft.Quantum.Simulation.Core
 
+#nowarn "44" // UserDefinedType.Range is deprecated.
```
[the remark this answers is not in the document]
bettinaheim: Looking more closely now that I can build locally, I think it would be better to just update the one test that is there and remove this.

src/QsCompiler/Tests.Compiler/TestUtils/Signatures.fs
```diff
@@ -3,6 +3,8 @@
 
 module Microsoft.Quantum.QsCompiler.Testing.Signatures
 
+#nowarn "44" // TypeParameter.Range and UserDefinedType.Range are deprecated.
```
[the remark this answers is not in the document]
bettinaheim: Same here; the tests can easily be updated to not use the old handles.

src/QsCompiler/Transformations/SearchAndReplace.cs
```diff
@@ -230,15 +230,19 @@ public TypeTransformation(SyntaxTreeTransformation<TransformationState> parent)
             public override QsTypeKind OnUserDefinedType(UserDefinedType udt)
             {
                 var id = Identifier.NewGlobalCallable(new QsQualifiedName(udt.Namespace, udt.Name));
+#pragma warning disable 618 // UserDefinedType.Range is obsolete.
```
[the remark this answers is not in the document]
bettinaheim: After looking through all code, I realize that you maintain the full support for the old handles and they are still populated (that makes sense, thank you!). Regardless, we need to update everything within the compiler to work only with non-deprecated items, otherwise this will just haunt us once we remove the support for the deprecated handles.

src/QsCompiler/SyntaxProcessor/SymbolTracker.fs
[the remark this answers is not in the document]
bettinaheim: Ok, I am good with making it a separate PR, so I'll unblock the PR. However, please address the comment around removing other suppressions of deprecated warnings where possible (they are all small to update).

src/QsCompiler/Transformations/ContentLifting.cs
```diff
@@ -439,7 +439,9 @@ public override ResolvedTypeKind OnTypeParameter(QsTypeParameter tp)
                     // Reroute a type parameter's origin to the newly generated operation
                     if (!this.SharedState.IsRecursiveIdentifier && this.SharedState.OldName.Equals(tp.Origin))
                     {
+#pragma warning disable 618 // QsTypeParameter.Range is obsolete.
```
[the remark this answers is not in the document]
samarsha: It doesn't have access to `ResolvedType.Range` because the parameter its given is only a `QsTypeParameter`. The update is non-trivial, so I will do it in a different PR.

src/QsCompiler/SyntaxProcessor/TypeInference/InferenceContext.fsi
samarsha: I updated the doc comments for `Unify` and `Intersect`. For `Unify`, `actual` must be a subtype of `expected`. `Intersect` returns a type that is a supertype of the given types.

The reason that the more general functions are private is simply because there is no reason for any other part of the compiler to ask for an invariant or contravariant unification or intersection. That only happens when recursing into an array or callable type, which only occurs in the inference module itself.

src/QsCompiler/DataStructures/SyntaxTree.fs
samarsha: Do you mean for the compiled name, the F# name, or both?

src/QsCompiler/BondSchemas/BondSchemaTranslator.cs
samarsha: Well, we could drop the ranges during serialization, but that is technically a breaking change if anyone is relying on them in a deserialized syntax tree (e.g. a rewrite step that looks at references in the AST). Probably no one would be affected by it, but I wanted to continue fully supporting the old ranges until they are removed.

samarsha: Or do you mean that since the old behavior was to strip range information during deserialization, then this wouldn't be a breaking change?

src/QsCompiler/DataStructures/SyntaxTree.fs
samarsha: I'm not sure what you mean by a pre-existing nowarn for access modifiers? There is one in Namespace.fs but not NamespaceManager.fs. I added the nowarn to NamespaceManager.fs deliberately, because it relies on the range of UserDefinedType for attribute diagnostics. I think fixing that is a non-trivial change that I would like to do in a separate PR.

src/QsCompiler/Core/Transformations/TypeTransformation.fs
samarsha: I think adding separate abstract methods for each case of `TypeRange` is too much, but I did deprecate `OnRangeInformation` and added a new `OnTypeRange`.

src/QsCompiler/DataStructures/SyntaxTree.fs
bettinaheim: Both, but just a thought.

src/QsCompiler/Core/Transformations/TypeTransformation.fs
bettinaheim: Ok, that works for me.

src/QsCompiler/BondSchemas/BondSchemaTranslator.cs
bettinaheim: I think this might not actually break anything if we remove it, and people are supposed to be using the compiler methods to load the syntax tree rather than rely on the serialization direction. So as long as that keeps working for old dlls, we are good to remove this.

src/QsCompiler/DataStructures/SyntaxTree.fs
bettinaheim: Yes, my desire was to update the diagnostics generation. Making this a separate PR makes sense.

src/QsCompiler/DataStructures/SyntaxTree.fs
samarsha: I don't think that's consistent with F# naming conventions; compare with `List.tryHead`, `List.tryLast`, etc. I generally think "get" is overused in function names since most functions are getting something.

PR #920, opened 2021-03-24
Generate namespace summaries as Markdown.

cgranade:
This PR fixes an issue reported offline by @tcNickolas by removing YAML versions of namespace summaries in favor of pure-Markdown summaries.

### DIFF
--- a/src/Documentation/Summarizer/summarize_documentation.py
+++ b/src/Documentation/Summarizer/summarize_documentation.py
@@ -44,6 +44,7 @@ def items_of_kind(items, kind):
     return [
         {
             "uid": item.uid,
+            "name": item.name,
             "summary": item.summary
         }
         for item in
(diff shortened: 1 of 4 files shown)


src/Documentation/Summarizer/summarize_documentation.py
```diff
@@ -53,6 +54,76 @@ def items_of_kind(items, kind):
...
+| Name | Summary |
+|------|---------|
+"""
+    def first_line(summary):
+        lines = summary.splitlines()
+        return lines[0] if lines else ""
+
+    return (
+        table_header + "\n".join(
+            f"|[{item['name']}](xref:{item['uid']}) |{first_line(item['summary'])} |"
```
samarsha, line 70: Is there no space after the first `|` of each column?

src/Documentation/Summarizer/summarize_documentation.py
```diff
@@ -53,6 +54,76 @@ def items_of_kind(items, kind):
         )
     ]
 
+def summaries_table(namespace, kind, header):
```
samarsha, line 57: Do you want to include type hints for `summaries_table` and `write_namespace_summary`?

src/Documentation/Summarizer/summarize_documentation.py
```diff
@@ -53,6 +54,76 @@ def items_of_kind(items, kind):
...
+        )
+        if kind in namespace and namespace[kind]
+        else ""
+    )
+
+def write_namespace_summary(target_file : Path, namespace):
+    """
+    Given the contents of a namespace, either creates a new Markdown file
+    summarizing those contents, or appends those contents to an existing
+    Markdown file for that namespace.
```
samarsha, line 81: This will replace an existing `summaries` section, not append to it, right?

src/Documentation/Summarizer/summarize_documentation.py
```diff
@@ -53,6 +54,76 @@ def items_of_kind(items, kind):
...
+            contents = contents[:contents.index(begin_region)] + summaries + contents[contents.index(end_region) + len(end_region):]
+        else:
+            contents += "\n" + summaries
+
+        with open(target_file, 'w', encoding='utf8') as f:
+            f.write(contents)
+
+    # If the file does not exist, go on and make it now.
+    else:
+        with open(target_file, 'w', encoding='utf8') as f:
```
samarsha, line 110: There's a time-of-check to time-of-use bug here. What if the file is created after `target_file.exists()` but before opening it for writing? Maybe it doesn't matter that much for this script though.

src/Documentation/Summarizer/summarize_documentation.py
cgranade: In the examples for formatting for docs.microsoft.com, non-heading cells did not have a space after the first `|`. It may well be that the examples are incorrect, but we'll need to check the reference rendering on docs.ms to check that.

src/Documentation/Summarizer/summarize_documentation.py
cgranade: Absolutely. My intent with this PR was to unblock the immediate issue with API docs, but I'll go on and improve type hinting now.

src/Documentation/Summarizer/summarize_documentation.py
cgranade: Agreed, and I did consider that but it made the flow of the code a bit less obvious to avoid that issue. Given that this script is intended to run on a static set of assets (e.g.: build outputs), I think that those assets changing during summarization is likely out of scope for this tool.

src/Documentation/Summarizer/summarize_documentation.py
samarsha: OK, I think it's fine without a space, just curious if there was a reason.

src/Documentation/Summarizer/summarize_documentation.py
cgranade: Those type hints have now been added. Thanks!

src/Documentation/Summarizer/summarize_documentation.py
```diff
@@ -53,11 +54,84 @@ def items_of_kind(items, kind):
...
+| Name | Summary |
+|------|---------|
+"""
+    def first_line(summary : str) -> str:
+        lines = summary.splitlines()
+        return lines[0] if lines else ""
+
+    items = items_of_kind(namespace.items, kind)
+    print(items)
+    print(namespace.items)
```
samarsha, line 70: Are the prints supposed to be there?

cgranade: Nope, thank you!

samarsha (approved):

PR #923, opened 2021-03-25
Separate packaging of QIR emission as a compiler extension

bettinaheim:
This PR splits the QIR emission pass into a separate compiler extension package. The change is mainly pursued due to the size of LlvmLib.


src/QsCompiler/QirGeneration/QirGeneration.nuspec
```diff
@@ -0,0 +1,31 @@
...
+      <dependency id="LLVMSharp" version="11.0.0-beta" />
+      <dependency id="libLLVM.runtime.win-x64" version="11.0.0" />
+      <dependency id="libLLVM.runtime.osx-x64" version="11.0.0" />
+      <dependency id="libLLVM.runtime.ubuntu.20.04-x64" version="11.0.0" />
+      <dependency id="libLLVM.runtime.ubuntu.18.04-x64" version="11.0.0" />
+    </dependencies>
+  </metadata>
+  <files>
+    <file src="QirGeneration.props" target="build\Microsoft.Quantum.QirGeneration.props" />
+    <file src="bin\$Configuration$\**\Microsoft.Quantum.LlvmBindings.dll" target="lib" />
```
swernli, line 29: Do we want this file to follow the patterns we have for other nuspecs, namely that the .nuspec is not checked in and instead a .nuspec.template is used along with a script to dynamically generate it? As annoying as that pattern can be sometimes, it ensures we get the right behavior from release builds where dependencies added to the .csproj are automatically added to the nuspec. With it checked in like this, the developer will have to remember to manually update the file with any new project or package references.

src/QsCompiler/QirGeneration/RewriteSteps.cs
```diff
@@ -4,14 +4,98 @@
...
+
+            PerformanceTracking.TaskStart(PerformanceTracking.Task.BitcodeGeneration);
+            var bcOutputFolder = this.AssemblyConstants.TryGetValue(ReservedKeywords.AssemblyConstants.OutputPath, out var path) && !string.IsNullOrWhiteSpace(path) ? path : "qir";
+            var bcFile = CompilationLoader.GeneratedFile(targetFile, Path.GetFullPath(bcOutputFolder), ".bc", "");
+            generator.Emit(bcFile, emitBitcode: true);
+            PerformanceTracking.TaskEnd(PerformanceTracking.Task.BitcodeGeneration);
+
+            // create the human readable version as well
+            var sourceOutputFolder = this.AssemblyConstants.TryGetValue(ReservedKeywords.AssemblyConstants.QirOutputPath, out path) && !string.IsNullOrWhiteSpace(path) ? path : "qir";
+            var llvmSourceFile = CompilationLoader.GeneratedFile(targetFile, Path.GetFullPath(sourceOutputFolder), ".ll", "");
```
swernli, line 89: With the other changes, does this mean that the .ll file will no longer be generated in the "qir" folder with a name matching the project, but instead be named "__qsharp_data_qir_v1__.ll" and end up in the regular tree? If so, that will break some logic in scripts in the runtime repo that expect to find the .ll file in a specific place. Now that we are producing bitcode we can probably update those scripts to expect bitcode instead, but we'd need to coordinate a cross-repo check in for this since it would be a breaking change.

src/QsCompiler/CommandLineTool/Options.cs
```diff
@@ -64,6 +64,20 @@ public class CompilationOptions : Options
...
+            Required = false,
+            Default = false,
+            HelpText = "Specifies whether to omit monomorphizing the AST for executables.")]
+        public bool SkipMonomorphization { get; set; }
+
+        [Option(
+            "force-rewrite-step-execution",
+            Required = false,
+            Default = false,
+            HelpText = "Specifies whether to execute rewrite steps even if their precondition is not satisfied. If this is the case, the transformation output will be ignored and the compilation fails.")]
```
swernli, line 78: I'm a little confused by this option, since it states that the compilation will fail. Why have an option that will always cause failure? I think I may be misunderstanding the comment. Either way, it looks like it is only used in a test so far, so perhaps mentioning in the help text that it is for testing purposes only would be helpful (assuming that is true)?

swernli (commented):
I see a few places that I'm worried may cause some headache for us. It's possible these can be addressed in a later PR, but I'd like to see what you think first before signing off.

src/QsCompiler/QirGeneration/RewriteSteps.cs
bettinaheim: No, it is still in the same place. The behavior of the Sdk (i.e. the use of the feature within a Q# project) did not change; the only things that is added is that the bc is also attached now.

src/QsCompiler/CommandLineTool/Options.cs
bettinaheim: Copy pasting the description of the original PR (https://github.com/microsoft/qsharp-compiler/pull/907):

The current behavior is that if the precondition verification fails for rewrite steps, then the step is skipped and the compilation succeeds with a warning. While overall this seems reasonable behavior given that rewrite steps may come from a 3rd party, I also found this rather suboptimal for testing: When I set up the test for qir emission for a targeted executable, the test could technically succeed when/because the targeting step is skipped (since we test for a specific .ll it will fail regardless, but still seems risky since it is easy to overlook that the transformation in question is not executed when e.g. updating the test).
I hence went ahead and added a command line flag to force execution of rewrite steps and generate an error when the precondition is not met (see the doc comments for more detail).

swernli: Ah, that makes perfect sense then. I think what confused me was the second sentence, "If this is the case, the transformation output will be ignored and the compilation fails." I thought that the "case" was the parameter being specified, meaning every time the parameter is specified the output would be ignored and compilation fails. Consider this a minor feedback on the phrasing, but the intent definitely makes sense now.

bettinaheim: I updated the phrasing to make it clearer.

src/QsCompiler/QirGeneration/QirGeneration.nuspec
bettinaheim: I made the nuspec a template and added the powershell script to walk and add the dependencies. I chose to add only package dependencies, and list projects to walk separately (which means when a project is added that should be included in the qir emission package one would need to edit both the nuspec template and the powershell script). This seemed nonetheless the best choice since which dependencies need to be included is not entirely straightforward in this chase; I suspect we might actually need to package the llvm libs with the step as well rather than having them be package dependencies (which would be nicer)...

swernli (approved):
Updates look great!

PR #928, opened 2021-03-26
Allow disabling generation properties explicitly in MSBuild SDK

cgranade:
This PR modifies the generation properties used to control QIR, perfdata, and API docs output so that:

- If the generation property is not explicitly turned on and output path is set, no files will be generated.
- If the generation property is set to true, and no output path is set, files will be output to a default path.
- If the generation property is not explicitly turned on, and an output path is set, files will be output to the specified path.
- If the generation property is set to false, then no files will be output, irrespective of any path settings.

This PR also updates the README accordingly and fixes a one-character typo preventing perfdata from being output in some conditions.


src/QuantumSdk/README.md
```diff
@@ -118,54 +118,54 @@ in the project that implements the compilation step.
 
 The Sdk defines the following properties for each project using it:
 
-- `QSharpLangVersion`:
```
bettinaheim, line 121: Please do not remove the whitespace in the markdown to keep the line breaks.

src/QuantumSdk/DefaultItems/DefaultItems.targets
```diff
@@ -59,10 +59,13 @@
   <PropertyGroup>
     <DefaultEntryPointGeneration>dotnet "$(MSBuildThisFileDirectory)../tools/utils/Microsoft.Quantum.Sdk.DefaultEntryPoint.Generation.dll"</DefaultEntryPointGeneration>
     <!-- true/false optional properties -->
-    <QirGeneration Condition="'$(QirOutputPath)' != ''">true</QirGeneration>
+    <QirGeneration Condition="'$(QirOutputPath)' != '' And '$(QirGeneration)' == 'default'">true</QirGeneration>
```
bettinaheim, line 62: I am a bit confused why that works; aren't you missing an update to the DefaultItems.props.template to set the value to default initially? Also, did you check that all three *Generation properties are used within the Sdk only after they are set to either true or false here?

bettinaheim (approved):

PR #929, opened 2021-03-26
Added c++ entrypoint template

puzzleshark:
* Added c++ entrypoint T4 template and updated test asset


src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
```diff
@@ -1,16 +1,342 @@
...
+    InteropRange(RangeTuple rangeTuple) :
+        Start(get<0>(rangeTuple)),
+        Step(get<1>(rangeTuple)),
+        End(get<2>(rangeTuple)){}
+};
+
+// This is the function corresponding to the QIR entry-point.
+extern "C" void <#= entryPointOperation.Name #>(
+    <# foreach (var arg in entryPointOperation.Arguments) { #>
+        <# if (arg.Type == DataType.IntegerType) { #>
```
cesarzc, line 57: It would be great to encapsulate some of the logic outside the template itself and keep the code in the template minimal.

One way this can be done is to extend the classes defined in the Bond schema since they are generated as partial classes. For example, an _EntryPointCpp.cs_ file can be created that contain logic specific for C++ generation. More specifically, something like this could be done in that file:
```csharp
    public partial class Argument
    {
        public string CppType() =>
            this.Type switch
            {
                DataType.BoolType => "char",
                DataType.IntType => "int64_t",
                // Add all the rest of the types here.
            };
    }
```
This can be used in the template as `arg.CppType()` which reduces the amount of logic needed in the template itself.

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
```diff
@@ -1,16 +1,342 @@
...
+#include "CLI11.hpp"
+
+#include "QirContext.hpp"
+#include "QirRuntime.hpp"
+#include "SimFactory.hpp"
+
+using namespace Microsoft::Quantum;
+using namespace std;
+
+struct InteropArray
```
cesarzc, line 26: Eventually we will move some of the generic logic and types to its own library and header, but in the meantime it would be great to not include types and logic that are not needed in the generated code.

For this case, extending the `EntryPointOperation` class to have a `ContainsArgumentType(DataType type)` method would be very useful.

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
```diff
@@ -1,16 +1,342 @@
...
+    // Add the --simulation-output options.
+    // N.B. This option should be present in all standalone drivers.
+    string simulationOutputFile;
+    CLI::Option* simulationOutputFileOpt = app.add_option(
+        "-s,--simulation-output", simulationOutputFile,
+        "File where the output produced during the simulation is written");
+
+    <# foreach (var arg in entryPointOperation.Arguments) { #>
+        <# if (arg.Type == DataType.IntegerType) { #>
+    // Add the options that correspond to the parameters that the QIR entry-point needs.
```
cesarzc, line 170: Some things in the reference C++ code can be changed to make the generated code clearer. In the case of adding options, the comment that precedes them could be something like:
```cpp
// Option for the argument named <# arg.Name #>.
```

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
```diff
@@ -1,16 +1,342 @@
...
+    // Option for a Q# Int type.
+    int64_t <#= arg.Name #> = 0;
+    app.add_option("--<#= arg.Name #>", <#= arg.Name #>, "An integer value")->required();
```
cesarzc, line 173: We should abstract the logic of both option name and option variable for an argument. For the case of option name, if the name of the argument is just a single character, it should be preceded by a single dash `-` instead of two dashes `--`.

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
```diff
@@ -1,16 +1,342 @@
...
+
+    // Initialize simulator.
+    unique_ptr<IRuntimeDriver> sim = CreateFullstateSimulator();
+    QirContextScope qirctx(sim.get(), false /*trackAllocatedObjects*/);
+
+    // Add the --simulation-output options.
+    // N.B. This option should be present in all standalone drivers.
+    string simulationOutputFile;
+    CLI::Option* simulationOutputFileOpt = app.add_option(
+        "-s,--simulation-output", simulationOutputFile,
```
cesarzc, line 165:
```suggestion
"--simulation-output", simulationOutputFile,
```

cesarzc: Removing the `-s` option to remove the chance of it colliding with an argument named `s`.

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
```diff
@@ -1,16 +1,342 @@
...
+    if (!simulationOutputFileOpt->empty())
+    {
+        simulationOutputFileStream.open(simulationOutputFile);
+        SetOutputStream(simulationOutputFileStream);
+        simulatorOutputStream = &simulationOutputFileStream;
+    }
+
+    // Run simulation and write the output of the operation to the corresponding stream.
+    <#= entryPointOperation.Name #>(
+    <# foreach (var arg in entryPointOperation.Arguments) { #>
```
cesarzc, line 307: To reduce the places where the code depends on the argument type, previous to this call, a final variable used as argument for the entry-point operation can be created with the `Interop` suffix and this logic wouldn't be dependent on argument type anymore.

cesarzc, line 307: To guarantee the order of arguments is correct when calling the entry-point, we should extend the `EntryPointOperation` class to have a `PositionallyOrderedArguments` that returns the ordered list according to the `Position` property of the arguments.

src/QsCompiler/Tests.Compiler/TestCases/QirEntryPointTests/TestUnitReceivesBool.cpp
```diff
@@ -2,14 +2,171 @@
 // This code was generated by a tool.
 // <auto-generated />
 //------------------------------------------------------------------------------
+//
```
cesarzc, line 5: Offline, we talked a bit about how to organize tests for this component, just adding a comment such that we don't forget about it.

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
puzzleshark: done

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
puzzleshark: removed if not used now

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
puzzleshark: removed these comments for now since they would only be marginally useful to the reader

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
puzzleshark: done

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
puzzleshark: done

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
puzzleshark: fixed this, now always get arguments after sorting

src/QsCompiler/Tests.Compiler/TestCases/QirEntryPointTests/TestUnitReceivesBool.cpp
puzzleshark: maybe I can do this in a follow up pr?

src/QsCompiler/Compiler/Templates/CppInterop.cs
```diff
@@ -0,0 +1,151 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.Quantum.QsCompiler.BondSchemas.EntryPoint;
+
+namespace Microsoft.Quantum.QsCompiler.Templates
+{
+    public static class CppInterop
```
cesarzc, line 8: I know that I originally mentioned that encapsulating logic in a `static class` was a good option but thinking a bit more about it, creating sub-classes seems like a better idea.

Sub-classes can inherit from the Bond-generated classes and implement methods that are specific for C++ interop:
```csharp
    public class ArgumentCpp : Argument
    {
        public ArgumentCpp(Argument argument)
        {
            this.Name = argument.Name;
            this.Position = argument.Position;
            this.Type = argument.Type;
            this.ArrayType = argument.ArrayType;
            this.Values = argument.Values.ToList();
        }

        public string InteropType() =>
            this.Type switch
            {
                DataType.BoolType => "char",
                DataType.IntegerType => "int64_t",
                DataType.DoubleType => "double",
                DataType.PauliType => "char",
                DataType.RangeType => "InteropRange*",
                DataType.ResultType => "char",
                DataType.StringType => "const char*",
                DataType.ArrayType => "InteropArray *",
            };
    }
 
    public class EntryPointOperationCpp : EntryPointOperation
    {
        public List<ArgumentCpp> InteropArguments;
 
        public EntryPointOperationCpp(EntryPointOperation entryPointOperation)
        {
            this.Name = entryPointOperation.Name;
            this.Arguments = entryPointOperation.Arguments.ToList();
            this.InteropArguments = entryPointOperation.Arguments.Select(arg => new ArgumentCpp(arg)).ToList();
        }
    }
```
Apologies for the churn.

src/QsCompiler/Tests.Compiler/TestCases/QirEntryPointTests/TestUnitReceivesBool.cpp
cesarzc: It is usually a good idea to merge implementation and tests in the same PR since it allows for validation before the code is merged. Having said that, I think this PR should contain the tests as well.

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
cesarzc: Can you please do the same for the other types defined here (e.g. `InteropRange` which is only used when a Q# `Range` type argument is present?

src/QsCompiler/Compiler/Templates/CppInterop.cs
cesarzc: You would need to update _src/QsCompiler/Compiler/Templates/QirDriverCppEx.cs_ to make the `entryPointOperation` member of type `EntryPointOperationCpp`.

src/QsCompiler/Compiler/Templates/CppInterop.cs
```diff
@@ -0,0 +1,151 @@
...
+                    DataType.DoubleType => null,
+                    DataType.PauliType => "PauliMap",
+                    DataType.RangeType => null,
+                    DataType.ResultType => "ResultAsCharMap",
+                    DataType.StringType => null,
+                },
+            };
+        }
+
+        public static bool AtLeastOneArrayArgument(EntryPointOperation op)
```
cesarzc, line 139: To make this more generic, this could receive a `DataType` and returns `true` if at least one argument is of that type.

Additionally, since it will be more generic a good name for this method would be `ContainsArgumentType`.

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
puzzleshark: done

src/QsCompiler/Tests.Compiler/TestCases/QirEntryPointTests/TestUnitReceivesBool.cpp
puzzleshark: added tests

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
puzzleshark: fixed

src/QsCompiler/Compiler/Templates/CppInterop.cs
puzzleshark: done

src/QsCompiler/Compiler/Templates/CppInterop.cs
puzzleshark: done

src/QsCompiler/Compiler/Templates/CppInterop.cs
```diff
@@ -0,0 +1,168 @@
...
+            }
+            else
+            {
+                return $"--{this.Name}";
+            }
+        }
+
+        public string? CliDescription()
+        {
+            return this.Type switch
```
cesarzc, line 49: I think this could include the name of the argument rather than the type. Something like the following would work:
```csharp
return $"Value for the {this.Name} argument"
```

puzzleshark: could this be redundant though since this is already the arguments name?

cesarzc: You're right, this could be redundant.

This is only relevant when looking at the help for the executable. For example:
```
> .\qir-input-reference-standalone.exe --help
QIR Standalone Entry Point Inputs Reference
Usage: C:\Microsoft\ReposA\qsharp-runtime\src\Qir\Samples\build\Debug\StandaloneInputReference\qir-input-reference-standalone.exe [OPTIONS]

Options:
  -h,--help                   Print this help message and exit
  -s,--simulation-output TEXT File where the output produced during the simulation is written
  --int-value INT REQUIRED    An integer value
  --integer-array INT ... REQUIRED
                              An integer array
  --double-value FLOAT REQUIRED
                              A double value
  --double-array FLOAT ... REQUIRED
                              A double array
```

I think it might be ok to have a bit of redundancy in the description. In the current implementation, all arguments of the same type would have the same description so having a bit of redundancy might be slightly better.

cesarzc: I see that you added the name of the argument to the description. I like how it turned out!

src/QsCompiler/Compiler/Templates/CppInterop.cs
```diff
@@ -0,0 +1,168 @@
...
+            {
+                return $"-{this.Name}";
+            }
+            else
+            {
+                return $"--{this.Name}";
+            }
+        }
+
+        public string? CliDescription()
```
cesarzc, line 47: I am curious about why this method returns a nullable string rather than just a string.

src/QsCompiler/Compiler/Templates/CppInterop.cs
cesarzc: Consider creating a `CliTypeDescription` method to simplify this method to be something like:
```csharp
public string? CliDescription() => $"A {this.CliTypeDescription()} value for the {this.Name} argument";
```

src/QsCompiler/Compiler/Templates/CppInterop.cs
```diff
@@ -0,0 +1,168 @@
...
+            return this.Type switch
+            {
+                DataType.BoolType => "char",
+                DataType.IntegerType => "int64_t",
+                DataType.DoubleType => "double",
+                DataType.PauliType => "char",
+                DataType.RangeType => "InteropRange*",
+                DataType.ResultType => "char",
+                DataType.StringType => "const char*",
+                DataType.ArrayType => "InteropArray *",
```
cesarzc, line 31: To cover all cases, can you please throw an exception for all not explicitly supported cases:
```csharp
_ => throw new NotSupportedException($"Unsupported argument type {this.Type}")
```

src/QsCompiler/Compiler/Templates/CppInterop.cs
```diff
@@ -0,0 +1,168 @@
...
+            return this.Type switch
+            {
+                DataType.BoolType => "char",
+                DataType.IntegerType => "int64_t",
+                DataType.DoubleType => "double_t",
+                DataType.PauliType => "PauliId",
+                DataType.RangeType => "RangeTuple",
+                DataType.ResultType => "char",
+                DataType.StringType => "string",
+                DataType.ArrayType => this.ArrayType switch
```
cesarzc, line 82: For all switch expressions, can you please throw an exception similar to the one in this [comment](https://github.com/microsoft/qsharp-compiler/pull/929/files#r609269397) when the argument type is not explicitly supported.

cesarzc: Also the for the sub-expressions that switch on `ArrayType`.

cesarzc:
@puzzleshark, can you please re-target this PR to the [feature/azure-quantum-simulator](https://github.com/microsoft/qsharp-compiler/tree/feature/azure-quantum-simulator) branch.

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
```diff
@@ -1,16 +1,268 @@
-﻿<#@ template language="C#" linePragmas="false" #>
+﻿<#@ template language="C#" #>
```
cesarzc, line 1:
```suggestion
<#@ template language="C#" linePragmas="false" #>
```

cesarzc: This is needed such that the the generated C# code does not have `#line` pragmas.

src/QsCompiler/Compiler/Templates/CppInterop.cs
```diff
@@ -0,0 +1,168 @@
...
+    public class EntryPointOperationCpp : EntryPointOperation
+    {
+        public List<ArgumentCpp> InteropArguments;
+
+        public EntryPointOperationCpp(EntryPointOperation entryPointOperation)
+        {
+            this.Name = entryPointOperation.Name;
+            this.Arguments = entryPointOperation.Arguments.ToList();
+            this.Arguments.Sort((a, b) => a.Position.CompareTo(b.Position));
+            this.InteropArguments = entryPointOperation.Arguments.Select(arg => new ArgumentCpp(arg)).ToList();
```
cesarzc, line 153:
```suggestion
            this.InteropArguments = this.Arguments.Select(arg => new ArgumentCpp(arg)).ToList();
```

cesarzc: This is to make sure that InteropArguments are sorted.

src/QsCompiler/Compiler/Templates/CppInterop.cs
puzzleshark: done

src/QsCompiler/Compiler/Templates/CppInterop.cs
puzzleshark: done

src/QsCompiler/Compiler/Templates/CppInterop.cs
puzzleshark: yep good call doesn't need to be, fixed

src/QsCompiler/Compiler/Templates/CppInterop.cs
puzzleshark: done

src/QsCompiler/Compiler/Templates/CppInterop.cs
puzzleshark: good catch! fixed

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
puzzleshark: I see, added back in

src/QsCompiler/Compiler/Templates/CppInterop.cs
```diff
@@ -0,0 +1,181 @@
...
+                DataType.ArrayType => this.ArrayType switch
+                {
+                    DataType.BoolType => $"bool array",
+                    DataType.IntegerType => $"integer array",
+                    DataType.DoubleType => $"double array",
+                    DataType.PauliType => $"Pauli array",
+                    DataType.RangeType => $"Range array",
+                    DataType.ResultType => $"Result array",
+                    DataType.StringType => $"String array",
+                    _ => throw new NotSupportedException($"Unsupported argument type {this.Type}")
```
cesarzc, line 70:
```suggestion
                    _ => throw new NotSupportedException($"Unsupported array type {this.Type}")
```

cesarzc: Please use "array" instead of "argument" when throwing an exception from the `ArrayType` switch expression.

puzzleshark: fixed

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
cesarzc: This can also be applied to auxiliary functions and data structures. More specifically:
- `InteropFalseAsChar`, `InteropTrueAsChar` and `BoolAsCharMap` are not needed if there are no Bool or Bool array arguments.
- `PauliMap` and `TranslatePauliToChar` are not needed if there are no Pauli or Pauli array arguments.
- `InteropResultZeroAsChar`, `InteropResultOneAsChar` and `ResultAsCharMap` are not needed if there are no Result or Result array arguments.
- `CreateInteropArray` and `TranslateVector` are not needed if there are no array arguments.
- `CreateInteropRange` and `TranslateRangeTupleToInteropRangePointer` are not needed if there are no Range or Range array arguments.
- `TranslateStringToCharBuffer` is not needed if there are no String or String array arguments.
- `FreePointerVector` is not needed if there are no Range array arguments.

src/QsCompiler/Tests.Compiler/EntryPointSchemaTests.fs
```diff
@@ -77,10 +79,10 @@ type EntryPointSchemaTests(output: ITestOutputHelper) =
     let sampleEntryPointOperations =
         Map
             .empty
-            .Add("UseNoArgs", (new EntryPointOperation(Name = "UseNoArgs"), "{\"Name\":\"UseNoArgs\"}"))
+            .Add("UseNoArgs", (new EntryPointOperation(Name = "UseNoArgs"), Path.Join(testCasesDirectory, "UseNoArgs.json") |> File.ReadAllText))
```
cesarzc: Instead of the value of this map being a tuple, it can just be the `EntryPointOperation` object and the test functions can read the text from the file using the key as the file name (plus the .json extension) in a known directory.

This way the `Path.Join` and `File.ReadAllText` operations are only performed in one place.

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
```diff
@@ -1,16 +1,268 @@
...
+        WriteLine("        ->required();");
+    } else {
+        WriteLine($"        ->required()->transform(CLI::CheckedTransformer({arg.TransformationType()}, CLI::ignore_case));");
+    }
+} #>
+
+    // With all the options added, parse arguments from the command line.
+    CLI11_PARSE(app, argc, argv);
+
+<# foreach (var arg in entryPointOperation.InteropArguments) {
```
cesarzc, line 169: What I meant originally with this [comment](https://github.com/microsoft/qsharp-compiler/pull/929#discussion_r603590674) was something like the following:

For every argument there are two variables. One that holds the value parsed from the command line whose suffix would be _CliValue_, and one that holds the interop value to be passed to the entry point operation whose suffix would be _Interop_.

In the previous loop, the _CliValue_ variable is declared and used for the option as you are currently doing it:
```
WriteLine($"{arg.CppCliValueType()} {arg.CliValueVariableName()};");
```

In this loop, the _Interop_ variable is declared and the _CliValue_ is used to assign a value to the _Interop_ variable depending on the argument type. For example, for the PauliType it would be something like this:
```
WriteLine($"{arg.CppType()} {arg.InteropVariableName()};");
switch (arg.Type) {
    case DataType.PauliType:
          WriteLine($"{arg.InteropVariableName()} =TranslatePauliToChar({arg.CliValueVariableName()});");
          break;
```

This way, in the next loop, a switch is not needed and something simpler can be done:
```
Write($"{arg.InteropVariableName()}");
```

cesarzc: I'll close the original comment and track the suggestion using this one.

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
cesarzc: Tracking this suggestion using this [comment](https://github.com/microsoft/qsharp-compiler/pull/929/files#r609958848).

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
puzzleshark: done

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
```diff
@@ -1,16 +1,278 @@
...
+        default:
+            Write($"        {arg.Name}");
+            break;
+    }
+    if (i < entryPointOperation.InteropArguments.Count-1 ) {
+        WriteLine(",");
+    }
+} #>
+
+);
```
cesarzc, line 265: Nit: there is a blank line before this closing parenthesis and it is not indented.

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
```diff
@@ -1,16 +1,278 @@
...
+    // Add the --simulation-output options.
+    // N.B. This option should be present in all standalone drivers.
+    string simulationOutputFile;
+    CLI::Option* simulationOutputFileOpt = app.add_option(
+        "--simulation-output", simulationOutputFile,
+        "File where the output produced during the simulation is written");
+    
+<# foreach (var arg in entryPointOperation.InteropArguments) {
+    WriteLine("");
+    WriteLine($"    {arg.CppVarType()} {arg.Name};");
```
cesarzc, line 166: For indentation, it would be a bit more readable to have a global `Indentation` string that you could reference.

src/QsCompiler/Compiler/Templates/CppInterop.cs
```diff
@@ -0,0 +1,193 @@
...
+                    DataType.StringType => $"String array",
+                    _ => throw new NotSupportedException($"Unsupported array type {this.Type}")
+                },
+                _ => throw new NotSupportedException($"Unsupported argument type {this.Type}")
+            };
+        }
+
+        public string CliDescription() => $"A {this.CliTypeDescription()} value for the {this.Name} argument";
+
+        public string? CppVarType()
```
cesarzc, line 78: Following the pattern from this [comment](https://github.com/microsoft/qsharp-compiler/pull/929#discussion_r609958848), a more descriptive name for this method would be `CppCliValueType`.

cesarzc, line 78: I think this method can return a non nullable `string`.

src/QsCompiler/Compiler/Templates/CppInterop.cs
```diff
@@ -0,0 +1,193 @@
...
+                    DataType.RangeType => null,
+                    DataType.ResultType => null,
+                    DataType.StringType => null,
+                    _ => throw new NotSupportedException($"Unsupported array type {this.Type}")
+                },
+                _ => throw new NotSupportedException($"Unsupported argument type {this.Type}")
+            };
+        }
+
+        public string? TransformationType()
```
cesarzc, line 130: Looking at the usage of this in the template, a slightly more descriptive name for this method would be `TransformerMapName`.

To avoid possible mismatches with the names given to the maps in the template, creating a `static` method called `DataTypeTransformerMapName`  that performs this operation and using it in the class method and in the template would help.

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
```diff
@@ -1,16 +1,278 @@
...
+        WriteLine(",");
+    }
+} #>
+
+);
+
+<# if (entryPointOperation.ContainsArgumentType(DataType.BoolType) || entryPointOperation.ContainsArrayType(DataType.BoolType)) { #>
+const char InteropFalseAsChar = 0x0;
+const char InteropTrueAsChar = 0x1;
+map<string, bool> BoolAsCharMap{
```
cesarzc, line 109: Related to this [comment](https://github.com/microsoft/qsharp-compiler/pull/929/files#r610028353).

Here the string "BoolAsCharMap" should be replaced by a call to the `Argument.DataTypeTransformerMapName` method.

cesarzc: This should also be done for all the other maps.

src/QsCompiler/Compiler/Templates/CppInterop.cs
```diff
@@ -0,0 +1,193 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Quantum.QsCompiler.BondSchemas.EntryPoint;
+
+namespace Microsoft.Quantum.QsCompiler.Templates
+{
+    public class ArgumentCpp : Argument
```
cesarzc, line 11: We should make this class `internal` since we don't want to expose it outside the binary.

src/QsCompiler/Compiler/Templates/CppInterop.cs
```diff
@@ -0,0 +1,193 @@
...
+                    DataType.ResultType => "ResultAsCharMap",
+                    DataType.StringType => null,
+                    _ => throw new NotSupportedException($"Unsupported array type {this.Type}")
+                },
+                _ => throw new NotSupportedException($"Unsupported argument type {this.Type}")
+            };
+        }
+    }
+
+    public class EntryPointOperationCpp : EntryPointOperation
```
cesarzc, line 157: This class should also be `internal` since we don't want to expose its functionality outside its binary.

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
```diff
@@ -1,16 +1,278 @@
...
+    {"0", InteropFalseAsChar},
+    {"false", InteropFalseAsChar},
+    {"1", InteropTrueAsChar},
+    {"true", InteropTrueAsChar}};
+
+<# } #>
+<# if (entryPointOperation.ContainsArgumentType(DataType.PauliType) || entryPointOperation.ContainsArrayType(DataType.PauliType)) { #>
```
cesarzc, line 116: It would be good to group global variables and auxiliary functions that share the same condition. I think the only ones left to consolidate are the ones related to Pauli and Pauli arrays.

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
```diff
@@ -1,16 +1,278 @@
...
+}
+
+template<typename S, typename D>
+void TranslateVector(vector<S>& sourceVector, vector<D>& destinationVector, function<D(S&)> translationFunction)
+{
+    destinationVector.resize(sourceVector.size());
+    transform(sourceVector.begin(), sourceVector.end(), destinationVector.begin(), translationFunction);
+
+<# } #>
+<# if (entryPointOperation.ContainsArgumentType(DataType.RangeType) || entryPointOperation.ContainsArrayType(DataType.RangeType)) { #>
```
cesarzc, line 51: For blocks of code that are conditional, adding a space at the beginning rather than at the end would result in a consistent separation in the generated C++ code.

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
```diff
@@ -1,16 +1,278 @@
...
+    unique_ptr<IRuntimeDriver> sim = CreateFullstateSimulator();
+    QirContextScope qirctx(sim.get(), false /*trackAllocatedObjects*/);
+
+    // Add the --simulation-output options.
+    // N.B. This option should be present in all standalone drivers.
+    string simulationOutputFile;
+    CLI::Option* simulationOutputFileOpt = app.add_option(
+        "--simulation-output", simulationOutputFile,
+        "File where the output produced during the simulation is written");
+
```
cesarzc, line 163: Nit: this blank line can be removed since a blank line is added as part of the loop.

src/QsCompiler/Compiler/Templates/CppInterop.cs
puzzleshark: done

puzzleshark: done

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
puzzleshark: done

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
puzzleshark: done

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
puzzleshark: The one hiccup is in some cases there is another intermediate variable

cesarzc: I think that's ok. It provides consistency on how the values for the entry-point operation are obtained from the command line and processed for interop with QIR. When the C++ source code is compiled, it's probably going to be optimized anyway.

puzzleshark: done

src/QsCompiler/Compiler/Templates/CppInterop.cs
cesarzc: I think this still needs to be changed to `internal`.

src/QsCompiler/Compiler/Templates/CppInterop.cs
puzzleshark: done

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
puzzleshark: done

src/QsCompiler/Compiler/Templates/CppInterop.cs
puzzleshark: done

puzzleshark: done

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
```diff
@@ -1,16 +1,278 @@
...
+    string simulationOutputFile;
+    CLI::Option* simulationOutputFileOpt = app.add_option(
+        "--simulation-output", simulationOutputFile,
+        "File where the output produced during the simulation is written");
+    
+<# foreach (var arg in entryPointOperation.InteropArguments) {
+    WriteLine("");
+    WriteLine($"    {arg.CppVarType()} {arg.Name};"); 
+    if (arg.CppVarInitialValue() != null) { WriteLine($"    {arg.Name} = {arg.CppVarInitialValue()};"); }
+    WriteLine($"    app.add_option(\"{arg.CliOptionString()}\", {arg.Name}, \"{arg.CliDescription()}\")");
```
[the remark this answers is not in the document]
puzzleshark: done

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
puzzleshark: this resolves to the correct spacing in cpp file

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
puzzleshark: not sure what is meant for this, in some cases multiple indents are necessary, so could look a little odd that way

cesarzc: Yes, this might be a matter of personal preference so it's ok to leave it as it is.

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
puzzleshark: fixed

cesarzc (approved):
Approving this PR to be merged to the feature branch. I'll open a subsequent PR to get the C++ driver generator ready to be used by the QIR controller and by integration tests in other repos.

PR #930, opened 2021-03-26
Remove line pragmas from template-generated C#

cesarzc:
This change sets the `linePragmas` directive for the T4 template to false and updates the generated C# code.


ScottCarda-MS (commented):
The changes do indeed seem to do what the PR says it does, but could you speak to why it is desirable to remove these pragmas?

cesarzc:
> 
> 
> The changes do indeed seem to do what the PR says it does, but could you speak to why it is desirable to remove these pragmas?

The C# code is generated by a tool provided through Visual Studio, which means it has to be generated locally rather that during a pipeline build and included in source control because the tool itself is not multi-platform. By default, the tool generates `#line` pragmas to be able to step through the code in the template when debugging. However, some of these pragmas contain absolute paths that are specific to the machine in which generation happens. This is undesirable because when someone changes the template, the C# will also change and will have many unrelated changes since it will replace the paths used by `#line` pragmas with the paths specific to the local machine that is generating the C# code. Moreover, this can also happen if the path of the local repo changes.

ScottCarda-MS (approved):

PR #934, opened 2021-03-30
 Suppress expected warnings in Q# test projects

samarsha:


swernli (approved):

src/QsCompiler/TestTargets/Libraries/Library1/Library1.csproj
```diff
@@ -1,8 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk" InitialTargets="RecompileOnChange">
-
   <PropertyGroup>
     <OutputType>Library</OutputType>
     <TargetFramework>netcoreapp3.1</TargetFramework>
+    <NoWarn>QS7010</NoWarn> <!-- Reference cannot be included in the generated DLL. -->
```
bettinaheim, line 5: Hu, good to know that this works as expected. I had thought someone had mentioned a bug on it, but looking in the compiler and runtime repo, I can't fine anything filed on nowarn. So I guess I misremembered. Thanks for disabling the warning.

bettinaheim (approved):

src/QsCompiler/TestTargets/Libraries/Library1/Library1.csproj
samarsha: I have mentioned the NoWarn bugs a couple times. :) There are bugs with it, since NoWarn is not actually passed to `qsc` as a `--no-warn` argument, causing it to claim there is "1 warning", even though MSBuild seems to hide that warning now. Also, the language server doesn't recognize NoWarn, so the warnings still show up in IDEs. I should file issues for those.

samarsha: Created #936.

PR #935, opened 2021-03-31
Avoid spawning new type-checking task when disposing CompilationUnitManager

samarsha:
This fixes a bug in `CompilationUnitManager.Dispose` that caused a new type-checking task to start when the compilation unit manager was being disposed. The result was that type checking always ran twice whenever a Q# project was built from the command line.

Some performance tests I ran on the standard library show build time (excluding reference loading, output generation and rewrite steps) reduced by 42% (from 47 seconds to 27 seconds).


bettinaheim (approved):
Yay, good find!!

Issue #916, opened 2021-03-22
Invalid LLVM code from unreachable Q# code


msoeken:
**Describe the bug**

Invalid LLVM code is generated in a Q# example with unreachable code.

**To Reproduce**

Run the following project:

```xml
<Project Sdk="Microsoft.Quantum.Sdk/0.15.2103131668-beta">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp5.0</TargetFramework>
    <PlatformTarget>x64</PlatformTarget>
    <QirGeneration>true</QirGeneration>
    <CSharpGeneration>false</CSharpGeneration>
    <IncludeQSharpCorePackages>false</IncludeQSharpCorePackages>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.Quantum.QSharp.Foundation" Version="0.15.2103131668-beta" />
  </ItemGroup>

</Project>
```

```qsharp
// Program.qs
namespace Example {
    open Microsoft.Quantum.Intrinsic;

    @EntryPoint()
    operation RunProgram() : Unit {
        let _ = Foo();
        return ();
        let _ = Foo();
    }

    internal operation Foo() : Bool {
        return true;
    }
}
```

```csharp
// Program.cs
return 0;
```

Then run `clang++ -c qir/project.ll`, which yields:

```
qir/project.ll:19:3: error: instruction expected to be numbered '%2'
  %1 = call i1 @Example__Foo__body()
  ^
1 error generated.
```

**Generated LLVM**

```llvm

%Result = type opaque
%Range = type { i64, i64, i64 }

@ResultZero = external global %Result*
@ResultOne = external global %Result*
@PauliI = constant i2 0
@PauliX = constant i2 1
@PauliY = constant i2 -1
@PauliZ = constant i2 -2
@EmptyRange = internal constant %Range { i64 0, i64 1, i64 -1 }

@Example__RunProgram = alias void (), void ()* @Example__RunProgram__body

define void @Example__RunProgram__body() #0 {
entry:
  %0 = call i1 @Example__Foo__body()
  ret void
  %1 = call i1 @Example__Foo__body()
  ret void
}

define i1 @Example__Foo__body() {
entry:
  ret i1 true
}

attributes #0 = { "EntryPoint" }
```

**System information**

- Version: 0.15.2103131668-beta
- macOS 11.2.2

**Additional context**

[...]

PR #941, opened 2021-03-31
Support QIR generation of unreachable code

kevinhartman:
The original motivation behind this changeset was to propose a fix for #916 at the QIR level. As @bettinaheim mentions there, this could instead be fixed by pruning unreachable code before generating QIR, but I wanted to explore this option.

To support this, the `FunctionContext` class owns/encapsulates the current function, current basic block, and current instruction builder. Access to the current instruction builder is (by convention) brokered by `FunctionContext.Emit`, which gives us a chance to perform validation before and after a sequence of instructions is emitted. This mechanism is used to detect what would be code emission to an already terminated `BasicBlock`, instead injecting a new block to hold the unreachable code. Perhaps it could be used for logging or other validation as well.

The ideal usage pattern for this would be to call `Emit` once per current block (change) per subtransformation of the AST. I.e. do all code generation inside a single `Emit` until you need to change the current block. However, you can also call `Emit` for each and any instruction, at the cost of running the validation logic for each.

To avoid putting too much work into this before getting opinions, I did not attempt to aggressively group calls to `Emit`. To do that cleanly would likely require adding an instruction builder parameter to many of the helper functions used for code generation. However, in cases where calls to `Emit` are nested (e.g. a helper function that calls `Emit` is itself called from inside an `Emit`), validation only occurs for the outer `Emit`, treating all code generated as part of the same instruction sequence.

It's probably worth noting that this pattern does not prevent someone from creating their own `InstructionBuilder` to avoid validation. I also recognize that this pattern might introduce more complexity than it's worth. If that is indeed the consensus, I'm happy to ditch the `Emit` stuff (I think `FunctionContext` itself is still a nice abstraction) and expose the current `InstructionBuilder` directly on the `FunctionContext`, but this will of course retract support for unreachable QIR generation.

#### Input program
```qs
namespace Microsoft.Quantum.Qir.Emission {

    @EntryPoint()
    operation RunProgram() : Unit {
        let _ = Foo();
        return ();
        let _ = Foo();
    }

    internal operation Foo() : Bool {
        return true;
    }
}
```

#### Before:
Note that this is not valid LLVM IR. See #916 for details.
https://gist.github.com/kevinhartman/2cc4fe52d207bdebf60a3a9a77e871df#file-main-emission-ll-L17

#### After:
https://gist.github.com/kevinhartman/4235e37bf48d8d0be80badbb639ac36b#file-functioncontext-emission-ll-L23


kevinhartman:
I will try to come back to this to add a bit more context and alternative approaches I considered to the description, but I do not have time right now.

I do not expect this to be merged in its current form, but am creating it as a draft for visibility / feedback.

Thanks!

kevinhartman:
All tests are now passing.

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestUnreachable.ll
```diff
@@ -0,0 +1,42 @@
...
+  ret i64 %c
+
+then0__1__unreachable__1:                         ; No predecessors!
+  %d = add i64 2, %a
+  %1 = call %String* @__quantum__rt__string_create(i8* getelementptr inbounds ([14 x i8], [14 x i8]* @0, i32 0, i32 0))
+  call void @__quantum__rt__fail(%String* %1)
+  unreachable
+
+then0__1__unreachable__2:                         ; No predecessors!
+  call void @__quantum__rt__string_update_reference_count(%String* %1, i32 -1)
```
kevinhartman, line 17: Note that the ref decrement calls in these unreachable blocks are generated because each of these blocks exit from the scope of the original allocation, either via a terminal statement or control reaching the end of the block.

This behavior is not new in this PR. The only difference is now we explicitly label these unreachable blocks to ensure that we end up with valid LLVM IR.

kevinhartman:
Now includes a nice shiny new QirTest for unreachable block validation.

swernli:
Apologies for the delay, I should be able to review this today!

kevinhartman:
No worries, @swernli!

Let me know if you need any more context for the approach. I was hoping to have a chance to improve this PR description and add some alternatives I'd considered.

In summary, this works by wrapping any instruction builder calls so we can detect situation where instructions are emitted to a block that has already been terminated, and instead create a new unreachable block to hold these instructions.

Alternatives considered:
1) I thought about processing the CFG for a function once it is complete, and just splitting blocks up on terminal instructions, but I couldn't find a way to do that with the LLVM C# wrapper API. This might also be complicated because control flow instructions reference blocks, so perhaps you would need to reconstruct those instructions (just a guess).
2) At first (I believe in an early commit here), I tried to take advantage of the fact that in cases where we start emitting unreachable code, the current block would be terminated, but no new call would happen to change the current block. My approach was to:
      - In `Context.AddReturn` and `StatementKindTransformation.OnFail`, always create a new unreachable block and use `SetCurrentBlock` to set it current.
      - On the next call to `SetCurrentBlock`, if the current block is empty, prune it (avoids empty unreachable blocks from cropping up everywhere).

After attempting 2), I ended up going with the approach in this PR instead, since it seemed kind of kludgy, and would only work if callsites that add terminal instructions to blocks (e.g. `OnFail`) are sure to swap in a new unreachable block afterward.

src/QsCompiler/QirGeneration/FunctionContext.cs
```diff
@@ -0,0 +1,251 @@
...
+                var currentInstCount = 0;
+
+                if (this.IsCurrentBlockTerminated)
+                {
+                    // Current block has already been terminated.
+                    // Create a new block to hold any unreachable instructions emitted.
+                    // If no instructions are emitted by func(), we will delete this block.
+                    // For this reason, we do not reserve a unique block name until we know
+                    // we'll keep the block.
+                    var unreachableBlock = this.AddBlockAfterCurrent("temp__unreachable", blockNameAllocate: null);
```
swernli, line 128: Is it possible that the temporary name may accidentally end up in the unreachable code? For example, if there is a for loop in the unreachable block, it will get generated before the name is updated. Does that end up needing to use the containing block's name?

swernli: Heh, unit tests speak louder than words :) Thanks for adding the test cases the demonstrate the behavior here! (https://github.com/microsoft/qsharp-compiler/pull/941/commits/8d858acc0b1283862538a3d9ba695db03cfc09b2)

swernli (approved):

swernli:
For the style failure, you can use the formatting tool Fantomas: https://github.com/fsprojects/fantomas/

src/QsCompiler/QirGeneration/FunctionContext.cs
kevinhartman: I'm not sure I'm following completely, but are you referring to the case that an instruction inside an `Emit` call references the current block (e.g. `PhiNode`)? For the temporary name to be the current block's name, it would have to be read inside an `Emit`. I think even in that case, instructions like `PhiNode` take the block by reference, and not name, so I'd wager that the emitted PhiNode instruction would end up with the updated name. I think that'd be worth testing prior to merge.

Something I didn't originally consider (maybe this is actually what you were getting at) is that callers need to know that `Emit` might change the current block.

For example, in this case, `precedingBlock` should probably be defined only after the first call to `Emit`, since the block might change, which would then result in an incorrect `PhiNode`.

```c#
// End the current block by branching into the header of the loop
BasicBlock precedingBlock = this.FunctionContext.CurrentBlock;

// We need to open a scope before starting the for-loop by creating the header block, since
// it is possible for the condition to perform an allocation that needs to get cleaned up.
this.ScopeMgr.OpenScope();
this.FunctionContext.Emit(b => b.Branch(headerBlock));

// Header block: create/update phi node representing the iteration variable and evaluate the condition
this.FunctionContext.SetCurrentBlock(headerBlock);
var loopVariable = this.FunctionContext.Emit(b => b.PhiNode(this.Types.Int));
loopVariable.AddIncoming(startValue, precedingBlock);
```

This specific case doesn't cause an issue in the (newly added :)) unit test, because a bit of code gets emitted before this method is called to calculate the loop end.

I think if we do go with this approach, it's very important to have unit tests that demonstrate that all statements after a terminal one work as expected, since it'd probably be pretty easy to forget that `Emit` might change the block.

That said, if this seems too difficult to reason about, I'd completely understand. There's always the possibility of just pruning the AST of unreachable code prior to QIR generation, but I just thought it'd be cool if the generation could support this.

kevinhartman: > Heh, unit tests speak louder than words :)

I promise I was working on the words too! I just thought I would write some tests to see if it'd help me understand what you were saying :)

kevinhartman:
Thanks for the review, @swernli! 

I think this one requires a little bit more discussion prior to merge. While this does fix handling of unreachable code, I worry that it might add a little bit too much complexity. Specifically, it might be somewhat unexpected to callers that `Emit` may change the current block, even if that's documented.

I think this depends on whether or not it's important for QIR generation to support unreachable code in the AST. If this isn't important, I'm happy to remove the `Emit` pattern from the PR. IMO, `FunctionContext` could still be a valuable factoring of control flow building functions and state management.

kevinhartman:
No problem. I'll reopen this for now, and depending on what you want to preserve, I'm happy to close this again and create a separate PR with those changes applied to main. I'll enable edits by maintainers on my branch as well in case it's easier to make any changes while deciding what to keep.

kuzminrobin:
Does it make sense to move forward or to close this PR? 
About 1 year old.

swernli:
@kuzminrobin Thanks for flagging this! I think this is far enough out of date that it makes sense to close the current version and revisit with the changes rebased onto the latest main.

@kevinhartman I'm going to close this PR. If you have the time and are still interested, we'd be happy to revisit a new version of this fix. If not, let us know and we can adapt your changes in a new working branch. Thanks again for working on this, apologies for letting it languish!

PR #942, opened 2021-04-01
Monomorphize intrinsics when keepAllIntrinsics is false

bettinaheim:
This PR is required to support DumpMachine and DumpRegister.


kuzminrobin (approved):
(I hardly understand the code ;-)

ScottCarda-MS (commented):
Unfortunately these changes conflict greatly with the changes for the Trim Rewrite Step, as the Monomorphization will no longer have the `keepAllIntrinsic` behavior. Because of that, I would prefer to add the fix to the issue to that PR instead so that it is well integrated. @bettinaheim could we chat offline about what is best to do here?

ScottCarda-MS (approved):
This looks good. Thanks for making the changes to retarget this to my branch!

PR #945, opened 2021-04-05
Create QIR controller input schema

achocron:
Creates the input schema.


cesarzc:
@achocron, can you please target this PR to the [feature/azure-quantum-simulator](https://github.com/microsoft/qsharp-compiler/tree/feature/azure-quantum-simulator) branch?

src/QsCompiler/BondSchemas/SandboxInput/SandboxInput.bond
```diff
@@ -0,0 +1,16 @@
...
+
+import "EntryPoint/EntryPoint.bond";
+namespace Microsoft.Quantum.QsCompiler.BondSchemas.SandboxInput;
+
+struct Bytecode
+{
+    5: blob Data;
+}
+
+struct Input
```
cesarzc, line 12: A more descriptive name for this `struct` would be `QirExecutionWrapper`. I would change the name of the folder containing the *.bond file to _Qir_ and the file to _QirExecutionWrapper.bond_.

src/QsCompiler/BondSchemas/SandboxInput/SandboxInput.bond
```diff
@@ -0,0 +1,16 @@
...
+{
+    5: Microsoft.Quantum.QsCompiler.BondSchemas.EntryPoint.EntryPointOperation EntryPoint;
+    10: bonded<Bytecode> QirBytes;
```
cesarzc, line 15: Was there a reason wrap the `Blob` inside a struct rather than using `blob` directly here?

src/QsCompiler/Tests.Compiler/SandboxInputSchemaTests.fs
```diff
@@ -0,0 +1,52 @@
...
+open Microsoft.Quantum.QsCompiler.BondSchemas.EntryPoint
+open Microsoft.Quantum.QsCompiler.BondSchemas.SandboxInput
+
+type SandboxInputSchemaTests(output: ITestOutputHelper) =
+
+    let createEntryPointOperation() =
+        let expectedEntryPointOperation = new EntryPointOperation()
+        let argument = new Argument()
+        argument.Name <- "name"
+        argument.Position <- 44
```
cesarzc, line 21: Nit: consider using more explicit names like "Test Argument" and a valid position in this case (which would be 0 since there is only 1 argument). Same for the name of the entry-point operation.

src/QsCompiler/Tests.Compiler/SandboxInputSchemaTests.fs
```diff
@@ -0,0 +1,52 @@
...
+        // Serialize input.
+        Protocols.SerializeToFastBinary(sandboxInput, memoryStream)
+
+        // Deserialize and confirm that it is correct.
+        let deserializedInput = Protocols.DeserializeFromFastBinary(memoryStream)
+        Assert.True(Extensions.ValueEquals(deserializedInput.EntryPoint, entryPoint))
+        let deserializedBytecode = deserializedInput.QirBytes.Deserialize()
+        Assert.Equal(deserializedBytecode.Data.Count, bytecode.Data.Count)
+        for i in 0..bytecode.Data.Count - 1 do
+            Assert.Equal(deserializedBytecode.Data.Item(i), bytecode.Data.Item(i))
```
cesarzc, line 52: This can be encapsulated in a `ValueEquals` extension method for the Bond generated class that contains the entry-point and the QIR bytecode.

src/QsCompiler/Tests.Compiler/Tests.Compiler.fsproj
```diff
@@ -17,6 +17,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <Compile Include="SandboxInputSchemaTests.fs" />
```
cesarzc, line 20: Once you rename the Bond-related files, it would be good to rename the test file too.

src/QsCompiler/BondSchemas/SandboxInput/SandboxInput.bond
achocron: Yeah, I guess you can only use `bonded` with structs. It might be fine to use a `blob` without `bonded` though? Not sure.

src/QsCompiler/BondSchemas/QirExecutionWrapper/Extensions.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// This class provides extension methods for objects in the Microsoft.Quantum.QsCompiler.BondSchemas.QirExecutionWrapper namespace.
+    /// </summary>
+    internal static class Extensions
+    {
+        /// <summary>
+        /// Determine whether the values of two object instances are equal.
+        /// </summary>
+        public static bool ValueEquals(this QirExecutionWrapper executionWrapperA, QirExecutionWrapper executionWrapperB)
+        {
+            if (executionWrapperA == null && executionWrapperB == null)
```
cesarzc, line 20: You can add a `#nullable enable` pragma at the beginning of this source file and you won't need to check for null here which would make the code cleaner.

src/QsCompiler/BondSchemas/QirExecutionWrapper/Protocols.cs
```diff
@@ -0,0 +1,44 @@
...
+            var reader = new FastBinaryReader<InputStream>(inputStream);
+            var deserializer = new Deserializer<FastBinaryReader<InputStream>>(typeof(QirExecutionWrapper));
+            var sandboxInput = deserializer.Deserialize<QirExecutionWrapper>(reader);
+            return sandboxInput;
+        }
+
+        /// <summary>
+        /// Serializes a sandbox input object to its fast binary representation.
+        /// </summary>
+        public static void SerializeToFastBinary(QirExecutionWrapper sandboxInput, Stream stream)
```
cesarzc, line 33:
```suggestion
        public static void SerializeToFastBinary(QirExecutionWrapper qirExecutionWrapper, Stream stream)
```

cesarzc, line 33: Nit: rename `sandboxInput` to `qirExecutionWrapper`.

src/QsCompiler/BondSchemas/QirExecutionWrapper/Protocols.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// </summary>
+    internal static class Protocols
+    {
+        /// <summary>
+        /// Deserializes a sandbox input object from its fast binary representation.
+        /// </summary>
+        public static QirExecutionWrapper DeserializeFromFastBinary(Stream stream)
+        {
+            stream.Flush();
+            stream.Position = 0;
```
cesarzc, line 22: When performing deserialization from Streams, I think it is a good idea to not perform any flush or change in position operation. The main reason is that it is the responsibility of the caller to do these operations. Another reason (which in our case is not that relevant) is that the caller might have multiple things in the same stream and it has set the position of the stream where it should start reading in order to deserialize a specific object.

src/QsCompiler/BondSchemas/QirExecutionWrapper/QirExecutionWrapper.bond
```diff
@@ -0,0 +1,16 @@
...
+
+struct Bytecode
+{
+    5: blob Data;
+}
+
+struct QirExecutionWrapper
+{
+    5: Microsoft.Quantum.QsCompiler.BondSchemas.EntryPoint.EntryPointOperation EntryPoint;
+    10: bonded<Bytecode> QirBytes;
```
cesarzc, line 15:
```suggestion
    10: bonded<Bytecode> QirBytecode;
```

src/QsCompiler/BondSchemas/SandboxInput/SandboxInput.bond
cesarzc: I think you are right. In this case there is nothing to deserialize from just a `blob` field since it just contains "raw" bytes.  Then here it should be just a `blob` field.

src/QsCompiler/BondSchemas/QirExecutionWrapper/QirExecutionWrapper.bond
cesarzc: This both a rename and a reference to [this comment](https://github.com/microsoft/qsharp-compiler/pull/945#discussion_r607437511).

cesarzc:
@achocron since the [PR](https://github.com/microsoft/qsharp-runtime/pull/610) in the qsharp-runtime repo meant to fix the build issues is already merged, I think we should revert the merge from `main` in this PR to be able to merge it to the feature branch. Afterwards we can merge from `main` to the feature branch as a separate PR.

src/QsCompiler/BondSchemas/QirExecutionWrapper/Extensions.cs
cesarzc: I think this is still not resolved.

cesarzc (approved):
Looks good, Thanks @achocron!

PR #946, opened 2021-04-05
Unit tests in Executables

ScottCarda-MS:
Updates call graph classes to process unit types as starting points to the call graph, similar to how entry points are used. This allows unit tests to be used in executables ensuring they will not be trimmed and their dependencies will be properly monomorphized.


src/QsCompiler/Tests.Compiler/TestCases/PopulateCallGraph.qs
```diff
@@ -134,6 +134,81 @@ namespace Microsoft.Quantum.Testing.PopulateCallGraph {
...
+
+    operation Foo() : Unit { }
+
+    operation Bar() : Unit { }
+
+    operation Zip() : Unit { }
+
+    operation Zap() : Unit { }
+
+    operation Unused() : Unit { }
```
kuzminrobin, line 207: (Minor)
Would be nice to see a code comment telling what this `Unused()` is for.

ScottCarda-MS: Unused is to test that unused callables are either included in the call graph or not, depending on the situation and unit test.

src/QsCompiler/Tests.Compiler/TestCases/PopulateCallGraph.qs
```diff
@@ -134,6 +134,81 @@ namespace Microsoft.Quantum.Testing.PopulateCallGraph {
...
+    open Microsoft.Quantum.Diagnostics;
+
+    @Test("QuantumSimulator")
+    operation Test() : Unit {
+        Foo();
+    }
+
+    operation Foo() : Unit { }
+
+    operation Bar() : Unit { }
```
kuzminrobin, line 169: (Minor)
It is somewhat suspicious to see multiple _identical_ definitions (for `Foo()`, `Bar()`, `Main1()`, `Main2()`) in the same namespace. It causes some redundant questions, e.g. "What if one of the two definitions of `Foo()` is updated, but the other is not? Will that cause a compiler error? If yes, then will that error be clear?".

What if in green lines 167 - 169 we remove the definitions for `Foo()` and `Bar()` (assuming that we reuse those from lines 150 - 152)?
Same for `Main1()` and `Main2()`.

src/QsCompiler/Tests.Compiler/TestCases/PopulateCallGraph.qs
ScottCarda-MS: I could add a comment to the corresponding test logic, but then commenting a line line `AssertNotInConcreteGraph graph Unused` seems a bit unnecessary to me. It's hard to be more descriptive than that 😊.

src/QsCompiler/Tests.Compiler/TestCases/PopulateCallGraph.qs
ScottCarda-MS: The testing logic breaks up the unit tests into totally separate compilations. The edges of the 'chunks' are the lines with the `// =================================`. So the code from lines 150 - 152 are not present in the compilation when we run the test case starting at line 158.

ScottCarda-MS: To answer the questions you posted, having multiple callables with the same name in the same namespace will cause an error in Q#. As I said this is not an issue here because the different 'chunks' are compiled completely separate from each other.

src/QsCompiler/Tests.Compiler/TestCases/PopulateCallGraph.qs
```diff
@@ -307,3 +388,75 @@ namespace Microsoft.Quantum.Testing.PopulateCallGraph {
...
+
+// Concrete Graph Entry Points and Unit Tests
+namespace Microsoft.Quantum.Testing.PopulateCallGraph {
+
+    open Microsoft.Quantum.Diagnostics;
+
+    @EntryPoint()
+    operation Main1() : Unit {
+        Foo<String>();
+    }
```
kuzminrobin, line 440: (I'm not confident in .qs, my note may be mistaken)

(At least for C/C++) It is suspicious to see the same name (`Main1`) for different function definitions in the same namespace. One definition is
```qsharp
    @EntryPoint()
    operation Main1() : Unit {
        Foo<Double>();
    }
```
another one is
```qsharp
    @EntryPoint()
    operation Main1() : Unit {
        Foo<String>();
    }
```
If some fragment calls `Main1()` then which definition will be called?

kuzminrobin (approved):
Approving with some questions to your discretion.
I only understand .qs. I recommend at least one more review/approval for .fs, .cs.

src/QsCompiler/Tests.Compiler/TestCases/PopulateCallGraph.qs
ScottCarda-MS: This file with Q# test cases is separated into chunks by the lines with `// =================================`. This is done by the unit test driving logic. Each chunk is compiled totally independent of the other chunks. That is to say that each test case is isolated from the other test cases. In the test case that you linked to above, there is an operation named `Main1` for the first entry point, and an operation for `Main2` for the second entry point. If they are named the same, then that is an accident and is incorrect. Q# does not allow multiple callables with the same name in the same namespace when it compiles the code.

kuzminrobin: Based on your other feedback, this is not applicable any more. Resolving.

PR #950, opened 2021-04-07
Make interface public

achocron:
Expose the deserializer so that it can be used in the QIR controller.


src/QsCompiler/BondSchemas/QirExecutionWrapper/Protocols.cs
```diff
@@ -28,7 +28,7 @@ public static QirExecutionWrapper DeserializeFromFastBinary(Stream stream)
         /// <summary>
         /// Serializes a QirExecutionWrapper object to its fast binary representation.
         /// </summary>
-        public static void SerializeToFastBinary(QirExecutionWrapper sandboxInput, Stream stream)
+        internal static void SerializeToFastBinary(QirExecutionWrapper sandboxInput, Stream stream)
```
cesarzc, line 31: I think this should be public. In the case of the entry-point, the serialization/deserialization methods were meant only for testing so they are marked as internal. In this case, these methods are meant to be used by other components so they should be public.

cesarzc (approved):

PR #953, opened 2021-04-08
Fix #949.

### LINKED ISSUE #949: Docs gen issue in AssertMeasurement[Probability](): The type of `result` param is `invalid`
It is confusing that the type of the `result` parameter is `invalid` [here](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.diagnostics.assertmeasurement#result--invalid) and [here](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.diagnostics.assertmeasurementprobability#result--invalid). 
Documentation rendering issue?

---
Same for [Measure() `Output`](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.intrinsic.measure#output--invalid) (and all entities of type `Result`?).

cgranade:
This PR fixes a regression in the documentation generation pipeline that causes some inputs of type `Result` to be formatted as `invalid` instead (#949).

### DIFF
--- a/src/Documentation/DocumentationGenerator/Extensions.cs
+++ b/src/Documentation/DocumentationGenerator/Extensions.cs
@@ -306,6 +306,7 @@ namespace Microsoft.Quantum.Documentation
                         ResolvedTypeKind.Tags.Range => "[Range](xref:microsoft.quantum.qsharp.valueliterals#range-literals)",
                         ResolvedTypeKind.Tags.String => "[String](xref:microsoft.quantum.qsharp.valueliterals#string-literals)",
                         ResolvedTypeKind.Tags.UnitType => "[Unit](xref:microsoft.quantum.qsharp.valueliterals#unit-literal)",
+                        ResolvedTypeKind.Tags.Result => "[Unit](xref:microsoft.quantum.qsharp.valueliterals#result-literal)",
                         ResolvedTypeKind.Tags.InvalidType => "__invalid__",
                         _ => $"__invalid<{type.Resolution.ToString()}>__",
                     },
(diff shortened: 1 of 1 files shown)


src/Documentation/DocumentationGenerator/Extensions.cs
```diff
@@ -306,6 +306,7 @@ item.Item.VariableName switch
                         ResolvedTypeKind.Tags.Range => "[Range](xref:microsoft.quantum.qsharp.valueliterals#range-literals)",
                         ResolvedTypeKind.Tags.String => "[String](xref:microsoft.quantum.qsharp.valueliterals#string-literals)",
                         ResolvedTypeKind.Tags.UnitType => "[Unit](xref:microsoft.quantum.qsharp.valueliterals#unit-literal)",
+                        ResolvedTypeKind.Tags.Result => "[Unit](xref:microsoft.quantum.qsharp.valueliterals#result-literal)",
```
samarsha, line 309:
```suggestion
                        ResolvedTypeKind.Tags.Result => "[Result](xref:microsoft.quantum.qsharp.valueliterals#result-literal)",
```

samarsha (approved):
If only C# had `switch` exhaustiveness checking...

cgranade:
> If only C# had `switch` exhaustiveness checking...

Weirdly, it does, but that feature is undermined by the fact that, in order to support bitflag-style enums, C# `enum`s can have any valid `int` value. Thus, switching requires a `_` pattern, even if the switch is already exhaustive without it, making it really easy to miss cases like this.

Anyway, thanks for reviewing and for catching the bad link!

PR #954, opened 2021-04-08
Added a code action for auto-fixing deprecated qubit binding syntax

filipw:
This PR introduces code action to fix the deprecated qubit binding syntax.
I believe this would be a helpful change for the users in the short term - as it allows them to easily deal with the recent changes to the binding syntax, and opens up a better path to fully remove the old syntax instead of just deprecating it.

Since Q# compiler doesn't have "round-trippable" ("lossless") syntax trees at the moment, since QsFmt is not yet integrated in the main compiler code and since the change here is a rather simple one (it is a glorified trim start/trim end) I implemented it using a simple regex (and even that is not really needed, the same could really be achieved with string operations).

In case the regex fails, the fallback is to simply replace the keyword `using -> use`, `borrowing -> borrow`, instead of trying to handle the entire fragment.

Example:

Bildschirmaufnahme 2021-04-03 um 10 07 00-2

The rule for the following curly brace are:
 - if it was on the same line as the binding keyword, it stays on the same line
 - if it was on its own line, it stays on its own line
 - if it was on a different line than the binding keyword, but shared that line with the closing `)` bracket, it gets pulled up

Note that the code action is only offered on the `WarningCode.DeprecatedQubitBindingKeyword` (so really on the `using`/`borrowing` keyword) - it could be offered on the `WarningCode.DeprecatedTupleBrackets` but I wanted to keep it simple.


src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
```diff
@@ -348,6 +351,26 @@ where DiagnosticTools.ErrorType(ErrorCode.UnknownType)(diagnostic)
             var orSuggestions = diagnostics.Where(DiagnosticTools.WarningType(WarningCode.DeprecatedORoperator))
                 .Select(d => ReplaceWith(Keywords.qsORop.op, d.Range));
 
+#pragma warning disable 0618
```
bettinaheim, line 354: One minor thing: I would prefer putting this only on the two lines that actually need it.

bettinaheim (approved):
Thanks for setting this up! 
I did a quick validation locally, and it seems to be working terrifically.

src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
bettinaheim: Thanks!

filipw:
cc @tcNickolas who asked about something like this in another issue

tcNickolas:
@filipw Thank you for doing this! Too bad we already went through the Katas to update this syntax by hand :-) I'm glad, though, that after the next release this will be a more automated process!

PR #956, opened 2021-04-08
Deprecate new array expression

samarsha:


src/QsCompiler/Tests.Compiler/GlobalVerificationTests.fs
```diff
@@ -552,7 +552,14 @@ type GlobalVerificationTests() =
         this.Expect "InvalidAttributes5" [ Error ErrorCode.UnknownTypeInNamespace ]
         this.Expect "InvalidAttributes6" [ Error ErrorCode.AttributeArgumentTypeMismatch ]
         this.Expect "InvalidAttributes7" [ Error ErrorCode.InvalidAttributeArgument ]
-        this.Expect "InvalidAttributes8" [ Error ErrorCode.ArgumentOfUserDefinedTypeInAttribute ]
+
+        this.Expect
```
swernli, line 556: nit: why is this one formatted differently? Earlier in the file on line 35 you have one that includes the error and warning together with a comma instead of separate lines.

swernli: actually, looking at the other files it appears that the single line one is the outlier, so I suppose the nit should go up there :)

swernli (approved):

src/QsCompiler/Tests.Compiler/GlobalVerificationTests.fs
samarsha: The F# formatting is completely automated, so I didn't make these decisions, Fantomas did. :) Fantomas has a threshold for when to split a list into multiple lines based on the length of the list literal; the single line one is shorter than the others.

PR #960, opened 2021-04-09
use "constant" completion item kind for Q# literals

filipw:
At the moment the Q# literals (`PauliI`, `PauliX`, `PauliY`, `PauliZ`, `One`, `Zero`) use "keyword" completion kind, the same as all other keywords. 

However, this feels weird, because they are used in a different way from "regular" language keywords. I think it will be more appropriate and helpful for the user to convey them as "constants". This ends up rendering a different glyph in the editor, which immediately helps distinguish their nature.

I left `true/false` as-is, even though they are literals too, in that case it makes sense IMHO.

### Before

Bildschirmfoto 2021-04-09 um 13 27 25
Bildschirmfoto 2021-04-09 um 13 27 05

### After

Bildschirmfoto 2021-04-09 um 13 26 27
Bildschirmfoto 2021-04-09 um 13 25 20


bettinaheim:
@filipw Thanks for the PR! I am curious about not treating true/false as literals; do other languages not mark them as literals in the IDE, or what is your reasoning here?

filipw:
To be fair, as far as I am aware, all editors treat and present keyword literals the same way as all other keywords (we certainly do that in O# too). However, most often keyword literals are roughly the same familiar set across languages - `true`, `false`, `null`, `self`, `nullptr` etc, and developers are very used to them.

In this case both `Paulis` and `One`/`Zero` are so much different than anything encountered in other languages, and at least my mental model of them was not keywords but rather "special values". So I thought it would make sense to amplify that and make them stand out in the intellisense selection (all the possible glyphs are listed here https://code.visualstudio.com/docs/editor/intellisense#_types-of-completions) and thus bring a bit more life into the intellisense.

However, if this doesn't make much sense we can of course close this PR.

bettinaheim:
> To be fair, as far as I am aware, all editors treat and present keyword literals the same way as all other keywords (we certainly do that in O# too). However, most often keyword literals are roughly the same familiar set across languages - `true`, `false`, `null`, `self`, `nullptr` etc, and developers are very used to them.
> 
> In this case both `Paulis` and `One`/`Zero` are so much different than anything encountered in other languages, and at least my mental model of them was not keywords but rather "special values". So I thought it would make sense to amplify that and make them stand out in the intellisense selection (all the possible glyphs are listed here https://code.visualstudio.com/docs/editor/intellisense#_types-of-completions) and thus bring a bit more life into the intellisense.
> 
> However, if this doesn't make much sense we can of course close this PR.

I am happy to have that change. I'd prefer having true and false treated the same way, since from a language perspective they are really no different than the other ones (just more common in other languages).

filipw:
that's fair - I changed true/false keyword literals to be handled the same way. Now all the keyword literals in the language are presented in the completion with the "constant" glyph which helps them stand out.

bettinaheim (approved):

PR #963, opened 2021-04-12
Handle SizedArray in code output and optimizations

samarsha:
Addresses issues brought up in #884.

In some cases, I replaced a catch-all match case with every individual expression kind to make the match exhaustive, which should prevent these matches from being missed when new expression kinds are added in the future.


bettinaheim (approved):
Looks fine.

PR #965, opened 2021-04-12
Generate command line arguments string based on model

achocron:
This adds logic to generate a string containing the command line arguments to the driver program based on the entry point's arguments.

### DIFF
--- a/src/QsCompiler/Compiler/QirDriverGeneration.cs
+++ b/src/QsCompiler/Compiler/QirDriverGeneration.cs
@@ -1,8 +1,10 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Text;
 using Microsoft.Quantum.QsCompiler.BondSchemas.EntryPoint;
 using Microsoft.Quantum.QsCompiler.Templates;
@@ -11,6 +13,8 @@ namespace Microsoft.Quantum.QsCompiler
 {
     public static class QirDriverGeneration
     {
+        private delegate string StringConversion(object? value);
+
         public static void GenerateQirDriverCpp(EntryPointOperation entryPointOperation, Stream stream)
         {
             QirDriverCpp qirDriverCpp = new QirDriverCpp(entryPointOperation);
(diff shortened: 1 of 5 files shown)


src/QsCompiler/Compiler/QirDriverGeneration.cs
```diff
@@ -18,5 +23,137 @@ public static void GenerateQirDriverCpp(EntryPointOperation entryPointOperation,
             stream.Flush();
             stream.Position = 0;
         }
+
+        public static string GenerateCommandLineArguments(IList<Argument> arguments)
```
cesarzc, line 27: Since this is a public function, can you please add documentation comments (summary, arguments, return value)?

src/QsCompiler/Compiler/QirDriverGeneration.cs
```diff
@@ -18,5 +23,137 @@ public static void GenerateQirDriverCpp(EntryPointOperation entryPointOperation,
...
+            static string ConvertArray<T>(IList<T> list, StringConversion conversion)
+            {
+                return list.Aggregate<T, string>(string.Empty, (aggregation, val) =>
+                {
+                    if (aggregation != string.Empty)
+                    {
+                        aggregation += ' ';
+                    }
+                    return aggregation + conversion.Invoke(val);
+                });
```
cesarzc, line 74: This looks a lot like `String.Join`. I think the following would be equivalent:
```csharp
return String.Join(" ", list.Select(item => conversion.Invoke(item));
```

src/QsCompiler/Compiler/QirDriverGeneration.cs
```diff
@@ -10,6 +13,8 @@ namespace Microsoft.Quantum.QsCompiler
 {
     public static class QirDriverGeneration
     {
+        private delegate string StringConversion(object? value);
```
cesarzc, line 16: It might be worth making this a generic delegate:
```csharp
private delegate string StringConversion<T>(T? value);
```

That way the `GetBoolValueString`, `GetDoubleValueString`, and the other similar methods can be typed so you don't have to do a dynamic cast inside the functions. It is also more type-safe that way because if a type error occurs, the compiler will catch it.

achocron: This doesn't actually work because you can't case these methods to a delegate with type arguments because they themselves do not have type arguments

achocron: Also, `T?` is not supported in this context until C# 9

src/QsCompiler/Tests.Compiler/QirDriverGenerationTests.fs
```diff
@@ -0,0 +1,114 @@
...
+
+open Xunit
+open Xunit.Abstractions
+open Microsoft.Quantum.QsCompiler
+open Microsoft.Quantum.QsCompiler.BondSchemas.EntryPoint
+
+type QirDriverGenerationTests(output: ITestOutputHelper) =
+
+    [<Fact>]
+    member this.GenerateCommandLineArguments() =
```
cesarzc, line 14: This approach seems difficult to scale. I would recommend to use an approach similar to the one used in `EntryPointSchemaTests`: https://github.com/microsoft/qsharp-compiler/blob/4b97de3ff0edf16969d918cf0bcf499697f9fea0/src/QsCompiler/Tests.Compiler/EntryPointSchemaTests.fs#L354

For the purposes of this PR, can you please add just this one test case in a setup similar to `EntryPointSchemaTests`? The reason for that is that this other [PR](https://github.com/microsoft/qsharp-compiler/pull/929) consolidates test cases in text files at `src\QsCompiler\Tests.Compiler\TestCases\QirEntryPointTests`. Once we have the test cases there, they can be used by multiple tests, including this one.

cesarzc: Also, maybe this test could be moved to [this](https://github.com/microsoft/qsharp-compiler/blob/feature/azure-quantum-simulator/src/QsCompiler/Tests.Compiler/EntryPointGenerationTests.fs) other source file.

cesarzc: This feedback will be addressed in a subsequent PR to unblock other on-going development in the feature branch.

cesarzc (approved):

src/QsCompiler/Tests.Compiler/QirDriverGenerationTests.fs
achocron: Discussed offline; will address in a separate PR

PR #969, opened 2021-04-15
Example for local development of qir emission

bettinaheim:
This PR adds an example project that is compiled using the source code version of the compiler and QIR generation. This is convenient for trouble shooting when it is not clear what the problematic code is and we merely have the Q# project to reproduce the issue.


examples/QIR/QIR.sln
```diff
@@ -5,6 +5,38 @@ VisualStudioVersion = 16.0.29613.14
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Emission", "Emission\Emission.csproj", "{A8D9184C-C211-42D2-9F48-4D2DBE810A1C}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Development", "Development\Development.csproj", "{94ECCC2D-96E7-4CFE-A37F-37266C091BC1}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Compiler", "Compiler", "{D1D740B2-1AC8-46EC-9192-4C8117AAE345}"
```
cesarzc, line 10: I am curious about why many of these projects have to be part of this solution if they are already part of the compiler solution?

bettinaheim: I would love to not have them, but the development project has a reference to the compiler and the QIR generation such that it can use the source code version to compile the project. Unfortunately, I have to have all the project dependencies in the same solution for the build in VS to work correctly; otherwise the build can fail with a somewhat cryptic error.

bettinaheim: It looks like what would be possible to do is to check in the .suo file in the hidden .vs folder to hide the solution folders (all compiler projects) from the solution view entirely. I am not sure if this is a good thing to do; if we do that and someone wants to add one of these projects to the solution (e.g. because they want to edit and test the QIR generation), then they get a somewhat confusing error of "the project already exists in the solution" despite that the solution doesn't look like it contains that project. It also seems that the solution folders (all compiler projects are in a solution folder, same for the qir generation projects) are closed by default such that the first view that people see is a solution with two folders and two project.

examples/QIR/Development/Main.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace Microsoft.Quantum.Sdk.Tools
+{
+    public static class DefaultEntryPoint
+    {
+        private static void Main(string[] args) =>
+            Console.WriteLine(
+                "Full support for executing projects compiled to QIR is not yet integrated into the Sdk. " +
+                "For more information about the feature, see https://github.com/microsoft/qsharp-compiler/tree/main/src/QsCompiler/QirGeneration. " +
+                "The generated QIR can be executed by manually linking the QIR runtime. " +
+                "See https://github.com/microsoft/qsharp-runtime/tree/main/src/QirRuntime for further instructions.");
```
swernli, line 12:
```suggestion
                "See https://github.com/microsoft/qsharp-runtime/tree/main/src/Qir/Runtime for further instructions.");
```
I just noticed that this path is wrong now that Cesar and I changed the folder structure. We should update that in the main.cs the compiler generates as well.

swernli (approved):
Changes look good, one minor path fix!

examples/QIR/Development/Main.cs
bettinaheim: Thanks for catching this - I think I actually need to update two more places.

PR #971, opened 2021-04-15
Puzzleshark/entry point driver fixes

puzzleshark:
Added some fixes for
* lower case variable names
* Array arguments
* Range Argument

### DIFF
--- a/src/QsCompiler/Compiler/Templates/CppInterop.cs
+++ b/src/QsCompiler/Compiler/Templates/CppInterop.cs
@@ -30,7 +30,7 @@ namespace Microsoft.Quantum.QsCompiler.Templates
                 DataType.RangeType => "InteropRange*",
                 DataType.ResultType => "char",
                 DataType.StringType => "const char*",
-                DataType.ArrayType => "InteropArray *",
+                DataType.ArrayType => "InteropArray*",
                 _ => throw new NotSupportedException($"Unsupported argument type {this.Type}")
             };
         }
--- a/src/QsCompiler/Compiler/Templates/CppInterop.cs
+++ b/src/QsCompiler/Compiler/Templates/CppInterop.cs
@@ -159,17 +159,17 @@ namespace Microsoft.Quantum.QsCompiler.Templates
 
         public string CliValueVariableName()
         {
-            return this.Name + "CliValue";
+            return + "v" + this.Name + "CliValue";
         }
 
         public string InteropVariableName()
         {
-            return this.Name + "InteropValue";
+            return "v" + this.Name + "InteropValue";
         }
 
         public string IntermediateVariableName()
         {
-            return this.Name + "IntermediateValue";
+            return "v" + this.Name + "IntermediateValue";
         }
     }
 
--- a/src/QsCompiler/Compiler/Templates/CppInterop.cs
+++ b/src/QsCompiler/Compiler/Templates/CppInterop.cs
@@ -159,17 +159,17 @@ namespace Microsoft.Quantum.QsCompiler.Templates
 
         public string CliValueVariableName()
         {
-            return "v" + this.Name + "CliValue";
+            return this.Name + "CliValue";
         }
 
         public string InteropVariableName()
         {
-            return "v" + this.Name + "InteropValue";
+            return this.Name + "InteropValue";
         }
 
         public string IntermediateVariableName()
         {
-            return "v" + this.Name + "IntermediateValue";
+            return this.Name + "IntermediateValue";
         }
     }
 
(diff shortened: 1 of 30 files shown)


src/QsCompiler/Compiler/Templates/CppInterop.cs
```diff
@@ -159,17 +159,17 @@ public string CppCliValueType()
 
         public string CliValueVariableName()
         {
-            return this.Name + "CliValue";
+            return "v" + this.Name + "CliValue";
```
cesarzc, line 162: What's the motivation to add a "v" at the beginning?

puzzleshark: the compiling seems to fail if variables start with uppercase. I think from a lint check or something?

achocron: Can we just disable linting? Seems wrong that it should be run on autogenerated code. And if not, the intent might be more clear by using `this.Name.ToLower()` and commenting the purpose behind it.

src/QsCompiler/Compiler/Templates/QirDriverCpp.cs
```diff
@@ -170,8 +171,9 @@ unique_ptr<InteropRange> CreateInteropRange(RangeTuple rangeTuple)
     if (arg.CppCliVariableInitialValue() != null) {
         WriteLine($"    {arg.CliValueVariableName()} = {arg.CppCliVariableInitialValue()};");
     }
-    WriteLine($"    app.add_option(\"{arg.CliOptionString()}\", {arg.CliValueVariableName()}, \"{arg.CliDescription()}\")->required()");
+    Write($"    app.add_option(\"{arg.CliOptionString()}\", {arg.CliValueVariableName()}, \"{arg.CliDescription()}\")->required()");
     if (arg.TransformerMapName() != null) {
+    WriteLine("");
         Write($"        ->transform(CLI::CheckedTransformer({arg.TransformerMapName()}, CLI::ignore_case))");
     }
     WriteLine(";");
```
achocron, line 179: Unrelated to your change, but why is this in a WriteLine rather than a Write?

src/QsCompiler/Compiler/Templates/QirDriverCpp.tt
```diff
@@ -167,8 +168,9 @@ int main(int argc, char* argv[])
     if (arg.CppCliVariableInitialValue() != null) {
         WriteLine($"    {arg.CliValueVariableName()} = {arg.CppCliVariableInitialValue()};");
     }
-    WriteLine($"    app.add_option(\"{arg.CliOptionString()}\", {arg.CliValueVariableName()}, \"{arg.CliDescription()}\")->required()");
+    Write($"    app.add_option(\"{arg.CliOptionString()}\", {arg.CliValueVariableName()}, \"{arg.CliDescription()}\")->required()");
     if (arg.TransformerMapName() != null) {
+    WriteLine("");
```
achocron, line 173: Does `WriteLine` without any arguments work the same? I can't recall.

src/QsCompiler/Tests.Compiler/TestCases/QirEntryPointTests/UseBoolArgWithValues.cpp
```diff
@@ -24,7 +24,7 @@ using namespace std;
 
 // This is the function corresponding to the QIR entry-point.
 extern "C" void UseBoolArgWithValues( // NOLINT
-    char BoolArgInteropValue
+    char vBoolArgInteropValue
```
achocron, line 27: Can this just be called `boolArgInteropValue`? Same with the others.

src/QsCompiler/Compiler/Templates/CppInterop.cs
cesarzc: Lint is only enabled when using a driver in a similar setup that the [reference driver](https://github.com/microsoft/qsharp-runtime/blob/main/src/Qir/Samples/StandaloneInputReference/qir-driver.cpp) uses. Having said that, C++ source code generated by this component will not be subject to linting so I think the "v" prefix should be removed.

For testing purposes, the [.clang-tidy](https://github.com/microsoft/qsharp-runtime/blob/main/src/Qir/Samples/StandaloneInputReference/.clang-tidy) file can be edited to have nothing and that should effectively disable linting.

puzzleshark: ok I will remove, the reason I didn't do this.Name.ToLower() is just because in theory it is possible to have name clashes if you do that

src/QsCompiler/Compiler/Templates/QirDriverCpp.cs
puzzleshark: it's just to get a newline

achocron: Looks like it left my comment in the wrong place. I was looking at a WriteLine(";"). But it's not a big deal at all, was just wondering.

cesarzc (approved):

PR #978, opened 2021-04-17
Added a code action for QS6303 (immutable binding)

filipw:
I sometimes run into this in my own development - when I forget a given local symbol was declared as immutable and then try to modify it. I thought a code action might be useful. It is of course optional.
If the user tries to modify an immutable local variable, the code action offers an automatic conversion to mutable binding.

Bildschirmaufnahme 2021-04-17 um 18 36 36


src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
```diff
@@ -307,6 +307,33 @@ where DiagnosticTools.ErrorType(ErrorCode.UnknownType)(diagnostic)
...
+            if (immutableBindingDiagnostics.Any())
+            {
+                var edits = new List<(string, WorkspaceEdit)>();
+                foreach (var diagnostic in immutableBindingDiagnostics)
+                {
+                    var refs = file.TryGetReferences(compilation, diagnostic.Range.Start.ToQSharp(), out var declarationLocation, out _);
+
+                    if (refs && declarationLocation != null)
+                    {
+                        var overlappingFragments = file.FragmentsOverlappingWithRange(declarationLocation.Range.ToQSharp()).Where(f => f.Text != null && f.Text.TrimStart().StartsWith(Keywords.qsImmutableBinding.id));
```
bettinaheim, line 323: Very minor nitpick (feel free to ignore), for all the people with limited screen width:
```suggestion
                        var overlappingFragments = file
                            .FragmentsOverlappingWithRange(declarationLocation.Range.ToQSharp())
                            .Where(f => f.Text != null && f.Text.TrimStart()
                            .StartsWith(Keywords.qsImmutableBinding.id));
```

bettinaheim (approved):
Thanks! Looks good; I validated it locally on my end.

src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
filipw: makes sense!

PR #979, opened 2021-04-17
Include AssemblyCommon.props into new SDK projects by default

AndreyTretyak:
Follow-up after https://github.com/microsoft/qsharp-compiler/pull/931

This change replaces manual import of `AssemblyCommon.props` file with automatic (using Directory.Build.props), so it would be automatically included in every SDK-based project in the `src` folder, so it's harder to forget it. 

`Directory.Build.props` also allows defining properties for both SDK and non-SDK projects, so I would suggest using it in the future instead of `AssemblyCommon.cs.v.template` (it might require some changes in build the logic that provides `#ASSEMBLY_VERSION#`).

An additional change in this PR is a move of `InternalsVisibleTo` attributes into project files directly. This allows using same approach for both C# and F# projects and replaces two helper files with public key to have a single place for it.

~~Some of the projects (`LanguageServer` and `Compiler`) have a lot of StyleCop warnings that would fail build, so I've disabled StyleCop for them (same as it was before), to avoid making PR too big.  Also, StyleCop disabled for `Parser` project since it looks like code in it autogenerated.~~

This change treats warnings as errors, so I had to silence some warnings for `Parser` project, since the code in it autogenerated.


samarsha:
> Some of the projects (`LanguageServer` and `Compiler`) have a lot of StyleCop warnings that would fail build, so I've disabled StyleCop for them (same as it was before)

Hmm, I thought the LanguageServer and Compiler projects already had StyleCop enabled in main already (see #513), so there shouldn't be any warnings. Are they using the same StyleCop settings as the other projects?

AndreyTretyak:
> > Some of the projects (`LanguageServer` and `Compiler`) have a lot of StyleCop warnings that would fail build, so I've disabled StyleCop for them (same as it was before)
> 
> Hmm, I thought the LanguageServer and Compiler projects already had StyleCop enabled in main (see #513), so there shouldn't be any warnings. Are they using the same StyleCop settings as the other projects?

Thank you for noticing it, and sorry for not checking it more thoroughly. Not sure why I was getting warnings about it before (possible some intermediate result got cached or due to other warnings I had) after enabling it back now I don't see those warnings.
I've pushed a change to enable it, let's see how PR build would go.
Thanks again for noticing it!

AndreyTretyak:
@samarsha I've removed unneeded StyleCop suppression. Could you please have a look once more?
Also, it would be great if you could validate that build that sings your dlls also works (since I've changed `InternalVisibleTo` attribute).

src/Directory.Build.props
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    <!--
+        This file would be included in all projects automatically.
+        AssemblyCommon.props included only in SDK projects, because non-SDK projects does not support PackageReferences and some othrer properties.
```
samarsha, line 5: nit:
```suggestion
        This file is included in all projects automatically.
        AssemblyCommon.props is included only in SDK projects, because non-SDK projects do not support PackageReferences and some other properties.
```

src/Directory.Build.props
```diff
@@ -0,0 +1,8 @@
...
+     -->
+    <Import Project="$(MSBuildThisFileDirectory)\Common\AssemblyCommon.props" Condition="'$(UsingMicrosoftNETSdk)' == 'true'" />
```
samarsha, line 7: Do we have non-.NET SDK project files that would break if AssemblyCommon.props was imported unconditionally?

src/QsCompiler/CompilationManager/CompilationManager.csproj
```diff
@@ -22,4 +21,13 @@
     <ProjectReference Include="..\Transformations\Transformations.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleTo">
+      <_Parameter1>TTests.Microsoft.Quantum.QsCompiler$(PublicKeySuffix)</_Parameter1>
```
samarsha, line 26: There's a typo (TTests instead of Tests), but since the build didn't break, maybe this InternalsVisibleTo isn't needed?

src/Directory.Build.props
AndreyTretyak: Non-SDK project is all of them that don't have `SDK` attribute in `Project` element. 
In this repository rough list of such projects are ([query](https://github.com/microsoft/qsharp-compiler/search?q=VSToolsPath)): `QSharpFileTemplate.csproj`, `QSharpAppTemplate.csproj`, `QSharpLibTemplate.csproj`, `QSharpTestTemplate.csproj`, `QSharpIonQAppTemplate.csproj`, `QSharpHoneywellAppTemplate.csproj`, `QSharpVsix.csproj`.

I think it's quite possible to change `AssemblyCommon.props` to work with those projects also with minor changes (like not including StyleCop reference by using condition). 
The reason I've done it this way is to keep behavior closer to what we had before since common properties were included only in SDK-based projects.

In the future, non-SDK project could be converted and this condition removed or `AssemblyCommon.props` could be updated to work with both. I think the first option is better because it sometimes hard to understand interactions between old and new properties (ex. `Platform` and `PlatfromTarget`) also old format quite verbose and hard to read.

src/QsCompiler/CompilationManager/CompilationManager.csproj
AndreyTretyak: You are right, it looks like solution building without it.

src/Directory.Build.props
samarsha: Ah, ok, right, the Visual Studio extension projects. That makes sense, thanks!

Issue #662, opened 2020-10-02
Some non-concretized callables are allowed to be called or given as arguments


samarsha:
**Describe the bug**

The Q# language specification says in [Type Parameterizations](https://github.com/microsoft/qsharp-language/blob/main/Specifications/Language/4_TypeSystem/TypeParameterizations.md):

> A type parametrized callable needs to be concretized before it can be assigned or passed as argument, meaning all type parameters need to be resolved to concrete types.

The compiler doesn't implement this correctly, because the following programs compile:

```qsharp
function Foo<'a>() : Unit { }

@EntryPoint()
function Main() : Unit {
    Foo(); // 'a is not concrete.
}
```

```qsharp
function Undefined<'a>() : 'a {
    fail "Undefined";
}

function Foo<'a>(f : (Unit -> 'a)) : String {
    return "Foo";
}

@EntryPoint()
function Main() : Unit {
    Message(Foo(Undefined)); // 'a is not concrete.
}
```

**Expected behavior**

1. The compiler should forbid the programs above to comply with the language spec; or
2. Since a type system that allows these programs would still be sound, and it may actually be useful to allow this in certain cases, we could consider updating the language spec to loosen the restriction that all type parameters must be concretized.

For an example of why we should allow non-concrete types, consider this program that is not supported currently but would be reasonable to support:

```qsharp
@EntryPoint()
function Main() : Unit {
    let xs = [];
    Message($"{xs}");
}
```

The program should clearly print the string `"[]"`. But the type of `xs` cannot be more specific than `'a[]`, which is not concrete. In the case of empty arrays, the type variable `'a` is not used, though, so it does not actually need to be concrete.

**Additional context**

The example programs cause monomorphization to fail using PR #654:

> System.ArgumentException: Couldn't resolve all type parameters for App.Foo
>    at Microsoft.Quantum.QsCompiler.Transformations.CallGraphWalker.BuildCallGraph.ConcreteCallGraphWalker.<>c__DisplayClass2_0.<GetSpecializationFromRequest>b__1(QsLocalSymbol typeParam) in C:\Users\samarsha\Source\Repos\microsoft\qsharp-compiler\src\QsCompiler\Transformations\CallGraph\ConcreteCallGraphWalker.cs:line 101

This issue is based on this discussion: https://github.com/microsoft/qsharp-compiler/pull/597#discussion_r490666610

bettinaheim:
[...]

PR #980, opened 2021-04-19
Improve HM type inference diagnostic messages

samarsha:
* Fix incorrect diagnostic ranges caused by `StripPositionInfoFromType` not stripping ranges from `TypeRange`.
* Improve phrasing and give more information in type mismatch, missing base type, and ambiguous type parameter diagnostic:
  - For ambiguous type parameter diagnostics, show the constraints applied to the type parameter.
  - Call attention to type mismatches with explicitly bound type parameters (i.e., an alternative to the original "constrains type parameter" error).
* Add test cases for issues #573 and #662 that are fixed by HM.

Note: @cesarzc reported an issue where multi-line diagnostic messages were cut off after the first line in Visual Studio, but I can't reproduce the issue. Here is what they look like for me:

image

### DIFF
--- a/src/QsCompiler/DataStructures/Diagnostics.fs
+++ b/src/QsCompiler/DataStructures/Diagnostics.fs
@@ -461,9 +461,9 @@ type DiagnosticItem =
             DiagnosticItem.ApplyArguments args
             << function
             | ErrorCode.TypeMismatch ->
-                "The expected type {0} does not match the actual type {1}.\nExpected type: {2}\n  Actual type: {3}"
+                "The type {1} does not match the type {0}. Actual type: {3}. Expected type: {2}."
             | ErrorCode.MissingBaseType ->
-                "The type {3} cannot be used with the type {4}, because {1} does not {0} {2}."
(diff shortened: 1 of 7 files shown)


src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -461,9 +461,9 @@ type DiagnosticItem =
             DiagnosticItem.ApplyArguments args
             << function
             | ErrorCode.TypeMismatch ->
-                "The expected type {0} does not match the actual type {1}.\nExpected type: {2}\n  Actual type: {3}"
+                "The type {1} does not match the type {0}.\nActual type:   {3}\nExpected type: {2}"
             | ErrorCode.MissingBaseType ->
-                "The type {3} cannot be used with the type {4}, because {1} does not {0} {2}."
```
swernli, line 466: I'm going to miss the madlib version of this message, but I agree the new phrasing is clearer.

swernli (approved):
Looks good to me!

Issue #970, opened 2021-04-15
[QIR] Array Concatenation is missing ref count increase


swernli:
When QIR emission handles array concatenation, a new array is created that includes the items from the two source arrays. However, this does not result in a ref count increase of the items in the newly combined array. As a result, one or both of the source arrays used in the concatenation goes out of scope while the combined array is still in use some of the items may be freed, resulting in a crash the next time they are accessed from the combined array.

I believe this code could result in the issue, but I haven't had the chance to confirm yet:
```qsharp
function MakeList() : (Int, Int)[] {
    let a = (1, 2);
    let b = (3, 4);
    return [a] + [b];
}

function UseList() : (Int, Int) {
    let myList = MakeList();
    return myList[0];
}
```

swernli:
I was able to confirm that the above Q# does reproduce the issue.

bettinaheim:
Reopening due to https://github.com/microsoft/qsharp-compiler/pull/989.

Issue #981, opened 2021-04-19
QIR interop wrapper for entry points with arguments of type String[] and Range[] is incorrect


bettinaheim:
Arguments of String[] or Range[] in entry points are not processed properly. The issue could either be in the generated wrapper, or in the generated entry point driver, or in the runtime - to be investigated.

PR #982, opened 2021-04-19
Ref count optimizations and some bug fixes

bettinaheim:
This PR switches to apply ref count changes only to the outer container and not the inner items unless the ref count changes are due to an assignment to a mutable variable. The reasoning is that I believe it is not possible for an inner item to ever go out of scope earlier than the container unless it is assigned to a mutable variable somewhere. 

This PR makes the adjustment in https://github.com/microsoft/qsharp-compiler/pull/859 unnecessary (i.e. incorrect), and fixes https://github.com/microsoft/qsharp-compiler/issues/970. I believe it may also possibly fix some of https://github.com/microsoft/qsharp-compiler/issues/981 (tbd, I think there are possibly multiple issues there).


swernli (approved):
The test QIR looks good and the logic is sound, so approving. We will have to rely on manual testing for runtime behavior for now.

bettinaheim:
> The test QIR looks good and the logic is sound, so approving. We will have to rely on manual testing for runtime behavior for now.

Yes, I did some manual validation locally, but of course not nearly as extensive as a proper CI setup should be - let's push to get the full CI up and running.

bettinaheim:
Very briefly outlining the issue that this PR ran into and that I overlooked initially: 
In order to ensure that a value stays alive when you return it, the ref counts for e.g. tuples and their items is 1 (so far so good), and the idea is that the calling function then releases them recursively. That is implemented this way with this PR and ok so far. The issue now is when we have a copy-and-update expression involving that. If that's the case, then we run into the same issue as described in https://github.com/microsoft/qsharp-compiler/pull/859; I had removed the adjustment introduced in 859, because I had thought it would no longer be necessary (or correct for that matter), which is incorrect.

PR #983, opened 2021-04-20
QEP 2: Enhanced Array Literals

### LINKED ISSUE #573: Type Argument Inference Fails if the same Type Parametrized Callable is used Multiple Times

### LINKED ISSUE #662: Some non-concretized callables are allowed to be called or given as arguments

### LINKED ISSUE #857: QEP 2: Enhanced Array Literals
This issue tracks the implementation of [QEP 2: Enhanced Array Literals](https://github.com/microsoft/qsharp-language/blob/main/Approved/2-enhanced-array-literals.md). The feature branch is [feature/qep2](https://github.com/microsoft/qsharp-compiler/tree/feature/qep2).

- [x] Support the empty array constructor, `[]` (#884)
  - [x] Add Hindley-Milner type inference module (#917)
  - [x] Replace existing type inference with Hindley-Milner (#940)
  - [x] Update usages of obsolete range fields in `QsTypeParameter` and `UserDefinedType` (#952)
- [x] Support the repeated constant array constructor, `[x, size = n]` (#856)
  - [x] Support `[x, size = n]` in attribute arguments (#884)
- [x] Add `SizedArray` to NamespaceManager.fs (`ExpressionHash`) and QsharpCodeOutput.cs (#963)
- [x] Update optimizations for `SizedArray`: ConstantPropagation.fs, Evaluation.fs, HelperFunctions.fs (`isLiteral`) (#963)
- [x] Address issues from bug bash (#980)
- [ ] Update QIR generation for `SizedArray` expression kind

Deferred:

* Deprecate the `new Type[n]` array constructor (#956)

samarsha:
Closes #857. Closes #662. Closes #573.

Depends on microsoft/qsharp-runtime#643.


src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -434,6 +463,11 @@ type DiagnosticItem =
         let ApplyArguments =
             DiagnosticItem.ApplyArguments args
             << function
+            | ErrorCode.TypeMismatch ->
+                "The type {1} does not match the type {0}.\nActual type:   {3}\nExpected type: {2}"
+            | ErrorCode.MissingBaseType ->
+                "The type {1} does not {0} the type {2}.\nLeft-hand type:  {3}\nRight-hand type: {4}"
```
bettinaheim, line 469: What is the argument 0 here?

bettinaheim (approved):
Great update to the type inference! 
The latest improvements to the error messages make a big difference.

src/QsCompiler/DataStructures/Diagnostics.fs
samarsha: A description of the relationship between the two types. Possible values are "share a subtype with", "equal", or "share a base type with". I think maybe the error name `MissingBaseType` should be renamed to something like `MissingIntersection`.

samarsha: Renamed to `TypeIntersectionMismatch`.

bettinaheim: Thanks.

Issue #974, opened 2021-04-15
[QIR] String interpolation with array results in use after free


swernli:
When displaying an array via string interpolation, the generated QIR includes a loop over string concatenation to assemble each of the items into a single string, such that `[1, 2]` becomes the string `"[1, 2]"`. However, the code incorrectly includes the addition of the opening `"["` inside the body of the loop instead of beforehand, and then tries to decrement the reference count on each iteration through the loop. So this works:
```qsharp
Message($"{[1]}");
```
because the loop over string concatenation only runs once, but this will crash:
```qsharp
Message($"{[1, 2]}");
```
Looking at the QIR directly, I can see where there are a few lines that should be moved up and out of the loop body:
```llvm
  ; THIS IS "["
  %21 = call %String* @__quantum__rt__string_create(i8* getelementptr inbounds ([2 x i8], [2 x i8]* @3, i32 0, i32 0))
  ; THIS IS ", "
  %22 = call %String* @__quantum__rt__string_create(i8* getelementptr inbounds ([2 x i8], [2 x i8]* @4, i32 0, i32 0))
  ; THIS IS "]"
  %23 = call %String* @__quantum__rt__string_create(i8* getelementptr inbounds ([2 x i8], [2 x i8]* @5, i32 0, i32 0))
  br label %header__2

header__2:                                        ; preds = %exiting__2, %exit__1
  %24 = phi i64 [ 0, %exit__1 ], [ %32, %exiting__2 ]
  %25 = icmp sle i64 %24, 1
  br i1 %25, label %body__2, label %exit__2

body__2:                                          ; preds = %header__2
  %26 = call i8* @__quantum__rt__array_get_element_ptr_1d(%Array* %16, i64 %24)
  %27 = bitcast i8* %26 to i64*
  %28 = load i64, i64* %27, align 4
  %29 = call %String* @__quantum__rt__int_to_string(i64 %28)
  ; "[" gets concatenated in loop body when it should be before it
  %30 = call %String* @__quantum__rt__string_concatenate(%String* %21, %String* %29)
  ; Ref count for "[" is decremented in each loop, so the next loop %21 is already freed
  call void @__quantum__rt__string_update_reference_count(%String* %21, i32 -1)
  call void @__quantum__rt__string_update_reference_count(%String* %29, i32 -1)
  %31 = call %String* @__quantum__rt__string_concatenate(%String* %30, %String* %22)
[...]

PR #985, opened 2021-04-21
Splitting out general purpose helpers for branches and loops that yield values

bettinaheim:
Fixes https://github.com/microsoft/qsharp-compiler/issues/974.
As part of fixing https://github.com/microsoft/qsharp-compiler/issues/974, I decided to split out the necessary logic into some general purpose methods. 
I also decided to get rid of the final comma after all for array printing (as previously suggested by @kuzminrobin), though it does come at the cost of having to evaluate a branching in each loop iteration (I am still not sure whether that is preferable or not).

### DIFF
--- a/src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
+++ b/src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
@@ -452,6 +452,9 @@ namespace Microsoft.Quantum.QsCompiler.QIR
                 }
             }
 
+            IValue CreateStringValue(Value str) =>
+                sharedState.Values.From(str, ResolvedType.New(ResolvedTypeKind.String));
+
             // Creates a string value that needs to be queued for unreferencing.
             Value CreateConstantString(string s)
             {
@@ -529,6 +532,7 @@ namespace Microsoft.Quantum.QsCompiler.QIR
                     return sharedState.CurrentBuilder.Call(createString, evaluated.Value);
                 }
 
+                // Creates a string value that needs to be queued for unreferencing.
                 Value TupleToString(TupleValue tuple)
                 {
                     var str = CreateConstantString($"{tuple.TypeName}(");
@@ -544,7 +548,7 @@ namespace Microsoft.Quantum.QsCompiler.QIR
                         }
                         else
                         {
-                            comma ??= CreateConstantString(",");
+                            comma ??= CreateConstantString(", ");
                             str = DoAppend(str, comma, unreferenceNext: false);
                         }
                     }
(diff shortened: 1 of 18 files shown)


src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
```diff
@@ -544,7 +548,7 @@ Value TupleToString(TupleValue tuple)
                         }
                         else
                         {
-                            comma ??= CreateConstantString(",");
+                            comma ??= CreateConstantString(", ");
                             str = DoAppend(str, comma, unreferenceNext: false);
                         }
                     }
```
kuzminrobin, line 552: FYI: Another way to construct a comma-separated list in parentheses is about like this:
```C++
// Output an opening parenthesis, or `$"{tuple.TypeName}("`.

for (var idx = 0; idx < tupleElements.Length; ++idx)
{
    if(idx != 0)
    {
        // Output comma (`, `).
    }
    // Output `element[i]`.
}

// Output the closing parenthesis `)`.
```

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
```diff
@@ -1199,7 +1157,10 @@ public override ResolvedExpressionKind OnConditionalExpression(TypedExpression c
             }
             else
             {
-                var evaluated = this.ConditionalEvaluation(cond.Value, onCondTrue: ifTrueEx, onCondFalse: ifFalseEx);
+                var evaluated = this.SharedState.ConditionalEvaluation(
+                    cond.Value,
+                    onCondTrue: () => this.SharedState.EvaluateSubexpression(ifTrueEx),
+                    onCondFalse: () => this.SharedState.EvaluateSubexpression(ifFalseEx));
```
kuzminrobin, line 1161: nit: I prefer aligning the parts of the similar strings such that it is immediately visible what's the difference between the two strings, like:
```c#
                    onCondTrue : () => this.SharedState.EvaluateSubexpression(ifTrueEx ),
                    onCondFalse: () => this.SharedState.EvaluateSubexpression(ifFalseEx));
```

kuzminrobin (approved):

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
bettinaheim: Personally I also like that very much. However, it is not compatible with the C# formatting rules that are enforced (or the F# ones for that matter...).

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
bettinaheim: Sure, either way is fine for me, so I changed it.

Issue #973, opened 2021-04-15
[QIR] Repeated invocation of partial application of operation using UDT fails


swernli:
I'm not exactly sure what is going on for this one, but I was at least able to isolate it to a relatively small example code:
```qsharp
namespace Microsoft.Quantum.Qir.Emission {
    open Microsoft.Quantum.Intrinsic;
    open Microsoft.Quantum.Canon;
    open Microsoft.Quantum.Synthesis;

    @EntryPoint()
    operation EntryPoint() : Unit {
        let array = [Identity1C(), Identity1C()];
        use q = Qubit();
        // ApplyToEach(Apply1C(_, q), array);
        for g in array {
            Apply1C(g, q);
        }
    }

}
```
The commented out line and the for-loop should be functionally equivalent, but the loop succeeds and the `ApplyToEach` version will crash. My best guess is that this is something to do with tuple passing when preparing the capture for the partial application, because the debugger breaks in on an Access Violation due to null dereference somewhere in the generated QIR with `QirCallable::Invoke` on the stack. Looking inside of `QirCallable::Invoke` I can't tell what callable is being invoked but I can see that the capture associated with it is null and the args are empty.

bettinaheim:
The issue is one indirection too many: The partial application for some reason expects the argtuple to be a tuple with a single item where the single item is a tuple of four items. When it is invoked, however, it gets simply a tuple with four items, resulting in undefined behavior. I'm looking into it.

Issue #964, opened 2021-04-12
[QIR] QIR Generation fails when a passing a generic callabe as a parameter with an inherited type specifier.


swernli:
The standard library includes two functions, `RectangularArrayFact` and `SquareArrayFact`, that make use of the generic `Length` function as a parameter while using an inherited type specifier:
`Any(Compose(NotEqualI(numColumns, _), Length<'T>), Rest(array))`
This fails QIR generation in the compiler with the error
```
System.NotImplementedException: Llvm type could not be constructed
     at Microsoft.Quantum.QsCompiler.QIR.GenerationContext.LlvmTypeFromQsharpType(ResolvedType resolvedType) in D:\a\1\s\submodules\qsharp-compiler\src\QsCompiler\QirGeneration\Context.cs:line 1310
```
when trying to process the argument tuple. It's unclear whether this is an issue specific to QIR generation or perhaps a corner case interaction with Monomorphization (maybe the generic doesn't get fully resolved in the callable tree?). We can work around this in the libraries for now, since it only occurs in those two functions, but we should investigate a fix to ensure QIR generation can handle this case.

bettinaheim:
This is a tricky one. I think the monomorphization itself is resolved with one of the latest PRs I did. However, I don't think that is actually what would prevent this from working; I believe that when building the callable value, we may actually need to make some special accommodations. I'll look into it with my next batch of fixing things.

bettinaheim:
Different manifestation of the same issue: https://github.com/microsoft/qsharp-compiler/issues/955

bettinaheim:
Fixed with #990.

Issue #955, opened 2021-04-08
[QIR] Exception when setting UDT named item to an intrinsic callable


cgranade:
**Describe the bug**

When setting a named item of a UDT to an intrinsic callable, the QIR generation rewrite step throws an exception, causing compilation to fail:

**To Reproduce**

```xml
<!-- IntrinsicUdts.csproj -->
<Project Sdk="Microsoft.Quantum.Sdk/0.15.2103133969">

    <PropertyGroup>
      <OutputType>Exe</OutputType>
      <TargetFramework>netcoreapp3.1</TargetFramework>
      <QirGeneration>true</QirGeneration>
      <QscVerbosity>D</QscVerbosity>
    </PropertyGroup>
  
</Project>
```

```qsharp
namespace IntrinsicUdts {
    open Microsoft.Quantum.Intrinsic;
    open Microsoft.Quantum.Canon;

    newtype Options = (
        VerboseMessage: (String -> Unit)
    );

    function DefaultOptions() : Options {
        return Options(
            Ignore<String>
        );
    }

    function IndirectMessage(s : String) : Unit {
        Message(s);
    }

    @EntryPoint()
    operation RunMain() : Unit {
        let options = DefaultOptions()
            w/ VerboseMessage <- IndirectMessage
            // Exception in QIR generation if the above line is replaced by:
            // w/ VerboseMessage <- Message
            ;
    }

}
```

Command to reproduce and observed output:

```
PS> dotnet build
Microsoft (R) Build Engine version 16.9.0-preview-21103-02+198f3f262 for .NET
Copyright (C) Microsoft Corporation. All rights reserved.

  Determining projects to restore...
  All projects are up-to-date for restore.
  You are using a preview version of .NET. See: https://aka.ms/dotnet-core-preview
  (1,1): info: Compiling with command line arguments
      ResponseFiles: obj/qsharp/config/qsc.rsp
      OutputFolder: (null)
      ProjectName: (null)
      EmitDll: False
      QirOutputFolder: (null)
      PerfOutputFolder: (null)
      TrimLevel: 1
      Plugins:
      TargetSpecificDecompositions:
      ExposeReferencesViaTestNames: False
      AdditionalAssemblyProperties:
      RuntimeCapabilityName: (null)
      MakeExecutable: False
      Verbosity: D
      OutputFormat: MsBuild
      Input:
      CodeSnippet: (null)
      WithinFunction: False
      References:
      NoWarn: System.Int32[]
[...]

PR #988, opened 2021-04-21
Fixing an issue when a partial application takes a udt as argument

bettinaheim:
Fixes https://github.com/microsoft/qsharp-compiler/issues/973, as well as https://github.com/microsoft/qsharp-compiler/issues/955 and https://github.com/microsoft/qsharp-compiler/issues/964. Confirmation and an additional test case will follow shortly.


bettinaheim:
Abandoned in favor of https://github.com/microsoft/qsharp-compiler/pull/991.

PR #991, opened 2021-04-22
Fixing issues with partial application and built-in intrinsics

### LINKED ISSUE #973: [QIR] Repeated invocation of partial application of operation using UDT fails

bettinaheim:
Fixes #973, as well as #955 and #964.


swernli (approved):

bettinaheim:
merging to beheim/revertMerge and adding the test case there

PR #993, opened 2021-04-22
Refactor argument and entry point types

achocron:


src/QsCompiler/BondSchemas/Argument/Argument.bond
```diff
@@ -1,7 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-namespace Microsoft.Quantum.QsCompiler.BondSchemas.EntryPoint;
```
cesarzc, line 4: I think it makes sense to have all the schemas in the same namespace. Having said that, maybe something like `Microsoft.Quantum.QsCompiler.BondSchemas.Execution`?

src/QsCompiler/Compiler/Compiler.csproj
```diff
@@ -31,12 +31,4 @@
   <ItemGroup>
     <Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" />
   </ItemGroup>
-
-  <ItemGroup>
```
achocron, line 35: Restore this

src/QsCompiler/Tests.Compiler/EntryPointGenerationTests.fs
```diff
@@ -7,49 +7,32 @@ open System.IO
...
 open Microsoft.Quantum.QsCompiler
-open Microsoft.Quantum.QsCompiler.BondSchemas.EntryPoint
 open System.Text
 
 type EntryPointGenerationTests(output: ITestOutputHelper) =
 
     let testCasesDirectory = Path.Combine("TestCases", "QirEntryPointTests")
 
     [<Theory>]
-    [<InlineData("UseNoArgs")>]
```
achocron, line 18: Re-add this test

src/QsCompiler/BondSchemas/Execution/Execution.cs
```diff
@@ -83,43 +83,38 @@ public enum ResultValue
 
     [global::Bond.Schema]
     [System.CodeDom.Compiler.GeneratedCode("gbc", "0.12.1.0")]
-    public partial class ArrayValue
+    public partial class Argument
```
ScottCarda-MS, line 86: Without values, and with its use in the Entry Point Operations, could we rename this type to `Parameter` instead? The term `argument` is more of a value that is bound to a parameter at run time, where as the term `parameter` is more related to the definition of the callable.

src/QsCompiler/BondSchemas/Execution/Execution.cs
```diff
@@ -141,68 +136,79 @@ public partial class ArgumentValue
...
-    public partial class Argument
+    public partial class EntryPointOperation
     {
         [global::Bond.Id(5)]
-        public DataType Type { get; set; }
+        public string Name { get; set; }
 
         [global::Bond.Id(10)]
-        public string Name { get; set; }
+        public List<Argument> Arguments { get; set; }
```
ScottCarda-MS, line 160: Along with my other comment, could this field be renamed as `Parameters`?

src/QsCompiler/BondSchemas/Execution/Execution.cs
achocron: Sure, that makes sense to me. I'll keep the name `ArgumentValue` as is, though, to make it very clear that it has a value (even though it is technically implicit)

ScottCarda-MS: Thanks! That will make it easier for me to reason about this.

src/QsCompiler/Compiler/Compiler.csproj
```diff
@@ -34,7 +38,7 @@
 
   <ItemGroup>
     <Compile Update="Templates\QirDriverCpp.cs">
-      <DesignTime>False</DesignTime>
+      <DesignTime>True</DesignTime>
```
cesarzc, line 41: Can you please set this to `False`?

cesarzc (approved):

Issue #937, opened 2021-03-31
Microsoft.Quantum.Sdk jump on size


anpaz:
**Describe the bug**

The Microsoft.Quantum.Sdk package size jump from about 9Mb to 127Mb
The 127Mb is a bit excessive and will hamper user experience as the first time a project compiled with this version will take a while to download for no apparent reason (as downloading happens on the background).

**To Reproduce**

* Version [0.15.2102129448](https://www.nuget.org/packages/Microsoft.Quantum.Sdk/0.15.2102129448) is 9Mb
* Version [0.15.2103133969](https://www.nuget.org/packages/Microsoft.Quantum.Sdk/0.15.2103133969) is 127Mb

PR #994, opened 2021-04-23
Don't reference the llvmlibs in the command line compiler

bettinaheim:
Instead of referencing the llvmlibs that are needed for QIR generation in the command line compiler, which has the consequence that they are packaged with the sdk, we make them a private asset in the Q# project for which to generate QIR, and use GeneratePathProperty to pass the path where to look for them to the command line compiler. This has the advantage that now the llvmlibs only need to be downloaded when either the nuget cache has been cleared, or we update the used llvm version.
This PR fixes https://github.com/microsoft/qsharp-compiler/issues/937.


src/QuantumSdk/Sdk/Sdk.props
```diff
@@ -44,6 +46,22 @@
...
                       IsQscReference="true" ExecutionTarget="Any" Priority="-1" PrivateAssets="All" />
+    <PackageReference Condition="$(QirGeneration) And $([MSBuild]::IsOsPlatform('Windows'))"
+                      Include="libLLVM.runtime.win-x64" Version="11.0.0"
+                      IsImplicitlyDefined="true" 
+                      PrivateAssets="All" GeneratePathProperty="true" />
+    <PackageReference Condition="$(QirGeneration) And $([MSBuild]::IsOsPlatform('OSX'))"
+                      Include="libLLVM.runtime.osx-x64" Version="11.0.0"
+                      IsImplicitlyDefined="true" 
+                      PrivateAssets="All" GeneratePathProperty="true" />
+    <PackageReference Condition="$(QirGeneration) And $([MSBuild]::IsOsPlatform('Linux')) And '$(UbuntuVersion)' == '18.04'"
```
bettinaheim, line 57: I haven't found a good way to figure out what the ubuntu version is (or if we are on ubuntu), and it is also unfortunate that we have to explicitly reference the os version for linux. At least the UbuntuVersion property gives a way to make it work on 18.04 in addition to 20.04 which is the default.

swernli: Yeah, this is an unfortunately necessity. From what I recall from building LLVM from source, there is a library provided by the OS that differs in some signatures from 18.04 to 20.04. While it probably isn't functionality we use directly, the LLVM binary will fail to load if it was compiled against one version but only the other version is present. In my experiments I found that I could install the newer library into 18.04 and get the LLVM compiled against 20.04 to work, but that seems like a weird dependency for an Ubuntu 18.04 user to have to manage. Further, I don't know what other workloads may get confused by having both versions of the library on the system, if any. More than that, I don't know if this library difference is something that would affect other versions of Linux such that we need to also grab the FreeBSD and SLES versions of the libLLVM package too (see https://github.com/microsoft/LLVMSharp/tree/a40e0ce6eefb2bed0b66a523611e306d5681e7cf/packages/libLLVM for the list of all native platforms the libLLVM package is produced for). I agree that for now supporting just Ubuntu 18.04 and Ubuntu 20.04 should probably be sufficient.

src/QsCompiler/LlvmBindings/LlvmBindings.csproj
```diff
@@ -9,11 +9,11 @@
   </PropertyGroup>
 
   <ItemGroup>
-      <PackageReference Include="LLVMSharp" Version="11.0.0-beta" />
-      <PackageReference Include="libLLVM.runtime.win-x64" Version="11.0.0" />
-      <PackageReference Include="libLLVM.runtime.osx-x64" Version="11.0.0" />
-      <PackageReference Include="libLLVM.runtime.ubuntu.20.04-x64" Version="11.0.0" />
-      <PackageReference Include="libLLVM.runtime.ubuntu.18.04-x64" Version="11.0.0" />
+    <PackageReference Include="LLVMSharp" Version="11.0.0-beta" />
+    <PackageReference Include="libLLVM.runtime.win-x64" Version="11.0.0" PrivateAssets="All" />
```
swernli, line 13: Since the SDK handles adding the libLLVM dependencies for any Q#->QIR projects, and the unit tests express a direct dependency on libLLVM as well, are these private assets even needed any more? I think the pattern we are settling on is that the consumer of the QirGeneration package is responsible for providing the right libLLVM native binaries, such that now these package references can be removed from here entirely.

swernli: You can consider this a nitpick comment that does not need to be resolved before merge. Having these as private assets doesn't really do anything; they aren't used directly by compilation, and they aren't treated as dependencies for consumers, and they will already be downloaded during restore of other packages, so they are effectively a no-op. You can either remove or we can clean them up another time, but no need to restart the build pipeline for removing them.

swernli (approved):
I think this approach will get us the behavior that we need! Signing off assuming we'll also test an alpha QDK package from the e2e build to confirm QIR generation still works as expected.

src/QsCompiler/LlvmBindings/LlvmBindings.csproj
bettinaheim: I removed them.

PR #996, opened 2021-04-23
LoadQirByteCode

ScottCarda-MS:
Adds a public function to the AssemblyLoader that can extract QIR byte code from dlls created by the compiler.

This method is used in runtime PR: https://github.com/microsoft/qsharp-runtime/pull/658


src/QsCompiler/CompilationManager/AssemblyLoader.cs
```diff
@@ -172,6 +172,53 @@ public static bool LoadReferencedAssembly(Uri asm, out References.Headers header
...
+        /// <summary>
+        /// Loads any QIR byte code included as a resource from <paramref name="assemblyFile"/>.
+        /// </summary>
+        /// <param name="assemblyFile">The reader for the byte stream of a dotnet DLL from which to load the compilation.</param>
+        /// <param name="qirByteCode">The QIR byte code included as a resource of <paramref name="assemblyFile"/>.</param>
+        /// <returns>
+        /// True if <paramref name="assemblyFile"/> includes a suitable resource, false otherwise.
+        /// </returns>
+        public static bool LoadQirByteCode(
+            PEReader assemblyFile,
```
cesarzc, line 184: I think this should be `FileInfo` instead of `PEReader`, and the `PEReader` object should be created inside.

src/QsCompiler/CompilationManager/AssemblyLoader.cs
```diff
@@ -172,6 +172,53 @@ public static bool LoadReferencedAssembly(Uri asm, out References.Headers header
...
+            [NotNullWhen(true)] out byte[]? qirByteCode)
```
cesarzc, line 185: Using `Stream` instead of `byte[]` would be more general and flexible.

src/QsCompiler/CompilationManager/AssemblyLoader.cs
cesarzc: Since this is something done in more than one place, it might make sense to create an auxiliary method.

src/QsCompiler/CompilationManager/AssemblyLoader.cs
```diff
@@ -172,6 +172,53 @@ public static bool LoadReferencedAssembly(Uri asm, out References.Headers header
...
+                !isResourcePresent ||
+                !resource.Implementation.IsNil)
+            {
+                return false;
+            }
+
+            // This is going to be very slow, as it loads the entire assembly into a managed array, byte by byte.
+            // Due to the finite size of the managed array, that imposes a memory limitation of around 4GB.
+            // The other alternative would be to have an unsafe block, or to contribute a fix to PEMemoryBlock to expose a ReadOnlySpan.
+            PerformanceTracking.TaskStart(PerformanceTracking.Task.LoadDataFromReferenceToStream);
```
cesarzc, line 210: No need to include performance tracking in this method since it is not used during compilation.

src/QsCompiler/CompilationManager/AssemblyLoader.cs
cesarzc: Actually, most of the method can be common. I would implement two generic methods:
One to get a `PEReader`
```csharp
private static PEReader GetAssemblyReader(FileInfo assembly);
```
And one to load a resource to a stream:
```csharp
private static void LoadResource(string resourceName, PEReader assemblyReader);
```

src/QsCompiler/CompilationManager/AssemblyLoader.cs
ScottCarda-MS: 🤔Do you know of any examples of extracting resources as streams from DLL's? I'm not really sure how.

src/QsCompiler/CompilationManager/AssemblyLoader.cs
ScottCarda-MS: Hmm, I'm not seeing a `FileInfo` to `PEReader` method being useful to anywhere else in this file. Did you mean somewhere else?

cesarzc: `LoadReferencedAssembly` does not have a `FileInfo` argument but it creates a `PEReader` object from a `string` argument. The creation of the `PEReader` would be common both to this method and the `LoadReferencedAssembly` one.

src/QsCompiler/CompilationManager/AssemblyLoader.cs
ScottCarda-MS: I've changed the type to a `Stream`, but I'm not sure if I've populated the stream correctly. Could you please double check that this is correct?

src/QsCompiler/CompilationManager/AssemblyLoader.cs
ScottCarda-MS: removed

src/QsCompiler/CompilationManager/AssemblyLoader.cs
cesarzc: In the latest commit you already have it already in a stream, I would just copy it to the stream passed as argument rather than setting its value:
```csharp
var resourceData = image.GetContent(absResourceOffset + sizeof(int), resourceLength).ToArray();
var tempStream = new MemoryStream(resourceData);
tempStream .CopyTo(qirBytecodeStream);
```

src/QsCompiler/CompilationManager/AssemblyLoader.cs
```diff
@@ -172,6 +160,79 @@ public static bool LoadReferencedAssembly(Uri asm, out References.Headers header
             return compilation != null && IsValidCompilation(compilation);
         }
 
+        /// <summary>
+        /// Loads any QIR byte code included as a resource from <paramref name="assemblyFileInfo"/>.
+        /// </summary>
+        /// <param name="assemblyFileInfo">The file info of a .NET DLL from which to load the QIR byte code.</param>
+        /// <param name="qirByteStream">A stream for reading the QIR byte code included as a resource of <paramref name="assemblyFileInfo"/>.</param>
```
cesarzc, line 167:
```suggestion
        /// <param name="qirByteStream">A stream to store the QIR byte code embedded to <paramref name="assemblyFileInfo"/> as a resource.</param>
```

src/QsCompiler/CompilationManager/AssemblyLoader.cs
```diff
@@ -172,6 +160,79 @@ public static bool LoadReferencedAssembly(Uri asm, out References.Headers header
...
+        /// <param name="resourceName">The name of the resource to load.</param>
+        /// <param name="assemblyReader">The Portable Executable Reader for the dll from which to load the resource.</param>
+        /// <param name="resourceStream">A stream for reading the resource as a bytes.</param>
+        /// <returns>
+        /// True if <paramref name="assemblyReader"/> includes the resource named by <paramref name="resourceName"/>, false otherwise.
+        /// </returns>
+        private static bool LoadResource(string resourceName, PEReader assemblyReader, Stream resourceStream)
+        {
+            var metadataReader = assemblyReader.GetMetadataReader();
+
```
cesarzc, line 207: nit: extra blank line.

src/QsCompiler/CompilationManager/AssemblyLoader.cs
```diff
@@ -172,6 +160,79 @@ public static bool LoadReferencedAssembly(Uri asm, out References.Headers header
...
+            bool isResourcePresent = false;
```
cesarzc, line 208: I think the `isResourcePresent` variable is no longer needed. This method can return `false` if `metadataReader.Resources().TryGetValue` is false and the condition in the second if statement would also be simplified.

src/QsCompiler/CompilationManager/AssemblyLoader.cs
```diff
@@ -172,6 +160,79 @@ public static bool LoadReferencedAssembly(Uri asm, out References.Headers header
...
+            ManifestResource resource;
+            if (metadataReader.Resources().TryGetValue(resourceName, out resource))
+            {
+                isResourcePresent = true;
+            }
+
+            // The offset of resources is relative to the resources directory.
+            // It is possible that there is no offset given because a valid dll allows for extenal resources.
+            // In all Q# dlls there will be a resource with the specific name chosen by the compiler.
```
cesarzc, line 217:
```suggestion
```

src/QsCompiler/CompilationManager/AssemblyLoader.cs
```diff
@@ -172,6 +160,79 @@ public static bool LoadReferencedAssembly(Uri asm, out References.Headers header
...
+                !isResourcePresent ||
+                !resource.Implementation.IsNil)
+            {
+                return false;
+            }
+
+            var image = assemblyReader.GetEntireImage(); // uses int to denote the length and access parameters
+            var absResourceOffset = (int)resource.Offset + directoryOffset;
+
+            // the first four bytes of the resource denote how long the resource is, and are followed by the actual resource data
```
cesarzc, line 229:
```suggestion
            // The first four bytes of the resource denote how long the resource is, and are followed by the actual resource data
```

src/QsCompiler/CompilationManager/AssemblyLoader.cs
```diff
@@ -172,6 +160,79 @@ public static bool LoadReferencedAssembly(Uri asm, out References.Headers header
...
+            var resourceLength = BitConverter.ToInt32(image.GetContent(absResourceOffset, sizeof(int)).ToArray(), 0);
+            new MemoryStream(image.GetContent(absResourceOffset + sizeof(int), resourceLength).ToArray()).CopyTo(resourceStream);
+
```
cesarzc, line 232:
```suggestion
            using var tempStream = new MemoryStream(image.GetContent(absResourceOffset + sizeof(int), resourceLength).ToArray());
            tempStream.CopyTo(resourceStream);
            resourceStream.Flush();
            resourceStrea.Position = 0;
```

cesarzc: Note that for the temporary memory stream the `using` keyword is introduced such that it is disposed after it has served its purpose.

src/QsCompiler/CompilationManager/AssemblyLoader.cs
ScottCarda-MS: done

src/QsCompiler/CompilationManager/AssemblyLoader.cs
ScottCarda-MS: removed

src/QsCompiler/CompilationManager/AssemblyLoader.cs
ScottCarda-MS: removed, the try-get is now part of the if-condition

src/QsCompiler/CompilationManager/AssemblyLoader.cs
ScottCarda-MS: removed

src/QsCompiler/CompilationManager/AssemblyLoader.cs
ScottCarda-MS: fixed

src/QsCompiler/CompilationManager/AssemblyLoader.cs
ScottCarda-MS: done

cesarzc (approved):
Looks great!

PR #1000, opened 2021-04-25
Fixing an issue with conditional expressions and open ended ranges

bettinaheim:
This fixes an issue that the simple Grover example currently ran into.
The issue was that certain branching in llvm (specifically for conditional expressions and for the correction in copy-and-update) were not properly registered with the caching mechanism in the QIR generation, such that the generated QIR tried to access values that were created only in certain branches.


swernli (approved):
Looks good to me! Lucky PR number One Thousand!

PR #1001, opened 2021-04-26
Support for sized array in qir generation and package number update

bettinaheim:
Taking care of some little things. The package number in the Sdk is set to the latest alpha package that contains the most recent bug fixes (only relevant for the local build).


cesarzc (approved):
LGTM!

PR #1004, opened 2021-04-28
Always generate function for QIR Interop

swernli:
Because aliases are a little hard to work with during JIT, we should always generate a full function for entry point interop, even when the signature matches exactly. This will also help differentiate when we use different linkages between functions for specializations and interop.

### DIFF
--- a/src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestBuiltInIntrinsics1.ll
+++ b/src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestBuiltInIntrinsics1.ll
@@ -1,4 +1,4 @@
-define void @Microsoft__Quantum__Testing__QIR__TestBuiltInIntrinsics__body() #0 {
+define void @Microsoft__Quantum__Testing__QIR__TestBuiltInIntrinsics__body() {
 entry:
   %0 = call { %Callable* }* @Microsoft__Quantum__Testing__QIR__DefaultOptions__body()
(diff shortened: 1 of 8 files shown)


src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestBuiltInIntrinsics1.ll
```diff
@@ -1,4 +1,4 @@
-define void @Microsoft__Quantum__Testing__QIR__TestBuiltInIntrinsics__body() #0 {
+define void @Microsoft__Quantum__Testing__QIR__TestBuiltInIntrinsics__body() {
```
kuzminrobin, line 1: FYI
In some earlier PR I also have seen the [`internal` keyword](https://llvm.org/docs/LangRef.html#linkage-types) used.
Looks like it will not harm adding that too if appropriate.

kuzminrobin (approved):
FYI: I didn't look in details to .cs code.

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestBuiltInIntrinsics1.ll
swernli: Exactly what I'm building towards :). You'll see that in the next PR, https://github.com/microsoft/qsharp-compiler/pull/1011.

PR #1007, opened 2021-04-29
Only allow QIR Generation on Exe projects

swernli:
This change suppresses QIR Generation for library projects. This is needed to unblok the pipeline now that we are testing QIR Generation for samples.

### DIFF
--- a/src/QuantumSdk/DefaultItems/DefaultItems.targets
+++ b/src/QuantumSdk/DefaultItems/DefaultItems.targets
@@ -59,9 +59,9 @@
   <PropertyGroup>
     <DefaultEntryPointGeneration>dotnet "$(MSBuildThisFileDirectory)../tools/utils/Microsoft.Quantum.Sdk.DefaultEntryPoint.Generation.dll"</DefaultEntryPointGeneration>
     <!-- true/false optional properties -->
-    <Warning Condition="'$(QirGeneration)' == 'true' And '$(ResolvedQSharpOutputType)' == 'QSharpLibrary'" Text="QIR Generation 'true' is not supported on library projects, returning to default behavior." />
+    <Warning Condition="'$(QirGeneration)' == 'true' And '$(ResolvedQSharpOutputType)' != 'QSharpExe'" Text="QIR Generation 'true' is not supported on library projects, returning to default behavior." />
     <QirGeneration Condition="'$(QirOutputPath)' != '' And '$(QirGeneration)' == 'default' And '$(ResolvedQSharpOutputType)' == 'QSharpExe'">true</QirGeneration>
-    <QirGeneration Condition="('$(QirOutputPath)' == '' And '$(QirGeneration)' == 'default') Or '$(ResolvedQSharpOutputType)' == 'QSharpLibrary'">false</QirGeneration>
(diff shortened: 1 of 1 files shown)


src/QuantumSdk/DefaultItems/DefaultItems.targets
```diff
@@ -59,8 +59,9 @@
   <PropertyGroup>
     <DefaultEntryPointGeneration>dotnet "$(MSBuildThisFileDirectory)../tools/utils/Microsoft.Quantum.Sdk.DefaultEntryPoint.Generation.dll"</DefaultEntryPointGeneration>
     <!-- true/false optional properties -->
-    <QirGeneration Condition="'$(QirOutputPath)' != '' And '$(QirGeneration)' == 'default'">true</QirGeneration>
-    <QirGeneration Condition="'$(QirOutputPath)' == '' And '$(QirGeneration)' == 'default'">false</QirGeneration>
+    <Warning Condition="'$(QirGeneration)' == 'true' And '$(ResolvedQSharpOutputType)' == 'QSharpLibrary'" Text="QIR Generation 'true' is not supported on library projects, returning to default behavior." />
+    <QirGeneration Condition="'$(QirOutputPath)' != '' And '$(QirGeneration)' == 'default' And '$(ResolvedQSharpOutputType)' == 'QSharpExe'">true</QirGeneration>
+    <QirGeneration Condition="('$(QirOutputPath)' == '' And '$(QirGeneration)' == 'default') Or '$(ResolvedQSharpOutputType)' == 'QSharpLibrary'">false</QirGeneration>
```
ScottCarda-MS, line 64: Would it be safer to have `'$(ResolvedQSharpOutputType)' != 'QSharpExe'` instead of `'$(ResolvedQSharpOutputType)' == 'QSharpLibrary'`?

src/QuantumSdk/DefaultItems/DefaultItems.targets
```diff
@@ -59,8 +59,9 @@
   <PropertyGroup>
     <DefaultEntryPointGeneration>dotnet "$(MSBuildThisFileDirectory)../tools/utils/Microsoft.Quantum.Sdk.DefaultEntryPoint.Generation.dll"</DefaultEntryPointGeneration>
     <!-- true/false optional properties -->
-    <QirGeneration Condition="'$(QirOutputPath)' != '' And '$(QirGeneration)' == 'default'">true</QirGeneration>
-    <QirGeneration Condition="'$(QirOutputPath)' == '' And '$(QirGeneration)' == 'default'">false</QirGeneration>
+    <Warning Condition="'$(QirGeneration)' == 'true' And '$(ResolvedQSharpOutputType)' == 'QSharpLibrary'" Text="QIR Generation 'true' is not supported on library projects, returning to default behavior." />
```
ScottCarda-MS, line 62: Looks like the comparison of `$QirGeneration` with 'true' is causing an issue when it is expecting `$QirGeneration` to have the value 'default'.

swernli: Ah, you're right, this is causing build failures here. I can get that fixed and try again.

ScottCarda-MS (approved):

swernli:
I don't think this works the way we need it to. So while we may want something like this eventually, we shouldn't rush it if it doesn't fix the current issues.

PR #1011, opened 2021-05-03
Mark all non-entrypoint functions in QIR as internal

swernli:
This change updates the QirGeneration rewrite step to always mark the linkage type of functions internal except when they are the generated functions for an EntryPoint (the `_Interop` variant and the message based variant). This better allows for LLVM passes to perform optimization, inling, and dead code elimination across these functions. Note that in LLVM, any global variable or function definition is considered `external` by default when no linkage is provided.

### DIFF
--- a/src/QsCompiler/QirGeneration/Context.cs
+++ b/src/QsCompiler/QirGeneration/Context.cs
@@ -672,6 +672,7 @@ namespace Microsoft.Quantum.QsCompiler.QIR
             }
 
             this.CurrentFunction = this.RegisterFunction(spec);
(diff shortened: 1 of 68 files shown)


swernli:
This PR should be merged after https://github.com/microsoft/qsharp-compiler/pull/1004, as it contains the same change in it.

swernli:
This change will not build e2e until calling conventions are updated on the Runtime repo. See https://github.com/microsoft/qsharp-runtime/pull/667

kuzminrobin (approved):

PR #1014, opened 2021-05-04
Move Bond Schemas to QIR

ScottCarda-MS:
- Moved the bond schemas for QIR into the QIR Generation Project.
- Removed all dependencies for these bond schemas in the Compiler and Test projects.
- Separated out static functions for name generation of QIR operations to their own static class.
- Added static class EntryPointOperationLoader for loading EntryPointOperation objects from dlls.


kuzminrobin (approved):
I'm not that familiar with .cs, I would recommend at least one more approval. Approving.

src/QsCompiler/Compiler/Compiler.csproj
```diff
@@ -21,22 +21,7 @@
...
-  <ItemGroup>
-    <None Update="Templates\QirDriverCpp.tt">
-      <Generator>TextTemplatingFilePreprocessor</Generator>
-      <LastGenOutput>QirDriverCpp.cs</LastGenOutput>
-    </None>
-  </ItemGroup>
-
   <ItemGroup>
     <Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" />
   </ItemGroup>
```
cesarzc, line 26: This item group can also be removed.

src/QsCompiler/QirGeneration/EntryPointOperationLoader.cs
```diff
@@ -0,0 +1,32 @@
...
+                throw new ArgumentException("Unable to read the Q# syntax tree from the given DLL.");
+            }
+            return GenerateEntryPointOperations(compilation);
+        }
+
+        private static IList<EntryPointOperation> GenerateEntryPointOperations(QsCompilation compilation)
+        {
+            return compilation.EntryPoints.Select(ep => new EntryPointOperation()
+            {
+                Name = NameGeneration.InteropFriendlyWrapperName(ep)
```
cesarzc, line 28: This method should also populate the `Parameters` list for each entry-point.

src/QsCompiler/Compiler/Compiler.csproj
ScottCarda-MS: removed

src/QsCompiler/QirGeneration/EntryPointOperationLoader.cs
ScottCarda-MS: I've populated the Parameters such that each parameter has a name (as it appears in Q#) and a type. I've had to write my own function for mapping the Q# types to the QIR types as I couldn't find anything else that would do this, but it seems kind of awkward. Let me know if there is a better way to do the type mapping or if there is something somewhere that does this already that I can use here.

src/QsCompiler/QirGeneration/BondSchemas/Execution.bond
```diff
@@ -1,7 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-namespace Microsoft.Quantum.QsCompiler.BondSchemas.Execution;
+namespace Microsoft.Quantum.QsCompiler.QIR.BondSchemas;
```
cesarzc, line 4: I think a more general name for this namespace can be `Microsoft.Quantum.Qir.Serialization`.

src/QsCompiler/QirGeneration/EntryPointOperationLoader.cs
```diff
@@ -0,0 +1,96 @@
...
+            return new Parameter()
+            {
+                Name = parameter.VariableName is QsLocalSymbol.ValidName name
+                    ? name.Item
+                    : throw new ArgumentException("Encountered invalid name for parameter."),
+                Type = MapResolvedTypeToDataType(parameter.Type)
+            };
+        }
+
+        private static DataType MapResolvedTypeToDataType(ResolvedType rt)
```
cesarzc, line 54: I believe you can use a switch expression if you use the `Tag` property:
```csharp
private static DataType MapResolvedTypeToDataType(ResolvedType rt) =>
    rt.Resolution.Tag switch {
        QsTypeKind.Tags.Bool => DataType.BoolType
        // The rest of the supported types.
    };
```

ScottCarda-MS: That's pretty cool! I didn't know you could use Tags like that. I've made the switch.

src/QsCompiler/QirGeneration/BondSchemas/Execution.bond
ScottCarda-MS: It seems like some of the namespaces under QirGeneration is under `Microsoft.Quantum.QsCompiler.QIR` while others are under `Microsoft.Quantum.QIR`. Since this is a separate package, it makes more sense to me that they shouldn't have the `QsCompiler` in their namespace, but I want to keep what I add consistent with what is there. Is there some way to tell which of the two namespaces I should add my content under? I can change all of the namespaces for the bond schemas to use what you suggest here, but which namespace should I have for the EntryPointOperationLoader.cs and the NameGeneration.cs?

cesarzc: I believe the way namespaces are setup after your latest change makes sense:
- Bond schemas under `Microsoft.Quantum.Qir.Serialization` because they are independent of Q# compiler data structures.
- EntryPointOperationLoader.cs and NameGeneration.cs under `Microsoft.Quantum.QsCompiler.Qir` because they depend on Q# compiler data structures.

src/QsCompiler/QirGeneration/EntryPointOperationLoader.cs
cesarzc: Resolving since this has been addressed.

src/QsCompiler/QirGeneration/EntryPointOperationLoader.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            QsTypeKind.Tags.Bool => DataType.BoolType,
+            QsTypeKind.Tags.Int => DataType.IntegerType,
+            QsTypeKind.Tags.Double => DataType.DoubleType,
+            QsTypeKind.Tags.Pauli => DataType.PauliType,
+            QsTypeKind.Tags.Range => DataType.RangeType,
+            QsTypeKind.Tags.Result => DataType.ResultType,
+            QsTypeKind.Tags.String => DataType.StringType,
+            QsTypeKind.Tags.ArrayType => DataType.ArrayType,
+            _ => throw new NotImplementedException("invalid type for entry point parameter")
```
cesarzc, line 67:
```suggestion
            _ => throw new ArgumentException($"Invalid type ({rt.Resolution.Tag}) for entry point parameter")
```

src/QsCompiler/QirGeneration/EntryPointOperationLoader.cs
```diff
@@ -0,0 +1,70 @@
...
+using Microsoft.Quantum.QsCompiler.SyntaxTokens;
+using Microsoft.Quantum.QsCompiler.SyntaxTree;
+
+namespace Microsoft.Quantum.QsCompiler.QIR
+{
+    using QsTypeKind = QsTypeKind<ResolvedType, UserDefinedType, QsTypeParameter, CallableInformation>;
+
+    public static class EntryPointOperationLoader
+    {
+        public static IList<EntryPointOperation> LoadEntryPointOperations(FileInfo assemblyFileInfo)
```
cesarzc, line 18: Can you please add a documentation comment to this method since it's public?

src/QsCompiler/QirGeneration/EntryPointOperationLoader.cs
ScottCarda-MS: done

src/QsCompiler/QirGeneration/EntryPointOperationLoader.cs
ScottCarda-MS: added

cesarzc (approved):

PR #1016, opened 2021-05-05
Splitting out the type transformation for Q# -> QIR type transformation

bettinaheim:
This separates the type transformation to get the QIR type for a Q# type from the GenerationContext. In particular, a method to map a Q# type to the entry point interop type that the corresponding QIR entry point wrapper has, can be accessed by constructing a Microsoft.Quantum.Qir.Types instance and then calling .InteropType on it.


src/QsCompiler/QirGeneration/Interop.cs
```diff
@@ -61,63 +62,69 @@ internal sealed class Interop
...
+        /// Maps the given type to an interop-friendly type.
+        /// Strips items of type unit inside tuples.
+        /// Returns null only if the given type is unit, or a tuple of unit values.
         /// </summary>
         /// <exception cref="ArgumentException">
         /// The given type is a pointer to a non-struct type,
         /// or the given type is not specified in the QIR format.
         /// </exception>
-        private ITypeRef? MapToInteropType(ITypeRef t)
+        internal static ITypeRef? MapToInteropType(Context context, ITypeRef t)
```
ScottCarda-MS, line 82: I was trying to avoid having a Context object. I don't have one in the place where I need to call this.

ScottCarda-MS: You can see where I need this here: https://github.com/microsoft/qsharp-compiler/pull/1014/files#diff-ca1be728ef0634194297443f83770d8360446808f72afe59669d9cdf4bf83c7eR53

bettinaheim: This is the context of the llvm bindings. You would construct it just with Context(). This is the piece that may require you to have a reference to the llvmlibs. The way around it is to have the entry point object construction not work with the Q# data structures at all but instead construct everything relying on just the information in the QIR itself (which is the better setup for other reasons/people as well).

ScottCarda-MS (approved):

Issue #637, opened 2020-09-23
Add debug information to QIR


alan-geller:
**Describe the solution you'd like**
It would be helpful if the QIR generator created LLVM debugging metadata. This is useful when consuming the QIR, and is strongly encouraged in the QIR specification.

kevinhartman:
I would love to work on this one!

As a first step, it looks like we need to add the Metadata and DebugInfo portions of Ubiquity LLVM to the LlvmBindings project, and port them to LLVMSharp.

I've just started doing that, but before I go too far, is there any technical reason why they weren't ported originally? Just curious if perhaps LLVMSharp is missing functionality we need to support this. Hopefully it was just a timing thing.

Perhaps @swernli would know?

swernli:
Hi @kevinhartman! We would definitely appreciate the help here. This case is going to be at least a little bit tricky, because of exactly what you were already worried about: Ubiquity LLVM makes use of additional functionality supported by their custom native wrapper that isn't exposed via the regular LLVM-C API and therefore isn't present in LLVMSharp. I did start trying to port the debug code, but gave up since there wasn't an immediate need. In recent conversations about debug info, this topic came up again, and the hope I had is that LLVM-C via LLVMSharp exposes enough of the debug API to port the most relevant portions of the Ubiquity functionality over such that we can get what we need for our purposes. That will require either writing new wrapper objects from scratch using the Ubiquity types as inspiration or teasing apart the components of the debug objects that are not dependent on custom APIs.

[...]

PR #1026, opened 2021-05-11
Port Debug and Metadata API to LLVMBindings

kevinhartman:
To support #637, this changeset ports the Debug and Metadata APIs of Ubiquity Llvm.NET to the `LLVMBindings` project using LLVMSharp.

As background, the vanilla Ubiquity Llvm.NET implementation of Debug and Metadata handling uses a custom native API to expose functionality that is not otherwise present in the LLVM C API. Because we only have access to the C API (via LLVMSharp), the goal of this changeset is to bring in as much of the Ubiquity Debug and Metadata object model as possible, while relying strictly on LLVM C via LLVMSharp.

I'm happy to report that I was able to port almost all of it, without major changes to the Ubiquity object model.

### Porting notes
To continue the same patterns used to originally port existing portions of Ubiquity to LLVMSharp:
- `Debug` and `Metadata` folders of Ubiquity were copied to `LLVMBindings`.
- Uses of `Ubiquity.NET.Llvm.Interop` were replaced with corresponding types and calls from `LLVMSharp.Interop`.
- While LLVMSharp exposes the entire unsafe LLVM C API, it does not contain safe wrapper functions for everything. For those cases, safe wrappers were added to `LLVMBindings/LLVMSharpExtensions` in the form of C# extension methods on the corresponding LLVM ref types.
- Any uses of `Ubiquity.NET.Llvm.Properties` or `Ubiquity.ArgValidators` were stripped out.

The next sections cover the difficulties posed by eliminating Ubiquity's custom native wrapper calls.

#### Challenges
- Operand access for `NamedMDNode` was moved to `BitcodeModule`, since corresponding LLVM C functions require a module ref, and LLVM C provides no way to get an `MDNode`'s parent module. This seemed like the most natural fit, since `NamedMDNode`s belong to a module, anyway. Clients can just pass the `NamedMDNode` name to the relevant operand function on `BitcodeModule`.
- In vanilla Ubiquity, `MDNode` operand access is done through custom native functions that operate in terms of metadata ref types. LLVM C provides `LLVMGetMDNodeOperands`, which is in terms of value refs. Luckily, `LLVMGetMDNodeOperands` expects a `MetadataAsValue` type, so we can first wrap our `MDNode` in a value via `LLVMMetadataAsValue`. The returned operands are value refs, but for operands that we know to be metadata (`MetadataAsValue`), the implementation of `LLVMValueAsMetadata` will actually extract and return the underlying metadata. For operands we know to be true values, a new method `GetOperandValue` is exposed which skips that conversion. A call to this method replaces previous situations where a value operand was obtained originally via the custom native call (as a `ValueAsMetadata`), and then unwrapped to a value via another custom native call. In summary, `MDNode` operand access was possible without Ubiquity's custom native lib via some clever chaining of LLVM C calls.
- We now depend on there being exactly one `Context` instance per C# .NET thread. This is managed via `ThreadContextCache` (previously `ContextCache`), and is implemented using thread local storage. Attempting to create an additional `Context` from the same thread without first disposing of the current one will throw an exception. The reason for this change is that LLVM C does not provide a way to get the context of an `MDNode`, but this is needed for `MDNode` operand access.
- Some enumeration types were not available via LLVMSharp, so in these cases, the proper values have been ported from LLVM sources.

#### Not ported
The vast majority of the non-portable Ubiquity API are `DI` class properties that depend directly on a custom Ubiquity native call. Some of these properties were used internally by Ubiquity to validate preconditions of certain operations. In these cases, the validation is simply removed. From the perspective of QIR generation, these missing properties probably aren't important.

This includes:
- `DISubrprogram.Describes(IrFunction)`
- `MDNode.IsTemporary`
- `MDNode.IsDistinct`
- `MDNode.IsResolved`
- `MDNode.IsUniqued`
- `DISubRange.GetLowerBound`
- `DILocation.InlinedAsScope`
- `DINode.Tag`
- `DebugInfoBuilder.Finish(DISubProgram)`
- `DIBasicType.Encoding`

Other:
- Setters for any `MDNode` operands. LLVM C doesn't expose any way to change the operands of an existing `MDNode`.
- `DebugInfoBuilder.ForwardDeclareFunction`. LLVM C does not expose `DIBuilder.createTempFunctionFwdDecl`.
- `ValueAsMetadata.Value`. LLVM C doesn't expose a way to get the value of a `ValueAsMetadata`. Note that using `LLVMMetadataAsValue` would simply wrap the `ValueAsMetadata` inside a `MetadataAsValue`.
- The source file name cannot be set on a BitcodeModule created via `BitcodeModule.InterningFactory.CreateBitcodeModule`.

### Testing
As a proof-of-concept, the below IR is the result of using the ported API to generate debug info for QIR compile units, namespaces, and function declarations ([deb781d0](https://github.com/microsoft/qsharp-compiler/pull/1026/commits/deb781d078b32959060d719a773761d8d25871fc)). Note that the `DISubroutineType` is stubbed.

The `MDNode` hierarchy was also tested to validate that operand access is working as intended via the VS debugger.

```ll
%Range = type { i64, i64, i64 }
%Tuple = type opaque
%String = type opaque

@PauliI = constant i2 0
@PauliX = constant i2 1
@PauliY = constant i2 -1
@PauliZ = constant i2 -2
@EmptyRange = internal constant %Range { i64 0, i64 1, i64 -1 }

define %Tuple* @Microsoft__Quantum__Core__Attribute__body() !dbg !2 {
entry:
  ret %Tuple* null
}

define %Tuple* @Microsoft__Quantum__Core__EntryPoint__body() !dbg !8 {
entry:
  ret %Tuple* null
}

define %Tuple* @Microsoft__Quantum__Core__Inline__body() !dbg !10 {
entry:
  ret %Tuple* null
}

define double @Microsoft__Quantum__Testing__QIR__TestDouble__body(double %x, double %y) !dbg !12 {
entry:
  %0 = fadd double %x, %y
  %a = fsub double %0, 2.000000e+00
  %1 = fmul double %a, 1.235000e+00
  %2 = call double @llvm.pow.f64(double %x, double %y)
  %b = fadd double %1, %2
  %3 = fcmp oge double %a, %b
  br i1 %3, label %condTrue__1, label %condFalse__1

condTrue__1:                                      ; preds = %entry
  %4 = fsub double %a, %b
  br label %condContinue__1

condFalse__1:                                     ; preds = %entry
  %5 = fadd double %a, %b
  br label %condContinue__1

condContinue__1:                                  ; preds = %condFalse__1, %condTrue__1
  %c = phi double [ %4, %condTrue__1 ], [ %5, %condFalse__1 ]
  %6 = fmul double %a, %b
  %7 = fmul double %6, %c
  ret double %7
}

; Function Attrs: nounwind readnone speculatable willreturn
declare double @llvm.pow.f64(double, double) #0

define { %String* }* @Microsoft__Quantum__Targeting__TargetInstruction__body(%String* %__Item1__) !dbg !15 {
entry:
  %0 = call %Tuple* @__quantum__rt__tuple_create(i64 ptrtoint (i1** getelementptr (i1*, i1** null, i32 1) to i64))
  %1 = bitcast %Tuple* %0 to { %String* }*
  %2 = getelementptr inbounds { %String* }, { %String* }* %1, i32 0, i32 0
  store %String* %__Item1__, %String** %2, align 8
  call void @__quantum__rt__string_update_reference_count(%String* %__Item1__, i32 1)
  ret { %String* }* %1
}

declare %Tuple* @__quantum__rt__tuple_create(i64)

declare void @__quantum__rt__string_update_reference_count(%String*, i32)

attributes #0 = { nounwind readnone speculatable willreturn }

!llvm.dbg.cu = !{!0}

!0 = distinct !DICompileUnit(language: DW_LANG_C, file: !1, producer: "Q# Compiler", isOptimized: false, runtimeVersion: 0, emissionKind: FullDebug, splitDebugInlining: false)
!1 = !DIFile(filename: "test-debug.qs", directory: "")
!2 = distinct !DISubprogram(name: "Microsoft__Quantum__Core__Attribute__body", scope: !3, file: !1, line: 6, type: !4, scopeLine: 6, flags: DIFlagPrototyped, spFlags: DISPFlagDefinition, unit: !0, retainedNodes: !7)
!3 = !DINamespace(name: "Microsoft.Quantum.Core", scope: !1, exportSymbols: true)
!4 = !DISubroutineType(types: !5)
!5 = !{!6, !6, !6}
!6 = !DIBasicType(name: "double", size: 64, encoding: DW_ATE_float)
!7 = <temporary!> !{}
!8 = distinct !DISubprogram(name: "Microsoft__Quantum__Core__EntryPoint__body", scope: !3, file: !1, line: 12, type: !4, scopeLine: 12, flags: DIFlagPrototyped, spFlags: DISPFlagDefinition, unit: !0, retainedNodes: !9)
!9 = <temporary!> !{}
!10 = distinct !DISubprogram(name: "Microsoft__Quantum__Core__Inline__body", scope: !3, file: !1, line: 9, type: !4, scopeLine: 9, flags: DIFlagPrototyped, spFlags: DISPFlagDefinition, unit: !0, retainedNodes: !11)
!11 = <temporary!> !{}
!12 = distinct !DISubprogram(name: "Microsoft__Quantum__Testing__QIR__TestDouble__body", scope: !13, file: !1, line: 5, type: !4, scopeLine: 5, flags: DIFlagPrototyped, spFlags: DISPFlagDefinition, unit: !0, retainedNodes: !14)
!13 = !DINamespace(name: "Microsoft.Quantum.Testing.QIR", scope: !1, exportSymbols: true)
!14 = <temporary!> !{}
!15 = distinct !DISubprogram(name: "Microsoft__Quantum__Targeting__TargetInstruction__body", scope: !16, file: !1, line: 28, type: !4, scopeLine: 28, flags: DIFlagPrototyped, spFlags: DISPFlagDefinition, unit: !0, retainedNodes: !17)
!16 = !DINamespace(name: "Microsoft.Quantum.Targeting", scope: !1, exportSymbols: true)
!17 = <temporary!> !{}
```


kevinhartman:
To review this PR, it might make sense to review and add comments [here](https://github.com/kevinhartman/Llvm.NET/pull/1/files) instead.

That is a PR of the relevant files from this branch into a fork of Ubiquity with existing Q# changes applied. That should capture exactly what porting changes I made.

kevinhartman:
Switching this out of draft, since only auto-formatting is needed, which I'll run only after review to keep the diff with Ubiquity more readable.

### TODO
- [x] Add a link to a diff between vanilla Ubiquity and ported code for reviewers.
https://github.com/kevinhartman/Llvm.NET/pull/1/files
- [x] Update method comments.
- [x] Switch to implicit ref conversions in `LLVMMetadataRefExtensions`.
- [x] Remove example QIR debug generation from PR (the full implementation should be a separate PR).
Reverted [deb781d0](https://github.com/microsoft/qsharp-compiler/pull/1026/commits/deb781d078b32959060d719a773761d8d25871fc).
- [x] Remove TODO section from PR description prior to merge.
- [ ] Run code formatting (avoided until now to make porting and code review easier).
- [ ] Add proper copyright header modifications to all imported files.

swernli:
Thanks for doing this, @kevinhartman ! I should hopefully have time to review in the next couple of days to provide any detailed feedback, but from the excellent description you gave above this sounds like exactly what we needed and the right approach to use to perform the port!

swernli:
Also, looks like at some point we enabled warnings as errors, which have flagged a ton of existing formatting oddities in this code as errors. Let me know how you'd like to proceed with that, but in the meantime I'll review for functionality and design rather than style.

kevinhartman:
Awesome! Let me know if there's anything you would like done differently, overall.

We can run the formatting once you have a chance to review the diff in the other repo I linked. The failing build errors should all be style related (I'm able to build and run locally).

src/QsCompiler/LlvmBindings/DebugInfo/DILocalVariable.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace Ubiquity.NET.Llvm.DebugInfo
+{
+    /// <summary>Debug information for a local variable</summary>
+    /// <seealso href="xref:llvm_langref#dilocalvariable">LLVM DILocalVariable</seealso>
+    public class DILocalVariable
+        : DIVariable
+    {
+        // ReSharper disable IdentifierTypo
+        /* TODO: non-operand properties
```
swernli, line 17: This is a bit of a nitpick, since these were present in the original Ubiquity code, but would you mind removing these TODO lists from the DI classes? In our case, there isn't really an intent to come back and fill these in, so they aren't true to-do items for this code base.

src/QsCompiler/LlvmBindings/LLVMSharpExtensions/LLVMMemoryBufferRefExtensions.cs
```diff
@@ -7,10 +7,10 @@
 
 namespace LLVMSharp.Interop
 {
-    /// <summary>Extensions for the LLVMSharp.Interop.LLVMMemoryBufferRef class.</summary>
+    /// <summary>Extensions for <see cref="LLVMMemoryBufferRef"/>.</summary>
```
swernli, line 10: Nice clean up, thanks!

src/QsCompiler/LlvmBindings/LLVMSharpExtensions/LLVMMetadataRefExtensions.cs
```diff
@@ -0,0 +1,160 @@
...
+    /// <summary>Extensions for <see cref="LLVMMetadataRef"/>.</summary>
+    public static unsafe class LLVMMetadataRefExtensions
+    {
+        /// <summary>Convenience wrapper for <see cref="LLVM.GetMetadataKind"/>.</summary>
+        public static LLVMMetadataKind GetMetadataKind(this LLVMMetadataRef self)
+        {
+            return (LLVMMetadataKind)LLVM.GetMetadataKind(self);
+        }
+
+	/// <summary>Convenience wrapper for <see cref="LLVM.MetadataReplaceAllUsesWith"/>.</summary>
```
swernli, line 19: Looks like there are a few places where tabs snuck in instead of spaces... hopefully the autoformat will take care of that, but if not please switch to spaces.

src/QsCompiler/LlvmBindings/LLVMSharpExtensions/LLVMDIBuilderRefExtensions.cs
```diff
@@ -0,0 +1,198 @@
...
+        {
+            return LLVM.DIBuilderCreateBasicType(self, name.AsMarshaledString(), (UIntPtr)name.Length, sizeInBits, encoding, diflags);
+        }
+
+        /// <summary>Convenience wrapper for <see cref="LLVM.DIBuilderCreatePointerType"/>.</summary>
+        public static LLVMMetadataRef CreatePointerType(this LLVMDIBuilderRef self, LLVMMetadataRef pointeeTy, ulong sizeInBits, uint alignInBits, uint addressSpace, string? name)
+        {
+            if (name == null)
+            {
+                name = string.Empty;
```
swernli, line 54: I haven't had my morning caffeine yet so may be interpreting this wrong, but wouldn't this have the side effect of changing the passed in string from null to empty string? If so, it would be better to use a local variable to avoid mutating the arguments. If not, please ignore!

src/QsCompiler/LlvmBindings/Metadata/MDNode.cs
```diff
@@ -0,0 +1,136 @@
...
+        {
+            if( IsDeleted )
+            {
+                throw new InvalidOperationException( "Cannot operate on a deleted node" );
+            }
+        }
+
+        private static Context GetMetadataContext( LLVMMetadataRef metadataHandle )
+        {
+            // TODO (Q#): we currently expect exactly one context, since we have no way
```
swernli, line 130: Glad to see this called out in code! I don't think this needs to be a "TODO" since as of now there isn't pending work associated with this, so that prefix can just be removed and the comment text kept as-is.

src/QsCompiler/LlvmBindings/Metadata/MDString.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    /// <summary>Stores a string in Metadata</summary>
+    public class MDString
+        : LlvmMetadata
+    {
+        /// <summary>Gets the string from the metadata node</summary>
+        /// <returns>String this node wraps</returns>
+        public override string ToString( )
+        {
+            var context = ContextCache.Single();
```
swernli, line 19: It might be worth reproducing the explanatory comment here too.

src/QsCompiler/LlvmBindings/Values/GlobalAlias.cs
```diff
@@ -17,7 +17,6 @@ public class GlobalAlias
         public Constant Aliasee
         {
             get => this.IndirectSymbol!;
-            set => this.IndirectSymbol = value;
```
swernli, line 20: Was this an intentional removal? Just curious why it cannot (or maybe should not) be supported.

src/QsCompiler/LlvmBindings/Values/GlobalIndirectSymbol.cs
```diff
@@ -17,7 +17,6 @@ public class GlobalIndirectSymbol
         public Constant IndirectSymbol
         {
             get => this.Operands.GetOperand<Constant>(0)!;
-            set => this.Operands[0] = value;
```
swernli, line 20: Similar question here, just wondering what the removal was motivated by.

src/QsCompiler/LlvmBindings/BitcodeModule.cs
```diff
@@ -63,18 +64,101 @@ public sealed class BitcodeModule
...
+
+                for (uint i = 0; i < len; ++i)
+                {
+                    var behavior = flags.GetFlagBehavior(i);
+                    string key = flags.GetKey(i);
+                    var metadata = LlvmMetadata.FromHandle<LlvmMetadata>(this.Context, flags.GetMetadata(i));
+                    retVal.Add(key, new ModuleFlag((ModuleFlagBehavior)behavior, key, metadata!));
+                }
+
+                flags.Dispose();
```
swernli, line 108: It might be worth a comment here calling out the need to explicitly clean up the flags. Just thinking out loud, is there any way this could be implemented via a `using` block?

swernli (commented):
This is great work, thanks again Kevin! I left a few comments here and there, nothing major. Once you've had the chance to merge the latest main and update the formatting to get rid of those build errors, I can do another quick pass and do a formal sign off.

src/QsCompiler/LlvmBindings/LLVMSharpExtensions/LLVMDIBuilderRefExtensions.cs
kevinhartman: In this case, we should be good, since without the `ref` or `out` keywords, our parameter will be copied-in / passed by value.

Though, it may very well be against the style guidelines. If it is, I'll be sure to adjust!

src/QsCompiler/LlvmBindings/Metadata/MDNode.cs
kevinhartman: Will do! Do you think it's worth keeping the `Q#` prefix to make it clear that the comment wasn't imported from Ubiquity?

swernli: That's a good idea; the Q# prefix will clarify origin!

src/QsCompiler/LlvmBindings/LLVMSharpExtensions/LLVMDIBuilderRefExtensions.cs
swernli: Makes sense, thanks!

src/QsCompiler/LlvmBindings/BitcodeModule.cs
kevinhartman: Good point. To support a `using` block, we'd need `flags` to be `IDisposable`, but I think the class comes directly from `LLVMSharp`. Instead, how about we used a `try...finally` here to guarantee that the flags are disposed?

swernli: Ah, right, I forgot about the `IDisposable` interface. Yes, a try-finally sounds like it would be a good fit.

src/QsCompiler/LlvmBindings/Values/GlobalAlias.cs
kevinhartman: Good catch! Yes this was accidental. I ripped out the setter methods from `IOperandCollection` since metadata operands cannot be set via the LLVM C API, but didn't notice there were other implementations. I'll add these back and probably split `IOperandCollection` into `IReadableOperandCollection` and `IWritableOperandCollection` (like C#'s `IReadOnlyList`).

kevinhartman: Should be all set! I actually didn't need to do anything other than add back the removed lines since `User.Operands` is actually exposed as the concrete type (`ValueOperandListCollection`), i.e. not `IOperandCollection`.

src/QsCompiler/LlvmBindings/Values/GlobalIndirectSymbol.cs
kevinhartman: Fixed in https://github.com/microsoft/qsharp-compiler/pull/1026/commits/7010cfd901ca34ffc8789dde97143ee9c195bb3a .

src/QsCompiler/LlvmBindings/Metadata/MDString.cs
kevinhartman: https://github.com/microsoft/qsharp-compiler/pull/1026/commits/6f0de6f347cf1b872daa29dc25d9e466d645288f

src/QsCompiler/LlvmBindings/Metadata/MDNode.cs
kevinhartman: https://github.com/microsoft/qsharp-compiler/pull/1026/commits/6f0de6f347cf1b872daa29dc25d9e466d645288f

src/QsCompiler/LlvmBindings/DebugInfo/DILocalVariable.cs
kevinhartman: https://github.com/microsoft/qsharp-compiler/pull/1026/commits/8505bc61cebaf92e8beca7fd2433edb56f459834

src/QsCompiler/LlvmBindings/BitcodeModule.cs
kevinhartman: https://github.com/microsoft/qsharp-compiler/pull/1026/commits/6211ce9423675baf1e039c98c118d43e79b96367

src/QsCompiler/LlvmBindings/ThreadContextCache.cs
```diff
@@ -0,0 +1,121 @@
...
+namespace Ubiquity.NET.Llvm
+{
+    /// <summary>
+    /// Manages a cache of <see cref="Context"/> instances using thread local storage.
+    /// </summary>
+    /// <remarks>
+    /// The public constructor <see cref="Context()"/> will register itself in the cache for its calling thread.
+    /// Since it is a new instance, that is a safe (LLVM) operation.
+    /// <para/>
+    /// Q# IMPORTANT: we currently expect exactly one context per thread. This was done because we have no way
```
swernli, line 21: Excellent call out, great fix!

swernli (approved):

PR #1038, opened 2021-05-17
Setting up Development project to also run

bettinaheim:
Adding support for the Development example that compiles using the source code version of the QIR generation to also run, using a fixed runtime version. For this project, the driver is hardcoded rather than generated. 

I also adjusted slightly the dependency handling for rewrite steps.


src/QsCompiler/Transformations/SyntaxTreeTrimming.cs
```diff
@@ -15,22 +13,18 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.SyntaxTreeTrimming
...
             {
                 var globals = compilation.Namespaces.GlobalCallableResolutions();
-                var augmentedEntryPoints = BuiltIn.AllBuiltIns
-                    .Where(bi => bi.Kind != BuiltInKind.Attribute && globals.ContainsKey(bi.FullName))
-                    .Select(bi => bi.FullName)
-                    .Concat(compilation.EntryPoints)
-                    .Distinct()
-                    .ToImmutableArray();
+                var dependenciesToKeep = dependencies?.Where(globals.ContainsKey) ?? ImmutableArray<QsQualifiedName>.Empty;
+                var augmentedEntryPoints = dependenciesToKeep.Concat(compilation.EntryPoints).Distinct().ToImmutableArray();
```
ScottCarda-MS, line 27: nit: A more typical syntax when using multiple LINQ methods is to have each on a new line. This reflects SQL syntax.

src/QsCompiler/Core/Dependencies.fs
```diff
@@ -39,58 +40,34 @@ type BuiltIn =
...
+            BuiltIn.ApplyIfOneC.FullName,
+            BuiltIn.ApplyIfOneCA.FullName,
+            BuiltIn.ApplyIfElseR.FullName,
+            BuiltIn.ApplyIfElseRA.FullName,
+            BuiltIn.ApplyIfElseRC.FullName,
+            BuiltIn.ApplyIfElseRCA.FullName
+        )
 
     /// The set of all built in callables and attributes
-    static member AllBuiltIns =
```
ScottCarda-MS, line 45: It seems to defeat the purpose of a built-in callable or attribute if it gets removed by the first rewrite step (syntax tree trimming) and is no longer available for any subsequent rewrite steps. Hence I think it would make the system more maintainable to save all of the built-ins through that rewrite step and not just the ones that we know to be used currently.

src/QsCompiler/Transformations/SyntaxTreeTrimming.cs
bettinaheim: done.

src/QsCompiler/Core/Dependencies.fs
bettinaheim: Briefly summarizing our discussion:
Both the RewriteStepsDependencies and AllBuiltIns are deprecated and will be replaced by letting/requiring each rewrite step to declare its dependencies as part of the interface.

src/QsCompiler/Core/Dependencies.fs
```diff
@@ -87,7 +112,7 @@ type BuiltIn =
             BuiltIn.ApplyIfElseRA
             BuiltIn.ApplyIfElseRC
             BuiltIn.ApplyIfElseRCA
-            // dependencies in other namespaces (e.g. things used for code actions)
+            // in other namespaces (e.g. things used for code actions)
```
ScottCarda-MS, line 115:
```suggestion
            // in Microsoft.Quantum.Arrays
```

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -574,7 +574,10 @@ public CompilationLoader(SourceLoader loadSources, ReferenceLoader loadReference
             {
                 // TODO: It would be nicer to trim unused intrinsics. Currently, this is not possible due to how the old setup of the C# runtime works.
                 // With the new setup (interface-based approach for target packages), it is possible to trim ununsed intrinsics.
-                var rewriteStep = new LoadedStep(new SyntaxTreeTrimming(keepAllIntrinsics: !qirEmissionEnabled), typeof(IRewriteStep), thisDllUri);
+#pragma warning disable CS0618 // Type or member is obsolete
+                // TODO: The dependencies for rewrite steps should be declared as part of IRewriteStep interface, and we should query those here.
+                var rewriteStep = new LoadedStep(new SyntaxTreeTrimming(keepAllIntrinsics: !qirEmissionEnabled, BuiltIn.AllBuiltIns.Select(e => e.FullName)), typeof(IRewriteStep), thisDllUri);
```
ScottCarda-MS, line 579: We had previously filtered out built-in attributes as well. Not sure if you want to keep those or not.

bettinaheim: It seems more consistent to keep them if we keep the rest with the argument of it being listed in BuiltIn.

ScottCarda-MS (approved):

PR #1040, opened 2021-05-19
Merge main to feature/azure-quantum-simulator

cesarzc:
Merge latest main changes into feature/azure-quantum-simulator


ScottCarda-MS (approved):
Looks like the build issue is with the styling on the AssemblyLoader.cs file. Should be an easy fix.

PR #1044, opened 2021-05-25
Update deprecated statements in unit test template

### LINKED ISSUE #607: Deprecated operation call in VS Code template
`Assert` is deprecated. I suggest to replace it with `AssertMeasurement`.

https://github.com/microsoft/qsharp-compiler/blob/20c020d889915f2f19d996b5e2dc6a23fe2b6075/src/VSCodeExtension/templates/unittest/Tests.qs#L11

dime10:
Edit: might close #607


bettinaheim (approved):
Thanks for updating this! Looks like this got missed.

PR #1045, opened 2021-05-27
Add API documentation linting to docgen step

cgranade:
This draft PR adds new functionality to the docgen step to check for common documentation errors and to expose them to the user. Each linting rule can be turned off, to warning level, or promoted to errors by setting the new `<QSharpDocsLintingRules>` MSBuild property. My goal with this would be to use the new rules to help improve API documentation in the libraries and runtime repos, with an eventual aim of promoting all docs linting rules to errors.

There's still a lot of code quality things left to be done (esp. adding API documentation to new types in this PR), but I wanted to open this up for early feedback and see if this could be a reasonable direction to help improve our API docs. Thanks!


samarsha:
@cgranade do you still want a review for this?

cgranade:
> @cgranade do you still want a review for this?

Oh, goodness, I forgot about this...! Would appreciate your thoughts on whether this is a useful way to go for improving our API docs, in any case. If it seems like a reasonable approach, happy to see what I can do to get it across the finish line.

PR #1054, opened 2021-06-09
Move CSharpGeneration to compiler repo

swernli:
This moves the Microsoft.Quantum.CSharpGeneration rewrite step package into this repo as part of addressing https://github.com/microsoft/qsharp-runtime/issues/618.


QsCompiler.sln
```diff
@@ -55,6 +55,18 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "QirGeneration", "src\QsComp
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "LlvmBindings", "src\QsCompiler\LlvmBindings\LlvmBindings.csproj", "{36FB68CC-533A-40D4-A2E9-597A35C6912C}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{E70908F4-CE50-4CDA-A5DE-0C02554FBBC6}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "QsCompiler", "QsCompiler", "{8088F458-4201-46A4-882A-047A35A8601D}"
+EndProject
```
samarsha, line 61: This shows up a little strangely in solution view. I'm not sure what the best organization is. There could be a folder for C# generation, but moving the C# generation tests to the Tests folder seems weird. Maybe we could get rid of Tests and in general have the test project adjacent to the project it's testing?

image

src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -457,614 +457,631 @@ type DiagnosticItem =
...
         try
             String.Format(str, args)
-        with _ -> str // let's fail silently for now
+        with
+        | _ -> str // let's fail silently for now
 
     static member Message(code: ErrorCode, args: IEnumerable<string>) =
         let ApplyArguments =
             DiagnosticItem.ApplyArguments args
             << function
```
samarsha, line 466: I think you could avoid most of the formatting changes in this file by using `match` here and in the other cases:

```suggestion
            match DiagnosticItem.ApplyArguments args with
```

swernli: I'll try that and see how the diff compares...

swernli: Yup, that worked, thanks!

QsCompiler.sln
swernli: Ah, right. I've run into this before, because I use `dotnet sln add` to handle adding things to the solution file, which follows the folder structure instead of the previous patterns that get created when someone uses the Visual Studio UI. Since I don't have Visual Studio installed, I'm going to try editing the file to get the new projects to show up in the other style.

swernli: Ok, I think I've got it back into the existing form. I've left the Test folder as-is; I'll leave it up to those who use VS regularly to decide what structure makes sense to that file. My two cents: we should be using the dotnet cli to manage solution files, since that's what anyone working cross-platform would need to use, even if that ends up creating a folder structure in the solution explorer pane for those who use VS.

src/QsCompiler/DataStructures/Diagnostics.fs
samarsha: Oh, I made a type error... I think you have to add a parameter:

```fsharp
let ApplyArguments code =
    match DiagnosticItem.ApplyArguments args code with
...
```
```

Comment (swernli) [reply src/QsCompiler/DataStructures/Diagnostics.fs]: oh, hmm... I think it still has a problem, and I need to have a `<<` somewhere. Let me see if I can figure it out, but it may end up that the large Fantomas update to the indentation is not avoidable...

Comment (swernli) [reply src/QsCompiler/DataStructures/Diagnostics.fs]: I think we've found a form that works and still minimizes the diff. Thanks!

**Review by samarsha — APPROVED** [review]
Without reviewing any of the files copied from qsharp-runtime, it looks good.

File: build/manifest.ps1 [hunk 3]
```diff
@@ -50,6 +51,7 @@ $artifacts = @{
         ".\src\QsCompiler\SyntaxProcessor\bin\$Env:BUILD_CONFIGURATION\netstandard2.1\Microsoft.Quantum.QsSyntaxProcessor.dll",
         ".\src\QsCompiler\TextProcessor\bin\$Env:BUILD_CONFIGURATION\netstandard2.1\Microsoft.Quantum.QsTextProcessor.dll",
         ".\src\QsCompiler\QirGeneration\bin\$Env:BUILD_CONFIGURATION\netstandard2.1\Microsoft.Quantum.QirGeneration.dll",
+        ".\src\QsCompiler\CSharpGeneration\bin\$Env:BUILD_CONFIGURATION\netstandard2.1\Microsoft.Quantum.CSharpGeneration.dll",
```
samarsha, line 54: I think you're missing

```
".\src\Simulation\RoslynWrapper\bin\$Env:BUILD_CONFIGURATION\netstandard2.1\Microsoft.Quantum.RoslynWrapper.dll",
```

swernli: Good catch! Added.

PR #1057, opened 2021-06-13
Sdk integration for full state simulator in Azure Quantum

bettinaheim:
If the ExecutionTarget property for a project is set to microsoft.simulator*, then there is no human readable QIR generated, but the bitcode representation is generated and attached to the dll, while C# generation is enabled at the same time. This means that the built executable will run locally via the C# setup, but can also be submitted for execution on the Azure Quantum simulator (which is via QIR).

Note: Enabling both C# and QIR generation is subject to the restrictions we had talked about but haven't enforced yet. Specifically, the C# generation will fail if a reference has been monomorphized; i.e. setting the ExecutionTarget to microsoft.simulator* for an Exe project that references another Q# executable will cause the C# compiler to fail.


bettinaheim:
The build is currently failing because we need to update the feature branches (or at least I need to update the matching branches on all repos).

src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -418,7 +418,8 @@ type WarningCode =
     | QirEmissionGeneratedWarning = 8002
 
     | InvalidAssemblyProperties = 8101
-    | MissingTargetInstructionName = 8102
+    | DuplicateAssemblyProperty = 8102
```
idavis, line 421: Can we change the number of existing error codes?

src/QsCompiler/CommandLineTool/Options.cs
```diff
@@ -117,20 +118,37 @@ public class CompilationOptions : Options
         /// Returns a dictionary with the specified assembly properties as out parameter.
         /// Returns a boolean indicating whether all specified properties were successfully added.
         /// </summary>
-        internal bool ParseAssemblyProperties(out Dictionary<string, string> parsed)
+        internal bool ParseAssemblyProperties(ILogger? logger, [MaybeNullWhen(false)] out Dictionary<string, string> asmProperties)
         {
-            var success = true;
-            parsed = new Dictionary<string, string>();
+            asmProperties = null;
+            var parsed = new Dictionary<string, string>();
```
idavis, line 124: Should an `IEqualityComparer<TKey>` be supplied in the ctor?

examples/QIR/Development/Development.csproj
```diff
@@ -64,7 +65,13 @@
       <Compile Include="Main.cs">
         <Visible>false</Visible>
       </Compile>
+      <_QirOutputFiles Include="$(QirOutputPath)**" />    
     </ItemGroup>
+    <!--
+    TODO: update Sdk version and uncomment these
```
idavis, line 71: Does this `TODO` need to be resolved before merge?

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -569,12 +569,11 @@ public CompilationLoader(SourceLoader loadSources, ReferenceLoader loadReference
             PerformanceTracking.TaskStart(PerformanceTracking.Task.RewriteSteps);
             var steps = new List<(int, string, Func<QsCompilation?>)>();
             var qirEmissionEnabled = this.externalRewriteSteps.Any(step => step.Name == "QIR Generation");
-
             if (this.config.IsExecutable && !this.config.SkipSyntaxTreeTrimming)
             {
                 // TODO: It would be nicer to trim unused intrinsics. Currently, this is not possible due to how the old setup of the C# runtime works.
                 // With the new setup (interface-based approach for target packages), it is possible to trim ununsed intrinsics.
-                var rewriteStep = new LoadedStep(new SyntaxTreeTrimming(keepAllIntrinsics: !qirEmissionEnabled), typeof(IRewriteStep), thisDllUri);
+                var rewriteStep = new LoadedStep(new SyntaxTreeTrimming(keepAllIntrinsics: true), typeof(IRewriteStep), thisDllUri);
```
idavis, line 576: Are the two booleans passed into `SyntaxTreeTrimming` and `Monomorphization` reversed?

src/QsCompiler/QirGeneration/RewriteSteps.cs
```diff
@@ -87,10 +88,13 @@ public bool Transformation(QsCompilation compilation, out QsCompilation transfor
             generator.Emit(bcFile, emitBitcode: true);
             PerformanceTracking.TaskEnd(PerformanceTracking.Task.BitcodeGeneration);
 
-            // create the human readable version as well
-            var sourceOutputFolder = this.AssemblyConstants.TryGetValue(ReservedKeywords.AssemblyConstants.QirOutputPath, out path) && !string.IsNullOrWhiteSpace(path) ? path : "qir";
-            var llvmSourceFile = CompilationLoader.GeneratedFile(targetFile, Path.GetFullPath(sourceOutputFolder), ".ll", "");
-            generator.Emit(llvmSourceFile, emitBitcode: false);
+            if (this.AssemblyConstants.TryGetValue(ReservedKeywords.AssemblyConstants.QirOutputPath, out path) && !string.IsNullOrWhiteSpace(path))
```
idavis, line 91: The behavior here is changing to no longer default to `qir`. If this is the desired behavior, the associated test code should probably have a different value than the old default in order to verify changed values.

src/QuantumSdk/DefaultItems/DefaultItems.targets
```diff
@@ -28,14 +28,15 @@
...
 
-  <!-- Initializing the ResolvedProcessorArchitecture to either HoneywellProcessor, IonQProcessor, QCIProcessor or Unspecified.-->
+  <!-- Initializing the ResolvedProcessorArchitecture to either HoneywellProcessor, IonQProcessor, QCIProcessor, AzureSimulator or Unspecified.-->
   <PropertyGroup>
     <ResolvedProcessorArchitecture Condition="('$(ResolvedQSharpOutputType)' == 'QSharpExe' Or '$(NoEntryPoint)' == 'true') And $(ExecutionTarget.StartsWith('honeywell'))">HoneywellProcessor</ResolvedProcessorArchitecture>
     <ResolvedProcessorArchitecture Condition="('$(ResolvedQSharpOutputType)' == 'QSharpExe' Or '$(NoEntryPoint)' == 'true') And $(ExecutionTarget.StartsWith('ionq'))">IonQProcessor</ResolvedProcessorArchitecture>
     <ResolvedProcessorArchitecture Condition="('$(ResolvedQSharpOutputType)' == 'QSharpExe' Or '$(NoEntryPoint)' == 'true') And $(ExecutionTarget.StartsWith('qci'))">QCIProcessor</ResolvedProcessorArchitecture>
+    <ResolvedProcessorArchitecture Condition="('$(ResolvedQSharpOutputType)' == 'QSharpExe' Or '$(NoEntryPoint)' == 'true') And $(ExecutionTarget.StartsWith('microsoft.simulator'))">AzureSimulator</ResolvedProcessorArchitecture>
```
cesarzc, line 36: Nit: I think that `MicrosoftSimulator` is a slightly better name for the `ResolvedProcessorArchitecture` than `AzureSimulator` because the intent is to allow QIR-based executables to be run locally at some point in the future.

cesarzc (approved):
I just left a suggestion in this [comment](https://github.com/microsoft/qsharp-compiler/pull/1057/files#r657572718) but other than that it looks good to me.

src/QuantumSdk/DefaultItems/DefaultItems.targets
bettinaheim: I renamed it to MicrosoftSimulator. In any case, this is only a name used internally within the SDK and not yet exposed beyond that; I haven't added it to the AssemblyConstants just yet.

src/QsCompiler/Compiler/CompilationLoader.cs
bettinaheim: Yes, they are. I would align them if one of them is new such that they are not reversed, but if both already exist I don't think it's worth making the breaking change. That's a good thing to watch out for when reviewing the feature branch as a whole.

examples/QIR/Development/Development.csproj
bettinaheim: Yes, it should, but I need a full set of packages first.

src/QsCompiler/DataStructures/Diagnostics.fs
bettinaheim: Nothing should be relying on the numbering and everything should use the DU case instead. 
At some point we should ideally revise the setup to allow for conversions between warnings and errors, and to facilitate documenting error codes online. I am not sure if we should try to keep the numbers stable before that but let's give it a try and that will help inform whether we think we can do that for 1.0 (and once we have done these changes).
I changed the numbering to add the new diagnostic as 8103.

src/QsCompiler/CommandLineTool/Options.cs
bettinaheim: The default comparer for strings sounds fine here. What would be the reason you had in mind for providing an argument instead of relying on the default behavior of string dictionaries?

examples/QIR/Development/Development.csproj
bettinaheim: Removed; updated the package numbers for the two QIR examples.

src/QsCompiler/QirGeneration/RewriteSteps.cs
bettinaheim: the fact that the current setup output something even when qiroutputpath was not found is an implementation detail that so far nobody could have relied on (unless they were compiling the command line compiler from source). It is hence fine to change that behavior to not outputting anything unless qiroutputpath is specified. The tests indeed were updated and now require that we set this value. Does that address this comment?

PR #1059, opened 2021-06-15
Finish implementing the parser visitors for Type in QsFmt

yjt98765:
## Issue
Before this PR, only the `IntType` and `UserDefinedType` defined in [QSharpParser.g4](https://github.com/microsoft/qsharp-compiler/blob/main/src/QsFmt/Parser/QSharpParser.g4) can be transferred into a syntax tree. The other types are [parsed](https://github.com/microsoft/qsharp-compiler/blob/main/src/QsFmt/Formatter/ParseTree/Type.fs#L49-L50) into `Unknown` types. This uncompleted implementation prevents formatting rules to be applied within the subtree nodes of a `Type`. For example, the extra spaces inside the brackets are not removed.
```
namespace Foo {
    function Bar () : Int [    ] {}
}
```

## This pull request
In this PR, I have implemented all the other `Type`s defined in [Type.fs](https://github.com/microsoft/qsharp-compiler/blob/main/src/QsFmt/Formatter/SyntaxTree/Type.fs#L36-L44), i.e., `Missing`, `Parameter`, `Tuple`, `Array`, `Callable`, and `BuiltIn` other than `IntType`. The visitors of their context are created. I also named some ANTLA fields to assist the implementation.

## Testing
After the implementation, the previous example is formatted as
```
namespace Foo {
    function Bar () : Int [ ] {}
}
```
Notice that the extra spaces inside the brackets are removed.

I have added several test cases in [FixedPoints.fs](https://github.com/microsoft/qsharp-compiler/blob/main/src/QsFmt/Formatter.Tests/FixedPoints.fs) to demonstrate that the types can be parsed without errors. Note that these test cases can **pass** even before the implementation.

### Question
Should I add the code example shown above as a test case? I notice that there is only one example in [Examples.fs](https://github.com/microsoft/qsharp-compiler/blob/main/src/QsFmt/Formatter.Tests/Examples.fs#L60-L78) to demonstrate that extra spaces are removed. That example serves the same purpose as the one above.

### DIFF
--- a/src/QsFmt/Formatter.Tests/Examples.fs
+++ b/src/QsFmt/Formatter.Tests/Examples.fs
@@ -60,7 +60,7 @@ let ``Adds newlines and indents`` =
 [<Example>]
 let ``Removes extraneous spaces`` =
     """namespace     Foo {
-    function Bar() : Int {
+    function Bar() : Int [     ] {
         let x= // Newlines are preserved.
             (7 -   1) // Comments too.
             + 4;
@@ -69,7 +69,7 @@ let ``Removes extraneous spaces`` =
 }""",
 
     """namespace Foo {
-    function Bar() : Int {
+    function Bar() : Int [ ] {
         let x = // Newlines are preserved.
             (7 - 1) // Comments too.
             + 4;
--- a/src/QsFmt/Formatter.Tests/FixedPoints.fs
+++ b/src/QsFmt/Formatter.Tests/FixedPoints.fs
@@ -141,15 +141,15 @@ let ``Type parameter`` = """namespace Foo {
 
 [<FixedPoint>]
 let ``Tuple type`` = """namespace Foo {
-    function Bar (arg : (BigInt, Bool, (Double, Int))) : Unit {}
+    function Bar (arg : (BigInt, Bool, (Double, Int))) : Pauli {}
 }"""
 
 [<FixedPoint>]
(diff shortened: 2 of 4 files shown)


samarsha:
Thanks. :)

> Should I add the code example shown above as a test case? I notice that there is only one example in [Examples.fs](https://github.com/microsoft/qsharp-compiler/blob/main/src/QsFmt/Formatter.Tests/Examples.fs#L60-L78) to demonstrate that extra spaces are removed. That example serves the same purpose as the one above.

I think it would make sense to update the tests to include extra whitespace in types. It seems fine to either add them to the existing "Removes extraneous spaces" test (making the example bigger) or adding separate examples for whitespace in types.

yjt98765:
> Thanks. :)
> 
> > Should I add the code example shown above as a test case? I notice that there is only one example in [Examples.fs](https://github.com/microsoft/qsharp-compiler/blob/main/src/QsFmt/Formatter.Tests/Examples.fs#L60-L78) to demonstrate that extra spaces are removed. That example serves the same purpose as the one above.
> 
> I think it would make sense to update the tests to include extra whitespace in types. It seems fine to either add them to the existing "Removes extraneous spaces" test (making the example bigger) or adding separate examples for whitespace in types.

I chose the first way. This is done with a minor modification in the test.

samarsha (approved):
Looks great! Thanks for your contribution, this is really helpful. :)

PR #1060, opened 2021-06-15
Doc Comments for Transformations

ScottCarda-MS:
Adds documentation comments for the Trim Syntax Tree transformation and the Validate Monomorphization transformation.


src/QsCompiler/Transformations/SyntaxTreeTrimming.cs
```diff
@@ -11,8 +11,19 @@
 
 namespace Microsoft.Quantum.QsCompiler.Transformations.SyntaxTreeTrimming
 {
+    /// <summary>
+    /// Removes unused callables from the syntax tree.
+    /// </summary>
     public static class TrimSyntaxTree
     {
+        /// <summary>
```
bettinaheim, line 19: It might be worth calling out what happens to types and type constructors. (I think the types may be preserved but the type constructors are subject to the same trimming rules as other callables?)

bettinaheim (approved):

src/QsCompiler/Transformations/SyntaxTreeTrimming.cs
ScottCarda-MS: I was looking for that. I thought we had updated the code to prevent type constructors from being removed at some point. Do you remember?

PR #1062, opened 2021-06-17
Add support for QIR Generation of Library projects

swernli:
This change adds support for generation of QIR from Library projects. When a Q# Library is compiled with `<QirGeneration>true</QirGeneration>` the compiler will make certain assumptions to enable output of usable QIR. First, it will treat every non-internal, non-generic callabe defined in the library as an effective entry point for the purpose of syntax tree trimming and monomorphization. Second, it will generate the QIR function for each specialization of these effective entry points without the `internal` linkage so that they can be invoked from outside of the compilation unit. And finally, it will NOT generate Interop or EntryPoint friendly wrappers for these effective entry points, to allow callers to instead invoke specific specializations directly while supporting any QIR-specific types, like `%Qubit*` or `%Tuple*`. This should enable new use cases of Q# libraries across FFI boundaries assuming the resulting library QIR gets compiled with the same runtime that it is eventually invoked with.


src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -568,14 +573,15 @@ public CompilationLoader(SourceLoader loadSources, ReferenceLoader loadReference
...
+            this.config.IsQirGenerationEnabled = this.externalRewriteSteps.Any(step => step.Name == "QIR Generation");
+
+            if ((this.config.IsExecutable || this.config.IsQirGenerationEnabled) && !this.config.SkipSyntaxTreeTrimming)
             {
                 // TODO: It would be nicer to trim unused intrinsics. Currently, this is not possible due to how the old setup of the C# runtime works.
                 // With the new setup (interface-based approach for target packages), it is possible to trim ununsed intrinsics.
 #pragma warning disable CS0618 // Type or member is obsolete
                 // TODO: The dependencies for rewrite steps should be declared as part of IRewriteStep interface, and we should query those here.
-                var rewriteStep = new LoadedStep(new SyntaxTreeTrimming(keepAllIntrinsics: true, BuiltIn.AllBuiltIns.Select(e => e.FullName)), typeof(IRewriteStep), thisDllUri);
+                var rewriteStep = new LoadedStep(new SyntaxTreeTrimming(keepAllIntrinsics: !this.config.IsQirGenerationEnabled, BuiltIn.AllBuiltIns.Select(e => e.FullName), !this.config.IsExecutable), typeof(IRewriteStep), thisDllUri);
```
swernli, line 584: @cesarzc @ScottCarda-MS I noticed this during merge conflict resolution and was wondering if you could explain the reason for hard-coding true here. In particular, I thought preserving intrinsics should not be necessary when generating QIR because it doesn't rely on any inheritance. Is there a new need to preserve all intrinsics even for QIR because of how some of them get used?

ScottCarda-MS: @bettinaheim first introduced the keepAllIntrinsics. She may have a better memory on this than me.

ScottCarda-MS: What callables (including intrinsics) should be kept and which ones should be trimmed should be determined based on usage (see the above TODOs), so setting this to true may have just been a temporary approach until we have a more robust dependency-based logic here.

src/QsCompiler/Compiler/RewriteSteps/Monomorphization.cs
```diff
@@ -34,17 +36,19 @@ internal class Monomorphization : IRewriteStep
...
+            this.isLibrary = isLibrary;
         }
 
-        public bool PreconditionVerification(QsCompilation compilation) => compilation.EntryPoints.Any();
+        public bool PreconditionVerification(QsCompilation compilation) => compilation.EntryPoints.Any() || this.isLibrary;
 
         public bool Transformation(QsCompilation compilation, out QsCompilation transformed)
         {
-            transformed = Monomorphize.Apply(compilation, this.monomorphizeIntrinsics);
+            transformed = Monomorphize.Apply(new QsCompilation(compilation.Namespaces, this.AugmentedEntryPoints(compilation)), this.monomorphizeIntrinsics);
```
ScottCarda-MS, line 51: This would actually change all relevant callables to entry points for all transformation here on out, beyond just for monomorphization. Is that a good idea? Maybe we could capture the true entry points beforehand, do what you do here, but then remember to set the entry points field back to its original value before returning the tree.

swernli: If the compiled project is an exe project, this will behave exactly as before (the original entry points just get passed through). For libraries, there are no original entrypoints (using that attribute in a library will cause a compile error). That said, you bring up a good point, in that right now the steps after this either don't care about entrypoints or know to treat them differently if it is a library. Any other rewrite steps could have the chance of getting confused by the compilation being updated this way. Originally I had considered updating the `QsCompilation` object to include a new field, but that would be a breaking change to the compiler.  I'm open to any thoughts on how to address this without affecting behavior for everyone. At least for now, any rewrite step that is active during QIR generation would believe that non-generic, non-internal callables in a library are actually entrypoints. That pretty much matches my intention, but I don't want to discount the possible side effects of this change.

ScottCarda-MS: Since the EntryPoints fields is just a list of names of the entry point callables, it is fairly lightweight. My thinking would be to just capture the value of the EntryPoints field in a temporary variable before calling Monomorphize.Apply. Then when you do call the transformation, you can do the same logic you did here already by updating the tree so that all the callables are listed as entry points. The return from the transformation will be the monomorphized tree with all the relevant callables as entry points. Lastly you can easily replace the EntryPoints field in that transformed tree with the captured value of the field from before you called Monomorphization.Apply before returning the transformed tree. You are guaranteed that all the original entry points will still be in the tree, so you wouldn't have to worry about listing a callable that is not present anymore.

swernli: That's a really good idea, thanks for suggesting! I can implement that approach and confirm it works for the target scenario.

swernli: This actually helped me find a bug: the QirGeneration rewrite step also invokes SyntaxTreeTrimming but was not passing the new `isLibrary` bool to ensure the tree is trimmed based on non-internal, non-generic callables instead of solely entry points. I was also able to remove a check later in QirGeneration that was avoiding the incorrect entry points leftover from the Monomorphization. Thanks again!

ScottCarda-MS: Glad it worked out 😁!

ScottCarda-MS (approved):

PR #1065, opened 2021-06-22
Fix remaining code-eating bugs in QsFmt

samarsha:
* Add more information to callable declaration nodes: attributes, access modifiers, type parameters, and specializations.
* Handle discard symbol in SymbolBindingVisitor.
* Handle tuples in ParameterVisitor.

The "Identity preserves original source code" test (which parses and unparses all of the Q# files in the compiler's TestCases folder) now passes.


ScottCarda-MS (approved):
It's a bit out of my area, so I don't understand most of this, but nothing stood out to me as problematic.

Issue #1069, opened 2021-06-29
Compilation fails when specifying microsoft.simulator as ExecutionTarget and using some operations provided by libraries


cesarzc:
When a Q# application that uses some operations like `ApplyToEach`, `ForEach` or `MResetZ` provided by Q# libraries, compilation fails.

**To Reproduce**

1. Create a Q# application that has the following project file:
```
<Project Sdk="Microsoft.Quantum.Sdk/0.17.2106147493-alpha">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp3.1</TargetFramework>
    <ExecutionTarget>microsoft.simulator</ExecutionTarget>
  </PropertyGroup>

</Project>
```
2. Implement a program that uses some library functions like ApplyToEach`, `ForEach` or `MResetZ`.
```
namespace Quantum.MicrosoftSimulatoExecutionTargetBug {

    open Microsoft.Quantum.Arrays;
    open Microsoft.Quantum.Canon;
    open Microsoft.Quantum.Intrinsic;
    open Microsoft.Quantum.Measurement;

    
    @EntryPoint()
    operation HelloQ () : Unit {
        use qubits = Qubit[10]
        {
            ApplyToEachA(H, qubits);
            let resuls = ForEach(MResetZ, qubits);
        }
    }
}
```
3. Compile the program (`dotnet build` was used to reproduce). Compilation fails with the following error:
```
> dotnet build
Microsoft (R) Build Engine version 16.9.0+57a23d249 for .NET
Copyright (C) Microsoft Corporation. All rights reserved.

  Determining projects to restore...
  All projects are up-to-date for restore.
  Determining projects to restore...
  All projects are up-to-date for restore.
  
  ____________________________________________

  Q#: Success! (0 errors, 0 warnings)

C:\Microsoft\Tests\MicrosoftSimulatoExecutionTargetBug\MicrosoftSimulatoExecutionTargetBug\Program.qs(42,199): error CS0234: The type or namespace name '_cc8e2bb1a3764c8b88eba7d6dd43903e_ApplyToEachA' does not exist in the namespace 'Microsoft.Quantum.Canon' (are you missing an assembly reference?) [C:\Microsoft\Tests\MicrosoftSimulatoExecutionTargetBug\MicrosoftSimulatoExecutionTargetBug\MicrosoftSimulatoExecutionTargetBug.csproj]
[...]

PR #1071, opened 2021-07-02
Microsoft Simulator Library Fix

ScottCarda-MS:
Ensures the Q# libraries get regenerated with monomorphized content when compiling with the AzureSimulator process architecture.

Addressed #1069

### DIFF
--- a/src/QsCompiler/CSharpGeneration/Context.fs
+++ b/src/QsCompiler/CSharpGeneration/Context.fs
@@ -152,6 +152,7 @@ type CodegenContext =
                 target = AssemblyConstants.HoneywellProcessor
                 || target = AssemblyConstants.IonQProcessor
                 || target = AssemblyConstants.QCIProcessor
+                || target = "AzureSimulator" // ToDo: We need to have an assembly constant for this.
             | _ -> false
 
         not (fileName.EndsWith ".dll") || targetsQuantumProcessor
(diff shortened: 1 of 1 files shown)


src/QsCompiler/CSharpGeneration/Context.fs
```diff
@@ -152,6 +152,7 @@ type CodegenContext =
                 target = AssemblyConstants.HoneywellProcessor
                 || target = AssemblyConstants.IonQProcessor
                 || target = AssemblyConstants.QCIProcessor
+                || target = "AzureSimulator" // ToDo: We need to have an assembly constant for this.
```
bettinaheim, line 155:
```suggestion
                || target = "MicrosoftSimulator" // ToDo: We need to have an assembly constant for this.
```

bettinaheim:
@ScottCarda-MS Please either merge this to cesarcz/merge-feature-azure-simulator-to-main, or directly to main after that one went in.

src/QsCompiler/CSharpGeneration/Context.fs
ScottCarda-MS: I'm not sure this is right. While debugging the name that came up was "AzureSimulator", not "MicrosoftSimulator".

bettinaheim: This is correct.

Issue #1043, opened 2021-05-23
Q# empty string becomes null in QIR, causing segmentation faults in runtime


swernli:
An empty string in Q# today seems to generate into an explicit null in QIR rather than remaining an empty string. For example, the code:
```qsharp
Message("");
```
becomes:
```llvm
  %0 = call %String* @__quantum__rt__string_create(i8* null)
  call void @__quantum__rt__message(%String* %0)
  call void @__quantum__rt__string_update_reference_count(%String* %0, i32 -1)
```
This code will cause a segmentation fault in the runtime when the null goes through string creation, as it eventually gets passed into the constructor for `std::string`.

It seems to me that since Q# works hard to not have null in the language, this should get generated into an empty string literal (ie: `internal constant [1 x i8] c"\00"`) rather than an explicit null. Looking at the [QIR spec for strings](https://github.com/microsoft/qsharp-language/blob/main/Specifications/QIR/Data-Types.md#strings), it does say that the input to `__quantum__rt__string_create` is expected to be a null terminated byte array, which implies that it cannot be null. It might be worth calling out that null is not supported as input.

bettinaheim:
Digging a bit about whether we can avoid passing a null pointer for empty strings, the issue that I haven't found my way around yet is that LLVM.ConstStringInContext returns a zero-initializer when the string to construct is empty.
Possibly related: https://reviews.llvm.org/D24904

bettinaheim:
Alright, here is the PR: https://github.com/microsoft/qsharp-compiler/pull/1073
It's not pretty but it's the best I can come up with to work around this.

PR #1073, opened 2021-07-06
Don't pass null pointer for empty strings

bettinaheim:
This PR fixes https://github.com/microsoft/qsharp-compiler/issues/1043. See the issue for further information.

### DIFF
--- a/src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestForLoop.ll
+++ b/src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestForLoop.ll
@@ -1,7 +1,7 @@
 define internal { double, %String* }* @Microsoft__Quantum__Testing__QIR__TestNestedLoops__body() {
 entry:
   %name = call %String* @__quantum__rt__string_create(i8* getelementptr inbounds ([7 x i8], [7 x i8]* @0, i32 0, i32 0))
-  %0 = call %String* @__quantum__rt__string_create(i8* null)
+  %0 = call %String* @__quantum__rt__string_create(i8* getelementptr inbounds ([1 x i8], [1 x i8]* @1, i32 0, i32 0))
   %1 = call { double, %String* }* @Microsoft__Quantum__Testing__QIR__Energy__body(double 0.000000e+00, %String* %0)
   %res = alloca { double, %String* }*, align 8
   store { double, %String* }* %1, { double, %String* }** %res, align 8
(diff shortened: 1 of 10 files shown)


src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestForLoop.ll
```diff
@@ -1,7 +1,7 @@
 define internal { double, %String* }* @Microsoft__Quantum__Testing__QIR__TestNestedLoops__body() {
 entry:
   %name = call %String* @__quantum__rt__string_create(i8* getelementptr inbounds ([7 x i8], [7 x i8]* @0, i32 0, i32 0))
-  %0 = call %String* @__quantum__rt__string_create(i8* null)
+  %0 = call %String* @__quantum__rt__string_create(i8* getelementptr inbounds ([1 x i8], [1 x i8]* @1, i32 0, i32 0))
```
kuzminrobin, line 4: Out of curiosity:
What is `@1` in this line (in the fragment `[1 x i8]* @1`?

I have an impression that this .ll file is a file template used to generate another .ll file. And `@1` is some sort of parameter for the file generation.

In any case I feel that the code is to generate an empty string (`""`) in this line, instead of passing `null` to `__quantum__rt__string_create()`. The empty string consists of a single terminating  null character `'\0'`. And the fragment in this line `i8* getelementptr inbounds ([1 x i8], [1 x i8]* @1, i32 0, i32 0)` gets the pointer to that null-terminating character.
But I don't see a code fragment where that terminating null character is generated/written to the memory.

So I expect `@1` to be a pointer pointing to where the terminating null character `'\0'`is located. But I don't see where that is.

kuzminrobin (approved):
Other than one question, the rest of the PR looks good to me.

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestForLoop.ll
bettinaheim: This file is a part of a test, and not used as a template or to generate the compiler output. The content of this file is what the compiler output is compared against for the purpose of testing whether the compiler output is correct. It does not contain all of the generated output, but only a part of it. The full output that the compiler generates contains the following definition for the global constant `@1`:
`@1 = internal constant [1 x i8] zeroinitializer`
`@1` is thus an internal global constant that represents the string ""  and more or less equivalent to
`@0 = internal constant [7 x i8] c"\00"`
which indeed means it contains only the zero-terminating character `\0`. The `getelementptr` line you mention above indeed creates a pointer to that global constant.

PR #1076, opened 2021-07-09
First Party LLVM Builds and PyQIR Wheel Generation

idavis:
This PR adds a new LLVM pipeline which builds and publishes LLVM builds for Windows/Linux/Mac tailored to our needs. To that end, the  main qsharp-compiler builds have been updated to pull these builds and install them locally. Pieces to review:

- Dev Container based on other Azure SDK dev containers which has almost all tools needed already installed.
- New submodule of llvm contained in `external/llvm-project`
- `Bootstrap.ps1`
  - has a `$Env:WHEEL_VERSION` for the PyQIR wheels. Rust requires semver instead of the nuget versioning scheme, so we may want to change how this is configured.
  - Generates the `Cargo.toml` with the `$Env:WHEEL_VERSION` set
- Dependency on Chocolatey has been removed. This is no longer needed and was causing random build failures as we were being rate limited.
- `QIR.sln` is compiled with vcvars configured on windows to get clang configured
- New `utils.ps1` contains build script utilities.
- `ci.yml` has new build matrix and jobs
- `libLLVM` the LLVM pipeline also assembles and renames the `libLLVM` libraries which we can leverage as alternative library sources to `LLVMSharp`
- LLVM is compiled in a ManyLinux 2014 container
- PyQIR has its own `build.ps1` which can be used on its own or from the main `build.ps1`


src/QirTools/pyqir/pyqir/__init__.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def x(self, qubit: str):
+        self.pyqir.x(qubit)
+
+    def y(self, qubit: str):
+        self.pyqir.y(qubit)
+
+    def z(self, qubit: str):
+        self.pyqir.z(qubit)
+
```
cgranade, line 51: Are methods planned for classical if blocks, loops, and subroutine calls? Those will be necessary for generating quantum programs (e.g.: teleportation) from a pure Python interface.

src/QirTools/pyqir/qirlib/build.rs
```diff
@@ -0,0 +1,30 @@
...
+
+fn main() -> Result<(), Box<dyn Error>> {
+    println!("cargo:rerun-if-env-changed=TARGET");
+    println!("cargo:rerun-if-changed=build.rs");
+    println!("cargo:rerun-if-changed=runtime.csproj");
+
+    let root_path = env::var("CARGO_MANIFEST_DIR")?;
+    let packages = PathBuf::from(format!("{}/packages", root_path.clone()));
+
+    let output = Command::new("dotnet")
```
cgranade, line 15: It looks like this is used to make sure the QIR runtime is made available by the relevant NuGet package, is that correct? If so, a comment indicating such may be helpful here.

src/QirTools/pyqir/qirlib/build.rs
```diff
@@ -0,0 +1,30 @@
+use std::env;
```
cgranade, line 1: Copyright header.

src/QirTools/pyqir/qirlib/runtime.csproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Quantum.Qir.Runtime" Version="0.18.2106148911-alpha" />
```
cgranade, line 8: Is there a way to not hardcode the package version numbers here? If not, we should make sure to exempt this from the version number update steps in bootstrap.ps1.

src/QirTools/pyqir/qirlib/src/emit/constants.rs
```diff
@@ -0,0 +1,49 @@
+use super::types::Types;
```
cgranade, line 1: Copyright header.

src/QirTools/pyqir/qirlib/src/emit/intrinsics.rs
```diff
@@ -0,0 +1,95 @@
+use inkwell::module::Module;
```
cgranade, line 1: Copyright header.

src/QirTools/pyqir/qirlib/src/emit/intrinsics.rs
```diff
@@ -0,0 +1,95 @@
...
+use inkwell::values::FunctionValue;
+
+pub const M: &str = "M";
```
cgranade, line 4: How will this work for instruction sets other than those represented in Microsoft.Quantum.Intrinsic? Will it be possible for users to emit (for example) calls to MS or IsingZZ gates on platforms that support those instructions?

src/QirTools/pyqir/qirlib/src/emit/intrinsics.rs
```diff
@@ -0,0 +1,95 @@
...
+pub const R_X: &str = "Rx";
```
cgranade, line 5: Can some of the duplicated definitions here possibly be reduced by a `macro_rules!` macro?

src/QirTools/pyqir/qirlib/src/emit/mod.rs
```diff
@@ -0,0 +1,276 @@
+use self::constants::Constants;
```
cgranade, line 1: Copyright header.

src/QirTools/pyqir/qirlib/src/emit/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+use self::runtime_library::RuntimeLibrary;
+use self::types::Types;
+use inkwell::values::BasicValueEnum;
+
+use super::interop::SemanticModel;
+use std::collections::BTreeMap;
+use std::path::Path;
+
+pub mod constants;
+    mod intrinsics;
```
cgranade, line 12: Is this standard `rustfmt`? I'd suggest adding a `rustfmt` check to CI passes regardless just to make sure we're using a consistent style.

src/QirTools/pyqir/qirlib/src/emit/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+pub mod qir;
+mod runtime_library;
+pub mod types;
+
+pub struct Emitter {}
```
cgranade, line 17: I love zero-size types. ♥

src/QirTools/pyqir/qirlib/src/emit/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+        context.builder.build_return(Some(&output.0));
+
+        if let Err(err) = context.module.verify() {
+            let message = err.to_string();
+            return Err(message);
+        }
+        Ok(())
+    }
+
+    fn free_qubits<'ctx>(context: &Context<'ctx>, qubits: &BTreeMap<String, BasicValueEnum<'ctx>>) {
```
cgranade, line 62: Purely out of curiosity, what motivates the use of a BTreeMap versus a HashMap here?

src/QirTools/pyqir/qirlib/src/emit/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+        model: &SemanticModel,
+        context: &Context<'ctx>,
+    ) -> BTreeMap<String, (BasicValueEnum<'ctx>, Option<u64>)> {
+        let mut registers = BTreeMap::new();
+        let number_of_registers = model.registers.len() as u64;
+        if number_of_registers > 0 {
+            let results =
+                qir::array1d::emit_array_allocate1d(&context, 8, number_of_registers, "results");
+            registers.insert(String::from("results"), (results, None));
+            let mut sub_results = vec![];
```
cgranade, line 91: Is it possible to avoid mutability here by using `.map(|reg| ...).collect()` on `model.registers.iter`?

src/QirTools/pyqir/qirlib/src/emit/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+        let intrinsics = Intrinsics::new(&module);
+        let constants = Constants::new(&module, &types);
+        Context {
+            builder: builder,
+            module: module,
+            types: types,
+            context: context,
+            runtime_library: runtime_library,
+            intrinsics: intrinsics,
+            constants: constants,
```
cgranade, line 148: Since each local variable here has the same name as the struct field being set, you can use the short-hand notation provided for struct initialization to reduce duplication. I believe this is also suggested by `cargo clippy` as well.

```suggestion
            builder,
            module,
            types,
            context,
            runtime_library,
            intrinsics,
            constants,
```

src/QirTools/pyqir/qirlib/src/emit/qir/array1d.rs
```diff
@@ -0,0 +1,237 @@
+use inkwell::values::{BasicValue, BasicValueEnum};
```
cgranade, line 1: Copyright header.

src/QirTools/pyqir/qirlib/src/emit/qir/array1d.rs
```diff
@@ -0,0 +1,237 @@
...
+use super::calls;
+
+pub(crate) fn emit_array_1d<'ctx>(
+    context: &Context<'ctx>,
+    name: &str,
+    size: u64,
+) -> (BasicValueEnum<'ctx>, Vec<(u64, BasicValueEnum<'ctx>)>) {
+    let sub_result_name = format!("{}", &name[..]);
+    let sub_result = emit_array_allocate1d(&context, 8, size, sub_result_name.as_str());
+    let mut items = vec![];
```
cgranade, line 16: Same question here, is it possible to avoid mutability by using `.collect()`?

src/QirTools/pyqir/qirlib/src/emit/qir/array1d.rs
```diff
@@ -0,0 +1,237 @@
...
+    );
+    cast
+}
+
+pub(crate) fn emit_empty_result_array_allocate1d<'ctx>(
+    context: &Context<'ctx>,
+    result_name: &str,
+) -> BasicValueEnum<'ctx> {
+    let results = emit_array_allocate1d(&context, 8, 0, &result_name[..]);
+    results
```
cgranade, line 148:
```suggestion
    emit_array_allocate1d(&context, 8, 0, &result_name[..])
```

src/QirTools/pyqir/qirlib/src/emit/qir/array1d.rs
```diff
@@ -0,0 +1,237 @@
...
+    }
+}
+
+pub(crate) fn create_ctl_wrapper<'ctx>(
+    context: &Context<'ctx>,
+    control_qubit: &BasicValueEnum<'ctx>,
+) -> BasicValueEnum<'ctx> {
+    let name = String::from("__controlQubits__");
+    let control_qubits = emit_array_allocate1d(&context, 8, 1, &name[..]);
+    wrap_value_in_array(
```
cgranade, line 214: Out of curiosity, how does this modify `control_qubits` given that local variable is immutable, and only an immutable reference is passed to `wrap_value_in_array`?

src/QirTools/pyqir/qirlib/src/emit/qir/basic_values.rs
```diff
@@ -0,0 +1,38 @@
+use inkwell::values::{BasicValue, BasicValueEnum};
```
cgranade, line 1: Copyright header.

src/QirTools/pyqir/qirlib/src/emit/qir/instructions.rs
```diff
@@ -0,0 +1,167 @@
...
+) -> () {
+    let intrinsics = &context.intrinsics;
+    let find_qubit = |name| get_qubit(name, qubits);
+    let ctl = |value| create_ctl_wrapper(context, value);
+    match inst {
+        Instruction::Cx(inst) => {
+            let control = ctl(&find_qubit(&inst.control));
+            let qubit = find_qubit(&inst.target);
+            controlled(context, intrinsics.x_ctl, control, qubit);
+        }
```
cgranade, line 37: Several of these look fairly similar, is it possible to consolidate the `Cx` and `Cz` arms, for example?

src/QirTools/pyqir/qirlib/src/emit/qir/instructions.rs
```diff
@@ -0,0 +1,167 @@
...
+        Instruction::S(inst) => {
+            calls::emit_void_call(context, intrinsics.s, &[find_qubit(&inst.qubit)])
+        }
+        Instruction::Sdg(inst) => {
+            calls::emit_void_call(context, intrinsics.s_adj, &[find_qubit(&inst.qubit)])
+        }
+        Instruction::T(inst) => {
+            calls::emit_void_call(context, intrinsics.t, &[find_qubit(&inst.qubit)])
+        }
+        Instruction::Tdg(inst) => {
```
cgranade, line 85: Elsewhere you've called this `t_adj` or `TAdj`, such that it may be good to use that terminology here as well. In particular, "dagger" is the notation used for the adjoint (e.g.: 𝑇⁺ is the adjoint of 𝑇), so using "adjoint" in preference to "dagger" is more explicit about the meaning here.

src/QirTools/pyqir/qirlib/src/emit/qir/mod.rs
```diff
@@ -0,0 +1,35 @@
...
+use super::Context;
+
+pub mod array1d;
+pub mod basic_values;
+pub mod calls;
+pub mod instructions;
+pub mod qubits;
+
+pub(crate) fn get_entry_function<'ctx>(context: &Context<'ctx>) -> FunctionValue<'ctx> {
+    let ns = "QuantumApplication";
```
cgranade, line 14: Is this namespace name fixed, then? Would it make sense to allow that to be user-configurable?

src/QirTools/pyqir/qirlib/src/interop.rs
```diff
@@ -0,0 +1,121 @@
...
+}
+
+impl Controlled {
+    pub fn new(control: String, target: String) -> Self {
+        Controlled { control, target }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct Rotated {
```
cgranade, line 52: What does this struct represent?

src/QirTools/pyqir/qirlib/src/interop.rs
```diff
@@ -0,0 +1,121 @@
...
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum Register {
+    Quantum(QuantumRegister),
+    Classical(ClassicalRegister),
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Controlled {
```
cgranade, line 40: Can this be used to control arbitrary instructions (similar to the `Controlled` keyword in Q#)? If not, what instruction does this represent?

src/QirTools/pyqir/qirlib/src/interop.rs
```diff
@@ -0,0 +1,121 @@
...
+    Cx(Controlled),
+    Cz(Controlled),
+    H(Single),
+    M { qubit: String, target: String },
+    Reset(Single),
+    Rx(Rotated),
+    Ry(Rotated),
+    Rz(Rotated),
+    S(Single),
+    Sdg(Single),
```
cgranade, line 86: It looks like `Sdg` and `Tdg` aren't used in Base-Profile.md, but rather `s_adj` and `t_adj` such that it should be possible to use `SAdj` and `TAdj` here while still being consistent with that spec.

src/QirTools/pyqir/qirlib/src/interop.rs
```diff
@@ -0,0 +1,121 @@
...
+    T(Single),
+    Tdg(Single),
+    X(Single),
+    Y(Single),
+    Z(Single),
+}
+
+pub struct SemanticModel {
```
cgranade, line 94: Is this the model for a quantum program being omitted? It looks like as is, it's limited to a single non-adaptive subroutine (that is, a circuit). You may need to extend the model to allow for examples like teleportation.

src/QirTools/pyqir/setup.cfg
```diff
@@ -0,0 +1,5 @@
+[aliases]
+test = pytest
+
+[tool:pytest]
+addopts = --verbose --capture=no
```
cgranade, line 5:
```suggestion
addopts = --verbose --capture=no

```

src/QirTools/pyqir/setup.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python
+import sys
+
+from setuptools import setup
+from setuptools_rust import Binding, RustExtension
+
+setup(
+    name="pyqir",
+    version="0.0.1",
```
cgranade, line 9: We should ideally pick up the version number from CI, similar to what's done for other QDK Python packages.

src/QirTools/pyqir/tests/test_api.py
```diff
@@ -0,0 +1,47 @@
...
+    builder.t_adj("q3")
+    builder.x("q0")
+    builder.y("q1")
+    builder.z("q2")
+
+    builder.m("q0", "c0")
+    builder.m("q1", "c1")
+    builder.m("q2", "c2")
+    builder.m("q3", "c3")
+    builder.build("pytest.ll")
```
cgranade, line 47: It doesn't look like these tests currently check that the built QIR matches what's expected, is that intended?

cgranade (commented):
Took a quick pass, but am happy to give a more detailed review later. Thanks for opening this, looks exciting!

src/QirTools/pyqir/qirlib/src/interop.rs
idavis: only circuits are targeted right now.

src/QirTools/pyqir/qirlib/src/emit/mod.rs
idavis:
```rust
        let qubits = model
            .qubits
            .iter()
            .map(|reg| {
                let indexed_name = format!("{}{}", &reg.name[..], reg.index);
                let value = qir::qubits::emit_allocate(&context, indexed_name.as_str());
                (indexed_name, value)
            })
            .collect();

        qubits
```

src/QirTools/pyqir/qirlib/src/emit/qir/array1d.rs
idavis: Separated on purpose to give debugging/inspection attachment point

src/QirTools/pyqir/README.md
```diff
@@ -0,0 +1,82 @@
...
+Once installed, the tests can be run:
+
+```bash
+pytest -v tests
+```
+
+## TODO
+
+ - Automate IR Runtime definition dependency.
+ - Currently the build only works for the single Python version installed. Future work may include:
```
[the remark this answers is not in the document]
idavis: Handled offline and in later commits.

src/QirTools/pyqir/pyqir/__init__.py
idavis: Handled offline and in later commits.

src/QirTools/pyqir/setup.py
idavis: Handled offline and in later commits.

src/QirTools/pyqir/tests/test_api.py
idavis: Handled offline and in later commits.

swernli:
One important question regarding the use of `Invoke-Expression` and child `pwsh` process in build.ps1. Can you confirm that errors in building get correctly reflected in the `$LASTEXITCODE`? I can see that the pipeline for this PR has failed in the past, but not in the build step since that change was introduced.

swernli (commented):
Somehow these comments didn’t get published, so doing that now. I see you already resolved the LASTEXITCODE issue, thanks!

build/build.ps1
```diff
@@ -134,23 +132,42 @@ function Build-VS() {
...
 
 $all_ok = $True
 
-Build-One '../QsCompiler.sln'
-Build-One '../examples/QIR/QIR.sln'
-Build-One '../src/QuantumSdk/Tools/Tools.sln'
-Build-One '../QsFmt.sln'
+if (Test-BuildCompiler) {
+    Build-One '../QsCompiler.sln'
+    Build-One '../examples/QIR/QIR.sln' -useVcVars
```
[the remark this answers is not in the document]
idavis: Commented, this also sets the ground if we move forward with formalizing build targets.

src/QirTools/qirlib/src/emit/constants.rs
```diff
@@ -1,22 +1,25 @@
...
 use super::types::Types;
 use inkwell::module::Module;
 use inkwell::values::GlobalValue;
-use inkwell::values::PointerValue;
+//use inkwell::values::PointerValue;
 
 pub(crate) struct Constants<'ctx> {
-    pub(crate) unit: PointerValue<'ctx>,
-    pub(crate) pauli_i: GlobalValue<'ctx>,
+    //pub(crate) unit: PointerValue<'ctx>,
```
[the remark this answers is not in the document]
idavis: Great question. I had implemented them as part of my initial work. When I started using an optimized bitcode base, these bits no longer worked. My plan was to keep them in for the moment as I wanted to have a conversation about profile support and what we need in the python library.

build/utils.ps1
```diff
@@ -0,0 +1,102 @@
...
+        $env:ENABLE_COMPILER -ne $false
+    }
+}
+
+function Test-BuildVsix {
+    ($Env:ENABLE_VSIX -ne "false") -and $IsWindows
+}
+
+function Test-BuildLlvmComponents {
+    ((Test-Path env:\ENABLE_LLVM_BUILDS) -and ($env:ENABLE_LLVM_BUILDS -eq $true))
```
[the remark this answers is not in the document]
idavis: I use `$true` as part of PowerShell's fuzzy falsy evaluation:

```
(base) PS > "$false" -eq $true
False
(base) PS > "false" -eq $false
True
(base) PS > "$false" -eq $false
True
(base) PS > "true" -eq $true
True
(base) PS > "$true" -eq $true
True
(base) PS > "TruE" -eq $true 
True
(base) PS > "nottrue" -eq $true
False
(base) PS > $true -eq $true
True
```

build/build.ps1
```diff
@@ -29,13 +24,16 @@ function Build-One {
...
-        /property:TreatWarningsAsErrors=true
+    $projectFilePath = $(Join-Path $PSScriptRoot $project)
+    $command = "dotnet build $projectFilePath -c $($Env:BUILD_CONFIGURATION) -v $($Env:BUILD_VERBOSITY) $($args) /property:Version=$($Env:ASSEMBLY_VERSION) /property:InformationalVersion=$($Env:SEMVER_VERSION) /property:TreatWarningsAsErrors=true /p:Platform='Any CPU'"
+
+    if ($useVcVars -and $IsWindows) {
+        $command = ". $(Join-Path $PSScriptRoot vcvars.ps1);$command"
+        Write-Host $command
+        pwsh -Command $command
+    } else {
+        Invoke-Expression $command
```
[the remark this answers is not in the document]
idavis: Yes, I'm sure. Here is the output from the build agent with the env applying:
```
Time Elapsed 00:02:40.65
##[info]Building ../examples/QIR/QIR.sln ...
. D:\a\1\s\build\vcvars.ps1;dotnet build D:\a\1\s\build\..\examples\QIR\QIR.sln -c Release -v n  /property:Version=0.16.2109.402 /property:InformationalVersion= /property:TreatWarningsAsErrors=true /p:Platform='Any CPU';exit $LASTEXITCODE
vs located at: C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise
**********************************************************************
** Visual Studio 2019 Developer Command Prompt v16.11.0
** Copyright (c) 2021 Microsoft Corporation
**********************************************************************
[vcvarsall.bat] Environment initialized for: 'x64'
setting env: agent.jobstatus = Succeeded
....
```
From [Build All](https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=32134&view=logs&j=ca395085-040a-526b-2ce8-bdc85f692774&t=319cc4f9-2fdf-5261-34c0-5e64584fddb0&l=6138) line 6138+

src/QirTools/qirlib/src/emit/constants.rs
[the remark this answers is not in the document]
swernli: Sounds good. Given this is for merging to the feature branch, that works fine.

swernli (approved):

swernli:
The failures in the QIR validation might be due to the feature branch not having this change: https://github.com/microsoft/qsharp-compiler/pull/1129.

build/build.ps1
```diff
@@ -29,13 +24,16 @@ function Build-One {
...
-        -v $Env:BUILD_VERBOSITY `
-        @args `
-        /property:Version=$Env:ASSEMBLY_VERSION `
-        /property:InformationalVersion=$Env:SEMVER_VERSION `
-        /property:TreatWarningsAsErrors=true
+    $projectFilePath = $(Join-Path $PSScriptRoot $project)
+    $command = "dotnet build $projectFilePath -c $($Env:BUILD_CONFIGURATION) -v $($Env:BUILD_VERBOSITY) $($args) /property:Version=$($Env:ASSEMBLY_VERSION) /property:InformationalVersion=$($Env:SEMVER_VERSION) /property:TreatWarningsAsErrors=true /p:Platform='Any CPU'"
+
+    if ($useVcVars -and $IsWindows) {
+        $command = ". $(Join-Path $PSScriptRoot vcvars.ps1);$command"
```
anpaz: Looks like this script is only used by pyqir so wondering if should really live in that folder.
Also, I don't see it mentioned in the README.md of that folder, so wondering how other developers will find it when trying to build locally.

build/build.ps1
```diff
@@ -134,23 +132,42 @@ function Build-VS() {
     Pop-Location
 }
 
+##
```
anpaz, line 135: This seems like a left-over...

build/libllvm.ps1
```diff
@@ -0,0 +1,105 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
```
anpaz, line 3: I think a summary explaning what this file does would be useful.

build/llvm.ps1
```diff
@@ -0,0 +1,128 @@
...
+    function Invoke-ContainerImage {
+        # Verify input files/folders and mounts exist
+        Assert (Test-Path $llvmCmakeFile) "$($llvmCmakeFile) is missing"
+        Assert (Test-Path $llvmDir) "$($llvmDir) is missing"
+        Assert (Test-Path $llvmBuildDir) "$($llvmBuildDir) is missing"
+
+        Assert (Test-Path $srcRoot) "$($srcRoot) is missing"
+
+        Write-Vso "Running container image:"
+        $command = "docker run --rm -t --user vsts -e PKG_NAME=$($Env:PKG_NAME) -e SOURCE_DIR=$srcRoot -e LLVM_CMAKEFILE=$llvmCmakeFile -e LLVM_DIR=$llvmDir -e LLVM_BUILD_DIR=$llvmBuildDir -e CCACHE_DIR=$cacheRoot -e CCACHE_CONFIGPATH=$cacheRoot -v $($srcRoot):$($srcRoot) -v $($cacheRoot):$($cacheRoot) manylinux-llvm-builder"
```
anpaz, line 76: please keep each argument in a different line so it's easier to read/diff.

build/llvm.ps1
```diff
@@ -0,0 +1,128 @@
...
+}
+
+Write-Vso "Generating package: $($Env:PKG_NAME)"
+
+if ($IsLinux) {
+    function Build-ContainerImage {
+
+        Write-Vso "Building container image manylinux-llvm-builder"
+        exec -wd $buildDir {
+            Get-Content manylinux.Dockerfile | docker build -t manylinux-llvm-builder --build-arg LLVM_BUILD_DIR="$llvmBuildDir" -
```
anpaz, line 63: Please keep args in a different line.

build/llvm.yml
```diff
@@ -0,0 +1,212 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+trigger: none
+
+variables:
+  CCACHE_DIR: $(Pipeline.Workspace)/ccache
+  AQ_LLVM_PACKAGE_GIT_VERSION: ""
```
anpaz, line 8: If `AQ` is for Azure Quantum, we're actually using `AZURE_QUANTUM` for variable names. Wondering if this really needs prefixing as it's only used by our internal build.

build/wrap-up.yml
```diff
@@ -19,6 +19,24 @@ steps:
...
+- task: CopyFiles@2
+  displayName: 'Copy Wheels to: $(Wheels.Outdir)'
+  condition: succeededOrFailed()
+  inputs:
+    SourceFolder: '$(System.DefaultWorkingDirectory)'
+    Contents: 'src/QirTools/pyqir/target/wheels/*.*'
+    TargetFolder: '$(Wheels.Outdir)'
+    FlattenFolders: true
+
+- task: CopyFiles@2
```
anpaz, line 31: duplicated.

src/QirTools/pyqir/Cargo.toml.template
```diff
@@ -0,0 +1,39 @@
+[package]
+authors = ["Microsoft"]
+name = "pyqir"
+version = "0.1.0"
+edition = "2018"
+license = "MIT"
+description = "Experimental Python based QIR generation library for use with the Quantum Development Kit."
```
anpaz, line 7: Still Experimental?

src/QirTools/pyqir/build.ps1
```diff
@@ -0,0 +1,302 @@
...
+
+function Get-TargetTriple {
+    $triple = "unknown"
+    if ($IsWindows) {
+        $triple = "x86_64-pc-windows-msvc-static"
+    }
+    elseif ($IsLinux) {
+        $triple = "x86_64-unknown-linux-gnu"
+    }
+    else {
```
anpaz, line 24: Maybe check explicitly for Mac, and fail if an unknown platform, otherwise strange bugs will appear.

src/QirTools/pyqir/pyqir/__init__.py
```diff
@@ -1,59 +1,214 @@
+# Copyright(c) Microsoft Corporation.
+# Licensed under the MIT License.
+
 from .pyqir import *
 
+
 class QirBuilder:
+    """
+    The QirBuilder objct describes quantum circuits and emits QIR
```
anpaz, line 9:
```suggestion
    The QirBuilder object describes quantum circuits and emits QIR
```

src/QirTools/pyqir/pyqir/__init__.py
```diff
@@ -1,59 +1,214 @@
...
         self.pyqir.y(qubit)
 
     def z(self, qubit: str):
+        """
+        Applies Z operation to the target qubit
+
+        :param qubit: name of the target qubit
+        :type qubit: str
+        """
         self.pyqir.z(qubit)
```
anpaz, line 164: I see a lot of intrinsics missing, along with their adjoints and controlled. Are we planning to add them later?

src/QirTools/pyqir/pyqir/__init__.py
```diff
@@ -1,59 +1,214 @@
...
+
+    :param module: name of the QIR module
+    :type module: str
+    """
+
     def __init__(self, module: str):
         self.pyqir = PyQIR(module)
 
-    def cx(self, control:str, target:str):
+    def cx(self, control: str, target: str):
```
anpaz, line 18: may i recommend using `x_ctr` or `ctr_x` so it's easy to define controlled versions of all intrinsics?

src/QirTools/pyqir/pyqir/__init__.py
swernli: I believe this is meant to map to the set of gates defined in the Base Profile draft here: https://github.com/microsoft/qsharp-language/blob/ageller/profile/Specifications/QIR/Base-Profile.md. Note that the set of gates is the same except for the Toffoli gate.

src/QirTools/pyqir/pyqir/__init__.py
swernli: In this case I think the naming is intentional, since only `cx` and `cz` are supported. This matches what we do for some target packages and the Base Profile where only specific controlled gates are supported and not arbitrary controls like in full Q#.

.devcontainer/Dockerfile
```diff
@@ -1,10 +1,19 @@
...
-RUN apt-get -y update && \
-    apt-get -y install dirmngr gnupg apt-transport-https ca-certificates && \
-    apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF && \
-    sh -c 'echo "deb https://download.mono-project.com/repo/ubuntu stable-focal main" > /etc/apt/sources.list.d/mono-official-stable.list' && \
-    apt-get -y update && \
-    apt-get -y install --no-install-recommends mono-complete nuget && \
-    apt-get clean && rm -rf /var/lib/apt/lists/
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/blob/v0.191.0/containers/codespaces-linux/.devcontainer/base.Dockerfile
+
+FROM mcr.microsoft.com/vscode/devcontainers/universal:1-linux
```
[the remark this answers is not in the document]
idavis: @anpaz Yes, it does: Python, Node.js, JavaScript, TypeScript, C++, Java, C#, F#, .NET Core, PHP, PowerShell, Go, Ruby, Rust

build/utils.ps1
```diff
@@ -0,0 +1,102 @@
...
+    return
+}
+
+# Fix temp path for non-windows platforms if missing
+if (!(Test-Path env:\TEMP)) {
+    $env:TEMP = [System.IO.Path]::GetTempPath()
+}
+
+function Test-BuildCompiler {
+    if (!(Test-Path env:\ENABLE_COMPILER)) {
```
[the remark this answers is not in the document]
idavis: I've updated the documentation on the method as below.
```pwsh
function Test-BuildCompiler {
    # By default we want to build the compiler
    # If no env var is defined we default to true
    if (!(Test-Path env:\ENABLE_COMPILER)) {
        $true
    }
    else {
        # The env var exists, we use its value
        # Coerce falsy values
        # negate to keep the defalt $true
        $env:ENABLE_COMPILER -ne $false
    }
}
```
If the comment is regarding the coercion of the string `"false"`:
```pwsh
PS > "$false" -eq $true
False
PS > "false" -eq $false
True
PS > "$false" -eq $false
True
PS > "true" -eq $true
True
PS > "$true" -eq $true
True
PS > "TruE" -eq $true
True
PS > "nottrue" -eq $true
False
PS > $true -eq $true
True
```

bootstrap.ps1
```diff
@@ -79,3 +89,60 @@ Pop-Location
...
+        $false
+    }
+}
+
+function Test-BuildLlvmComponents {
+    ((Test-Path env:\ENABLE_LLVM_BUILDS) -and ($env:ENABLE_LLVM_BUILDS -eq $true))
+}
+
+function Edit-TomlContent ($inputFile, $outputFile) {
+    $outFile = New-Item -ItemType File -Path $outputFile
```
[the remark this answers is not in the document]
idavis: I renamed `Edit-TomlContent` to `Restore-CargoTomlWithVersionInfo` and moved it to `utils.ps1`. Other methods were leftover from refactoring and I deleted them.

`$WheelVersion` and `$Env:WHEEL_VERSION` follow the same rules as every other version default and are configured earlier in the file.

```pwsh
if ("$WheelVersion".Trim().Length -eq 0) {
    $WheelVersion = "$MajorVersion.$MinorVersion.$patch-alpha";
}
# ...
If ($Env:WHEEL_VERSION -eq $null) { $Env:WHEEL_VERSION ="$WheelVersion" }
```

On you final note, let's discuss. I don't think `bootstrap.ps1` should be doing anything that it is doing and it pollutes your shell creating inconsistent dev environments unless bootstrap is run every time one opens the repo.

src/QirTools/pyqir/pyqir/__init__.py
idavis: I am targeting the base profile as @swernli mentioned. We may add some others, but to start everything off we stayed with just the base. The QIR output also uses the higher level methods instead of the `__quantum__qis__??__body` calls.

interop.rs lines 89-107
``` rust
// https://github.com/microsoft/qsharp-language/blob/ageller/profile/Specifications/QIR/Base-Profile.md
#[derive(Clone, Debug, PartialEq)]
pub enum Instruction {
    Cx(Controlled),
    Cz(Controlled),
    H(Single),
    M(Measured),
    Reset(Single),
    Rx(Rotated),
    Ry(Rotated),
    Rz(Rotated),
    S(Single),
    SAdj(Single),
    T(Single),
    TAdj(Single),
    X(Single),
    Y(Single),
    Z(Single),
}
```

src/QirTools/pyqir/pyqir/__init__.py
idavis: The naming was based on the base profile qubit names except I expanded `q` `=>` `qubit`

bootstrap.ps1
```diff
@@ -79,3 +89,17 @@ Pop-Location
 Push-Location (Join-Path $PSScriptRoot 'src/QsCompiler/QirGeneration')
 .\FindNuspecReferences.ps1;
 Pop-Location
+
+try {
+    Push-Location (Join-Path $PSScriptRoot "src" "QirTools" "pyqir")
```
cgranade, line 94: Nice to have this wrapped in a try/finally so that exceptions don't corrupt the location stack. I keep meaning to write an `Invoke-AtLocation` cmdlet that takes a scriptblock and a path to do this...

```pwsh
Invoke-AtLocation (Join-Path $PSScriptRoot "src" "QirTools" "pyqir") {
     ...
}
```

build/libllvm.ps1
```diff
@@ -0,0 +1,105 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
cgranade, line 1:
```suggestion
# Copyright (c) Microsoft Corporation.
```

build/libllvm.ps1
```diff
@@ -0,0 +1,105 @@
...
+Assert (Test-Path env:\BLOBS_OUTDIR) "BLOBS_OUTDIR was not defined"
+Assert (Test-Path env:\BUILD_ARTIFACTSTAGINGDIRECTORY) "BUILD_ARTIFACTSTAGINGDIRECTORY was not defined"
+
+$llvmArchiveFiles = @(
+    "$($env:LINUX_PKG_NAME)/external/llvm-project/build/$($env:LINUX_PKG_NAME).tar.gz",
+    "$($env:WINDOWS_PKG_NAME)/external/llvm-project/build/$($env:WINDOWS_PKG_NAME).zip",
+    "$($env:DARWIN_PKG_NAME)/external/llvm-project/build/$($env:DARWIN_PKG_NAME).tar.gz"
+)
+
+for ($i = 0; $i -lt $llvmArchiveFiles.Count; $i++) {
```
cgranade, line 20: Out of curiousity, what motivates a `for` loop here versus `foreach` or `ForEach-Object`?

build/libllvm.ps1
```diff
@@ -0,0 +1,105 @@
...
+    $file = $llvmArchiveFiles[$i]
+    Write-Vso "Raw: $file"
+    $path = Join-Path (Get-Location) '..' $file
```
cgranade, line 23: If I understand this correctly, the paths at lines 14–18 are relative to the folder above the current working directory? If so, it may be worth adding that to a `.SYNOPSIS` or `.DESCRIPTION` comment.

build/libllvm.ps1
```diff
@@ -0,0 +1,105 @@
...
+Write-Vso "Move-Item -Path $darwinPath -Destination $(Join-Path $packageDir 'libLLVM.dylib')"
+Move-Item -Path $darwinPath -Destination (Join-Path $packageDir "libLLVM.dylib")
+
+$libLlvmPackage = Join-Path "$($env:BUILD_ARTIFACTSTAGINGDIRECTORY)" "$($env:PKG_NAME).zip"
+$compress = @{
+    Path             = $packageDir
+    CompressionLevel = "Fastest"
+    DestinationPath  = $libLlvmPackage
+}
+Compress-Archive @compress
```
cgranade, line 86: If this script should be run on hosts other than Windows, you may need a fallback; not all versions of `pwsh` on Linux / macOS support `Compress-Archive`. (Perhaps related, the tricks I had to pull back in the day: https://github.com/cgranade/posh-tex/blob/master/src/cmdlets/Core.ps1#L97)

build/libllvm.ps1
```diff
@@ -0,0 +1,105 @@
...
+
+foreach ($file in $llvmArchiveFiles) {
+    Move-Item -Path $file -Destination $artifactsDir
```
cgranade, line 89: IIRC, `$Path` has type `string[]` such that you should be able to use `$llvmArchiveFiles` directly here.

build/libllvm.ps1
```diff
@@ -0,0 +1,105 @@
...
+}
+
+# Add it the job's artifacts
```
cgranade, line 92:
```suggestion
# Add it to the job's artifacts
```

build/llvm.cmake
```diff
@@ -0,0 +1,72 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
cgranade, line 1:
```suggestion
# Copyright (c) Microsoft Corporation.
```

build/llvm.ps1
```diff
@@ -0,0 +1,128 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
cgranade, line 1:
```suggestion
# Copyright (c) Microsoft Corporation.
```

build/llvm.ps1
cgranade: What motivates piping `manylinux.Dockerfile` into `docker build` instead of using `docker build -f manylinux.Dockerfile`?

build/llvm.ps1
```diff
@@ -0,0 +1,128 @@
...
+            Invoke-Expression $command
+        }
+    }
+
+    Build-ContainerImage
+    Invoke-ContainerImage
+}
+else {
+
+    if ($IsWindows) {
```
cgranade, line 92: If this script is intended to work under 5.1, you may need to also check `PSEdition`.
```suggestion
    if ($PSVersionTable.PSEdition -eq "Desktop" -or $IsWindows) {
```

build/llvm.ps1
```diff
@@ -0,0 +1,128 @@
...
+        # find VS root
+        $vswhere = "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe"
+        $visualStudioPath = & $vswhere -prerelease -latest -property installationPath
+        Write-Output "vs located at: $visualStudioPath"
+
+        # Call vcvars64.bat and write the set calls to file
+        cmd.exe /c "call `"$visualStudioPath\VC\Auxiliary\Build\vcvars64.bat`" && set > %temp%\vcvars.txt"
```
cgranade, line 99: Is the `vcvars.txt` file used again, or can this be piped straight into `ForEach-Object`?

build/llvm.yml
```diff
@@ -0,0 +1,212 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
cgranade, line 1:
```suggestion
# Copyright (c) Microsoft Corporation.
```

build/utils.ps1
```diff
@@ -0,0 +1,131 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
cgranade, line 1:
```suggestion
# Copyright (c) Microsoft Corporation.
```

build/utils.ps1
```diff
@@ -0,0 +1,131 @@
...
+    }
+    else {
+        # The env var exists, we use its value
+        # Coerce falsy values and negate to keep the defalt $true
+        $env:ENABLE_COMPILER -ne $false
+    }
+}
+
+function Test-BuildVsix {
+    ($Env:ENABLE_VSIX -ne "false") -and $IsWindows
```
cgranade, line 28: Similarly here, may need to check `PSEdition` if this script is intended to be supported for use on 5.1.

build/utils.ps1
```diff
@@ -0,0 +1,131 @@
...
+function Test-BuildLlvmComponents {
+    ((Test-Path env:\ENABLE_LLVM_BUILDS) -and ($env:ENABLE_LLVM_BUILDS -eq $true))
+}
+
+# Utilities
+function exec {
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory = $true)]
+        [scriptblock]$cmd,
```
cgranade, line 40: [style nit]

```suggestion
        [scriptblock]$Command,
```

Sources:
- https://docs.microsoft.com/en-us/powershell/scripting/developer/cmdlet/activity-parameters?view=powershell-7.1
- https://docs.microsoft.com/en-us/powershell/scripting/developer/cmdlet/strongly-encouraged-development-guidelines?view=powershell-7.1#use-pascal-case-for-parameter-names

build/vcvars.ps1
```diff
@@ -0,0 +1,17 @@
+if ($IsWindows) {
```
cgranade, line 1: Copyright header, PSEdition on Windows.

src/QirTools/pyqir/README.md
```diff
@@ -15,68 +15,148 @@ The Rust project uses:
...
+
+In a command prompt:
+
+```bash
+conda install -y -c conda-forge llvm-tools=11.1.0 llvmdev=11.1.0 clang=11.1.0 cmake=3.20.4 ninja=1.10.2
+```
+
+### Windows
+
+Install Python 3.6+ from [Python.org](https://www.python.org/downloads/).
```
cgranade, line 88: Surprising that this requires both a stock Python and a `conda`-managed Python environment; can maturin and tox be used from the same environment as at line 103?

src/QirTools/pyqir/README.md
```diff
@@ -15,68 +15,148 @@ The Rust project uses:
...
 
 ```bash
-python3 setup.py install --user
+conda install -y -c conda-forge llvm-tools=11.1.0 llvmdev=11.1.0 clang=11.1.0 cmake=3.20.4 ninja=1.10.2
 ```
 
-Once installed, the tests can be run:
+## Development
+
+If this is the first time you've loaded the code we need to install the python dev packages:
```
cgranade, line 108:
```suggestion
If this is the first time you've loaded the code we need to install the Python dev packages:
```

src/QirTools/pyqir/README.md
```diff
@@ -15,68 +15,148 @@ The Rust project uses:
...
-After which we can build everything and install the module locally:
+Install Rust from [rustup](https://rustup.rs/).
+
+Install [miniconda](https://docs.conda.io/en/latest/miniconda.html#latest-miniconda-installer-links).
+
+In a command prompt:
 
 ```bash
-python3 setup.py install --user
+conda install -y -c conda-forge llvm-tools=11.1.0 llvmdev=11.1.0 clang=11.1.0 cmake=3.20.4 ninja=1.10.2
```
cgranade, line 103: It may also be helpful to provide an environment.yml file with these versions locked.

src/QirTools/pyqir/README.md
```diff
@@ -15,68 +15,148 @@ The Rust project uses:
...
+
+Two targets are available for tox:
+- `python -m tox -e test`
+  - Runs the python tests in an isolated environment
+- `python -m tox -e pack`
+  - Packages all wheels in an isolated environment
+
+### Packaging
+
+Note: For Windows, the packaging will look for python installations available and build for them. More information on [supporting multiple python versions on Windows](https://tox.readthedocs.io/en/latest/developers.html?highlight=windows#multiple-python-versions-on-windows)
```
cgranade, line 138:
```suggestion
Note: For Windows, the packaging will look for python installations available and build for them. More information on [supporting multiple Python versions on Windows](https://tox.readthedocs.io/en/latest/developers.html?highlight=windows#multiple-python-versions-on-windows)
```

src/QirTools/pyqir/build.ps1
```diff
@@ -0,0 +1,302 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
cgranade, line 1:
```suggestion
# Copyright (c) Microsoft Corporation.

```

src/QirTools/pyqir/build.ps1
```diff
@@ -0,0 +1,302 @@
...
+
+function Get-CommitHash {
+    # rev-parse changes length based on your git settings, use length = 9
+    # to match azure devops
+    exec { git rev-parse --short=9 HEAD }
+}
+
+function Get-TargetTriple {
+    $triple = "unknown"
+    if ($IsWindows) {
```
cgranade, line 18: PSEdition or require 6.0.

src/QirTools/pyqir/pyqir/__init__.py
```diff
@@ -1,59 +1,214 @@
...
+        """
+        Applies controlled X operation to the target qubit
```
cgranade, line 20:
```suggestion
        Applies a singly controlled X operation to the target qubit.
```

src/QirTools/pyqir/pyqir/__init__.py
```diff
@@ -1,59 +1,214 @@
+# Copyright(c) Microsoft Corporation.
+# Licensed under the MIT License.
+
 from .pyqir import *
 
+
 class QirBuilder:
+    """
+    The QirBuilder object describes quantum circuits and emits QIR
```
cgranade, line 9:
```suggestion
    The QirBuilder object describes quantum circuits and emits QIR.
```

It may also be helpful to reference the QIR Base Profile explicitly here instead of implicitly via the word "circuit." (Typically, a "quantum circuit" is a nonadaptive quantum program, but I think we mean something more specific in this context?)

src/QirTools/pyqir/pyqir/__init__.py
```diff
@@ -1,59 +1,214 @@
...
+        :type control: str
+        :param target: name of the target qubit
+        :type target: str
+        """
         self.pyqir.cx(control, target)
 
-    def cz(self, control:str, target:str):
+    def cz(self, control: str, target: str):
+        """
+        Applies controlled Z operation to the target qubit
```
cgranade, line 31: Similarly here.

src/QirTools/pyqir/pyqir/__init__.py
```diff
@@ -1,59 +1,214 @@
...
+        :param control: name of the control qubit
+        :type control: str
+        :param target: name of the target qubit
+        :type target: str
+        """
         self.pyqir.cz(control, target)
 
     def h(self, target: str):
+        """
+        Applies H operation to the target qubit
```
cgranade, line 42:
```suggestion
        Applies a Hadamard (H) operation to the target qubit.
```

src/QirTools/pyqir/pyqir/__init__.py
```diff
@@ -1,59 +1,214 @@
...
+
+        :param target: name of the target qubit
+        :type target: str
+        """
         self.pyqir.h(target)
 
     def m(self, qubit: str, target: str):
+        """
+        Applies measurement operation or the source qubit into the target register
```
cgranade, line 51: What basis is this measurement in? I would guess 𝑍, but it's good to spell that out explicitly.

src/QirTools/pyqir/pyqir/__init__.py
```diff
@@ -1,59 +1,214 @@
...
+        :param qubit: name of the source qubit
+        :type qubit: str
+        :param target: name of the target register
+        :type target: str
+        """
         self.pyqir.m(qubit, target)
 
     def reset(self, target: str):
+        """
+        Applies Reset operation to the target qubit
```
cgranade, line 62: It may be good to explicitly specify what the reset operation does here (i.e.: resets the target qubit into the |0⟩ state), or to link to an explicit definition elsewhere.

src/QirTools/pyqir/pyqir/__init__.py
```diff
@@ -1,59 +1,214 @@
...
+
+        :param target: name of the target qubit
+        :type target: str
+        """
         self.pyqir.reset(target)
 
     def rx(self, theta: float, qubit: str):
+        """
+        Applies Rx operation to the target qubit
```
cgranade, line 71: Similarly here, there's a lot of different conventions for rotation operations such that either explicitly listing the matrix form of the unitary that can be used to simulate this operation or linking to Base Profile docs would be really helpful in disambiguiating.

src/QirTools/pyqir/pyqir/__init__.py
```diff
@@ -1,59 +1,214 @@
...
+        :param theta: rotation value for target qubit
+        :type theta: float
+        :param qubit: name of the target qubit
+        :type qubit: str
+        """
         self.pyqir.rz(theta, qubit)
 
     def s(self, qubit: str):
+        """
+        Applies S operation to the target qubit
```
cgranade, line 104: I take it that the S operation is the phase gate simulated by 𝑆 = diag(1, 𝑖)?

src/QirTools/pyqir/pyqir/__init__.py
```diff
@@ -1,59 +1,214 @@
...
+        :type qubit: str
+        """
         self.pyqir.z(qubit)
 
-    def add_classical_register(self, name, size):
+    def add_classical_register(self, name: str, size: int):
+        """
+        Models a classical register of the given size. The individual values
+        are accessed by name "<name><index>" with 0 based indicies.
+        Example:
```
cgranade, line 169: Should this example be formatted as an RST code block?
```suggestion
        Example::

```

src/QirTools/pyqir/tests/test_api.py
```diff
@@ -18,6 +18,40 @@ def test_bell_no_measure():
...
+    builder.h("input3")
+    builder.h("input4")
+
+    builder.m("input0", "output0")
+    builder.m("input1", "output1")
+    builder.m("input2", "output2")
+    builder.m("input3", "output3")
+    builder.m("input4", "output4")
+
+    builder.build("bernstein_vazirani.ll")
```
cgranade, line 53: Does the output of this build step get checked in another test?

src/QirTools/qirlib/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "qirlib"
+version = "0.1.0"
+edition = "2018"
+description = "QIR generation library for use with the Quantum Development Kit."
```
cgranade, line 5: It may be helpful to note in the description that this is currently limited to Base Profile / nonadaptive programs (informally, "circuits").

src/QirTools/qirlib/src/emit/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+use std::collections::HashMap;
+use std::path::Path;
+
+pub mod constants;
+mod intrinsics;
+pub mod qir;
+mod runtime_library;
+pub mod types;
+
+pub struct Emitter {}
```
cgranade, line 20: A Rustdoc comment here as to the purpose of the `Emitter` struct may be helpful.

src/QirTools/qirlib/src/emit/qir/basic_values.rs
```diff
@@ -18,6 +21,17 @@ pub(crate) fn u64_to_i32<'ctx>(context: &Context<'ctx>, value: u64) -> BasicValu
         .as_basic_value_enum()
 }
 
+pub(crate) fn i64_to_i32<'ctx>(context: &Context<'ctx>, value: i64) -> BasicValueEnum<'ctx> {
```
cgranade, line 24: Odd thought, but would a new `trait IntoBasicValue<'ctx, T> { fn into_basic_value(context: &Context<'ctx>) -> BasicValueEnum<'ctx>; }` be helpful? Could then group these functions into something like `impl<'ctx> IntoBasicValue<'ctx, i32> for i64` blocks?

src/QirTools/qirlib/src/emit/qir/instructions.rs
```diff
@@ -0,0 +1,170 @@
...
+};
+use crate::{emit::Context, interop::Instruction};
+use inkwell::values::{BasicValueEnum, FunctionValue};
+use std::collections::HashMap;
+
+fn get_qubit<'ctx>(
+    name: &String,
+    qubits: &HashMap<String, BasicValueEnum<'ctx>>,
+) -> BasicValueEnum<'ctx> {
+    qubits.get(name).unwrap().to_owned()
```
cgranade, line 16: Does this `unwrap` fail when an invalid `name` is passed? If so, it may be worth using `?` instead to bubble that out into a user-actionable `Result` rather than `panic`ing when `name` is invalid.

cgranade (commented):
Left a few more comments, thanks for tagging me in!

bootstrap.ps1
```diff
@@ -63,10 +72,11 @@ Get-ChildItem -Recurse *.v.template `
             | Set-Content $Target
     }
 
-If ($Env:ASSEMBLY_VERSION -eq $null) { $Env:ASSEMBLY_VERSION ="$AssemblyVersion" }
-If ($Env:NUGET_VERSION -eq $null) { $Env:NUGET_VERSION ="$NuGetVersion" }
-If ($Env:SEMVER_VERSION -eq $null) { $Env:SEMVER_VERSION ="$SemverVersion" }
-If ($Env:VSVSIX_VERSION -eq $null) { $Env:VSVSIX_VERSION ="$VsVsixVersion" }
+    If ($Env:ASSEMBLY_VERSION -eq $null) { $Env:ASSEMBLY_VERSION ="$AssemblyVersion" }
```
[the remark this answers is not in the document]
idavis: Not sure how that happened. Reformatting now.

build/libllvm.ps1
idavis: @cgranade This is the format used in 13 powershell files, 7 C#, 91 F# files

build/libllvm.ps1
idavis: I wanted to save the fully resolved path back into the original collection. Stylistic choice, I didn't like the alternative
```pwsh
$llvmArchiveFiles = $llvmArchiveFiles | ForEach-Object {
    Write-Vso "Raw: $_"
    $path = Join-Path (Get-Location) '..' $_
    Write-Vso "Joined: $path"
    $resolvedPath = Resolve-Path $path
    Write-Vso "Resolved: $resolvedPath"
    $resolvedPath
}
```

src/QirTools/pyqir/tests/test_api.py
idavis: These are API smoke tests at the moment.

build/libllvm.ps1
cgranade: Yes, and policy used to require it, but that policy was updated ~two years ago, and we are now asked to not include the phrase "all rights reserved" in new code. A lot of older code still uses the old policy, and sometimes the phrase slips through code reviews.

bootstrap.ps1
idavis: We have `exec` now that I added to `utils.ps1` and use ` -ErrorAction SilentlyContinue` if you don't want to fail if the command fails.

src/QirTools/qirlib/src/emit/mod.rs
idavis: Removed as it is no longer needed

anpaz (approved):

PR #1077, opened 2021-07-09
Package number updates

bettinaheim:
... so that the local repo build also contains the simulator features and everything is in a good state for development.


src/QsCompiler/TestTargets/Libraries/Library1/Types.qs
```diff
@@ -1,11 +1,9 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-namespace Microsoft.Quantum.Testing
-{
-    open Microsoft.Quantum.Intrinsic;
+namespace Microsoft.Quantum.Testing{
```
cesarzc, line 4: nit: maybe the opening bracket was inadvertently moved to the same line that the namespace declaration.

cesarzc (approved):

src/QsCompiler/TestTargets/Libraries/Library1/Types.qs
bettinaheim: Technically, I believe our style guide intends for them to be on the same line, though I don't think it is that prescriptive when it comes to whitespace.

PR #1080, opened 2021-07-09
Adds the Rough Draft of the Spec Doc for Formatter

ScottCarda-MS:
Rough Draft of the Specification Document for the Q# Formatter. There are things that still need to be filled out and details that need adding.


src/QsFmt/SPECIFICATION.md
```diff
@@ -0,0 +1,58 @@
...
+
+## How to use the command line formatter
+
+After installing the tool the command `qsfmt` is used to run the formatter.
+There are two commands supported for the tool: 'format' and 'update'.
+The 'format' command runs the tool as a formatter for Q# code, running only those transformation rules that update whitespace and affect indentation. The underlying code will not be changed.
+The 'update' command allows the tool to be used as a means of updating old Q# code, replacing deprecated syntax with newer supported syntax. These transformation rules change the actual code given to them, and may occasionally fail.
+These two commands work separately and can't be run together. If the user wishes to both format and update their code, they would need to run the tool twice, first with 'update' and then with 'format'.
+
+The tool will initially only support the 'update' command, as the formatting functionality is still being worked on. The 'format' command will be unavailable until it is supported.
```
kuzminrobin, line 18: Is there any default if the command ('update', 'format') is not specified?

src/QsFmt/SPECIFICATION.md
```diff
@@ -0,0 +1,58 @@
...
+## Input and Output
+Input files to the formatter will be the last argument(s) to the tool. Zero to many arguments may be given.
+ - The default behavior, which is used if no input arguments are provided, is for the formatter to take as input all files with the '.qs' extension in the current working directory.
+ - Inputs can be given as one to many paths to directories, in which all files with the '.qs' extension found will be taken as input.
+ - Inputs can be given as one to many paths to individual files with the '.qs' extension.
+ - Inputs can be any combination of paths to files and paths to directories.
+ - The `--recurse` option can be specified to process the input directories recursively for input files. You can specify this for all input directories, not per-directory.
+
+The output of the formatter is to overwrite the input files that it processes.
+
```
kuzminrobin, line 29: Will the formatter provide the transaction semantics (either successful completion or no change to the file)? 
Or, if the formatter updates the input file on the fly, and hits an error or crashes, then what happens to the input file? Will the input file remain partially overwritten (or removed/corrupted because its file handle has not been closed properly)?

src/QsFmt/SPECIFICATION.md
```diff
@@ -0,0 +1,58 @@
...
+The tool will parse the input into a concrete syntax tree using ANTLR. This tree will then have rules applied to it in the form of transformations on the tree.
+The transformations are separated into two groups: transformations run by the 'format' command, and transformations run by the 'update' command.
+Listed here are some of the transformations we intend to use.
+
+### Formatting Transformations
+
+Collapsed Spaces - removes duplicate spaces so that there is only a single space where spaces are used
+Operator Spacing - ensure that operators have the appropriate spaces in them
+New Lines - removes excessive new lines
+Indentation - ensures that there are appropriate indentation for the text
```
kuzminrobin, line 41: This all is written in a single line. You can see this by pressing the button
image

src/QsFmt/SPECIFICATION.md
```diff
@@ -0,0 +1,58 @@
...
+
+Array Syntax - [proposal](https://github.com/microsoft/qsharp-language/blob/main/Approved/2-enhanced-array-literals.md)
+Using Syntax - [proposal](https://github.com/microsoft/qsharp-language/blob/main/Approved/1-implicitly-scoped-qubit-allocation.md)
+
+## How errors are handled
+
+If an exception is thrown, that will be surfaced to the user and is a bug.
+
+Parsing errors may occur if the formatter is given text that is not proper Q# code. In this case ANTLR gives the appropriate error message.
+Errors should not occur during a formatting transformation. All formatting transformations should never encounter the syntax they are expected to change and be unable to perform the change.
```
kuzminrobin, line 55: I suspect you meant to write
> All formatting transformations should never encounter **an error in** the syntax they are expected to change

src/QsFmt/SPECIFICATION.md
```diff
@@ -0,0 +1,58 @@
...
+### Updating Transformations
+
+Updating Transformations remove outdated syntax that is deprecated and will be no longer supported in the future.
+
+Array Syntax - [proposal](https://github.com/microsoft/qsharp-language/blob/main/Approved/2-enhanced-array-literals.md)
+Using Syntax - [proposal](https://github.com/microsoft/qsharp-language/blob/main/Approved/1-implicitly-scoped-qubit-allocation.md)
+
+## How errors are handled
+
+If an exception is thrown, that will be surfaced to the user and is a bug.
```
kuzminrobin, line 52: I suspect you meant to write
> If an **_unhandled_** exception is thrown **_by the Formatter_**, that will be surfaced to the user

What extra info is carried by the fragment "and is a bug"? How the meaning will change if that fragment is removed?

src/QsFmt/SPECIFICATION.md
ScottCarda-MS: Hmm, not quite. I'm not sure the best way to word this, but the formatting transformations should not (see the syntax they need to change AND not be able to change it). If they see the syntax they are targeting, they should always be able to perform their change.

src/QsFmt/SPECIFICATION.md
ScottCarda-MS: I think, maybe in the future, the `format` command will be the default, but because we are not initially shipping with the `format` command, I think for now we go with no default. The tool will say it is missing a command if none is given. I don't think we want the `update` to be the default.

src/QsFmt/SPECIFICATION.md
ScottCarda-MS: I've rewritten this per your suggestion, although I opted for 'the tool' over 'the formatter' as I'm still a little unsure if this should be called a formatter when not specifically talking about the `format` command. I've also tried to add some clarity around the use of the word 'bug'. Its purpose is to indicate that if an unhandled exception is surfaced, this is a bug with the tool, and the user should file a bug against us so we can address it.

src/QsFmt/SPECIFICATION.md
ScottCarda-MS: Thanks, I've fixed it now

src/QsFmt/SPECIFICATION.md
ScottCarda-MS: Ideally if an error or unhandled exception is encountered, the file currently being process does not get overwritten. The file would only get overwritten if the formatting was successful. I suspect the behavior should be different for updating errors, though I still don't really understand what that means. @samarsha can you better explain what an error means in the context of an update transformation?

samarsha: For example, `new Qubit[5]` has no equivalent in the new array syntax, because there is no default value for `Qubit`. It can't be updated automatically without refactoring the code.

ScottCarda-MS: @samarsha so what should be the expected response from the tool? It shouldn't transform those instances, but does it need to do anything else?

samarsha: I don't think that has been decided yet, but I think displaying a warning for every instance that it couldn't update would make sense.

ScottCarda-MS: Alright, thanks @samarsha. Let's put a pin in this for now, and come back to it once it becomes more relevant.

src/QsFmt/SPECIFICATION.md
kuzminrobin: Then I assume that the sentence means the following:
> All formatting transformations should never encounter the syntax **_that_** they are expected to change **_but are unable to_** change

src/QsFmt/SPECIFICATION.md
```diff
@@ -0,0 +1,58 @@
...
+ - Array Syntax - [proposal](https://github.com/microsoft/qsharp-language/blob/main/Approved/2-enhanced-array-literals.md)
+ - Using Syntax - [proposal](https://github.com/microsoft/qsharp-language/blob/main/Approved/1-implicitly-scoped-qubit-allocation.md)
+
+## How errors are handled
+
+If an unhandled exception is thrown by the tool, that will be surfaced to the user and is a bug with the tool that should be filed and addressed.
+
+Parsing errors may occur if the formatter is given text that is not proper Q# code. In this case ANTLR gives the appropriate error message.
+Errors should not occur during a formatting transformation. All formatting transformations should never encounter the syntax they are expected to change and be unable to perform the change.
+Errors may occur during an updating transformation of the concrete syntax tree if an updating transformation encounters an issue, such as syntax that it would be expected to updated, but can't for some reason.
```
kuzminrobin, line 56: I assume 
> <p> that it would be expected to update<del>d</del></p>
(redundant 'd' in the word update**d**)

ScottCarda-MS: removed

kuzminrobin (approved):

src/QsFmt/SPECIFICATION.md
ScottCarda-MS: applied

src/QsFmt/SPECIFICATION.md
```diff
@@ -0,0 +1,68 @@
...
+There are two commands supported for the tool: 'format' and 'update'.
+The 'format' command runs the tool as a formatter for Q# code, running only those transformation rules that update whitespace and affect indentation. The underlying code will not be changed.
+The 'update' command allows the tool to be used as a means of updating old Q# code, replacing deprecated syntax with newer supported syntax. These transformation rules change the actual code given to them, and may occasionally fail.
+These two commands work separately and can't be run together. If the user wishes to both format and update their code, they would need to run the tool twice, first with 'update' and then with 'format'.
+Both of these commands should be idempotent, meaning that iterated uses of the commands will not continue to have effects on the inputs.
+
+The tool will initially only support the 'update' command, as the formatting functionality is still being worked on. The 'format' command will be unavailable until it is supported.
+
+## Input and Output
+Input files to the formatter will be the last argument(s) to the tool.
```
cgranade, line 22: Can stdin be passed as an input file, e.g. as `qsfmt format -`?

src/QsFmt/SPECIFICATION.md
```diff
@@ -0,0 +1,68 @@
...
+At least one input argument is expected. If no inputs are given, usage information is printed.
+Inputs can be given as one to many paths to directories, files, or any combination. In directories specified, all files with the '.qs' extension found will be taken as input.
+Basic wild cards may be used in the paths.
+The `--recurse` option can be specified to process the input directories recursively for input files. You can specify this for all input directories, not per-directory.
```
cgranade, line 26: It looks like some of these have different names than the corresponding `dotnet-format` options documented at https://github.com/dotnet/format, would it perhaps make sense to align there? Also, should there be a `--check` mode that can be used in CI to enforce formatting?

src/QsFmt/SPECIFICATION.md
```diff
@@ -0,0 +1,68 @@
...
+
+The output of the formatter is to overwrite the input files that it processes.
+
+## Rules that are run
+
+The tool will parse the input into a concrete syntax tree using ANTLR. This tree will then have rules applied to it in the form of transformations on the tree.
+The transformations are separated into two groups: transformations run by the 'format' command, and transformations run by the 'update' command.
+Listed here are some of the transformations we intend to use.
+
+### Formatting Transformations
```
cgranade, line 37: Will any of these transformations be configurable, perhaps by project file options or a new config file?

src/QsFmt/SPECIFICATION.md
ScottCarda-MS: I was using Fantomas as an example for this, found here : https://github.com/fsprojects/fantomas/blob/master/docs/Documentation.md.
Could you be more specific with what changes you are looking for?
The `--check` option is a good possibility.

src/QsFmt/SPECIFICATION.md
ScottCarda-MS: From the [Fantomas documentation](https://github.com/fsprojects/fantomas/blob/master/docs/Documentation.md) I see they have an option `--stdin` as well as an option `--stdout` for inputting and outputting to the standard streams. I hadn't seen much of an interest in this feature as of yet, the focus being on the formatting of files, so I didn't want to commit to having these options yet. They can always be added on afterwards if there is enough time and interest. Would you say that these options are pretty important to you? If they are important enough to people they can be added up front instead of later.

src/QsFmt/SPECIFICATION.md
```diff
@@ -0,0 +1,68 @@
...
+
+The tool will initially only support the 'update' command, as the formatting functionality is still being worked on. The 'format' command will be unavailable until it is supported.
+
+## Input and Output
+Input files to the formatter will be the last argument(s) to the tool.
+At least one input argument is expected. If no inputs are given, usage information is printed.
+Inputs can be given as one to many paths to directories, files, or any combination. In directories specified, all files with the '.qs' extension found will be taken as input.
+Basic wild cards may be used in the paths.
+The `--recurse` option can be specified to process the input directories recursively for input files. You can specify this for all input directories, not per-directory.
+The `--backup` option can be specified to create backup files for all input files with the original content in them.
```
efratshabtai, line 27: I think backup files should be created by default so the user can rollback the changes if needed

src/QsFmt/SPECIFICATION.md
ScottCarda-MS: Currently this is only configurable by selecting which of the two commands to use. The `update` command will run all of the updating rules, and the `format` command will run all of the formatting rules. I believe there was some talk of having these be more configurable in the future, probably through a config file, but this has yet to be designed through, so the current system is not configurable.

src/QsFmt/SPECIFICATION.md
ScottCarda-MS: So instead of a `--backup` option, we would have a `--no-backup` option to instruct the tool to not create backup files it would otherwise normally create? That seems reasonable to me. Let's see what others think on this.

samarsha: Creating a backup file for every formatted file is a fairly intrusive operation since it effectively doubles the size of every source folder, which can make navigating harder. With version control, backup files are also mostly redundant. Also, my experience with other formatters is that they don't create backup files by default.

src/QsFmt/SPECIFICATION.md
viktorveis: I recommend modeling after https://github.com/dotnet/format because other Q# commands such as  build and run use .NET.

src/QsFmt/SPECIFICATION.md
bettinaheim: Closing this thread since the immediate question seems to be answered - further conversations about the functionality in general can be part of the spec review later this months; we will merge a draft spec earlier, and update it based on that review as needed.

src/QsFmt/SPECIFICATION.md
bettinaheim: Closing this thread since the immediate question seems to be answered - further conversations about the functionality in general can be part of the spec review later this months; we will merge a draft spec earlier, and update it based on that review as needed.

src/QsFmt/SPECIFICATION.md
ScottCarda-MS: Moving this conversation to the spec document comments section. Please pick up this conversation there.

src/QsFmt/SPECIFICATION.md
ScottCarda-MS: Moving this conversation to the spec document comments section. Please pick up this conversation there.

bettinaheim (approved):
Looks like a good start. Let's pull it in; we'll update it as needed.

PR #1082, opened 2021-07-12
Fix IsTargetPackage override

swernli:
This change updates the loading of references to ignore the dll identified as the target package. This allows the `IsTargetPackage` flag to work as intended such that the callables defined in that reference override those from Microsoft.Quantum.QSharp.Core without first causing a conflict.


src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -989,7 +989,21 @@ private References LoadAssemblies(IEnumerable<string> refs, bool loadTestNames,
 
             void OnException(Exception ex) => this.LogAndUpdate(ref this.compilationStatus.ReferenceLoading, ex);
             void OnDiagnostic(Diagnostic d) => this.LogAndUpdateLoadDiagnostics(ref this.compilationStatus.ReferenceLoading, d);
-            var headers = ProjectManager.LoadReferencedAssembliesInParallel(refs ?? Enumerable.Empty<string>(), OnDiagnostic, OnException, ignoreDllResources);
+
+            // Skip loading any assemblies referenced as target packages. These will be included in a later
+            // override step.
+            var filteredRefs = refs ?? Enumerable.Empty<string>();
+            if (this.config.TargetPackageAssemblies != null)
```
idavis, line 996: nit:

using idiomatic not null check, and materializing assemblies. The code were calling into doesn't benefit from deferred evaluation. 
```csharp
            if (this.config.TargetPackageAssemblies is object)
            {
                var targetAssemblies = this.config.TargetPackageAssemblies.Select(a => Path.GetFullPath(a)).ToArray();
                filteredRefs = filteredRefs.Except(targetAssemblies).ToArray();
            }
```

swernli: Fair enough. That will make debugging the intermediate results of the Linq invocations easier.

idavis (approved):

PR #1083, opened 2021-07-16
Making it possible to manually invoke the QIR generation logic...

bettinaheim:
...without loading it as a rewrite step. Ensures that QIR and C# generation can run at the same time.

### DIFF
--- a/src/QsCompiler/Compiler/CompilationLoader.cs
+++ b/src/QsCompiler/Compiler/CompilationLoader.cs
@@ -189,9 +189,9 @@ namespace Microsoft.Quantum.QsCompiler
             public IEnumerable<string>? TargetPackageAssemblies { get; set; }
 
             /// <summary>
-            /// Whether or not QIR Generation was requested for this compilation unit.
+            /// Indicates whether the necessary compiler passes are executed for the compilation to be compatible with QIR generation.
             /// </summary>
-            public bool IsQirGenerationEnabled { get; set; }
+            public bool PrepareQirGeneration { get; set; }
 
             /// <summary>
             /// Indicates whether a serialization of the syntax tree needs to be generated.
@@ -199,6 +199,13 @@ namespace Microsoft.Quantum.QsCompiler
             /// </summary>
             internal bool SerializeSyntaxTree =>
                 this.BuildOutputFolder != null || this.DllOutputPath != null;
+
+            /// <summary>
+            /// Indicates whether the compilation needs to be monomorphized.
+            /// This value is never true if SkipMonomorphization is specified.
+            /// </summary>
+            internal bool Monomorphize =>
+                (this.IsExecutable || this.PrepareQirGeneration) && !this.SkipMonomorphization;
 
             /// <summary>
             /// Indicates whether the compiler will remove if-statements and replace them with calls to appropriate intrinsic operations.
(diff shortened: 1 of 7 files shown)


src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -573,16 +580,17 @@ public CompilationLoader(SourceLoader loadSources, ReferenceLoader loadReference
...
 
-            if ((this.config.IsExecutable || this.config.IsQirGenerationEnabled) && !this.config.SkipSyntaxTreeTrimming)
+            if (this.config.IsExecutable && !this.config.SkipSyntaxTreeTrimming)
             {
                 // TODO: It would be nicer to trim unused intrinsics. Currently, this is not possible due to how the old setup of the C# runtime works.
                 // With the new setup (interface-based approach for target packages), it is possible to trim ununsed intrinsics.
 #pragma warning disable CS0618 // Type or member is obsolete
                 // TODO: The dependencies for rewrite steps should be declared as part of IRewriteStep interface, and we should query those here.
-                var rewriteStep = new LoadedStep(new SyntaxTreeTrimming(keepAllIntrinsics: !this.config.IsQirGenerationEnabled, BuiltIn.AllBuiltIns.Select(e => e.FullName), !this.config.IsExecutable), typeof(IRewriteStep), thisDllUri);
+                var trimming = new SyntaxTreeTrimming(keepAllIntrinsics: true, BuiltIn.AllBuiltIns.Select(e => e.FullName));
```
bettinaheim, line 591: This needs to be true! See the comment above:
"TODO: It would be nicer to trim unused intrinsics. Currently, this is not possible due to how the old setup of the C# runtime works."

bettinaheim: I unfortunately missed the conversation here at that time: https://github.com/microsoft/qsharp-compiler/pull/1062#discussion_r668125520

src/QsCompiler/Compiler/RewriteSteps/Monomorphization.cs
```diff
@@ -71,7 +71,7 @@ private ImmutableArray<QsQualifiedName> AugmentedEntryPoints(QsCompilation compi
         {
             // If this compilation is for a library project, there are no defined entry points. Instead,
             // treat every public, non-generic callable as a possible entry point into the library.
-            return this.isLibrary ?
+            return compilation.EntryPoints.Length == 0 ?
```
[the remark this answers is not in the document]
swernli: I suppose the presence or lack of entry points is sufficient to identify a library, since a library by definition can't have entry points and an executable must have at least one. If that's the case and we also do intend to run monorphization for libraries even when QIR is not explicitly enabled, then maybe we no longer need `enabledForLibraries` at all?

src/QsCompiler/Compiler/CompilationLoader.cs
swernli: Ah, I didn't realize that we already have scenarios that need to support C# generation and QIR generation simultaneously. I think that since QirGenration runs it's own syntax tree trimming step, it will still clean up the unused intrinsics, right?

swernli (commented):
It looks like this isn't working for some of the compilation of libraries, and it's failing some QIR Generation tests as well. It seems like something in here broke an assumption in way that is having cascading effects.

src/QsCompiler/Compiler/CompilationLoader.cs
bettinaheim: Yes, the QirGeneration runs its own syntax tree trimming.

src/QsCompiler/Compiler/RewriteSteps/Monomorphization.cs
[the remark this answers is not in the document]
bettinaheim: I decided to remove the precondition and the parameter for both the monomorphization and the syntax tree trimming.

swernli (approved):

Issue #867, opened 2021-02-13
Formatter should compare input source with unparsed tree as a sanity test


samarsha:
As a sanity test to prevent data loss when formatting, the formatter should unparse the CST (without applying any formatting rules) and make sure that it's equal to the original input source code. If it's not, then the formatter shouldn't continue (ideally with an error message asking the user to report a bug).

samarsha:
Hi @yjt98765, you mentioned you were interested in working on this, so I can explain my idea some more.

The type of bug that this is trying to prevent is when there's a silent error in the translation from the parse tree to the syntax tree - so the syntax tree appears to have been constructed successfully, but data may have accidentally been lost or rearranged. For example, maybe the open and close parentheses of an expression accidentally swapped places in the `ExpressionVisitor`.

I think it would be better to stop the formatter with an error message in this case instead of continuing and outputting an incorrect Q# file for the user. So when the `Formatter.format` function is called, before it runs any formatting rules, it could call `printer.Document` and make sure the output matches the original source code exactly, and return an error if it doesn't.

yjt98765:
Hi @samarsha, thank you for the hint. I have initiated a pull request. I think it generally meets the target, but maybe some details need to be further discussed, such as the line-ending matching. I can continue improving it based on your feedback.

PR #1084, opened 2021-07-17
Add a sanity test before formatting Q# code

yjt98765:
This pull request implements issue #867, i.e., it unparses the syntax tree and compares the result to the original code before applying format rules.

## Modification
To use the `parse` variable, I have to change the accessibility of `Document`. In addition, the F# compiler also requires me to change the accessibility of `Terminal`, which I do not quite understand.

When the formatter fails the sanity test, it reports an error message. The message is composed based on the [README](https://github.com/microsoft/qsharp-compiler/blob/main/README.md#reporting-security-issues) file.

## Testing
I have to change the `Formats standard input` test after implementing the above feature. Because I am working on Windows, the unparsing result ends with `\r\n`, which is different from the '\n' ending in the test case. This difference makes the test fail.

The above incident inspires me to add a new test named `Shows unparsed tree not equal to source error` to [Tests.qs](https://github.com/microsoft/qsharp-compiler/blob/main/src/QsFmt/App.Tests/Tests.fs). In the Q# code attached to this test, I inserted both `\r\n` and `\n`. It ensures that the unparsed result in all environments will be different from the source. In this way, it can trigger the newly added error message.

I am not sure whether not being able to handle mixed line endings is a bug. In most cases, it will not exist in users' code. However, I tried similar things to test F#'s formatter. It can still format the file without reporting errors. Maybe we should omit the line-ending differences in the sanity test?

### DIFF
--- a/src/QsFmt/App.Tests/Tests.fs
+++ b/src/QsFmt/App.Tests/Tests.fs
@@ -90,7 +90,8 @@ let ``Formats file`` path output =
     )
 
 [<Theory>]
-[<InlineData("namespace Foo { function Bar() : Int { return 0; } }\n",
+[<InlineData("namespace Foo { function Bar() : Int { return 0; } }
+",
              "namespace Foo {
     function Bar() : Int {
         return 0;
@@ -128,3 +129,17 @@ let ``Shows file not found error`` path =
     Assert.Equal(3, result.Code)
     Assert.Empty result.Out
     Assert.NotEmpty result.Error
+
+[<Theory>]
+[<InlineData("namespace Foo { function Bar() : Int { return 0; }\n }\r\n",
+             "The formater does not work properly. Please let us know by filing a new issue in https://github.com/microsoft/qsharp-compiler/issues/new/choose.
+")>]
+let ``Shows unparsed tree not equal to source error`` input output =
+    Assert.Equal(
+        {
+            Code = 5
+            Out = ""
+            Error = output
+        },
+        run [| "-" |] input
+    )
--- a/src/QsFmt/App.Tests/Tests.fs
+++ b/src/QsFmt/App.Tests/Tests.fs
@@ -129,17 +129,3 @@ let ``Shows file not found error`` path =
     Assert.Equal(3, result.Code)
     Assert.Empty result.Out
     Assert.NotEmpty result.Error
-
-[<Theory>]
-[<InlineData("namespace Foo { function Bar() : Int { return 0; }\n }\r\n",
-             "The formater does not work properly. Please let us know by filing a new issue in https://github.com/microsoft/qsharp-compiler/issues/new/choose.
-")>]
-let ``Shows unparsed tree not equal to source error`` input output =
-    Assert.Equal(
-        {
-            Code = 5
-            Out = ""
-            Error = output
-        },
-        run [| "-" |] input
-    )
--- a/src/QsFmt/App.Tests/Tests.fs
+++ b/src/QsFmt/App.Tests/Tests.fs
@@ -90,11 +90,14 @@ let ``Formats file`` path output =
     )
 
 [<Theory>]
-[<InlineData("namespace Foo { function Bar() : Int { return 0; }\n }\r\n",
+[<InlineData("namespace Foo { function Bar() : Int { return 0; } }
+",
              "namespace Foo {
     function Bar() : Int {
         return 0;
-    }\n}\r\n")>]
+    }
+}
+")>]
 let ``Formats standard input`` input output =
     Assert.Equal(
         {
--- a/src/QsFmt/App.Tests/Tests.fs
+++ b/src/QsFmt/App.Tests/Tests.fs
@@ -90,14 +90,11 @@ let ``Formats file`` path output =
     )
 
 [<Theory>]
-[<InlineData("namespace Foo { function Bar() : Int { return 0; } }
-",
+[<InlineData("namespace Foo { function Bar() : Int { return 0; }\n }\r\n",
              "namespace Foo {
     function Bar() : Int {
         return 0;
-    }
-}
-")>]
+    }\n}\r\n")>]
 let ``Formats standard input`` input output =
     Assert.Equal(
         {
--- a/src/QsFmt/App/Program.fs
+++ b/src/QsFmt/App/Program.fs
@@ -29,13 +29,24 @@ let main args =
         let input = results.GetResult Input
         let source = if input = "-" then stdin.ReadToEnd() else File.ReadAllText input
 
-        match Formatter.format source with
-        | Ok result ->
-            printf "%s" result
-            0
+        match Formatter.parse source with
+        | Ok document ->
+            // Test whether there is data loss during parsing and unparsing
+            if Formatter.unparse document = source then
+                // The actuall format process
+                printf "%s" (Formatter.formatDocument document)
+                0
+
+            // Report error if the unparsing result does match the original source
+            else
+                failwith (
+                    "The formater does not work properly. "
+                    + "Please let us know by filing a new issue in https://github.com/microsoft/qsharp-compiler/issues/new/choose."
+                )
         | Error errors ->
             errors |> List.iter (eprintfn "%O")
             1
+
     with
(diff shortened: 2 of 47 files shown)


src/QsFmt/App/Program.fs
```diff
@@ -29,13 +29,24 @@ let main args =
...
         let source = if input = "-" then stdin.ReadToEnd() else File.ReadAllText input
 
-        match Formatter.format source with
-        | Ok result ->
-            printf "%s" result
-            0
+        match Formatter.parse source with
+        | Ok document ->
+            // Test whether there is data loss during parsing and unparsing
+            if Formatter.unparse document = source then
```
samarsha, line 35: It might be simpler to move this check inside the `Formatter.format` function. I think that you wouldn't need to expose new functions or types in that case.

samarsha (commented):
> I have to change the `Formats standard input` test after implementing the above feature. Because I am working on Windows, the unparsing result ends with `\r\n`, which is different from the '\n' ending in the test case. This difference makes the test fail.

Oh, I see. I guess that is because the input string uses `\n` but the expected output string uses an actual newline in the source code. I think your change to use the same type of newline in both strings makes sense.

> I am not sure whether not being able to handle mixed line endings is a bug. In most cases, it will not exist in users' code. However, I tried similar things to test F#'s formatter. It can still format the file without reporting errors. Maybe we should omit the line-ending differences in the sanity test?

Hmm, I think that if there are mixed newlines in the original source code, then parsing and unparsing should preserve the original newlines. So we should still be able to compare the strings for an exact match. Is that not what is happening?

src/QsFmt/App/Program.fs
yjt98765: Agree. I have rolled back most changes in my last commit, i.e., I have added back `internal` labels and removed new exposing functions. Now, the check is within `Formatter.format` function.

yjt98765:
> > I am not sure whether not being able to handle mixed line endings is a bug. In most cases, it will not exist in users' code. However, I tried similar things to test F#'s formatter. It can still format the file without reporting errors. Maybe we should omit the line-ending differences in the sanity test?
> 
> Hmm, I think that if there are mixed newlines in the original source code, then parsing and unparsing should preserve the original newlines. So we should still be able to compare the strings for an exact match. Is that not what is happening?

No. In the [original code](https://github.com/microsoft/qsharp-compiler/pull/1084/files#diff-d25afa6d6770df0d965521635ea3fbb306ae5c6faaedd4d359ba950ee6f7de8fL16), all newlines are printed as `Environment.NewLine` instead of original ones. However, I have changed it so that the original newlines can be preserved now.

The "downside" of this change is that the [test case](https://github.com/microsoft/qsharp-compiler/pull/1084/commits/75c0b780fdd904ac23038699bb328d4793b794dc#diff-0b2979608a6ed36fc432c25f756a899b5fe480ac324b322ae79e1f21f463f7e8L133-L145) I added last time is invalid now. I cannot think of a code example to trigger the bug [reporting part](https://github.com/microsoft/qsharp-compiler/pull/1084/commits/75c0b780fdd904ac23038699bb328d4793b794dc#diff-87778067becd07ddecda46d7e5a1a3fb1ef7bb32095b43f07f514e57d4e4a359L41-L45). Even if we find one, it means there is a bug in our code, and thus it should be resolved quickly. How can we test the sanity test? Currently, all the test cases pass.

src/QsFmt/App/Program.fs
```diff
@@ -46,3 +47,6 @@ let main args =
     | :? UnauthorizedAccessException as ex ->
         eprintfn "%s" ex.Message
         4
+    | Failure (msg) ->
+        eprintfn "%s" msg
```
samarsha, line 51: This hides the stack trace which may be useful for debugging (and to include in a bug report). I would be fine with leaving this as an unhandled exception, since the default behavior is to print the message, stack trace, and (on Windows) show a message box saying the program stopped working.

src/QsFmt/Formatter/Formatter.fs
```diff
@@ -33,14 +33,24 @@ let parse (source: string) =
...
+            |> curry collapsedSpaces.Document ()
+            |> curry operatorSpacing.Document ()
+            |> curry newLines.Document ()
+            |> curry indentation.Document 0
+            |> printer.Document
+        // Report error if the unparsing result does match the original source
+        else
+            failwith (
+                "The formater does not work properly. "
+                + "Please let us know by filing a new issue in https://github.com/microsoft/qsharp-compiler/issues/new/choose."
```
samarsha, line 50: It would be nice if the error message was a little more specific - for example it could say that the formatter's syntax tree is not consistent with the input source code. Also, including the unparsed document in the message could help with debugging.

src/QsFmt/Formatter/SyntaxTree/Node.fsi
```diff
@@ -37,6 +37,9 @@ module internal Trivia =
     /// </summary>
     val newLine: Trivia
```
samarsha, line 38: It might be good to update the documentation for `newLine` to clarify that it's a new line node containing the default new line character for the current platform.

samarsha (commented):
> No. In the [original code](https://github.com/microsoft/qsharp-compiler/pull/1084/files#diff-d25afa6d6770df0d965521635ea3fbb306ae5c6faaedd4d359ba950ee6f7de8fL16), all newlines are printed as `Environment.NewLine` instead of original ones. However, I have changed it so that the original newlines can be preserved now. See the newly edited [test case](https://github.com/microsoft/qsharp-compiler/pull/1084/files#diff-0b2979608a6ed36fc432c25f756a899b5fe480ac324b322ae79e1f21f463f7e8L93).

You're right, I forgot about that, thanks for fixing it. :)

> I cannot think of a code example to trigger the bug [reporting part](https://github.com/microsoft/qsharp-compiler/pull/1084/commits/75c0b780fdd904ac23038699bb328d4793b794dc#diff-87778067becd07ddecda46d7e5a1a3fb1ef7bb32095b43f07f514e57d4e4a359L41-L45). Even if we find one, it means there is a bug in our code, and thus it should be resolved quickly. How can we test the sanity test? Currently, all the test cases pass.

Yes, I think I'm OK with not having a test for this error message. By design there should be no input that triggers the error, and it doesn't seem like a good idea to add a way to trigger the error manually.

src/QsFmt/App.Tests/Tests.fs
```diff
@@ -90,13 +90,11 @@ let ``Formats file`` path output =
     )
 
 [<Theory>]
-[<InlineData("namespace Foo { function Bar() : Int { return 0; } }\n",
+[<InlineData("namespace Foo { function Bar() : Int { return 0; }\n }\r\n",
```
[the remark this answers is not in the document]
yjt98765: Sure. I have added the [new test](https://github.com/microsoft/qsharp-compiler/pull/1084/files#diff-27cec60c6dde2dea728bb9a23831ee402cd9639782130ba7112f37acc56bf1e4R46-R50) in FixedPoints.fs.

src/QsFmt/Formatter/Printer.fs
```diff
@@ -5,15 +5,14 @@ module Microsoft.Quantum.QsFmt.Formatter.Printer
...
-open System
 
 /// <summary>
 /// Prints a <see cref="Trivia"/> node to a string.
 /// </summary>
 let printTrivia =
     function
     | Whitespace ws -> ws
-    | NewLine -> Environment.NewLine
+    | NewLine nl -> nl
```
ScottCarda-MS, line 15: I think should keep this as Environment.NewLine to keep this flexible with different local setups. My local setup, for example is a mix of different new line characters, and having this be a specific new line, one way or the other, breaks most of my tests. My recommendation is to leave this as Environment.NewLine and instead use a function like the following on inputs before string-comparison:
```
/// <summary>
/// Replaces New Line characters in source string for Environment.NewLine characters.
/// </summary>
let private standardizeNewLines (source: string) =
    source.Replace("\r", "").Replace("\n", Environment.NewLine)
```

samarsha (approved):
Looks great, thanks!

PR #1085, opened 2021-07-20
Setting LSP document version in Language Server

ricardo-espinoza:
Each time that a document is change in an editor, the Language Server Protocol includes a document version number as part of its `textNotification/didChange` updates. This document version can be specified as part of WorkspaceEdit objects when setting up code actions.

Up to Visual Studio 16.9, it was optional by the LSP protocol to include a matching version, and the extension defaulted to a version of 1 in all cases. However, starting from 16.10, the protocol expects the version of the document to match the latest one provided by the LSP, otherwise code actions will be ignored.

This change keeps track of the last received version from the LSP and uses it when setting up code actions.

### DIFF
--- a/src/QsCompiler/CompilationManager/FileContentManager.cs
+++ b/src/QsCompiler/CompilationManager/FileContentManager.cs
@@ -32,6 +32,11 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
 
         public string FileName { get; }
 
+        /// <summary>
+        /// Indicates the current version of the document according to the Language Server Protocol.
+        /// </summary>
+        public int? Version { get; set; }
+
         private readonly ManagedList<CodeLine> content;
         private readonly ManagedList<ImmutableArray<CodeFragment>> tokens;
         private readonly FileHeader header;
@@ -113,6 +118,7 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
             this.SyncRoot = new ReaderWriterLockSlim(LockRecursionPolicy.SupportsRecursion);
             this.Uri = uri;
             this.FileName = fileName;
+            this.Version = 1;
 
             this.content = new ManagedList<CodeLine>(this.SyncRoot);
             this.content.Add(CodeLine.Empty()); // each new file per default has one line without text
--- a/src/QsCompiler/CompilationManager/FileContentManager.cs
+++ b/src/QsCompiler/CompilationManager/FileContentManager.cs
@@ -33,7 +33,8 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
         public string FileName { get; }
 
         /// <summary>
-        /// Indicates the current version of the document according to the Language Server Protocol.
+        /// An arbitrary integer representing the current version number of the file, or null if no version number is available.
+        /// The version number may change at any time to any other integer, including a lower number than its current value.
         /// </summary>
         public int? Version { get; set; }
 
--- a/src/QsCompiler/CompilationManager/FileContentManager.cs
+++ b/src/QsCompiler/CompilationManager/FileContentManager.cs
@@ -118,7 +118,7 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
             this.SyncRoot = new ReaderWriterLockSlim(LockRecursionPolicy.SupportsRecursion);
             this.Uri = uri;
             this.FileName = fileName;
-            this.Version = 1;
+            this.Version = null;
 
             this.content = new ManagedList<CodeLine>(this.SyncRoot);
             this.content.Add(CodeLine.Empty()); // each new file per default has one line without text
--- a/src/QsCompiler/CompilationManager/CompilationUnitManager.cs
+++ b/src/QsCompiler/CompilationManager/CompilationUnitManager.cs
@@ -362,6 +362,10 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
 
                 this.changedFiles.Add(docKey);
                 var publish = false;
+
+                // We should keep the file version in sync with the reported by the LSP.
+                file.Version = param.TextDocument.Version;
+
                 foreach (var change in param.ContentChanges)
                 {
                     file.PushChange(change, out publish); // only the last one here is relevant
--- a/src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
+++ b/src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
@@ -30,7 +30,7 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
         /// </summary>
         private static WorkspaceEdit GetWorkspaceEdit(this FileContentManager file, params TextEdit[] edits)
         {
-            var versionedFileId = new VersionedTextDocumentIdentifier { Uri = file.Uri, Version = 1 }; // setting version to null here won't work in VS Code ...
+            var versionedFileId = new VersionedTextDocumentIdentifier { Uri = file.Uri, Version = file.Version };
             return new WorkspaceEdit
             {
                 DocumentChanges = new[] { new TextDocumentEdit { TextDocument = versionedFileId, Edits = edits } },
--- a/src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
+++ b/src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs
@@ -30,7 +30,7 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
         /// </summary>
         private static WorkspaceEdit GetWorkspaceEdit(this FileContentManager file, params TextEdit[] edits)
         {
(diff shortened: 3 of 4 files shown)


cgranade (approved):
Looks good to me, but it may be helpful to get a second opinion given that I'm not as familiar with this part of the language server.

src/QsCompiler/CompilationManager/FileContentManager.cs
```diff
@@ -32,6 +32,11 @@ public class FileContentManager : IDisposable
 
         public string FileName { get; }
 
+        /// <summary>
+        /// Indicates the current version of the document according to the Language Server Protocol.
+        /// </summary>
+        public int? Version { get; set; }
```
samarsha, line 38: I don't think `Version` can ever be `null`, can it?

samarsha, line 38: I think it would be nice to enforce the invariant that the file version is monotonically increasing.

src/QsCompiler/CompilationManager/FileContentManager.cs
```diff
@@ -113,6 +118,7 @@ internal FileContentManager(Uri uri, string fileName)
             this.SyncRoot = new ReaderWriterLockSlim(LockRecursionPolicy.SupportsRecursion);
             this.Uri = uri;
             this.FileName = fileName;
+            this.Version = 1;
```
samarsha, line 121: Instead of initializing this to 1 and setting it later in OpenFileAsync, can the initial version be passed into the constructor?

samarsha (changes requested):
Since the new `Version` property is part of the public API for `FileContentManager`, I'd like to make sure we are clear about its definition. @bettinaheim for API review.

src/QsCompiler/CompilationManager/FileContentManager.cs
ricardo-espinoza: This field is storing a value that comes from `VersionedTextDocumentIdentifier.Version` which is a an `int?` ... I don't expect it to normally be `null`, but I wanted to preserve the type used by the Language Server Protocol. If we want to enforce it to not be null, we'll need to decide a behavior in case we ever receive a null. With this implementation, we're recording the version as reported by the LSP in all cases.

ricardo-espinoza: Should we log an error or fail if that happens? Also, I'm not sure how to enforce it on our side, since we're just recording the value sent in 'didChange' notifications, and in the unlikely situation in which the IDE restarted the numbering, that would still be the version that we should use with the protocol. Assuming that it happens, I don't know if erroring out on our side would be better experience, than attempting to follow the protocol with the new value.

Would it be ok if we simply log this condition for debugging purposes?

src/QsCompiler/CompilationManager/FileContentManager.cs
ricardo-espinoza: There are many instances of usage of FileContentManager where there is no associated version. Since this is part of the public API, I'd rather not change the constructor and just limit the usage of this new field to those cases in which we want to record the property value. For this one I'm a bit on the fence, I'd like to hear @bettinaheim 's opinion as well.

The advantage of keeping it outside the constructor is that the change is scoped to only this particular situation, and we don't need to change existing usage of FileContentManager objects outside of the context of interaction with the LSP.

ricardo-espinoza:
Testing completed for VS and VS Code with those changes.

src/QsCompiler/CompilationManager/FileContentManager.cs
samarsha: Yeah, I saw that the LSP package we're using has it as `int?`, but the [LSP specification says that is it a non-nullable integer](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#versionedTextDocumentIdentifier). If it is null, I think we are free to reject the message since it seems like it is invalid.

samarsha: > in the unlikely situation in which the IDE restarted the numbering, that would still be the version that we should use with the protocol

According to the LSP specification, the client is not allowed to do this. If this happens, then I think it would indicate that the client and server are out of sync, and maybe reloading the file from disk (which would reset the version number) would be best.

src/QsCompiler/CompilationManager/FileContentManager.cs
samarsha: One reason I would like to see it as a constructor parameter is that it makes it easier to enforce monotonicity. From the LSP specification, it doesn't seem like the client is prevented from opening a file starting with a negative version number - but from the FileContentManager's perspective, this would look like the version is going from 1 to a negative number, which is not allowed.

bettinaheim: Given that the server itself does not actually need to do anything with this version and simply stores it temporarily to use it when as part of sending messages back to the client, let's leave it up to the client what it wants to do - independent on what the version is (null, not null, monotonous or not), nothing will negatively impact the server code. Let's keep it nullable, and set/keep the version at null here. We only blindly give back versions we got.

samarsha: > we don't need to change existing usage of FileContentManager objects outside of the context of interaction with the LSP

My perspective is that FileContentManager is not LSP-specific, so it should make sense outside the context of an LSP server. If we are adding a monotonically increasing version number to FileContentManager, then I think we should make sure it's well-defined and integrated with the rest of the class.

In my opinion, that would mean enforcing monotonicity and ensuring that version number is incremented whenever the file contents change. That way, other parts of the code or other programs that use FileContentManager (which may not necessarily know about LSP) can rely on the invariants of the version property.

samarsha: > let's leave it up to the client what it wants to do - independent on what the version is (null, not null, monotonous or not)

I'm not sure why to allow null though - that does not fit the JSON schema provided by the spec? (It seems like the choice of `int?` by the Visual Studio LSP package might be a mistake?)

I believe we currently reject messages that are not `int` or `null`, but if we really want to blindly give back what the client sends us, we should loosen the type to `object` since the client may send us any JSON value...

src/QsCompiler/CompilationManager/FileContentManager.cs
ricardo-espinoza: We should use the same data type that the LSP package, since we don't know if they intend to update the implementation or the specification. But this is a very good point, and something I'll ask them in our next sync.

ricardo-espinoza: It is an interesting discussion to decide how permissive our server should be, given that the history of this bug started with the editors being very relaxed with the versioning protocol and then VS started enforcing it in 16.10. The risk of actively enforcing the version from our side is not minimal, and that would make the change out of the scope of a bug fix. However, we're discussing this outside of this PR, and we can decide what our future plans are.

src/QsCompiler/CompilationManager/FileContentManager.cs
ricardo-espinoza: Per @bettinaheim 's comment, the initialized value in the constructor is now `null`. I can't speak for the VS SDK team on their choice of `int?` for `VersionedTextDocumentIdentifier.Version`, but my guess is that it was a special value for absent or non-initialized versions in some contexts. Let's not open it for now to something as wide as `object`, but per Bettina's suggestion, we'll keep it null until it's set by the LSP on a particular file. See commit [76793c67](https://github.com/microsoft/qsharp-compiler/pull/1085/commits/76793c67478d68ee19bfebe70ad8ce31319bbc41) on this.

src/QsCompiler/CompilationManager/FileContentManager.cs
samarsha: As far as I know, the VS LSP package is not an official implementation of the LSP protocol - it is just the implementation that Visual Studio uses. I would trust the specification in case of any differences, especially since I recall that there have been other problems with nullable types in this package.

bettinaheim (approved):
Let's move forward with this minimal chance as is, so that we fix the issue before contemplating additional improvements.

src/QsCompiler/CompilationManager/FileContentManager.cs
```diff
@@ -32,6 +32,11 @@ public class FileContentManager : IDisposable
 
         public string FileName { get; }
 
+        /// <summary>
+        /// Indicates the current version of the document according to the Language Server Protocol.
```
samarsha, line 36: Since FileContentManager isn't LSP-specific, maybe it would be better not to refer to LSP here - something like this?

```suggestion
        /// An arbitrary integer representing the current version number of the file, or null if no version number is available.
        /// The version number may change at any time to any other integer, including a lower number than its current value.
```

samarsha (approved):
I'm OK with not having any invariants for now, but I'd like to be explicit about that in the documentation.

src/QsCompiler/CompilationManager/FileContentManager.cs
ricardo-espinoza: Sounds good. Applying.

PR #1088, opened 2021-07-21
Fixing an issue with division...

bettinaheim:
What should have been a signed division was accidentally an unsigned division.

### DIFF
--- a/src/QsCompiler/LlvmBindings/Instructions/InstructionBuilder.cs
+++ b/src/QsCompiler/LlvmBindings/Instructions/InstructionBuilder.cs
@@ -8,7 +8,6 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
-using System.Globalization;
 using System.Linq;
 using LLVMSharp.Interop;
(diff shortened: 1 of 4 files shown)


src/QsCompiler/LlvmBindings/Instructions/InstructionBuilder.cs
```diff
@@ -106,7 +105,6 @@ public void PositionAtEnd(BasicBlock basicBlock)
         /// is not required to come from the same block the instruction builder is
         /// currently referencing.</note>
         /// </remarks>
-        [SuppressMessage("Microsoft.Design", "CA1011:ConsiderPassingBaseTypesAsParameters", Justification = "Specific type required by interop call")]
```
cgranade, line 109: Out of curiosity, what motivated removing these attributes?

cgranade (approved):

src/QsCompiler/LlvmBindings/Instructions/InstructionBuilder.cs
bettinaheim: Visual studio, which told me they were unnecessary.

PR #1092, opened 2021-07-22
Feature/llvm pass proposal

troelsfr:
This PR contains three major items:

1. A structure for creating loadable LLVM passes in C++
2. A draft style guide for contributing to QsPasses code
3. Continuous integration and style enforcement scripts

# CMake structure for passes

An important part of this PR is that it proposes a structure for passes: It is suggested that each pass has their own subcode base. The reason for this proposal is that it makes it very easy to add and remove passes as well as decide which passes to link against. Each pass is kept in its own subdirectory under `libs`:

```
libs
├── CMakeLists.txt
└── OpsCounter
    ├── OpsCounter.cpp
    └── OpsCounter.hpp
```
Adding  a new pass is easy using the `manage` tool developed in this PR:
```
% ./manage create-pass HelloWorld
Available templates:

1. Function Pass

Select a template:1
```
which results in a new pass code in the `libs`:
```
libs
├── CMakeLists.txt
├── HelloWorld
│   ├── HelloWorld.cpp
│   ├── HelloWorld.hpp
│   └── SPECIFICATION.md
└── OpsCounter
    ├── OpsCounter.cpp
    └── OpsCounter.hpp
```
A full example of how to create a basic function pass is included in the README.md file for anyone interested.

# Draft style guide

This PR contains a very early stage draft guideline for contributing C++ code to this part of the code. These are included in the file `CONTRIBUTING.md`. This is not meant as a style dictation. Rather everything is up for change and as such it is just a starting point. Criticisms and feedback is welcome.

# Continuous integration

This component is the largest part of this PR. The continuous integration component includes:

1. Style formatting to ensure that everything looks the same. This includes checking that relevant copyrights are in place.
2. Static analysis
3. Unit testing

The automatic style enforcement is configurable with the ability to easily add or remove rules. Currently the source pipelines are defined as:

```python
SOURCE_PIPELINES = [
    {
        "name": "C++ Main",
        "src": path.join(PROJECT_ROOT, "libs"),

        "pipelines": {
            "hpp": [
                require_pragma_once,
                enforce_cpp_license,
                enforce_formatting
            ],
            "cpp": [
                enforce_cpp_license,
                enforce_formatting
            ]
        }
    },
    # ...
]
```
This part defines pipelines for `.hpp` files and `.cpp` files allowing the developer to add such requirements such as having copyright in the op of the source file and ensure that formatting follows that given by `.clang-format`.

Each of these CI stages can executed individually using `./manage` or you can run the entire CI process by invoking `./manage runci`. An example of what this may look like is here:

```zsh
% ./manage runci

2021-07-21 14:38:04,896 - FormatChecker - ERROR - /Users/tfr/Documents/Projects/qsharp-compiler/src/QsPasses/src/OpsCounter/OpsCounter.cpp was not correctly formatted.
2021-07-21 14:38:04,899 - FormatChecker - ERROR - Your code did not pass formatting.

% ./manage stylecheck --fix-issues
% ./manage runci

-- Found LLVM 11.1.0
-- Using LLVMConfig.cmake in: /usr/local/opt/llvm@11/lib/cmake/llvm
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/tfr/Documents/Projects/qsharp-compiler/src/QsPasses/Debug
Consolidate compiler generated dependencies of target QSharpPasses
[ 50%] Building CXX object CMakeFiles/QSharpPasses.dir/src/OpsCounter/OpsCounter.cpp.o
[100%] Linking CXX shared library libQSharpPasses.dylib
ld: warning: directory not found for option '-L/usr/local/opt/llvm/lib'
[100%] Built target QSharpPasses
/Users/tfr/Documents/Projects/qsharp-compiler/src/QsPasses/src/OpsCounter/OpsCounter.cpp:29:7: error: invalid case style for class 'LegacyOpsCounterPass' [readability-identifier-naming,-warnings-as-errors]
class LegacyOpsCounterPass : public FunctionPass
      ^~~~~~~~~~~~~~~~~~~~
      CLegacyOpsCounterPass
113345 warnings generated.
Suppressed 113345 warnings (113344 in non-user code, 1 NOLINT).
Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.
1 warning treated as error
2021-07-21 14:38:40,191 - Linter - ERROR - /Users/tfr/Documents/Projects/qsharp-compiler/src/QsPasses/src/OpsCounter/OpsCounter.cpp failed static analysis

# ISSUES FIXED MANUALLY
% ./manage runci

-- Found LLVM 11.1.0
-- Using LLVMConfig.cmake in: /usr/local/opt/llvm@11/lib/cmake/llvm
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/tfr/Documents/Projects/qsharp-compiler/src/QsPasses/Debug
Consolidate compiler generated dependencies of target QSharpPasses
[ 50%] Building CXX object CMakeFiles/QSharpPasses.dir/src/OpsCounter/OpsCounter.cpp.o
[100%] Linking CXX shared library libQSharpPasses.dylib
ld: warning: directory not found for option '-L/usr/local/opt/llvm/lib'
[100%] Built target QSharpPasses
-- Found LLVM 11.1.0
-- Using LLVMConfig.cmake in: /usr/local/opt/llvm@11/lib/cmake/llvm
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/tfr/Documents/Projects/qsharp-compiler/src/QsPasses/Debug
Consolidate compiler generated dependencies of target QSharpPasses
[100%] Built target QSharpPasses
*********************************
No test configuration file found!
*********************************
```
The key idea here is to make it extremely easy to be complaint with the style guide, correct any issues that might come as a result of static analysis and at the same time enforce this when a PR is made.


bettinaheim:
@troelsfr Let's merge that to a feature branch until all pieces here are fully functional and not longer considered wip. We can/should merge to main frequently, but let's make sure main is always in a clean state without any code that doesn't yet have a clear functionality associated with it.

troelsfr:
> @troelsfr Let's merge that to a feature branch until all pieces here are fully functional and not longer considered wip. We can/should merge to main frequently, but let's make sure main is always in a clean state without any code that doesn't yet have a clear functionality associated with it.

Sounds good.

src/Passes/.clang-tidy
```diff
@@ -0,0 +1,76 @@
...
+  - key:    readability-identifier-naming.EnumConstantCase
+    value:  'CamelCase'
+  - key:    readability-identifier-naming.GlobalConstantPrefix
+    value:  'G_'
+  - key:    readability-identifier-naming.ConstantCase
+    value: 'UPPER_CASE'
+  - key:    readability-identifier-naming.MacroDefinitionCase
+    value:  'UPPER_CASE'
+  - key:    readability-identifier-naming.TypeAliasCase
+    value:  'CamelCase'
```
kuzminrobin, line 68: FYI: I see `readability-identifier-naming.TypeAliasCase` twice. You likely meant once.

src/Passes/.clang-tidy
```diff
@@ -0,0 +1,76 @@
...
+  - key:    readability-identifier-naming.NamespaceCase
+    value:  'lower_case'
+  - key:    readability-identifier-naming.GlobalConstantCase
+    value:  'UPPER_CASE'
+  - key:    readability-identifier-naming.EnumCase
+    value:  'CamelCase'
+  - key:    readability-identifier-naming.EnumConstantCase
+    value:  'CamelCase'
+  - key:    readability-identifier-naming.GlobalConstantPrefix
+    value:  'G_'
```
kuzminrobin, line 62: So far I didn't look at the `.clang-tidy` files carefully and I can say that I have _no experience_ in this area (so my opinion is not strong here).

I'm not sure if the order of directives here matters, but if we can freely change the order then the following comes to my mind.
I would feel comfortable if closely related items were close together, e.g. 
`GlobalConstantCase` and `GlobalConstantPrefix`.
When I write the Global Constant, I write the prefix first, then the name. That is why I would place `GlobalConstantPrefix` first, then `GlobalConstantCase`.

If the same idea is applied to the while code writing stage, I would consider the following order:
```
NamespaceCase

AbstractClassPrefix

ClassCase
StructCase
UnionCase

FunctionCase
// Are member functions covered by `FunctionCase`?
ParameterCase

// Member types, enums, aliases/typedefs, member constants, member variables (including static). Public first, then protected, then private. Prefix first, then name.

// Non-member funcs, their params, local vars (including local statics).

// Global consts and vars (including global statics).

// Macros last (since they are not recommended, see [Item 2: Prefer consts, enums, and inlines to #defines](http://wavelino.coffeecup.com/pdf/Effective%20Cshapshap.pdf))
```

src/Passes/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+include(CheckCXXCompilerFlag)
+
+message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}")
+message(STATUS "Using LLVMConfig.cmake in: ${LLVM_DIR}")
+
+# Setting the standard configuration for the C++ compiler
+# Rather than allowing C++17, we restrict the ourselves to 
+# C++14 as this is the standard currently used by LLVM. While
+# there is a very small change that the difference in standard 
+# would break things, it is a possibility nonetheless.
```
kuzminrobin, line 15: (Just curiosity)
What does it mean: "C++14 .. this is the standard currently used by LLVM"? As I know LLVM deals with IR rather than C++ code (and Clang is the one who deals with C and C++ code). 

Did you mean "C++14 .. is the standard currently used by _Clang_"? If so, then what does that phrase mean again? I thought that [Clang fully supports C++17 since Clang 5](https://clang.llvm.org/cxx_status.html#cxx17), and the stable version of Clang is at least Clang 12 (or maybe even newer).

To summarize I would recommend to consider C++20 and C++17 first, and if there is a strong reason, then C++14. For now I just don't see the reasons for limiting ourselves with C++14.

kuzminrobin, line 15:
```suggestion
# Setting the standard configuration for the C++ compiler.
# Rather than allowing C++17, we restrict ourselves to 
# C++14 as this is the standard currently used by LLVM. While
# there is a very small chance that the difference in standard 
# would break things, it is a possibility nonetheless.
```

src/Passes/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+# are triggered if llvm-tutor is built without this flag (though otherwise it
+# builds fine). For consistency, add it here too.
+check_cxx_compiler_flag("-fvisibility-inlines-hidden" SUPPORTS_FVISIBILITY_INLINES_HIDDEN_FLAG)
+if (${SUPPORTS_FVISIBILITY_INLINES_HIDDEN_FLAG} EQUAL "1")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fvisibility-inlines-hidden")
+endif()
+
+# We export the compile commands which are needed by clang-tidy 
+# to run the static analysis
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
```
kuzminrobin, line 37: FYI: I assume that with "static analysis" you mean the code formatting tooling (clang-tidy and clang-format), but not the topic discussed [here](https://clang-analyzer.llvm.org/).

src/Passes/CONTRIBUTING.md
```diff
@@ -0,0 +1,83 @@
...
+
+Always add copyrights at the top of the file.
+
+```text
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+```
+
+For header files, prefer to put `#prama once` before the copyright.
+
```
kuzminrobin, line 80: Nit: `#prama once` does not relate to copyrights. Consider either removing or placing in a separate section.

src/Passes/CONTRIBUTING.md
```diff
@@ -0,0 +1,83 @@
...
+(...)
+QsPasses/src/OpsCounter/OpsCounter.cpp:39:21: error: missing username/bug in TODO [google-readability-todo,-warnings-as-errors]
+                    // TODO: Fails to load if this is present
+                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+                    // TODO(tfr): Fails to load if this is present
+```
+
+## Always add copyrights
+
+Always add copyrights at the top of the file.
```
kuzminrobin, line 72: (I'm not strong here, more opinions are recommended)
Consider either excluding this section or something like
> Add MS copyrights at the top of the newly-created files and files consisting of fragments from MS-owned sources. For non-MS-owned sources consider adding those sources without changes and preserving their copyrights.

(And I would request specifically for more opinions about the case when a non-MS employee (or even the whole company) adds a new file, what copyright should be used there, if any)

src/Passes/README.md
```diff
@@ -0,0 +1,322 @@
+# Q# Passes for LLVM
+
+This library defines LLVM passes used for analysing, optimising and transforming the IR. The Q# pass library is a dynamic library that can be compiled and ran separately from the
+rest of the project code. While it is not clear whether this possible at the moment, we hope that it will be possible to write passes that enforce the QIR standard.
+
```
kuzminrobin, line 5: Consider making the fragment "LLVM passes" a web link to something like [this](https://llvm.org/docs/WritingAnLLVMPass.html#introduction-what-is-a-pass). Such that the reader immediately understands that the "LLVM passes" is a special term with a strict meaning (and gets the right intro/tutorials before diving into the topic).

src/Passes/README.md
```diff
@@ -0,0 +1,322 @@
+# Q# Passes for LLVM
+
+This library defines LLVM passes used for analysing, optimising and transforming the IR. The Q# pass library is a dynamic library that can be compiled and ran separately from the
+rest of the project code. While it is not clear whether this possible at the moment, we hope that it will be possible to write passes that enforce the QIR standard.
```
kuzminrobin, line 4: Consider making "QIR standard" a link to something like [this](https://github.com/microsoft/qsharp-runtime/tree/main/src/Qir/Runtime) or [this](https://github.com/microsoft/qsharp-language/tree/main/Specifications/QIR).

(I'm not sure that QIR is strong enough to use the word "standard" to it. For me personally the word "standard" is an extremely strong word. Consider replacing with "specification" or similar)

src/Passes/README.md
```diff
@@ -0,0 +1,322 @@
...
+
+## What does LLVM passes do?
```
kuzminrobin, line 6:
```suggestion
## What do LLVM passes do?
```
or
```suggestion
## What does LLVM pass do?
```

src/Passes/README.md
```diff
@@ -0,0 +1,322 @@
...
+call            2
+---------------------------
+```
+
+**Example 3: Code validation**. A third use case is code validation. For example, one could write a pass to check whether bounds are exceeded on static arrays [2].
+Note that this is a non-standard usecase as such analysis is usually made using the AST rather than at the IR level.
+
+**References**
+[1] https://github.com/banach-space/llvm-tutor#analysis-vs-transformation-pass
+[2] https://github.com/victor-fdez/llvm-array-check-pass
```
kuzminrobin, line 146: [FYI](https://github.com/microsoft/qsharp-compiler/pull/1080#discussion_r674313289)

src/Passes/README.md
```diff
@@ -0,0 +1,322 @@
...
+store           1
+call            2
+---------------------------
+```
+
+**Example 3: Code validation**. A third use case is code validation. For example, one could write a pass to check whether bounds are exceeded on static arrays [2].
+Note that this is a non-standard usecase as such analysis is usually made using the AST rather than at the IR level.
+
+**References**
+[1] https://github.com/banach-space/llvm-tutor#analysis-vs-transformation-pass
```
kuzminrobin, line 145: FYI: I see a reference to [2], but I don't see any reference to [1].

src/Passes/README.md
```diff
@@ -0,0 +1,322 @@
...
+
+```sh
+cmake ..
+```
+
+and then make your target
+
+```sh
+make [target]
+```
```
kuzminrobin, line 185: What are possible targets and which is the default?

src/Passes/README.md
```diff
@@ -0,0 +1,322 @@
...
+
+## Running a pass
+
+You can run a pass using `opt` as follows:
+
```
kuzminrobin, line 190: Consider making `opt` a web link to something like [this](https://llvm.org/docs/CommandGuide/opt.html) or [this](https://llvm.org/docs/WritingAnLLVMPass.html#running-a-pass-with-opt) or both or other .

src/Passes/examples/OptimisationUsingOpt/README.md
```diff
@@ -0,0 +1,58 @@
...
+  %0 = tail call %String* @__quantum__rt__int_to_string(i64 0)
+  tail call void @__quantum__rt__message(%String* %0)
+  tail call void @__quantum__rt__string_update_reference_count(%String* %0, i32 -1)
+  ret void
+}
+```
+
+plus a few extra delcarations.
+
+## Applying a pass
```
kuzminrobin, line 58: Did you intend to add some text after this heading?

src/Passes/libs/OpsCounter/OpsCounter.hpp
```diff
@@ -0,0 +1,69 @@
...
+        OpsCounterAnalytics(OpsCounterAnalytics const&) = delete;
+        OpsCounterAnalytics(OpsCounterAnalytics&&)      = default;
+        ~OpsCounterAnalytics()                          = default;
+        /// @}
+
+        /// Operators
+        /// @{
+        OpsCounterAnalytics& operator=(OpsCounterAnalytics const&) = delete;
+        OpsCounterAnalytics& operator=(OpsCounterAnalytics&&) = delete;
+        /// @}
```
kuzminrobin, line 29: (I have no experience with this (`llvm::`) code base, feel free to ignore if my note is not applicable)

It looks somewhat suspicious that for copy functions (copy constructor, copy assignment operator) both are `delete`d,
but for the move functions (move constructor and move assignment operator) one is `default` and the other is `delete`.

I also have an impression that explicitly providing the destructor (`~OpsCounterAnalytics() = default;`) but deleting some other special member functions somewhat violates "the rule of all or nothing".

Just in case some more info:
* [Item 5: Know what functions C++ silently writes and calls](http://wavelino.coffeecup.com/pdf/Effective%20Cshapshap.pdf)
* What is move semantics? See _both_ [here](https://stackoverflow.com/a/3109981/6362941) and [here](https://stackoverflow.com/a/11540204/6362941)
* [Item 17: Understand special member function generation](https://moodle.ufsc.br/pluginfile.php/2377667/mod_resource/content/0/Effective_Modern_C__.pdf)
* [Know What Happens If Rule of Three Is Violated](https://github.com/kuzminrobin/code_review_notes/blob/master/cpp_design_bookmarks.md#know-what-happens-if-rule-of-three-is-violated)

src/Passes/manage
```diff
@@ -0,0 +1,12 @@
...
+import sys
+
+# Adding the site-packages directory to our Python path
+# in order to access the TasksCI module
+ROOT = os.path.dirname(__file__)
+sys.path.insert(0, os.path.join(ROOT, "site-packages"))
+
+# Loading the CLI tool and running it
+from TasksCI.cli import cli  # noqa: E402
+cli()
```
kuzminrobin, line 12: Would be nice to have a comment with a web link here providing the introduction to what is going on here.

src/Passes/requirements.txt
```diff
@@ -0,0 +1 @@
+click==8.0.1
```
kuzminrobin, line 1: Would be nice to have some info/link telling which entity uses this file.

kuzminrobin (approved):
Minor notes, but in general looks good to me.
(I didn't check the Python code, recommend one more opinion about that)
The info in the PR description is nice to be saved in a .md file :-)

src/Passes/.clang-tidy
troelsfr: Yes, this is a mistake. I've remove the second entry.

src/Passes/.clang-tidy
troelsfr: This is a good suggestion. I've updated the `.clang-tidy` file with sections with closely related items.

troelsfr: "Are member functions covered by `FunctionCase`?" <- Yes, they are, but you can specialise for member functions if you like: https://clang.llvm.org/extra/clang-tidy/checks/readability-identifier-naming.html

I've added the link to the source for future reference.

src/Passes/CMakeLists.txt
troelsfr: LLVM is written in C++ and by default they use `C++14` when the compile the library. Therefore, it is advisable to use `C++14` to avoid any linking problems that may arise from using different standards. This is unrelated to what kind of compilers are currently implemented by LLVM. I guess a more accurate formulation would be `C++14 as this is the standard currently used by LLVM project for compilation.`. I've updated the text - let me know if it is still unclear.

troelsfr: 👍

src/Passes/CMakeLists.txt
troelsfr: Clang-tidy is a static analyser more than just a formatter - it also does static analysis and, in fact, it uses the project that you are linking. You can activate different parts of static analyser using flags in the top section of `.clang-tidy`. Here are some details:

https://clang.llvm.org/docs/analyzer/checkers.html

This means that besides checking for correct casing and naming, `clang-tidy` also checks that loops does not use floats `clang-analyzer-security.FloatLoopCounter` for instance.

src/Passes/CONTRIBUTING.md
troelsfr: You are correct: `#pragma once` does not relate directly to the copyright. However, since all functions should have copyrights and all headers should have `#pragma once`, the suggestion is to always format your section like:
```
#pragma once
// Copyright ...
```
rather than
```
// Copyright ...
#pragma once
```
This is somewhat pedantic and I am happy to remove this.  The reason it is here is to add consistency to the files. Let me know what you think

src/Passes/CONTRIBUTING.md
troelsfr: The more important part is that if you _do_ choose to add a copyright it should be at the top of the file (and not in bottom or in the middle of the code). It is about consistency more than it is about MS copyright practices. 

Right now, the scripts are set up to enforce a very strict requirement that all files must have MS copyright notice at the top. We should probably lose that restriction as I didn't really consider the case of non-MS copyrights. That is definitely a case that we also need to deal with. I am happy to remove this section.  Let me know your thoughts.

A separate, but important part of the discussion is what the copyright notice should look like. I understand that we may want to move this part of the code into a collaborative repository which will be built with our partners. In this case, it may be more appropriate to have a copyright statement along the lines of 
```
// Copyright (c) The __PROJECT NAME__ Authors
```
with a `AUTHORS` file at the root of the repository. This removes MS as being the perceived dominant copyright holder to more of a collaboration effort. This practice is for instance used in the `Chromium` project. @bettinaheim what are your thoughts?

A third discussion to be had: Do we allow copy/paste of external code (if appropriate copyright is preserved)? My hunch is that this should only be allowed (if at all) if it is not possible to include it as a library.

Note, I have no strong preference for either of these issues, the only thing that is important for me is that we have some level of consistency across files.

src/Passes/README.md
troelsfr: Done!

src/Passes/README.md
troelsfr: Both good suggestions. Done.

src/Passes/README.md
troelsfr: Fixed.

src/Passes/README.md
troelsfr: Fixed.

src/Passes/README.md
troelsfr: There is no default target. I have added a line explaining what valid targets are.

src/Passes/README.md
troelsfr: Done.

src/Passes/examples/OptimisationUsingOpt/README.md
troelsfr: No, that is a mistake that the heading is there. Thanks

src/Passes/libs/OpsCounter/OpsCounter.hpp
troelsfr: I am not entirely sure that I understand this comment, so if the answer is a bit off, please ask again.

As such there is nothing wrong in deleting the copy constructor and copy assignment while preserving the move constructor and move assignment. This is the typical behaviour of a file or network connection where you do not want to make a copy of the actual connection, but you do want the ability to do the construction in one part of the code and use it in another part.

The specific configuration chosen here is LLVM specific and dictated by the parent class - the code just explicitly declares this behaviour such that it is apparent to any one who uses the code that this is the behaviour of the class. With any other configuration (using delete/default) your compiler will complain that it is not compatible with the parent class. The only thing that imo is a bit odd about the configuration is having a move constructor without a move assignment. This may be due to some optimisation or to enforce a specific behaviour in the LLVM library.

When you refer to the "rule of all or nothing", you mean the rule of three/five? Note that rule-of-X only applies to user-defined constructors/destructors/operations:

> If a class requires a user-defined destructor, a user-defined copy constructor, or a user-defined copy assignment operator, it almost certainly requires all three.

See [rule of three](https://en.cppreference.com/w/cpp/language/rule_of_three)

There is no user-defined constructors/destructors/operations in the the code. The code merely states that we are using the default or prohibiting the use of default (the ones the compiler would have used anyway) constructors/destructor/operations. My personal preference is that all constructors are explicitly declared every time. This makes it easy to see what the class can and cannot do, hence, the explicit declarations. In this way, it makes it easier to read the code, as you immediately know that this class will only default construct and move construct (without having to read the LLVM code base). As such you could just delete all of these and the class would fall back to using its parents constructors/operators/destructor.

Please let me know if this did not answer your concern.

src/Passes/requirements.txt
troelsfr: This file is used to setup the environment. It is used in the [README.md](https://github.com/troelsfr/qsharp-compiler/blob/feature/llvm-pass-proposal/src/Passes/README.md) under the section `CI` to setup the environment to executing the manage command for the CI processes.

src/Passes/manage
troelsfr: This part of the code, imports the function `cli` from the [TasksCI](https://github.com/troelsfr/qsharp-compiler/tree/feature/llvm-pass-proposal/src/Passes/site-packages/TasksCI) which is a custom written Python package to deal with style checking and static analysis using `clang-format`, `clang-tidy` etc. 

I've added a comment on where to look for the details of the function `cli`. Let me know if you would like further changes.

src/Passes/CONTRIBUTING.md
bettinaheim: "Right now, the scripts are set up to enforce a very strict requirement that all files must have MS copyright notice at the top. " That is indeed the requirement we have to stick to as long as the code lives on this repo. A chance in that will be done (only) as part of moving (contributing/donating) the code in the future. Please do not add any instructions on the copyright to the readme beyond maybe stating that copy right headers go at the top, since guidance on the copy right may change and it's our job upon code review to make sure the correct copy right is there.

src/Passes/.clang-tidy
```diff
@@ -0,0 +1,103 @@
...
+  - key:    readability-identifier-naming.GlobalConstantCase
+    value:  'UPPER_CASE'
+  - key:    readability-identifier-naming.GlobalConstantPrefix
+    value:  'G_'
+  - key:    readability-identifier-naming.ConstantCase
+    value: 'UPPER_CASE'
+  - key:    readability-identifier-naming.EnumCase
+    value:  'CamelCase'
+  - key:    readability-identifier-naming.EnumConstantCase
+    value:  'CamelCase'
```
kuzminrobin, line 99: Thank you for providing the [link](https://clang.llvm.org/extra/clang-tidy/checks/readability-identifier-naming.html)!
It encouraged me to take a look at what's in there.

(Below is is an insane "pedanticness", feel free to ignore :-D)
After which I asked myself: "Do `EnumCase` and `EnumConstantCase` talk about the _name of the variable_ of the enumerated type or about the _name of the enumerated type_ itself?" The examples in the link (like `enum pre_foo_post { One, Two, Three }`) told me that it is the _data type itself_ (not the var name). That is why I'd suggest to move the `EnumCase` and `EnumConstantCase` to the `# Alias` section (and probably rename the section to `# Types and Type Aliases` or similar).

src/Passes/CMakeLists.txt
kuzminrobin: Aha... The passes are compiled to dynamic libraries that are dynamically loaded by an executable written in C++14. That is why it is desirable for the dynamic libraries to be also written in C++14. Totally makes sense now. Thank you for clarification. Marking this discussion as resolved.

src/Passes/CMakeLists.txt
kuzminrobin: Thank you for explaining!

src/Passes/CONTRIBUTING.md
kuzminrobin: OK, makes sense.

src/Passes/README.md
kuzminrobin: FYI: Doing that _for the first-most occurrence only_ of the "LLVM passes", was sufficient. ;-)

src/Passes/README.md
kuzminrobin: The fragment `make [target]` tells me that specifying the `target` is optional (we can just run `make`). 
What will happen if we run just `make`? (that's what I mean with "default")
E.g. "If `target` is not specified then ...".

src/Passes/manage
kuzminrobin: OK. I just didn't notice that TasksCI goes together with this PR (I thought it was some commonly known tool). 
Thanks.

src/Passes/requirements.txt
kuzminrobin: Thank you.

src/Passes/libs/OpsCounter/OpsCounter.hpp
kuzminrobin: This fragment 
> The only thing that imo is a bit odd about the configuration is _having a move constructor without a move assignment_. This may be due to some optimisation or to enforce a specific behaviour in the LLVM library.

answers the first half of my note 
(In my note I shouldn't have mentioned the copy constructor and copy assignment, I didn't mean to discuss them, I'm all right that they are deleted. I just mentioned them for comparison with the move constructor and move assignment, one of which is deleted and the other one is not, which looked strange).

As for the second half of my note - the destructor... The topic is rather tricky and I'm not sure I should have mentioned that. Any way I consider this discussion as resolved.

If you are interested then I can try to explain what I was thinking about in the second half.
(All the special member functions are mentioned explicitly, that is why the "rule of all or nothing" (aka "rule of 3", "rule of 5", "rule of 0") should be ok here, I shouldn't have mentioned that rule, sorry, my bad)

As for the fragment `~OpsCounterAnalytics() = default;`. I believe it provides the user-declared destructor and makes a difference from a situation where this fragment is skipped. I played with the examples a bit but failed to create an example that demonstrates the difference. Looks like the required code has been optimized out in my examples.

Sorry for taking your time. Marking as resolved.

src/Passes/CONTRIBUTING.md
kuzminrobin: Marking as resolved.

kuzminrobin:
Reapproving.

src/Passes/.clang-tidy
troelsfr: I am happy to incorporate these changes. I will do this on the new branch I am working if you are happy with that?

Just a side note: Enums are a bit of a funny kind of entity - I see your point and why you'd want them together with Types and Type Aliases. The reason why they are with consts is that anonymous enums, i.e.
```
enum {
  One, 
  Two,
  Three
}
```
gives you enumerated constants which are not typed as a separated class.

troelsfr: I've updated this in #1098.

src/Passes/README.md
troelsfr: It was incorrect what I wrote. The default target is `all` - that is if you just run make, it will make all. I've updated this in #1098 with following line after the code block: 

> The default target is `all`. Other valid targets are the name of the folders in `libs/` found in the passes root.

src/Passes/README.md
```diff
@@ -0,0 +1,327 @@
+# Q# Passes for LLVM
+
+This library defines [LLVM passes](https://llvm.org/docs/Passes.html) used for analysing, optimising and transforming the IR. The Q# pass library is a dynamic library that can be compiled and ran separately from the
+rest of the project code. While it is not clear whether this possible at the moment, we hope that it will be possible to write passes that enforce the [QIR specification](https://github.com/microsoft/qsharp-language/tree/main/Specifications/QIR).
+
```
swernli, line 5: This is a bit of a nitpick, but something I do think we need to start being careful with differentiating: these are QIR Passes for LLVM, not Q# passes. Since they process LLVM that is using QIR conventions for representing quantum runtime and intrinsic capabilities, they don't rely on anything Q# specific, and should be usable on QIR from sources other than the Q# compiler.

swernli:
Maybe I'm just missing it, but I don't see where the new CI elements are added to the existing CI scripts (no .ps1 or .yml changes in the PR). Is that something that is going to be part of a later change?

src/Passes/README.md
bettinaheim: Yes, agreed. Even if we have passes that rely on certain restrictions having been imposed by the source language rather than being enforced across the board by the QIR spec, we should still call them QIR passes and document the restrictions it is relying on (and at some point we will need to build validators).

src/Passes/libs/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+  
+
+  # Linking
+  target_link_libraries(${pass_plugin}
+    "$<$<PLATFORM_ID:Darwin>:-undefined dynamic_lookup>")
+
+endforeach()
+
+
+# add_library(passes SHARED ${ALL_PASSES})
```
idavis, line 46: add newline at end of file

src/Passes/examples/OptimisationUsingOpt/SimpleExample/Makefile
```diff
@@ -0,0 +1,5 @@
+clean:
+	rm -rf bin
+	rm -rf obj
+	rm -rf qir
+
```
idavis, line 5: add newline at end of file

src/Passes/examples/OptimisationUsingOpt/SimpleExample/Makefile
```diff
@@ -0,0 +1,5 @@
+clean:
```
idavis, line 1: why are we only cleaning in make?

src/Passes/examples/ClassicalIrCommandline/Makefile
```diff
@@ -0,0 +1,18 @@
...
+
+
+debug-ng-pass-mac: emit-llvm-bc
+	opt -load-pass-plugin ../../Debug/libQSharpPasses.dylib -debug --passes="operation-counter" -disable-output classical-program.bc
+
+
+
+clean:
+	rm -f classical-program.ll
+	rm -f classical-program.bc
```
idavis, line 18: add newline at end of file

src/Passes/examples/ClassicalIrCommandline/classical-program.c
```diff
@@ -0,0 +1,17 @@
...
+  foo(x + y);
+}
+
+int main()
+{
+  foo(2);
+  bar(3, 2);
+
+  return 0;
+}
```
idavis, line 17: add newline at end of file

src/Passes/site-packages/TasksCI/linting.py
```diff
@@ -0,0 +1,135 @@
...
+    print("")
+    print(config)
+    print("")
+    print("Clang tidy checks:")
+
+    for check in sorted(checks):
+        print(" -", check)
+
+
+def run_clang_tidy(build_dir, filename, fix_issues: bool = False):
```
idavis, line 41: add return type hint

src/Passes/site-packages/TasksCI/linting.py
```diff
@@ -0,0 +1,135 @@
...
+            sys.stderr.write(err)
+
+            logger.error("{} failed static analysis".format(filename))
+            return False
+
+    logger.info("All good!")
+    return True
+
+
+def main_cpp(fix_issues: bool):
```
idavis, line 95: add return type hint

src/Passes/site-packages/TasksCI/settings.py
```diff
@@ -0,0 +1,17 @@
...
+
+from os import path
+import multiprocessing
+
+PROJECT_ROOT = path.abspath(path.dirname(path.dirname(path.dirname(__file__))))
+
+MAX_CONCURRENCY = 7
+
+
+def get_concurrency():
```
idavis, line 12: `get_degree_of_concurrency` ? Not sure how python apis typically handle this

src/Passes/Makefile
```diff
@@ -0,0 +1,7 @@
+nothing:
```
idavis, line 1: The other makefile which only cleans doesn't have this protection

idavis (changes requested):

src/Passes/README.md
troelsfr: I've removed any mentioning of the Q# compiler.

troelsfr:
> Maybe I'm just missing it, but I don't see where the new CI elements are added to the existing CI scripts (no .ps1 or .yml changes in the PR). Is that something that is going to be part of a later change?

It is not integrated yet and this is not a PR into main but rather into `feature/llvm-passes`. To ensure that it will not be an overwhelming PR tens of thousands of line, the passes implementation is split into many smaller PRs.

src/Passes/README.md
```diff
@@ -0,0 +1,327 @@
...
+You can run a pass using [opt](https://llvm.org/docs/CommandGuide/opt.html) as follows:
+
+```sh
+cd examples/ClassicalIrCommandline
+make emit-llvm-bc
+opt -load-pass-plugin ../../{Debug,Release}/libOpsCounter.{dylib,so} --passes="print<operation-counter>" -disable-output classical-program.bc
+```
+
+For a gentle introduction, see examples.
```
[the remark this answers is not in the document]
troelsfr: That's a good point. I've updated the phrase.

src/Passes/docs/index.md
```diff
@@ -0,0 +1,3 @@
+# Q# pass documentation
```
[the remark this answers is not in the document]
troelsfr: Changed.

src/Passes/examples/ClassicalIrCommandline/Makefile
```diff
@@ -0,0 +1,18 @@
+emit-llvm-cpp:
```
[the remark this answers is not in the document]
troelsfr: Yes, that target needs to go as it is not there. Done.

src/Passes/examples/ClassicalIrCommandline/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+```sh
+  clang -O1 -c -emit-llvm classical-program.c -o classical-program.bc
+```
+
+This generates a nice and short IR which makes not too overwhelming to understand what is going on.
+
+## Legacy passes
+
+This part assumes that you have build the QsPasses library.
```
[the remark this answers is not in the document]
troelsfr: Corrected.

src/Passes/examples/ClassicalIrCommandline/README.md
```diff
@@ -0,0 +1,36 @@
...
+a pass to a IR using commandline tools only.
+
+IRs can be represented either by a human readible language or through bytecode. For
+C programs former is generated by
+
+```sh
+  clang -O1 -S -emit-llvm classical-program.c -o classical-program.ll
+```
+
+where as the latter is generated writing:
```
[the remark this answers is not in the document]
troelsfr: Changed.

src/Passes/examples/ClassicalIrCommandline/README.md
```diff
@@ -0,0 +1,36 @@
...
+```
+
+This generates a nice and short IR which makes not too overwhelming to understand what is going on.
+
+## Legacy passes
+
+This part assumes that you have build the QsPasses library.
+
+```sh
+opt -load ../../{Debug,Release}/libQSharpPasses.{dylib,so} -legacy-operation-counter -analyze classical-program.ll
```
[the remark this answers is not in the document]
troelsfr: Thanks, yes, we should be consistent. In other `.md` files I use `% ` to indicate that it is a command being executed (as opposed to output) and I have changed all here with no leading spaces. Out of curiosity, is there a preferred style for terminal input/output?

src/Passes/site-packages/TasksCI/cli.py
```diff
@@ -0,0 +1,238 @@
...
+                options.append(template_name)
+
+                # Printing option
+                pretty_template_name = re.sub(r'(?<!^)(?=[A-Z])', ' ', template_name)
+                print("{}. {}".format(len(options), pretty_template_name))
+
+        # Getting a selection
+        n = 0
+        print("")
+        while n < 1 or n > len(options) + 1:
```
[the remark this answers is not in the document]
troelsfr: I've added the possibility of type 'q' or 'quit' to exit without an error.

src/Passes/site-packages/TasksCI/cli.py
```diff
@@ -0,0 +1,238 @@
...
+    # a catastrophy
+    if fix_issues:
+        if not force:
+            print("""Fixing isssues using Clang Tidy will break your code.
+Make sure that you have committed your changes BEFORE DOING THIS.
+Even so, this feature is experimental and there have been reports of
+clang-tidy modying system libraries - therefore, USE THIS FEATURE AT
+YOUR OWN RISK.
+
+Write 'I understand' to proceed.""")
```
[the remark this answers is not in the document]
troelsfr: This feature is experimental and you will need to reinstall LLVM after running it. If it was a stable feature, I agree it should just ask `y/n`, but at the moment, `clang-tidy` chooses to (that is, is mal-configured) destroy your LLVM headers. The damage is irreversible in the sense that the only solution is to delete you LLVM installation and do a fresh install. Hence, I am reluctant to have a simple yes/no guard as this could give impression that the worst that can happen is that you need to reset your git repo to head and the damage is far beyond what you should possibly expect from a CI script.

If it is a strong preference on your side to that this should just be a yes/no answer, then we can change it.

src/Passes/examples/OptimisationUsingOpt/SimpleExample/Makefile
troelsfr: There is no good reason for that. I've added a target for the QIR generation.

src/Passes/site-packages/TasksCI/settings.py
troelsfr: I do not believe that Python has any standard way dealing with this. Regardless, we do want to have the ability to limit the concurrency as you may end up in the situation that your compilation process is bound by memory more than processing power. Hence a custom function to define the concurrency for the CI tasks. In time, we may expand this different functions for build, test and other tasks.

src/Passes/Makefile
troelsfr: Any Makefile that only cleans should have this protection imo and it was a mistake that some did not. It is corrected now.

Note: The only purpose of the root Makefile is to perform cleans at the moment. Arguably this functionality should be moved to `./manage` which does all of the other build and test related tasks. Likely this Makefile will be deprecated in the future.

swernli:
> It is not integrated yet and this is not a PR into main but rather into feature/llvm-passes. To ensure that it will not be an overwhelming PR tens of thousands of line, the passes implementation is split into many smaller PRs.

Yup, that makes sense and certainly helps with smaller incremental reviews into the feature branch! What I mean is that it is still nonetheless valuable to have the CI updates integrated as early as possible into the feature branch, that way subsequent PRs go through the new CI as part of the PR. Otherwise, the only way for reviewers to know if the code compiles and passes any tests is to download locally and run it. I think it would be fine to merge this PR without including CI, but that should be done before #1098 gets merged.

src/Passes/manage
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env python
```
swernli, line 1: This mechanism for running tests and verifying the code seems very Linux specific. Is development of passes on Windows not supported by this infrastructure yet?

src/Passes/README.md
```diff
@@ -0,0 +1,327 @@
...
+Implement your pass here: bar
+Implement your pass here: main
+```
+
+## CI
+
+Before making a pull request with changes to this library, please ensure that style checks passes, that the code compiles,
+unit test passes and that there are no erros found by the static analyser.
+
+To setup the CI environment, run following commands
```
swernli, line 261: Here too the instructions are Linux specific. If we do not support dev/test on Windows right now, we should explicitly call that out and direct readers to set up WSL or containers to use a Linux development environment (ideally pointing to the VSCode devcontainer at the top of the repo and then updating that to have the necessary prerequisites).

src/Passes/manage
bettinaheim: That's a question to answer with a Windows machine handy. :) If the infrastructure for custom passes works well on Windows, then we should certainly keep the scripts working for it as well. If not, we need to make a call regarding our pass infrastructure.

troelsfr: Given that it is a Python script, you should be able to run this on a Windows machine. This is not too different to how frameworks like `Django` works and those runs on Windows for sure :). That said, the CI tool does rely on `shutil` and `subprocess` which may act different on different OSes. Finally, we will probably need to run CI inside a docker container regardless to ensure that `clang-format` is reproducible across all systems.

I am happy to set up a Windows VM, to test and document Windows behaviour if this is considered high priority.

troelsfr:
> > It is not integrated yet and this is not a PR into main but rather into feature/llvm-passes. To ensure that it will not be an overwhelming PR tens of thousands of line, the passes implementation is split into many smaller PRs.
> 
> Yup, that makes sense and certainly helps with smaller incremental reviews into the feature branch! What I mean is that it is still nonetheless valuable to have the CI updates integrated as early as possible into the feature branch, that way subsequent PRs go through the new CI as part of the PR. Otherwise, the only way for reviewers to know if the code compiles and passes any tests is to download locally and run it. I think it would be fine to merge this PR without including CI, but that should be done before #1098 gets merged.

That's fine. I am happy to include that update as part of #1098 or do another PR with just the CI integration added before #1098.

src/Passes/manage
swernli: We discussed this a bit offline, and need to do a separate PR and discussion around the build. Right now the CI only builds the compiler repo on Windows, which made sense when the repo was managed code only. Now that more native code components are being introduced, we should update the CI to be cross-platform accordingly.

src/Passes/README.md
swernli: Will be addressed in a later PR that is CI focused. No need to update that here now!

src/Passes/examples/ClassicalIrCommandline/README.md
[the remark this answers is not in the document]
idavis: @kuzminrobin @bettinaheim  @swernli  Any thoughts here?

idavis (approved):

src/Passes/examples/ClassicalIrCommandline/README.md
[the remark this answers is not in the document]
swernli: I looks like our other readme files just represent cmdline invocations using code blocks without any special leading characters. For example, the top level readme.md for the repo just has:
```
dotnet build MyProject.csproj
```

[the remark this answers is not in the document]
troelsfr: OK, I will revise accordingly.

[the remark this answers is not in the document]
troelsfr: It is updated now.

PR #1102, opened 2021-07-28
Generate arguments in the correct order

cesarzc:
This change generates arguments for QIR submission in the correct order (the one that the API expects them to be).

### DIFF
--- a/src/QsCompiler/CSharpGeneration/EntryPoint.fs
+++ b/src/QsCompiler/CSharpGeneration/EntryPoint.fs
@@ -200,7 +200,9 @@ let private qirArguments parameters parseResult =
         |> qirArgumentValue param.QSharpType
         |> Option.map (fun value -> ``new`` (ident argumentType) ``(`` [ literal param.Name; value ] ``)``)
 
+    // The parameters sequence has to be reversed here because otherwise the QirSubmission is created with a list of arguments in reverse order.
     parameters
+    |> Seq.rev
     |> Seq.fold (fun state param -> Option.map2 (fun xs x -> x :: xs) state (argument param)) (Some [])
     |> Option.map (fun args -> ident listType <.> (sprintf "Create<%s>" argumentType |> ident, args))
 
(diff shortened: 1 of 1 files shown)


src/QsCompiler/CSharpGeneration/EntryPoint.fs
```diff
@@ -200,7 +200,9 @@ let private qirArguments parameters parseResult =
         |> qirArgumentValue param.QSharpType
         |> Option.map (fun value -> ``new`` (ident argumentType) ``(`` [ literal param.Name; value ] ``)``)
 
+    // The parameters sequence has to be reversed here because otherwise the QirSubmission is created with a list of arguments in reverse order.
     parameters
+    |> Seq.rev
```
samarsha, line 205: Does this mean that the parameters given to `qirArguments` are in reverse order (right-to-left from the original parameter tuple), and this is un-reversing them? Is it possible to fix the initial order of the parameters when the sequence is created? I think that happens here:

https://github.com/microsoft/qsharp-compiler/blob/0ca3a80a502421073f17047ebced196fde60f6bc/src/QsCompiler/CSharpGeneration/EntryPoint.fs#L30-L31

samarsha: Oh, or is `parameters` in the right order initially, but the list cons in the fold on the next line is reversing it? If so, it might help to mention that fold specifically in the comment.

cesarzc: Yes, the parameters given to `qirArguments` is given in reverse order. My main concern regarding changing it when the sequence is created was that it might affect how QSharpSubmission is constructed, but thinking about it a bit more it might not be a problem. I'll try to reverse the order when the sequence is created as you suggest.

cesarzc: @samarsha I just saw your second comment. I don't know why I missed it the first time. Yes, I confirmed that `parameters` is in the right order initially but fold is reversing it. I made the comment more detailed :)

samarsha (approved):

Issue #1101, opened 2021-07-28
Formatter Updates Deprecated Unit Type Syntax


ScottCarda-MS:
The formatter should have a rule for updating the `()` syntax for describing the Unit type to the supported `Unit` syntax. This should not affect the use of the `()` for the Unit value literal, which is still valid.

For example:
```qsharp
operation Foo() : () {
    return ();
}
```
should be updated to
```qsharp
operation Foo() : Unit {
    return ();
}
```

ScottCarda-MS:
Closed with #1103.

PR #1103, opened 2021-07-29
Update Unit Syntax

ScottCarda-MS:
Adds rule to update the unit type syntax to replace the deprecated `()` syntax for the `Unit` syntax. This will not replace the `()` syntax for unit value literals, as that is still valid. Adds a unit test for the new rule. Addresses issue #1101.

Since this branch uses code developed in the unmerged sccarda/UpdateUnitSyntax branch, it will target that branch instead of main to create a better diff. When #1099 is merged in, this will be redirected toward main. It will be in draft until then.


src/QsFmt/Parser/QSharpParser.g4
```diff
@@ -210,7 +210,8 @@ expression
     | value=boolLiteral # BoolExpression
     | value=resultLiteral # ResultExpression
     | value=pauliLiteral # PauliExpression
-    | openParen='(' (items+=expression (commas+=',' items+=expression)* commas+=','?)? closeParen=')' # TupleExpression
+    | openParen='(' (items+=expression (commas+=',' items+=expression)* commas+=','?) closeParen=')' # TupleExpression
+    | openParen='(' closeParen=')' # UnitExpression
```
samarsha, line 214: Why is this needed - isn't `UnitExpression` a special case of `TupleExpression` where `items` is empty?

ScottCarda-MS: Thanks for the question. My thinking here is that I see the Unit type and the Tuple type as being on divergent development paths in Q#, both in our code and in the design of Q#. One thing that stick out for me as evidence of this is that the Unit type is a separate type in Q# from the Tuple type, having its own keyword in the language and its own separate set of handles in the compiler and here in the formatter. To me, separating out the unit value literal expressions as having their own handles, distinct from the Tuple handles, is a reflection of the different directions these two separate types are taking. Having them separate, I think, tells a more consistent story for developers about these two types and will, in a small way, make writing and reading rules and transformation in the code base clearer and cleaner. When developers want to do something special with the unit value literals, we can use the special handle for it, instead of having to have a more complicated Tuple handle with a special case in it for an empty tuple.

But I'd like to know your thoughts on this topic. Do you agree with my thoughts on this? Or do you have a different perspective about the unit and tuples types that you can share?

samarsha: I don't think unit and tuples are "two types" - there are an infinite number of tuple types, and they can be categorized by how many items they contain. Unit is a natural consequence of setting the number of items to 0, and a separate rule isn't needed. For similar reasons we allow `[]` and `""` to naturally occur as the empty cases of arrays and strings without needing separate rules for them, as well.

> When developers want to do something special with the unit value literals, we can use the special handle for it, instead of having to have a more complicated Tuple handle with a special case in it for an empty tuple.

I think overall this increases complexity instead of reducing it. When there isn't a separate case for unit, it's possible to write formatting rules that apply to all tuples, and they will automatically apply to unit as well for free. With a separate case, this is no longer possible - you need to write one handler for unit and a separate one for other tuples. On the other hand, if you do want to target unit specifically, a condition that checks if a tuple is empty is not that complicated to write.

samarsha: Another issue with having two cases is that there's now two ways to represent `()` in the CST:

1. `{ OpenParen = ..., Items = [], CloseParen = ... } : 'a Tuple`
2. `{ OpenParen = ..., CloseParen = ... } : UnitExpression`

That means that if you handle tuples in a tree transformer, you must still handle the empty tuple in addition to the separate unit expression. To avoid that you could change the type of tuples to require the list to be non-empty, but I still think that this would overall add complexity as it would need separate handling from other sequences like arrays.

ScottCarda-MS: > Another issue with having two cases is that there's now two ways to represent `()` in the CST:
> To avoid that you could change the type of tuples to require the list to be non-empty

I believe I addressed that in the code already by updating the parsing for tuples to be non-empty. Was there something more that needed to be done to address this?

samarsha: The parser isn't the only way to create syntax tree nodes - I mean that `Tuple` is still defined with a possibly empty list of items here:

https://github.com/microsoft/qsharp-compiler/blob/6b23852f92c27d0edf66d0dca2b902153a33b366/src/QsFmt/Formatter/SyntaxTree/Node.fsi#L82-L83

To express the property that a `Tuple` must be non-empty, `Items` should be a non-empty list type. (However, I don't think that's the right direction.)

ScottCarda-MS: Is it particularly complicated to enforce a list to be non-empty like that in F#? Is there a `non-empty list` type of some kind we could use here?

samarsha: I don't think it comes with F# but the typical definition is

```fsharp
 type 'a NonEmpty = NonEmpty of 'a * 'a list
```

ScottCarda-MS: I've removed the Unit Expression handles.

swernli (approved):

PR #1104, opened 2021-07-30
Passes light roadmap document

troelsfr:
This pull request contains a document that discusses the pass requirements to implement an adder in Q# and have compiled all the way to OpenQASM using the base [profile described in the QIR specification](https://github.com/microsoft/qsharp-language/blob/ageller/profile/Specifications/QIR/Base-Profile.md).


src/Passes/docs/base-profile-transformations.md
```diff
@@ -0,0 +1,233 @@
...
+This document sets out to motivate and demonstrate feasibility of building such a tool.
+
+## Motivation
+
+It is anticipated that most usages of the QIR specificaiton will need to only use a subset of it. These subsets
+may further be subject to constraints such as how one allocate or aquire a qubit handle. We refer subset with
+constraints as a profile. For instance, it is likely that early versions of quantum hardware will have a limited
+set of classical instructions available. With this in mind, the vendor or user of said hardware would define a profile
+that only contains a specified subset. One example of such a profile, is the [base profile](https://github.com/microsoft/qsharp-language/blob/ageller/profile/Specifications/QIR/Base-Profile.md)
+which only allows function calls and branching, but no arithmetic, classical memory or classical registers.
```
bettinaheim, line 19: I find this sentence a bit confusing, since the profile does allow to store measurement results as classical bits such that it is possible to branch based on them. Additionally, the profile also allows for integer and double literals (constants). Saying there is no classical memory is hence a bit odd to me.

src/Passes/docs/base-profile-transformations.md
```diff
@@ -0,0 +1,233 @@
...
+  ret void
+}
+```
+
+After applying the our [demo profile transformation](https://github.com/troelsfr/qsharp-compiler/tree/feature/profile-pass/src/Passes/examples/QubitAllocationAnalysis), the QIR is reduced to:
+
+```
+define void @Feasibility__QubitMapping__Interop() local_unnamed_addr #0 {
+entry:
+  call void @__quantum__qis__x__body(%Qubit* null)
```
bettinaheim, line 114: Hmm, it is a bit suboptimal that qubit index 0 becomes a null pointer. What happens if the pointer is cast back to int64 again? Does it give you 0L?

src/Passes/docs/base-profile-transformations.md
```diff
@@ -0,0 +1,233 @@
...
+│                      │
+│     QIR Profile      │
+│                      │
+└──────────────────────┘
+```
+
+We propose that such a reduction could be done using the LLVM passes infrastructure to compose a profile
+which would map the QIR to a subset of the available instructions with any required constraints.
+
+## Feasibility Study
```
bettinaheim, line 53: For a general feasibility study, the two things that would worry me most are 1) that information properly propagates across function boundaries, and 2) how callable values (and captures) can be dealt with. Captured arrays might be interesting. In any case, the work as started makes sense - there are just a bunch of hard cases down the line that need some work to deal with.

src/Passes/docs/base-profile-transformations.md
```diff
@@ -0,0 +1,233 @@
...
+  call void @__quantum__qis__x__body(%Qubit* null)
+  call void @__quantum__qis__x__body(%Qubit* nonnull inttoptr (i64 1 to %Qubit*))
+  call void @__quantum__qis__x__body(%Qubit* nonnull inttoptr (i64 2 to %Qubit*))
+  ret void
+}
+```
+
+We note that we successfully have eliminated loops, arithmetic operations, dynamic qubit allocation and alias counting - all operations which are not supported by the base profile.
+
+## Goal
```
bettinaheim, line 123: I like the goal. :)

src/Passes/docs/base-profile-transformations.md
```diff
@@ -0,0 +1,233 @@
...
+
+Every profile is specified through a YAML file which defines an object at the top-level. This object must contain the fields `name` and `displayName`:
+
+```yaml
+name: profile-name
+displayName: Profile Name
+# ...
+```
+
+Additionally, top level also contains the fields `version` and `mode`. The version refers to the QIR version which forms the basis for the specification and `mode` explains how the profile is defined.
```
bettinaheim, line 153: What is your thinking regarding versioning; would you expect that the same library handles all supported versions, or would you imaging that we release a library (and updates) per version? Would people download the correct version or would you imaging the tool to pull the right version (if not already on disk) depending on what is needed?

src/Passes/docs/base-profile-transformations.md
```diff
@@ -0,0 +1,233 @@
...
+    - __quantum__qis__ry__body
+    - __quantum__qis__rz__body
+    - __quantum__qis__s__body
+    - __quantum__qis__s_adj
+    - __quantum__qis__t__body
+    - __quantum__qis__t__adj
+    - __quantum__qis__x__body
+    - __quantum__qis__y__body
+    - __quantum__qis__z__body
+  instructions:
```
bettinaheim, line 183: What about specifying which classical values/types are supported? Within the base profile, int and double values are supported, but not anything else. How would one specify e.g. the availability of fixed point numbers with a certain precision (only)?

src/Passes/docs/base-profile-transformations.md
```diff
@@ -0,0 +1,233 @@
...
+specification:
+  functions:
+    - __quantum__rt__array_update_alias_count
+    - __quantum__rt__array_update_reference_count
+  instructions:
+    - br
+# ...
+```
+
+This profile specifies a system that does not allow reference and alias counting and neither have support for branching, but otherwise has the full QIR vesion 1.0 available.
```
bettinaheim, line 207: I am not sure that any combination of limitations actually will lead to a well-defined profile - I believe there are essentially dependencies/implications between certain limitations. I can also imaging that we have certain features be exclusive (you can have either feature A or feature B but never both at the same time). What is your thinking here? Are these features and limitations in any way validated?

src/Passes/docs/base-profile-transformations.md
```diff
@@ -0,0 +1,233 @@
...
+  - passName: useStaticQubitAllocation
+  - passName: eliminateClassicalMemoryUsage
+  - passName: ignoreCall
+    config:
+      names:
+        - __quantum__rt__array_update_alias_count
+        - __quantum__rt__array_update_reference_count
+```
+
+For those passes which are defined specifically for QIR we we allow configuration to be passed to them. This will allow the end-user to fine-tune the behaviour of profile generator.
```
bettinaheim, line 229: I like the inclusion of a config from the get-go.

src/Passes/docs/base-profile-transformations.md
```diff
@@ -0,0 +1,233 @@
...
+    - __quantum__rt__array_update_alias_count
+    - __quantum__rt__array_update_reference_count
+  instructions:
+    - br
+# ...
+```
+
+This profile specifies a system that does not allow reference and alias counting and neither have support for branching, but otherwise has the full QIR vesion 1.0 available.
+
+## Generation specification
```
bettinaheim, line 209: It seems to me that there is some partial redundancy between the profile specification and the generation. I take it that the profile specification does not actually impact at all which passes are run? Is it correct that the profile specification is only used for validation where as the generation specification is used only for generation?

src/Passes/docs/base-profile-transformations.md
```diff
@@ -0,0 +1,233 @@
...
+
+The default `mode` of specification is by `feature` which means that specification describes the feature set available. Alternatively, one can specify a profile by `limitation`. As an example profile for
+
+```yaml
+name: profile-name
+displayName: Profile Name
+version: 1.0
+mode: feature
+specification:
+  functions:
```
bettinaheim, line 166: Why does the profile specification need to know about the list of acceptable qis functions - purely for validation purposes? The signatures are unspecified here; is the idea here that the validation will only work if the profile contains only the recognized functions? Alternatively, is it possible to list any name here, but the signature is not validated in that case? Is the signature validated for recognized functions, and if so, is there a list of recognized functions?

src/Passes/docs/base-profile-transformations.md
```diff
@@ -0,0 +1,233 @@
...
+
+This document sets out to motivate and demonstrate feasibility of building such a tool.
+
+## Motivation
+
+It is anticipated that most usages of the QIR specificaiton will need to only use a subset of it. These subsets
+may further be subject to constraints such as how one allocate or aquire a qubit handle. We refer subset with
+constraints as a profile. For instance, it is likely that early versions of quantum hardware will have a limited
+set of classical instructions available. With this in mind, the vendor or user of said hardware would define a profile
+that only contains a specified subset. One example of such a profile, is the [base profile](https://github.com/microsoft/qsharp-language/blob/ageller/profile/Specifications/QIR/Base-Profile.md)
```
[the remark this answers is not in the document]
troelsfr: Good point. I think the link would only break if you delete the feature branch which there seems be a tradition not to do.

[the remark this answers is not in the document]
bettinaheim: The branch is not even a feature branch - at this time it is really just a private branch and very much work in progress.

src/Passes/docs/base-profile-transformations.md
troelsfr: I few points here:
1. It does not allow you to store or even look at measurement results as far as I can see. There is exactly one function to perform measurements which allows you to flush the result(with an assigned id) but no means of retrieving the result. See the function signature of `__quantum__qis__mz__body`. You would therefore not be able to branch on any measurement result. I would argue that there is a classical sink.
2. Constants are neither registers nor memory. Yes, literals are allowed
3. Given the classical instruction set `call`, `br` and `ret`, but no means of retrieving, storing or in any way manipulating data. Consequently, only sensible function you can have will be `Unit` functions

src/Passes/docs/base-profile-transformations.md
troelsfr: The thought is that at some point the spec will need to mature to cover what the industry will need for the next 3 - 5 years. Once that happens we are at version 1.0 and freeze all specs such that they will not have to build tool chains around work-in-progress. We can then look ahead for what is need 5 - 10 years down the road and start on version 2.0 if we believe that is needed. You can then still make minor changes with minor version increases such as 1.1

I believe this will be needed as building the full hardware setup is a slow and expensive process and will require some level of stability.

src/Passes/docs/base-profile-transformations.md
troelsfr: No, I realised this after discovering that QIR does not specify all the functions, so this will need to be rethought. It can still be helpful to have spec by exclusion (even if i don't really like this method of defining what something is. But for instance, you could "include all integer operations, but exclude shift operations" rather than specifying all operations. 

We could have mutually exclusive features. I have not given it much thought, so I will think a bit about it. Do you have a good example?

src/Passes/docs/base-profile-transformations.md
troelsfr: That is the thought. Specification is essentially an analysis whereas generation is a transformation.  Specification has a predetermined set of passes it will run and the specification will tell these passes what is allowed and not allowed. This is mostly due to that I could not think of any case where you needed to truly compose passes to define the profile. 

Generation, on the other hand, is a far more complex and require certain level of flexibility and configurability since it transforms (generates) the IR.

src/Passes/docs/base-profile-transformations.md
troelsfr: This current proposal was with an understanding that the QIR specification would hold an extensive list of intrinsic classical and/or quantum functions with their corresponding signature. However, without this being true, we may indeed  need the function signatures as well. I will update with a proposal for this today.

src/Passes/docs/base-profile-transformations.md
troelsfr: I think this is an interesting question that possibly requires a wider discussion.

src/Passes/docs/base-profile-transformations.md
```diff
@@ -0,0 +1,233 @@
...
+displayName: Profile Name
+# ...
+```
+
+Additionally, top level also contains the fields `version` and `mode`. The version refers to the QIR version which forms the basis for the specification and `mode` explains how the profile is defined.
+The final two top level fields are lists named `specification` and `generation`. These contains the specification and generation producedure, respectively.
+
+## Specification
+
+The default `mode` of specification is by `feature` which means that specification describes the feature set available. Alternatively, one can specify a profile by `limitation`. As an example profile for
```
alan-geller, line 158: I can imagine profiles where you have a mix of `feature` and `limitation`. For instance, I could imagine specifying the list of quantum instructions by `feature` but wanting to rule out specific LLVM instructions by `limitation`. Might it be more flexible to get rid of `mode` and instead have top-level `feature` and `limitation` tags?

src/Passes/docs/base-profile-transformations.md
```diff
@@ -0,0 +1,233 @@
...
+specification:
+  functions:
+    - __quantum__rt__array_update_alias_count
+    - __quantum__rt__array_update_reference_count
+  instructions:
+    - br
+# ...
+```
+
+This profile specifies a system that does not allow reference and alias counting and neither have support for branching, but otherwise has the full QIR vesion 1.0 available.
```
alan-geller, line 207: On a slightly different note, it might be useful to have well-known names for certain features. Alias counting and reference counting both feel like good examples for this, since there may be subtle differences between code that uses counting and code that doesn't, especially for alias counting. Otherwise the tool has to recognize "magic" runtime function names and infer from their removal that the entire supporting feature is removed.

Strings or bigints are other good examples of features that really come and go as a unit. For these, you probably can't do anything to adapt the code, but at least you can verify it. It would be nice to have a shorthand way to remove features like these without listing all of the related runtime functions (which seems error-prone) and without designating one of the functions as a magic keyword that disables the entire feature.

alan-geller (approved):
A few minor wording tweaks and a couple of suggestions. I like the idea a lot!

src/Passes/docs/base-profile-transformations.md
troelsfr: You will need to expand on this as it is not entirely clear what you mean here with regards to 1. Could you give an example?

For callables, it is unclear whether you mean lambdas or function pointers. Regardless, if the implementation is a first-class LLVM implementation, it should not be difficult to do as you would just inline lambdas the same way you would do with any other function call. The same for captures: These would follow well defined structures that can be parsed. If on the other hand, callables are implemented with opaque objects `Callable`  with custom functions for invoking say `__invoke_callable(%Callable %fnc, %Tuple %args)` it becomes a lot harder since the implementational details depends on the runtime and you cannot provide many guarantees on the correctness.

src/Passes/docs/base-profile-transformations.md
troelsfr: Yes. `null` is LLVMs doing and I do not think there is much to do about it, but I have not looked.

src/Passes/docs/base-profile-transformations.md
troelsfr: Yes, I agree with this. I was thinking exactly the same. I was thinking that this could be a later extension where we would have `mixed` and that we would start with `feature` and `limitation`. It is clear though that in many case it will be beneficial to for instance say "No floating point operations, all integer operation except for division and multiplication" instead either specifying all support or all not supported.

src/Passes/docs/base-profile-transformations.md
troelsfr: Yes, it would definitely be preferable if there are rules specifying how naming should happen. I also think that reference counting is a good example, but also more complex things as "what does an array load pattern look like in the IR?". If you should have any possibility of transforming this:
```
  %0 = call i8* @__quantum__rt__array_get_element_ptr_1d(%Array* %leftPreshared, i64 0)
  %1 = bitcast i8* %0 to %Qubit**
  %2 = load %Qubit*, %Qubit** %1, align 8
 ```
into 
```
%0 = inttoptr i64 0 to %Qubit*
```
you do need most of the pattern structure to work out the that this is regarding a qubit type at index 0 in the array `leftPreshared`.

troelsfr:
I've merge this document and we can continue the discussion in the MVP implementation branch PR #1110

PR #1106, opened 2021-08-02
Adding Python to the CI process

troelsfr:


build/test.ps1
```diff
@@ -38,6 +38,8 @@ function Test-One {
     }
 }
 
+python
```
swernli, line 41: This looks like it might be leftover from something?

src/Passes/site-packages/TasksCI/toolchain.py
```diff
@@ -2,31 +2,113 @@
...
+
+
+def discover_cmake_impl(ext: str) -> str:
     """
     Finds the cmake executable
     """
-    return shutil.which("cmake")
+    ret = shutil.which("cmake{}".format(ext))
 
+    if ret is not None:
```
swernli, line 81: This pattern seems a little odd... if `ret` is not `None` them we return it, but if it is `None` then we explicitly return `None`. Could this just be `return ret`?

build/passes-linux.yml
```diff
@@ -0,0 +1,14 @@
+
+steps:
+- script: |
```
swernli, line 3: It would be good to give this a friendly name so that it is labeled in the build details page. Right now it just shows up with the generic name "CmdLine" (see https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=30149&view=logs&j=743bc1cd-037d-5150-fc69-2953b556a708&t=ac832828-8db5-5d4a-de7c-12645fac0708&l=1)

swernli (changes requested):
It looks like the `manage runci` invocation is not actually finding the tests to run. From [the logs](https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=30149&view=logs&j=743bc1cd-037d-5150-fc69-2953b556a708&t=ac832828-8db5-5d4a-de7c-12645fac0708&l=160):
```
*********************************
No test configuration file found!
*********************************
Usage

  ctest [options]
```
This might have to do with the current working directory when ctest is invoked, as ctest needs to be run while the working directory is the cmake build output folder.

src/Passes/site-packages/TasksCI/toolchain.py
troelsfr: This is simply to follow the pattern of the previous function. The function attempts to find the command and if it succeeds, it returns. Most likely though, as we (eventually) get the Windows up running, we may need to add more tests. The significance of reaching the end of the function is that we have failed at finding `cmake` and we explicitly return `None` to make it clear that `None` is used as a error signal (see for instance the clang-tidy search). I am happy to simplify the function if you prefer that or add a placeholder comment say something like `# Test more locations`

build/test.ps1
troelsfr: Yes, this is clearly a leftover from testing if Python was available on Windows.
```suggestion
```

build/passes-linux.yml
troelsfr:
```suggestion
- name: Linux Passes Build & Tests
   script: |
```

src/Passes/site-packages/TasksCI/toolchain.py
```diff
@@ -2,31 +2,113 @@
...
+        return ret
 
-def discover_ctest() -> str:
+    return None
```
troelsfr, line 84:
```suggestion
    # Test more locations here
    
    # Finally returns None to indicate failed attempt at finding CMake
    return None
```

troelsfr:
> It looks like the `manage runci` invocation is not actually finding the tests to run. From [the logs](https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=30149&view=logs&j=743bc1cd-037d-5150-fc69-2953b556a708&t=ac832828-8db5-5d4a-de7c-12645fac0708&l=160):
> 
> ```
> *********************************
> No test configuration file found!
> *********************************
> Usage
> 
>   ctest [options]
> ```
> 
> This might have to do with the current working directory when ctest is invoked, as ctest needs to be run while the working directory is the cmake build output folder.
> 
> EDIT: Looking at the other PRs, it may be that this behavior is expected for now. If so, my apologies! I'll update the review accordingly.

Indeed there are no tests yet and this is the reason for the output. That is, everything is working as expected.

src/Passes/site-packages/TasksCI/toolchain.py
swernli: Ah, I see. That makes sense, thanks for explaining!

swernli (approved):

build/passes-linux.yml
```diff
@@ -0,0 +1,14 @@
+
+steps:
+- name: Linux Passes Build & Tests
```
troelsfr, line 3:
```suggestion
```

build/passes-linux.yml
```diff
@@ -0,0 +1,13 @@
+
+steps:
+  script: |
```
troelsfr, line 3:
```suggestion
  displayName: Linux build and CI for passes
  script: |
```

build/passes-linux.yml
```diff
@@ -0,0 +1,14 @@
+
+steps:
+  displayName: Linux build and CI for passes
```
troelsfr, line 3:
```suggestion
-  displayName: "Linux build and CI for passes"
```

build/passes-linux.yml
```diff
@@ -0,0 +1,14 @@
+
+steps:
+-  displayName: "Linux build and CI for passes"
```
troelsfr, line 3:
```suggestion
- displayName: "Linux build and CI for passes"
```

build/passes-linux.yml
```diff
@@ -0,0 +1,14 @@
+
+steps:
+- displayName: "Linux build and CI for passes"
```
bettinaheim, line 3:
```suggestion
```

build/passes-linux.yml
bettinaheim:
```suggestion
- script: |
```

build/passes-linux.yml
```diff
@@ -0,0 +1,14 @@
...
+      sudo apt update -y
+      sudo apt install -y clang-11 cmake clang-format-11 clang-tidy-11 
+      sudo apt-get install -y llvm-11 lldb-11 llvm-11-dev libllvm11 llvm-11-runtime
+      export CC=clang-11
+      export CXX=clang++
+      cd src/Passes/
+
+      pip install -r requirements.txt
+      chmod +x manage
+      ./manage runci
```
bettinaheim, line 14:
```suggestion
      ./manage runci
  displayName: Linux build and CI for passes      
```

PR #1110, opened 2021-08-04
Profile adoption tool

troelsfr:
# Profile adoption tool

Note PR is WiP.

## Goal
The overall goal of this PR is to create a first version of a tool that allows transformation of a general QIR to a base profile.

## Quick start

Once the project is built (see next sections), you can generate a new QIR as follows:

```sh
./Source/Apps/qat --generate --profile baseProfile -S ../examples/QubitAllocationAnalysis/analysis-example.ll
```

Likewise, you can validate that a QIR follows a specification by running:

```sh
./Source/Apps/qat --validate --profile baseProfile -S ../examples/QubitAllocationAnalysis/analysis-example.ll
```

## Example

In this example, we start with a QIR generated by the Q# frontend. Rather than giving the full 3445 lines of QIR, we instead give the frontend code:

```qsharp
namespace TeleportChain {
    open Microsoft.Quantum.Canon;
    open Microsoft.Quantum.Measurement;
    open Microsoft.Quantum.Preparation;

    operation PrepareEntangledPair(left : Qubit, right : Qubit) : Unit is Adj + Ctl {
        H(left);
        CNOT(left, right);
    }

    operation ApplyCorrection(src : Qubit, intermediary : Qubit, dest : Qubit) : Unit {
        if (MResetZ(src) == One) { Z(dest); }
        if (MResetZ(intermediary) == One) { X(dest); }
    }

    operation TeleportQubitUsingPresharedEntanglement(src : Qubit, intermediary : Qubit, dest : Qubit) : Unit {
        Adjoint PrepareEntangledPair(src, intermediary);
        ApplyCorrection(src, intermediary, dest);
    }

    operation TeleportQubit(src : Qubit, dest : Qubit) : Unit {
        use intermediary = Qubit();
        PrepareEntangledPair(intermediary, dest);
        TeleportQubitUsingPresharedEntanglement(src, intermediary, dest);
    }

    operation DemonstrateEntanglementSwapping() : (Result, Result) {
        use (reference, src, intermediary, dest) = (Qubit(), Qubit(), Qubit(), Qubit());
        PrepareEntangledPair(reference, src);
        TeleportQubit(src, dest);
        return (MResetZ(reference), MResetZ(dest));
    }

    @EntryPoint()
    operation DemonstrateTeleportationUsingPresharedEntanglement() : Unit {
        let nPairs = 2;
        use (leftMessage, rightMessage, leftPreshared, rightPreshared) = (Qubit(), Qubit(), Qubit[nPairs], Qubit[nPairs]);
        PrepareEntangledPair(leftMessage, rightMessage);
        for i in 0..nPairs-1 {
            PrepareEntangledPair(leftPreshared[i], rightPreshared[i]);
        }

        TeleportQubitUsingPresharedEntanglement(rightMessage, leftPreshared[0], rightPreshared[0]);
        for i in 1..nPairs-1 {
            TeleportQubitUsingPresharedEntanglement(rightPreshared[i-1], leftPreshared[i], rightPreshared[i]);
        }

        let _ = MResetZ(leftMessage);
        let _ =  MResetZ(rightPreshared[nPairs-1]);
    }
}
```

Once compiled and the initial QIR is generated and save in the file `analysis-example.ll`, we execute the command

```
./Source/Apps/qat --generate --profile baseProfile ./analysis-example.ll
```

The QAT tool will now attempt to map the QIR in `analysis-example.ll` into a QIR which is compatible with the base format. Removing type and function declarations, the correspoding code reads:

```
; ModuleID = './analysis-example.ll'
source_filename = "./analysis-example.ll"

define internal fastcc void @TeleportChain__DemonstrateTeleportationUsingPresharedEntanglement__body() unnamed_addr {
entry:
  call void @__quantum__qis__h(%Qubit* null)
  call void @__quantum__qis__cnot(%Qubit* null, %Qubit* nonnull inttoptr (i64 1 to %Qubit*))
  call void @__quantum__qis__h(%Qubit* null)
  call void @__quantum__qis__cnot(%Qubit* null, %Qubit* nonnull inttoptr (i64 2 to %Qubit*))
  call void @__quantum__qis__h(%Qubit* nonnull inttoptr (i64 1 to %Qubit*))
  call void @__quantum__qis__cnot(%Qubit* nonnull inttoptr (i64 1 to %Qubit*), %Qubit* nonnull inttoptr (i64 3 to %Qubit*))
  call void @__quantum__qis__cnot(%Qubit* nonnull inttoptr (i64 1 to %Qubit*), %Qubit* null)
  call void @__quantum__qis__h(%Qubit* nonnull inttoptr (i64 1 to %Qubit*))
  %0 = call i1 @__quantum__qir__read_result(%Result* nonnull inttoptr (i64 4 to %Result*))
  call void @__quantum__qis__mz__body(%Qubit* nonnull inttoptr (i64 1 to %Qubit*), %Result* nonnull inttoptr (i64 4 to %Result*))
  call void @__quantum__qis__reset__body(%Qubit* nonnull inttoptr (i64 1 to %Qubit*))
  br i1 %0, label %then0__1.i.i, label %continue__1.i.i

then0__1.i.i:                                     ; preds = %entry
  call void @__quantum__qis__z(%Qubit* nonnull inttoptr (i64 2 to %Qubit*))
  br label %continue__1.i.i

continue__1.i.i:                                  ; preds = %then0__1.i.i, %entry
  %1 = call i1 @__quantum__qir__read_result(%Result* nonnull inttoptr (i64 5 to %Result*))
  call void @__quantum__qis__mz__body(%Qubit* null, %Result* nonnull inttoptr (i64 5 to %Result*))
  call void @__quantum__qis__reset__body(%Qubit* null)
  br i1 %1, label %then0__2.i.i, label %TeleportChain__TeleportQubitUsingPresharedEntanglement__body.1.exit

then0__2.i.i:                                     ; preds = %continue__1.i.i
  call void @__quantum__qis__x(%Qubit* nonnull inttoptr (i64 2 to %Qubit*))
  br label %TeleportChain__TeleportQubitUsingPresharedEntanglement__body.1.exit

TeleportChain__TeleportQubitUsingPresharedEntanglement__body.1.exit: ; preds = %continue__1.i.i, %then0__2.i.i
  call void @__quantum__qis__cnot(%Qubit* nonnull inttoptr (i64 2 to %Qubit*), %Qubit* nonnull inttoptr (i64 1 to %Qubit*))
  call void @__quantum__qis__h(%Qubit* nonnull inttoptr (i64 2 to %Qubit*))
  %2 = call i1 @__quantum__qir__read_result(%Result* nonnull inttoptr (i64 6 to %Result*))
  call void @__quantum__qis__mz__body(%Qubit* nonnull inttoptr (i64 2 to %Qubit*), %Result* nonnull inttoptr (i64 6 to %Result*))
  call void @__quantum__qis__reset__body(%Qubit* nonnull inttoptr (i64 2 to %Qubit*))
  br i1 %2, label %then0__1.i.i2, label %continue__1.i.i3

then0__1.i.i2:                                    ; preds = %TeleportChain__TeleportQubitUsingPresharedEntanglement__body.1.exit
  call void @__quantum__qis__z(%Qubit* nonnull inttoptr (i64 3 to %Qubit*))
  br label %continue__1.i.i3

continue__1.i.i3:                                 ; preds = %then0__1.i.i2, %TeleportChain__TeleportQubitUsingPresharedEntanglement__body.1.exit
  %3 = call i1 @__quantum__qir__read_result(%Result* nonnull inttoptr (i64 7 to %Result*))
  call void @__quantum__qis__mz__body(%Qubit* nonnull inttoptr (i64 1 to %Qubit*), %Result* nonnull inttoptr (i64 7 to %Result*))
  call void @__quantum__qis__reset__body(%Qubit* nonnull inttoptr (i64 1 to %Qubit*))
  br i1 %3, label %then0__2.i.i4, label %TeleportChain__TeleportQubitUsingPresharedEntanglement__body.2.exit

then0__2.i.i4:                                    ; preds = %continue__1.i.i3
  call void @__quantum__qis__x(%Qubit* nonnull inttoptr (i64 3 to %Qubit*))
  br label %TeleportChain__TeleportQubitUsingPresharedEntanglement__body.2.exit

TeleportChain__TeleportQubitUsingPresharedEntanglement__body.2.exit: ; preds = %continue__1.i.i3, %then0__2.i.i4
  call void @__quantum__qis__mz__body(%Qubit* null, %Result* null)
  call void @__quantum__qis__reset__body(%Qubit* null)
  call void @__quantum__qis__mz__body(%Qubit* nonnull inttoptr (i64 3 to %Qubit*), %Result* nonnull inttoptr (i64 1 to %Result*))
  call void @__quantum__qis__reset__body(%Qubit* nonnull inttoptr (i64 3 to %Qubit*))
  ret void
}
```

We note the absence of loops, and that quantum registers are "allocated" at compile time meaning that each qubit instance is assigned a unique ID. As some code may be dead and optimised away, the qubit allocation is not garantueed to be sequential at this point in time. Future work will include writing a qubit ID remapper which will allow qubits.

We also note that the function `TeleportChain__TeleportQubitUsingPresharedEntanglement__body` was cloned twice. This is due to the allocation of qubits and the function being called twice. At present, the analyser does not take qubit release into account and just assumes that it will never be released due to the complicated nature for dealing with nested functions at compile time.

Current TODOs include getting LLVM to remove dead code, do better constant folding and function inlining. Once this is performed correctly, next steps is the remapper and finally a better analysis on what call paths potentially create problems in terms of qubit allocation.

## Dependencies

This library is written in C++ and depends on:

- LLVM

Additional development dependencies include:

- CMake
- clang-format
- clang-tidy

## Building the passes

To build the passes, create a new build directory and switch to that directory:

```sh
mkdir Debug
cd Debug/
```

To build the library, first configure CMake from the build directory

```sh
cmake ..
```

and then make your target

```sh
make [target]
```

The default target is `all`. Other valid targets are the name of the folders in `libs/` found in the passes root.

# Profile adoption tool

## Building QAT

First

```sh
cd Debug
make qat
```

then

```sh
./Source/Apps/qat
```

## Implementing a profile pass

As an example of how one can implement a new profile pass, we here show the implementational details of our example pass which allows mapping the teleportation code to the base profile:

```c++
        pb.registerPipelineParsingCallback([](StringRef name, FunctionPassManager &fpm,
                                              ArrayRef<PassBuilder::PipelineElement> /*unused*/) {
          // Base profile
          if (name == "restrict-qir<base-profile>")
          {
            RuleSet rule_set;

            // Defining the mapping
            auto factory = RuleFactory(rule_set);

            factory.useStaticQuantumArrayAllocation();
            factory.useStaticQuantumAllocation();
            factory.useStaticResultAllocation();

            factory.optimiseBranchQuatumOne();
            //  factory.optimiseBranchQuatumZero();

            factory.disableReferenceCounting();
            factory.disableAliasCounting();
            factory.disableStringSupport();

            fpm.addPass(TransformationRulePass(std::move(rule_set)));
            return true;
          }

          return false;
        });
      }};
```

Transformations of the IR will happen on the basis of what rules are added to the rule set. The purpose of the factory is to make easy to add rules that serve a single purpose as well as making a basis for making rules unit testable.

## Implementing new rules

Implementing new rules consists of two steps: Defining a pattern that one wish to replace and implementing the corresponding replacement logic. Inside a factory member function, this look as follows:

```c++
  auto get_element =
      Call("__quantum__rt__array_get_element_ptr_1d", "arrayName"_cap = _, "index"_cap = _);
  auto cast_pattern = BitCast("getElement"_cap = get_element);
  auto load_pattern = Load("cast"_cap = cast_pattern);

  addRule({std::move(load_pattern), access_replacer});
```

where `addRule` adds the rule to the current rule set.

### Capturing patterns

The pattern defined in this snippet matches IR like:

```c++
  %0 = call i8* @__quantum__rt__array_get_element_ptr_1d(%Array* %leftPreshared, i64 0)
  %1 = bitcast i8* %0 to %Qubit**
  %2 = load %Qubit*, %Qubit** %1, align 8
```

In the above rule, the first and a second argument of `__quantum__rt__array_get_element_ptr_1d` is captured as `arrayName` and `index`, respectively. Likewise, the bitcast instruction is captured as `cast`. Each of these captures will be available inside the replacement function `access_replacer`.

### Implementing replacement logic

After a positive match is found, the lead instruction alongside a IRBuilder, a capture table and a replacement table is passed to the replacement function. Here is an example on how one can access the captured variables to perform a transformation of the IR:

```c++
  auto access_replacer = [qubit_alloc_manager](Builder &builder, Value *val, Captures &cap,
                                               Replacements &replacements) {
    // ...
    auto cst = llvm::dyn_cast<llvm::ConstantInt>(cap["index"]);
    // ...
    auto llvm_size = cst->getValue();
    auto offset    = qubit_alloc_manager->getOffset(cap["arrayName"]->getName().str());

    auto idx = llvm::APInt(llvm_size.getBitWidth(), llvm_size.getZExtValue() + offset);
    auto new_index = llvm::ConstantInt::get(builder.getContext(), idx);
    auto instr = new llvm::IntToPtrInst(new_index, ptr_type);
    instr->takeName(val);

    // Replacing the lead instruction with a the new instruction
    replacements.push_back({llvm::dyn_cast<Instruction>(val), instr});

    // Deleting the getelement and cast operations
    replacements.push_back({llvm::dyn_cast<Instruction>(cap["getElement"]), nullptr});
    replacements.push_back({llvm::dyn_cast<Instruction>(cap["cast"]), nullptr});

    return true;
  };
```


swernli:
@troelsfr I'm about halfway through the review, and I should be able to finish it before Monday morning your time so you have the chance to respond and we can get this ready to merge by EOD!

src/Passes/Source/Rules/Factory.cpp
```diff
@@ -0,0 +1,327 @@
...
+
+                // Replacing the instruction with new instruction
+                replacements.push_back({llvm::dyn_cast<Instruction>(val), instr});
+
+                return true;
+            };
+        addRule({call("__quantum__rt__qubit_allocate"), allocation_replacer});
+
+        // Removing release calls
+        removeFunctionCall("__quantum__rt__qubit_release");
```
swernli, line 164: In the QIR qubit array allocation replacement, the `allocation_replacer` includes a call to the `qubit_alloc_manager->allocate();` method and a corresponding call to `qubit_alloc_manager->release()` in the lambda for removing the QIR qubit array release instructions. Here the `allocation_replacer` still calls allocate on the qubit manager but the removal here of the release instruction doesn't call `qubit_alloc_manager->release()`. Is this because the array allocation is done with a name? The asymmetry seems odd, so I want to make sure it's intentional and not a subtle bug.

src/Passes/Source/Rules/Notation/Notation.cpp
```diff
@@ -0,0 +1,37 @@
...
+    {
+
+        std::function<bool(
+            ReplacementRule::Builder&,
+            ReplacementRule::Value*,
+            ReplacementRule::Captures&,
+            ReplacementRule::Replacements&)>
+        deleteInstruction()
+        {
+            return [](ReplacementRule::Builder&, ReplacementRule::Value* val, ReplacementRule::Captures&,
```
swernli, line 28: For some of these rules it might be helpful to have a comment with an example of the pattern it is meant to recognize. This on in particular is a little trickier to understand at first glance.

src/Passes/Source/Rules/OperandPrototype.cpp
```diff
@@ -0,0 +1,93 @@
...
+            {
+                auto v = user->getOperand(static_cast<uint32_t>(i));
+                if (!children_[i]->match(v, captures))
+                {
+                    return false;
+                }
+                ++i;
+            }
+
+            //    llvm::errs() << "SUCCESS MATCH: " << *value << "\n";
```
swernli, line 38: This debug line can be removed.

src/Passes/examples/QubitAllocationAnalysis/ConstSizeArray/ConstSizeArray.qs
```diff
@@ -1,30 +1,40 @@
-namespace Example {
-    @EntryPoint()
-    operation Main() : Int
-    {
+namespace TeleportChain {
```
swernli, line 1: I think it is worth renaming this file now to avoid confusion, since it is no longer just about array handling but rather a full example algorithm. Then in the PR diff it should show up like a removal of the old file and an add of the new file.

swernli (commented):
The changes mostly look good, just a few questions before I'm ready to sign off! There are also some suggestions about minor updates (file renames or comments) that you can either address here or in a follow up PR that we work on Monday to get the feature branch ready for sharing.

src/Passes/README.md
```diff
@@ -1,156 +1,159 @@
...
-
-- [1] https://github.com/banach-space/llvm-tutor#analysis-vs-transformation-pass
-- [2] https://github.com/victor-fdez/llvm-array-check-pass
-
-## Out-of-source Pass
-
-This library is build as set of out-of-source-passes. All this means is that we will not be downloading the LLVM repository and modifying this repository directly. You can read more [here](https://llvm.org/docs/CMake.html#cmake-out-of-source-pass).
-
-# Getting started
+Current TODOs include getting LLVM to remove dead code, do better constant folding and function inlining. Once this is performed correctly, next steps is the remapper and finally a better analysis on what call paths potentially create problems in terms of qubit allocation.
```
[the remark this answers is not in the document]
troelsfr: Yes, the documentation is outdated here.

src/Passes/Source/Passes/ResourceRemapper/ResourceRemapper.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+namespace microsoft
+{
+namespace quantum
+{
+    llvm::PreservedAnalyses ResourceRemapperPass::run(llvm::Function& function, llvm::FunctionAnalysisManager& /*fam*/)
+    {
+        // Pass body
+
+        llvm::errs() << "Implement your pass here: " << function.getName() << "\n";
```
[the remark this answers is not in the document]
troelsfr: Yes, this is the beginning of the resource remapper pass. I will delete it for now as it is just a placeholder.

src/Passes/Source/Rules/Factory.cpp
troelsfr: Yes, this is because of the name. The allocation manager will throw if the same name is allocated twice within the same scope and hence you need to "release" it. I agree that the asymmetry is odd and, in fact, the cleaner thing to do would be to also call release (which would not do anything at this point).

src/Passes/Source/Rules/OperandPrototype.cpp
troelsfr:
```suggestion
```

src/Passes/examples/QubitAllocationAnalysis/ConstSizeArray/ConstSizeArray.qs
troelsfr: Yes, I agree.

src/Passes/Source/Rules/Notation/Notation.cpp
troelsfr: I've added a comment explaining the purpose of this function: It is not a rule, but a shorthand notation the produces the logic to delete and instruction. That is to say, it is a replacer function factory that does not contain any patterns. An example use is following: ```c++
addRule({callByNameOnly(name), deleteInstruction()});
```

File: src/Passes/Source/Passes/ExpandStaticAllocation/ExpandStaticAllocation.hpp [hunk 7]
```diff
@@ -2,9 +2,9 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-#include "Llvm.hpp"
+#include "Passes/QubitAllocationAnalysis/QubitAllocationAnalysis.hpp"
 
-#include "QubitAllocationAnalysis/QubitAllocationAnalysis.hpp"
+#include "Llvm/Llvm.hpp"
```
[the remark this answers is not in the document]
troelsfr: Yes, order sort-of-matters. When you include in C++ it is the same as just copy/pasting the function into the place where you do the include. This means that if I have a header file `foo.hpp` that depends on iostream and a main file that includes the header and iostream:
```c++
#include<iostream>

#include "foo.h"
int main() { /* ... */ }
```
but I forgot to include `iostream` in `foo.hpp` itself, then everything will compile just fine, but `foo.hpp` will not compile on it's own. Hence if I make another program where I include `foo.hpp` and `iostream` afterwards
```c++
#include "foo.h"
#include<iostream>

int main() { /* ... */ }
```
everything will break. Therefore, the style guide (clang-format) is setup to order includes according to the latter as this will immediately reveal if you have forgotten to include a dependency in `foo.hpp`. As such, you could do without this requirement, but as you scale up your code base these types of missing headers will end up causing you a great deal of headaches.

I am not sure whether it is worthwhile to put a comment about the ordering in this specific file as clang format will automatically rearrange everything according to its liking. And probably, for this particular instance, it would not matter at all - it is just clang-format that insists.

src/Passes/Source/Passes/QubitAllocationAnalysis/QubitAllocationAnalysis.cpp
```diff
@@ -0,0 +1,89 @@
...
+                {
+                    return {true};
+                }
+
+                if (name == "__quantum__qis__m__body")
+                {
+                    return {true};
+                }
+
+                if (name == "__quantum__qis__z__body")
```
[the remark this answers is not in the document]
troelsfr: `__quantum__qis__z__body` "allocates" a result type.

src/Passes/Source/Passes/QubitAllocationAnalysis/QubitAllocationAnalysis.cpp
```diff
@@ -0,0 +1,89 @@
...
+                {
+                    return {true};
+                }
+
+                if (name == "__quantum__rt__qubit_allocate_array")
+                {
+                    return {true};
+                }
+
+                if (name == "__quantum__qis__m__body")
```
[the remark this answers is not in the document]
troelsfr: Yes, I agree that that is a better name. It started out just scanning for qubit allocations until I realised that I also needed to capture result allocations.

src/Passes/Source/Passes/QubitAllocationAnalysis/QubitAllocationAnalysis.cpp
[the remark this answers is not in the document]
troelsfr: This is clearly a mistake.

src/Passes/Source/Passes/QubitAllocationAnalysis/QubitAllocationAnalysis.cpp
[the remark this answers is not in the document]
troelsfr: I've renamed it to `QirAllocationAnalysis` as this does not suggest which types are meant to be captured, but just that it relates to the Qir. I.e. it is not just qubits and it is not all opaque types (although it might be).

src/Passes/Source/Passes/ExpandStaticAllocation/LibExpandStaticAllocation.cpp
```diff
@@ -1,9 +1,9 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-#include "Llvm.hpp"
+#include "Passes/ExpandStaticAllocation/ExpandStaticAllocation.hpp"
 
-#include "ExpandStaticAllocation/ExpandStaticAllocation.hpp"
+#include "Llvm/Llvm.hpp"
```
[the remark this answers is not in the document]
troelsfr: See answer above. In short, the ordering is dictated by `clang-format`.

src/Passes/Source/Commandline/ParameterParser.hpp
```diff
@@ -0,0 +1,81 @@
...
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+
+namespace microsoft
+{
+namespace quantum
+{
+
+    class ParameterParser
```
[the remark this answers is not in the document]
troelsfr: Great, I will keep that in mind - per request of @bettinaheim I am keeping dependencies to an absolute minimum at this stage, but at some point we should review what dependencies are worth including and which not.

src/Passes/README.md
```diff
@@ -325,3 +440,49 @@ If you forget to instantiate this variable in your corresponding `.cpp` file,
...
+
+## Loading IR
+
+https://stackoverflow.com/questions/22239801/how-to-load-llvm-bitcode-file-from-an-ifstream/22241953
+
+## Load passes LLVM passes
+
+https://llvm.org/docs/tutorial/MyFirstLanguageFrontend/LangImpl04.html
+
+## Load custom passes
```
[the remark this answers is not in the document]
troelsfr: Actually, all of these should go as these are just my personal notes for things to look into.

src/Passes/Source/Commandline/ParameterParser.cpp
```diff
@@ -0,0 +1,100 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#include "Commandline/ParameterParser.hpp"
```
bettinaheim, line 4: Feel free to use the same command line package we are already using here: https://github.com/microsoft/qsharp-runtime/tree/main/src/Qir/Tools/Externals/CLI11. Command line parsing is one thing where taking a dependency makes sense, since the work for nice handling including help functions etc accummulates.

src/Passes/Source/Commandline/ParameterParser.hpp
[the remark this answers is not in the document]
bettinaheim: Ah, I see Stefan has already commented on that. :)

src/Passes/Source/Commandline/Settings.hpp
```diff
@@ -0,0 +1,68 @@
+#pragma once
```
bettinaheim, line 1: I like the use of pragma once when possible.

src/Passes/Source/Llvm/Llvm.hpp
```diff
@@ -41,11 +41,22 @@
...
 #include "llvm/IR/Verifier.h"
 
+// Reader tool
+#include "llvm/IRReader/IRReader.h"
+#include "llvm/Support/SourceMgr.h"
+#include "llvm/Transforms/IPO/Inliner.h"
+#include "llvm/Transforms/Scalar/LoopUnrollPass.h"
+
+// Profiles
+#include "llvm/IR/LegacyPassManager.h"
```
bettinaheim, line 55: Are the passes currently written using the new pass manager, the legacy one, or both?

src/Passes/Source/Passes/QubitAllocationAnalysis/QubitAllocationAnalysis.hpp
```diff
@@ -15,29 +15,13 @@ namespace quantum
     class QubitAllocationAnalysisAnalytics : public llvm::AnalysisInfoMixin<QubitAllocationAnalysisAnalytics>
```
bettinaheim: A comment here on the class regarding what it is/does would be good to have. 
In general, the policy (enforced upon code review for PRs to master) is that the public API should be fully documented - in the future, I'd expect we will ultimately automatically generate the API docs for C++ code.
The same goes for other files/classes.

src/Passes/Source/Passes/TransformationRule/LibTransformationRule.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include "Rules/Factory.hpp"
+
+#include "Llvm/Llvm.hpp"
+
+#include <fstream>
+#include <iostream>
+
+namespace
+{
+llvm::PassPluginLibraryInfo getTransformationRulePluginInfo()
```
bettinaheim, line 14: I seem to be missing something here; neither the folder structure nor the naming within this file gives me any indication that this is specific to a particular profile, even though the source code seems to implement a particular set of rules that would apply for specific profile configurations only. Is this code expected to change in the future and if so, how?

src/Passes/Source/Passes/TransformationRule/LibTransformationRule.cpp
```diff
@@ -0,0 +1,56 @@
...
+{
+    using namespace microsoft::quantum;
+    using namespace llvm;
+
+    return {LLVM_PLUGIN_API_VERSION, "TransformationRule", LLVM_VERSION_STRING, [](PassBuilder& pb) {
+                // Registering the pass
+                pb.registerPipelineParsingCallback(
+                    [](StringRef name, FunctionPassManager& fpm, ArrayRef<PassBuilder::PipelineElement> /*unused*/) {
+                        // Base profile
```
bettinaheim, line 23: Could you elaborate a bit more regarding what is at this time configurable, and what is still hardcoded? Is any kind of profile definition as yml file supported at this time, or is this PR containing only the hardcoded handling of the base profile specifically?

src/Passes/Source/Passes/TransformationRule/TransformationRule.cpp
```diff
@@ -0,0 +1,95 @@
...
+                {
+                    instr1->replaceAllUsesWith(llvm::UndefValue::get(instr1->getType()));
+                }
+
+                // And finally we delete the instruction
+                instr1->eraseFromParent();
+            }
+        }
+
+        /*
```
bettinaheim, line 72: In general, there shouldn't be any commented out code blocks. 
If having a detailed log while the tool runs, then it would be good to think about setting up a proper log level that allows to optionally enable a verbose output.

src/Passes/Source/Passes/TransformationRule/TransformationRule.hpp
```diff
@@ -0,0 +1,52 @@
...
+#include "Llvm/Llvm.hpp"
+
+#include <vector>
+
+namespace microsoft
+{
+namespace quantum
+{
+
+    class TransformationRulePass : public llvm::PassInfoMixin<TransformationRulePass>
```
bettinaheim, line 16: A comment would be helpful.

src/Passes/Source/Rules/Notation/Call.ipp
```diff
@@ -0,0 +1,41 @@
+#pragma once
```
bettinaheim, line 1: The file name should be Call.cpp.

src/Passes/Source/Rules/Notation/Notation.hpp
```diff
@@ -0,0 +1,65 @@
...
+    namespace notation
+    {
+
+        using IOperandPrototypePtr = std::shared_ptr<IOperandPrototype>;
+
+        class Capture
+        {
+          public:
+            explicit Capture(std::string const& name);
+            // Note that this operator is delibrately unconventional
```
bettinaheim, line 28: Ok - it would be nice to have a bit more insight the deliberation to the comment; what was the reason for that?

src/Passes/Source/Rules/OperandPrototype.cpp
```diff
@@ -0,0 +1,92 @@
...
+                }
+                ++i;
+            }
+
+
+            return true;
+        }
+
+        //  llvm::errs() << "SUCCESS MATCH: " << *value << " " << user->getNumOperands() << "\n";
+        // TODO(tfr): Check other possibilities for value
```
bettinaheim, line 42: It would be good to track todo's in the code with tasks in ADO (usually GitHub, but given that this code base is still very much in flux, just a quick note in ADO makes more sense).
Same comment as above regarding control prints and commented out code.

src/Passes/Source/Rules/OperandPrototype.hpp
```diff
@@ -0,0 +1,87 @@
...
+        /// @}
+
+        /// Shared functionality
+        /// @{
+
+        /// Adds a child to be matched against the matchees children. Children
+        /// are matched in order and by size.
+        void addChild(Child const& child);
+
+        /// Flags that this operand should be captured. This function ensures
```
bettinaheim, line 46: The comment alone here is not sufficient to understand what the purpose is of this flag; what operands can and should be captured? What happens with captured operands? A see also reference to the piece of code that uses this information might help to clarify as well.

src/Passes/Source/Rules/Operands/Instruction.cpp
```diff
@@ -0,0 +1,50 @@
...
+        auto ret = std::make_shared<InstructionPattern<T>>();
+        ret->copyPropertiesFrom(*this);
+        return std::move(ret);
+    }
+
+// TODO(tfr): This seems to be a bug in LLVM. Template instantiations in
+// a single translation unit is not supposed to reinstantiate across other
+// translation units.
+//
+// However, it is suspecious that htis problem has been around since Clang 8.
```
bettinaheim, line 35: The links to the bug reports are very helpful.

src/Passes/Source/Rules/ReplacementRule.hpp
```diff
@@ -0,0 +1,68 @@
...
+#include <unordered_map>
+#include <vector>
+
+namespace microsoft
+{
+namespace quantum
+{
+
+    /// Rule that describes a pattern and how to make a replacement of the matched values.
+    /// The class contians a OprandPrototype which is used to test whether an LLVM IR value
```
bettinaheim, line 20:
```suggestion
    /// The class contains a OperandPrototype which is used to test whether an LLVM IR value
```

src/Passes/Source/Passes/ExpandStaticAllocation/ExpandStaticAllocation.hpp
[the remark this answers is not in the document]
swernli: Makes sense. Since it was clang-format that rearranged them I'm less worried.

As an aside, I've seen corner cases where ordering matters in C++ includes such that changing the order of headers doesn't cause compilation to fail but does change behavior in a way that introduces a subtle bug. That's actually a technique that is used to frightening effect in some [obfuscated C contests](https://en.wikipedia.org/wiki/International_Obfuscated_C_Code_Contest).

swernli (approved):
Thanks for addressing my questions! We should be ready to merge this once the builds pass.

src/Passes/examples/ClassicalIrCommandline/README.md
```diff
@@ -19,18 +19,12 @@ whereas the latter is generated by executing:
 
 This generates a nice and short IR which makes not too overwhelming to understand what is going on.
 
-## Legacy passes
```
bettinaheim, line 22: Is legacy passes here referring to the pass manager? If so, are all passes now currently written only for the new pass manager? I saw an include for the legacy pass manager further above I believe, so the answer I assume is no.

src/Passes/libs/QubitAllocationAnalysis/SPECIFICATION.md
```diff
@@ -1,9 +0,0 @@
-# Qubit Allocation Analysis
```
bettinaheim, line 1: It might be worth moving this file to the new location rather than deleting it - alternatively this is short enough that it might make more sense as a doc comment on the code. 

Regarding the renamed .ll files (I can't comment on those files in GitHub), what is the reason that they are kept but renamed to .deprecated? Are they used for any tests?

src/Passes/Source/Passes/QirAllocationAnalysis/QirAllocationAnalysis.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+#include <unordered_map>
+#include <vector>
+
+namespace microsoft
+{
+namespace quantum
+{
+
+    class QirAllocationAnalysisAnalytics : public llvm::AnalysisInfoMixin<QirAllocationAnalysisAnalytics>
```
bettinaheim, line 15: nit: The naming here seems a bit redundant; maybe QirAllocationAnalytics is sufficiently descriptive?

src/Passes/Source/Profiles/BaseProfile.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+        // auto &cgpm = inliner_pass.getPM();
+        // cgpm.addPass(llvm::ADCEPass());
+
+        // CGPM.addPass(createCGSCCToFunctionPassAdaptor(createFunctionToLoopPassAdaptor(LoopFooPass())));
+        // CGPM.addPass(createCGSCCToFunctionPassAdaptor(FunctionFooPass()));
+
+        ret.addPass(createModuleToFunctionPassAdaptor(std::move(function_pass_manager)));
+
+        // TODO(tfr): Not available in 11
```
bettinaheim, line 73: The comment here is not very clear; what is the purpose of keeping the commented out line of code - should it be uncommented if we updated the llvm version? What should be removed instead?
I have similar questions for the other commented out code. If there is any commented out code left here, then it should be clear when it should be uncommented and if anything else should be removed instead.

src/Passes/Source/Profiles/BaseProfile.cpp
```diff
@@ -0,0 +1,98 @@
...
+        function_pass_manager.addPass(TransformationRulePass(std::move(rule_set)));
+
+        // Eliminate dead code
+        function_pass_manager.addPass(llvm::DCEPass());
+        function_pass_manager.addPass(llvm::ADCEPass());
+
+        //  function_pass_manager.addPass(llvm::createCalledValuePropagationPass());
+        // function_pass_manager.addPass(createSIFoldOperandsPass());
+
+        // Legacy passes:
```
bettinaheim, line 58: Which of these lines of code apply to using the legacy pass manager instead?

src/Passes/Source/Rules/Factory.cpp
```diff
@@ -0,0 +1,334 @@
...
+                auto llvm_size = cst->getValue();
+                auto offset    = qubit_alloc_manager->getOffset(cap["arrayName"]->getName().str());
+
+                // Creating a new index APInt that is shifted by the offset of the allocation
+                auto idx = llvm::APInt(llvm_size.getBitWidth(), llvm_size.getZExtValue() + offset);
+
+                // Computing offset
+                auto new_index = llvm::ConstantInt::get(builder.getContext(), idx);
+
+                // TODO(tfr): Understand what the significance of the addressspace is in relation to the
```
bettinaheim, line 92: I am not sure this todo still applies, if the static qubit allocations are using the default address space as well.

src/Passes/Source/Rules/Factory.cpp
```diff
@@ -0,0 +1,334 @@
...
+                auto offset = qubit_alloc_manager->allocate();
+
+                // Creating a new index APInt that is shifted by the offset of the allocation
+                // TODO(tfr): Get the bitwidth size from somewhere
+                auto idx = llvm::APInt(64, offset);
+
+                // Computing offset
+                auto new_index = llvm::ConstantInt::get(builder.getContext(), idx);
+
+                // TODO(tfr): Understand what the significance of the addressspace is in relation to the
```
bettinaheim, line 149: Same as above. There also seems to be some code duplication with the above logic. It would be good to avoid the redundancy.

src/Passes/Source/Rules/Factory.cpp
```diff
@@ -0,0 +1,334 @@
...
+            cond->replaceAllUsesWith(new_call);
+
+            // Deleting the previous condition and function to fetch one
+            replacements.push_back({cond, nullptr});
+            replacements.push_back({cap["one"], nullptr});
+
+            return false;
+        };
+
+        /*
```
bettinaheim, line 286: That should be removed.

src/Passes/Source/Rules/Operands/Instruction.cpp
troelsfr: Yes, but I need to investigate this more as the error does not make any sense.

src/Passes/examples/ClassicalIrCommandline/README.md
troelsfr: This section can go. We only use the new pass manager.

src/Passes/Source/Passes/ExpandStaticAllocation/ExpandStaticAllocation.cpp
```diff
@@ -0,0 +1,180 @@
...
+#include "Llvm/Llvm.hpp"
+
+#include <fstream>
+#include <iostream>
+
+namespace microsoft
+{
+namespace quantum
+{
+    llvm::PreservedAnalyses ExpandStaticAllocationPass::run(
```
bettinaheim, line 15: A doc comment would be helpful, especially for larger functions that are intended to be more widely used.

src/Passes/libs/QubitAllocationAnalysis/SPECIFICATION.md
troelsfr: Hmm... that file was supposed to be deleted.

The `.ll` files are tests for the old QirAllocationAnalysis. They are kept as `.deprecated` to prevent them from running, but they are intended to be reinstated once I write tests for the full system.

src/Passes/README.md
```diff
@@ -1,156 +1,157 @@
-# QIR Passes for LLVM
```
bettinaheim, line 1: What was the reason for removing the general intro and links to the docs? Also the link in ## Out-of-source Pass might be useful to keep.

src/Passes/Source/AllocationManager/AllocationManager.cpp
```diff
@@ -0,0 +1,105 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#include "AllocationManager/AllocationManager.hpp"
```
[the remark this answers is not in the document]
troelsfr: This is a good suggestion. I will not make any comments on this at this point in time as I do not really know how the runtime allocator works. I think it is still very early days for the compiler allocations so it is hard to estimate what exactly is needed here in the long run. It may be that we can reuse parts of the runtime, or might make sense to make a common interface. 

I will leave this for now as I think this is a wider discussion to have with others involved as well.

bettinaheim (commented):
I went over the code relatively quick, and I am sure there are places where I didn't look carefully enough. I did a quick read through everything, though.

src/Passes/Source/AllocationManager/AllocationManager.cpp
```diff
@@ -0,0 +1,105 @@
...
+        auto ret = start_;
+        ++start_;
+        return ret;
+    }
+
+    void AllocationManager::allocate(String const& name, Index const& size, bool value_only)
+    {
+        if (resources_.find(name) != resources_.end())
+        {
+            throw std::runtime_error("Resource with name " + name + " already exists.");
```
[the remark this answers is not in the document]
troelsfr: At the top level of qat, yes. Inside the pass logic and/or LLVM, no. This is something that we may wish to revise once we have developed a good framework for handling errors in passes.

src/Passes/examples/ClassicalIrCommandline/README.md
bettinaheim: Are the passes themselves at this time only written for the new pass manager, or is it just the tool integration that does not enable using the legacy manager? I saw commented out code related to the legacy pass manager in the tool and I saw a (not commented out) include for the legacy pass manager - is that still needed?

src/Passes/Source/Commandline/ParameterParser.cpp
troelsfr: Yes, I agree. And in the long run, we will add a few dependencies, but I think this will be part of a separate PR that takes the tool closer to being fully configurable. This will also require a YAML parser so we should probably agree on what dependencies we should include overall there. I am happy to use CLI11 for parameter parsing and have a suggestion for a YAML library.

src/Passes/Source/Commandline/Settings.hpp
troelsfr: This is actually enforced by CI - you are not allowed to make headers that does not have this atm.

src/Passes/Source/Llvm/Llvm.hpp
troelsfr: We only use the new one and I believe we can remove this line.

src/Passes/Source/Passes/TransformationRule/LibTransformationRule.cpp
troelsfr: This is just an example pass that can be used with `opt`. In general, all files that starts with `Lib{name}` under passes implement a loadable pass. As we cannot (at least I believe so) provide a pass configuration except for O1, O2 or O3, the logic is hard coded. We may choose to to make multiple passes using the rule engine in the future, or we may choose to deprecate this alltogether.

src/Passes/Source/Passes/TransformationRule/LibTransformationRule.cpp
troelsfr: Everything that is done with our own tool `qat` is configurable and everything with `opt` is hard coded. As such, I suspect we will deprecate the usage of `opt` over time but it is quite instructive to have both to test certain aspects of a pass and to run passes individually (even if they are not configurable). Whether we want to make a pass for every profile is questionable though as it provides a lot of extra maintenance with little extra benefit.

src/Passes/Source/Passes/TransformationRule/TransformationRule.cpp
troelsfr: Yes, eventually, we will need a logger. We do not have this at the moment, but that is something to be considered when introducing dependencies.

src/Passes/Source/Rules/Notation/Call.ipp
troelsfr: No, it should called `.ipp` as in contains the implementation of a header library.

src/Passes/Source/Rules/Notation/Notation.hpp
troelsfr: The reason is that we want to allow notation like:
```c++
"foo"_cap = call("bar")
```
The structure `"foo"_cap` creates a `Capture` object which copies the pointer on the right-handside and returns a R-value that has a specified capture. Since it is an R-value we cannot return a reference as you would do under normal circumstances and the comment is to prevent anybody from removing the "// NOLINT" comment in the following line and return a reference as this would result in a incorrect implementation that would segfault in many cases.

src/Passes/Source/Rules/OperandPrototype.cpp
troelsfr: Yes, I agree. I haven't been able to do this until now as I had no place to do so.

src/Passes/Source/Rules/OperandPrototype.hpp
troelsfr: I think what you possible need is some high level description of the `OperandPrototype`, its relation to creating rules and in that context. what captures are?

src/Passes/examples/ClassicalIrCommandline/README.md
troelsfr: We only use the new pass manager at this point in time.

src/Passes/Source/Passes/QirAllocationAnalysis/QirAllocationAnalysis.hpp
troelsfr: As such I agree. The reason is that analysis typically comes with two components: The analytics component and the printer.

src/Passes/Source/Profiles/BaseProfile.cpp
troelsfr: Yes, it means, this code does not work in LLVM 11.  I will try to clarify this a bit, but it relates to my question about whether we would soon upgrade to 12 as the way to get a CG pass manager is somewhat hacky in 11 whereas it looks nice and clean in 12.

src/Passes/Source/Profiles/BaseProfile.cpp
troelsfr: This is just a link to all legacy passes which I use to figure out what the pass name is and subsequently derive the pass name for the new pass manager.

src/Passes/Source/Rules/Factory.cpp
troelsfr: No it does not apply any longer - we covered that.

src/Passes/Source/Rules/Factory.cpp
troelsfr: I actually don't think it should be removed. Instead it should be expanded as it is quite helpful to have an example IR that it is supposed to match next to the pattern definition.

src/Passes/README.md
troelsfr: As such we can keep it as training material, but it does not belong at the top-level README as this file should contain the "user" documentation such as "What does the tool do", "How to build the tool", "How to run the tool" and "How do I create new profiles". Passes and the usage of passes  is an implementation detail that only has relevance for those contributing to the implementation of the library.

src/Passes/Source/Profiles/BaseProfile.cpp
troelsfr: I've made it more clear.

src/Passes/Source/Profiles/BaseProfile.cpp
troelsfr: There is none of it that is for the legacy pass manager. We do not use that one.

src/Passes/Source/Rules/Factory.cpp
troelsfr: It is not!

src/Passes/Source/AllocationManager/AllocationManager.cpp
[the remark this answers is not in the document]
bettinaheim: Sounds good.

src/Passes/Source/Rules/Notation/Notation.hpp
bettinaheim: I think in that case this should be made more explicit in the comment. 
Something like "do not remove the //NOLINT since returning a reference will result in an incorrect implementation".

src/Passes/Source/Rules/OperandPrototype.hpp
bettinaheim: Yes.

src/Passes/Source/Rules/Factory.cpp
bettinaheim: Then it needs to be morphed into a proper comment. As is, it is not clear what to take away from this comment.

src/Passes/README.md
bettinaheim: I'd keep the info in the same place as any instructions for contributing to the passes.

PR #1118, opened 2021-08-13
Use stdin/stdout instead of TCP/IP.

cgranade:
This draft PR (open for early feedback) adds a new option to the language server allowing for stdin/stdout, and ports the VSCode extension to use the new option. This allows for avoiding firewall prompts when using the QDK on Windows.

### DIFF
--- a/src/QsCompiler/LanguageServer/Program.cs
+++ b/src/QsCompiler/LanguageServer/Program.cs
@@ -20,6 +20,7 @@ namespace Microsoft.Quantum.QsLanguageServer
             // Note: items in one set are mutually exclusive with items from other sets
             protected const string ConnectionViaSocket = "connectionViaSocket";
             protected const string ConnectionViaPipe = "connectionViaPipe";
+            protected const string ConnectionViaStdInOut = "connectionViaStdInOut";
 
             [Option(
                 'l',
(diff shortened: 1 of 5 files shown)


src/QsCompiler/LanguageServer/Program.cs
```diff
@@ -136,9 +145,11 @@ private static int Run(Options options)
             QsLanguageServer server;
             try
             {
-                server = options.ReaderPipeName != null && options.WriterPipeName != null
-                    ? ConnectViaNamedPipe(options.WriterPipeName, options.ReaderPipeName, options.LogFile)
-                    : ConnectViaSocket(port: options.Port, logFile: options.LogFile);
+                server = options.UseStdInOut
```
ricardo-espinoza, line 148: This order of precedence makes sense, standard input -> named pipe -> socket, but since we have independent options set for each, I wonder if we should return an error instead if a user specifies more than one option at the same time.

ricardo-espinoza:
LGTM. I'm only a bit surprised that so many changes happened in the package-lock.json file. Before merging the PR, I'd recommend re-creating it after merging main to see if the delta becomes smaller.

ricardo-espinoza (approved):

src/QsCompiler/LanguageServer/Program.cs
cgranade: That could make sense, yeah. My understanding was that that's handled by the option set functionality itself, but I could be wrong there.

cgranade:
> LGTM. I'm only a bit surprised that so many changes happened in the package-lock.json file. Before merging the PR, I'd recommend re-creating it after merging main to see if the delta becomes smaller.

Thanks for the review! As you suspected, the package-lock file did indeed become significantly smaller after merging in most recent changes on main.

jruales (approved):

PR #1120, opened 2021-08-17
Adding new custom simulator example

dime10:
Adds a new example demonstrating how to write a custom simulator and connect it to the QIR Runtime.

- A trace simulator sample printing straight-line quantum programs demonstrates the simplest implementation of a simulator backend, which can also be used to hook up hardware prototypes.
- A full state simulator sample demonstrates the elements of actually simulating quantum operations and how to use the QIR provided qubit manager.

~~Ideally I'd use a different method of including some of the headers defined in [qsharp-runtime/src/Qir/Runtime/public](https://github.com/microsoft/qsharp-runtime/tree/main/src/Qir/Runtime/public) needed to compile the simulator samples (currently just copied over).~~

Preview sample [here](https://github.com/microsoft/qsharp-compiler/tree/dime-qir-sim/examples/QIR/Simulation).


examples/QIR/Simulation/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+A qubit manager (see next section) is also provided in [lib/QIR/QubitManager.cpp](https://github.com/microsoft/qsharp-runtime/blob/main/src/Qir/Runtime/lib/QIR/QubitManager.cpp).
+
+## Qubit Management
+
+Generally, we can distinguish three layers of qubit representation or abstractions.
+Two of those are frequently encountered in quantum computing literature, the *logical qubit* and *the physical* qubit.
+
+A logical qubit is an idealized entity that represents the Qubit from quantum computation theory: a perfect two-level system that behaves as expected when applying quantum gates or measurements.
+The physical qubit on the other hand is a real physical system that *approximately* behaves as a two-level system, and is imperfectly manipulated and isolated from the environment, leading to a limited reliability and information life-time.
```
cgranade, line 59: I'm not sure that I agree with this definition of "logical" and "physical qubits"? Normally, I'd think of a logical qubit as being a two-level degree of freedom in a decoherence free subsystem, error correcting code, or other encoding into a larger register of physical qubits. Similarly, excepting errors such as leakage and coherent memory in the environment, physical qubits can be true two-level systems while still experiencing noise.

examples/QIR/Simulation/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+There are various techniques to create logical qubits from physical qubits via the use of error correction schemes and fault-tolerant systems.
+Whether the logical qubit is implemented by 1 or 1000 physical qubits, which type of physical system is used, or even whether one logical qubit remains on the same physical qubits throughout the entire computation, is not relevant at the logical level and is thus abstracted away.
+Quantum computing algorithms are typically formulated at the logical level unless they specifically deal with error correction or other aspects of physical qubits.
```
cgranade, line 63: I agree that this is definitely true of scalable devices, but may be worth calling out that this is _not_ true of near-term devices in general?

examples/QIR/Simulation/README.md
```diff
@@ -0,0 +1,193 @@
...
+virtual Result Measure(long numBases, PauliId bases[], long numTargets, Qubit targets[]) = 0;
+```
+
+---
+
+`IDiagnostics`
+
+```cpp
+// Internal state
+virtual void GetState(TGetStateCallback callback) = 0; // deprecated
```
cgranade, line 144: It may be worth defining `TGetStateCallback` here.

examples/QIR/Simulation/README.md
```diff
@@ -0,0 +1,193 @@
...
+---
+
+The requirements on a simulator are then:
+
+- Qubit management, which can include:
+  - virtual qubits
+  - logical qubits
+  - physical qubits
+
+  Generally, the physical qubit layer is not present in simulators, as they typically directly implement logical qubits, unless the simulator is built to simulate noise in real quantum computers.
```
cgranade, line 162: It may be worth mentioning the noisy simulators as a counterexample.

examples/QIR/Simulation/SimulatorTemplate.cpp
```diff
@@ -0,0 +1,176 @@
+#include <memory>
```
cgranade, line 1: Copyright header.

examples/QIR/Simulation/SimulatorTemplate.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+#include "QirRuntimeApi_I.hpp"
+#include "QSharpSimApi_I.hpp"
+
+namespace Microsoft
+{
+namespace Quantum
+{
```
cgranade, line 9: What version of C++ is being targeted? May be able to collapse this to `namespace Microsoft::Quantum`, depending on whether C++17 can be assumed.

examples/QIR/Simulation/StateSimulator/README.md
```diff
@@ -0,0 +1,336 @@
...
+Implementation of the `IRuntimeDriver` interface is straightforward.
+Qubit management is delegated to the respective `QubitManager` functions, taking care to add or remove qubits from the compute register and update the state vector accordingly.
+Note that new qubits are simply appended to the end of the register:
+
+```cpp
+Qubit StateSimulator::AllocateQubit()
+{
+    Qubit q = this->qbm->Allocate();
+    this->computeRegister.push_back(q);
+    UpdateState(this->numActiveQubits++);  // |Ψ'> = |Ψ> ⊗ |0>
```
cgranade, line 116:
```suggestion
    UpdateState(this->numActiveQubits++);  // |Ψ'⟩ = |Ψ> ⊗ |0⟩
```

examples/QIR/Simulation/StateSimulator/README.md
```diff
@@ -0,0 +1,336 @@
...
+{
+    // When adding a qubit, the state vector can be updated with: |Ψ'> = |Ψ> ⊗ |0>.
+    // When removing a qubit, it is traced out from the state vector: ρ' = tr_i[|Ψ><Ψ|].
+    if (!remove) {
+        this->stateVec = kroneckerProduct(this->stateVec, Vector2cd(1,0)).eval();
+    } else {
+        Operator densityMatrix = this->stateVec * this->stateVec.adjoint();
+        densityMatrix = PartialTrace(densityMatrix, qubitIndex, this->numActiveQubits);
+
+        // Ensure state is pure tr(ρ^2)=1, meaning the removed qubit was in a product state.
```
cgranade, line 267: It'd be good to sync with @swernli on this to make sure that release semantics agree with those on other simulators.

examples/QIR/Simulation/StateSimulator/RuntimeManagement.cpp
```diff
@@ -0,0 +1,59 @@
+#include "StateSimulator.hpp"
```
cgranade, line 1: Copyright header.

examples/QIR/Simulation/StateSimulator/StateSimulation.cpp
```diff
@@ -0,0 +1,338 @@
+#include <complex>
```
cgranade, line 1: Copyright header.

examples/QIR/Simulation/TraceSimulator/README.md
```diff
@@ -0,0 +1,229 @@
+# The Trace Simulator
+
+A trace simulator is a quick way to provide a backend to the QIR Runtime system.
+It's function is to print out all quantum instructions it receives.
+While this can useful for debugging, it is also a simple way to connect QIR to hardware backends that do not support control flow.
```
cgranade, line 5: In that application, it's worth noting that the output from trace simulation isn't a representation of the entire quantum program, but a trajectory through the program.

examples/QIR/Simulation/TraceSimulator/RuntimeManagement.cpp
```diff
@@ -0,0 +1,74 @@
+#include <memory>
```
cgranade, line 1: Copyright header.

examples/QIR/Simulation/TraceSimulator/README.md
dime10: This is true, which is why I specify this applies to "straight-line" programs with no control flow.

examples/QIR/Simulation/README.md
dime10: Are you referring to one of our existing simulators?

examples/QIR/Simulation/README.md
dime10: Is it worth it given the method is deprecated in favour of the others?

examples/QIR/Simulation/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+- a state-less [trace simulator](TraceSimulator): prints each quantum instructions it receives, useful for debugging or simple hardware backend hookup
+- a full state [quantum simulator](StateSimulator): simulates ideal quantum computer, inefficient but simple implementation that maps directly to mathematical description
+
+The file `SimulatorTemplate.cpp` in this directory is also good starting point for a custom simulator implementation, as it provides a template that just needs to be filled in with the bodies of required methods.
+
+## Understanding the QIR Runtime system
+
+The [QIR Runtime](https://github.com/microsoft/qsharp-runtime/tree/main/src/Qir/Runtime) implements the [QIR language specification](https://github.com/microsoft/qsharp-language/tree/main/Specifications/QIR) in LLVM.
+A quantum program written in QIR (or translated from Q# e.g.) will expect specific runtime functions to be available, which can include data type management or console output, but also qubit allocation, callable creation, and functor application, such as these QIR/LLVM functions below:
```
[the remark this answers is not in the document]
dime10: Yes exactly, thanks for providing this better phrasing!

examples/QIR/Simulation/SimulatorTemplate.cpp
dime10: I was actually looking for this feature but didn't know it was now possible in C++ 17, thank you for pointing that out.
However, I'm now considering it might be best to retain a larger compatibility, unless the QDK/Runtime/... is already exclusively using C++ 17?

examples/QIR/Simulation/StateSimulator/README.md
dime10: Nice formatting!

examples/QIR/Simulation/README.md
```diff
@@ -0,0 +1,193 @@
...
+void @__quantum__qis__t__adj(%Qubit*)
+void @__quantum__qis__z__ctl(%Array*, %Qubit*)
+%Result* @__quantum__qis__measure__body(%Array*, %Array*)
+```
+
+In order to communicate with the QIR Runtime, hardware backends or simulators can implement the following interfaces defined by the Runtime:
+
+- the driver `IRuntimeDriver` : this is critical for the Runtime to operate, implementing qubit allocation and other backend-specific functionality
+- the gate set `IQuantumGateSet` : only used by those backends that want to provide the Q# instruction set
+- the diagnostics `IDiagnostics` : optional but helpful component to provide debug information of the backend state
```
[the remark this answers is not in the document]
dime10: Is there other information that could be provided via the diagnostics interface that I can mention here?

examples/QIR/Simulation/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+```llvm
+%String* __quantum__rt__string_create(i8*)
+void __quantum__rt__message(%String*)
+%Qubit* __quantum__rt__qubit_allocate()
+void __quantum__rt__callable_invoke(%Callable*, %Tuple*, %Tuple*)
+void __quantum__rt__callable_make_adjoint(%Callable*)
+```
+
+Some are directly implemented by the Runtime, such as `__quantum__rt__message` or `__quantum__rt__string_create`, while others will require a call to the backend system, such as `__quantum__rt__qubit_allocate`.
```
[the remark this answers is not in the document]
dime10: That is how the runtime is currently written if you look at [bridge-rt.ll](https://github.com/microsoft/qsharp-runtime/blob/main/src/Qir/Runtime/lib/QIR/bridge-rt.ll) and [Output.cpp](https://github.com/microsoft/qsharp-runtime/blob/main/src/Qir/Runtime/lib/QIR/Output.cpp). I suppose the backend may provide an output stream, but the Runtime does not call any interface methods implemented by backends like it does for qubit allocation for example.

examples/QIR/Simulation/README.md
dime10: I'm not quite sure how our descriptions diverge:
- "a two-level degree of freedom in a decoherence free subsystem" to me sounds equivalent to "a perfect two-level system that behaves as expected [in reference to the theory, i.e. without noise/decoherence]", but maybe this could be phrased better.
- "physical qubits can be true two-level systems while still experiencing noise" I guess here you are getting at the possibility of a perfect two-level physical qubit, but as far as I know, for practical qubit systems the "two-levelness" is achieved by sufficiently separating two energy eigenstates from the rest of the spectrum. I could be wrong on this though, so please feel free to correct me on this.

dime10:
Thank you Chris (@cgranade) for providing feedback on this! :)

swernli:
Apologies to for the continued delay... I'm working on this review and should have it ready tomorrow!

examples/QIR/Simulation/StateSimulator/README.md
swernli: Reading through this, I think it works fine. It doesn't match the strict statement of released qubits needing to be measured or zero, but it does ensure that qubits are unentangled, which I think matches the minumum guarantee that releaseing a qubit should not lead to side effects.

examples/QIR/Simulation/StateSimulator/README.md
```diff
@@ -0,0 +1,375 @@
...
+    this->computeRegister.push_back(q);
+    UpdateState(this->numActiveQubits++);  // |Ψ'⟩ = |Ψ⟩ ⊗ |0⟩
+    return q;
+}
+
+void StateSimulator::ReleaseQubit(Qubit q)
+{
+    UpdateState(GetQubitIdx(q), /*remove=*/true);  // ρ' = tr_i[|Ψ⟩〈Ψ|]
+    this->numActiveQubits--;
+    this->computeRegister.erase(this->computeRegister.begin() + GetQubitIdx(q));
```
swernli, line 124: One thing to note about this implementation for qubit management is that it assumes a given qubit is never released before a qubit that was allocated after it (qubits are strictly First-In-Last-Out ordered). That is true for Q# due to enforced qubit scoping, but may not be true for QIR generally when generated from arbitrary languages.

swernli: I think it would be worth calling this out in the description above along with the note that newly allocated qubits are appended to the end of the register.

examples/QIR/Simulation/TraceSimulator/QubitManager.hpp
```diff
@@ -0,0 +1,39 @@
...
+        }
+
+      public:
+        Qubit AllocateQubit()
+        {
+            return reinterpret_cast<Qubit>(this->nextQubitId++);
+        }
+
+        void ReleaseQubit(Qubit qubit)
+        {
```
swernli, line 28: It might be worth noting here that not decrementing the `nextQubitId` means that no id will ever be reused. This may be intentional, but should probably be called out, as it differs from the behavior in the State Simulator.

examples/QIR/Simulation/TraceSimulator/README.md
```diff
@@ -0,0 +1,281 @@
...
+In this case, the internal qubit ID representation is used:
+
+```cpp
+std::string GetQubitName(Qubit qubit)
+{
+    return std::to_string(GetQubitId(qubit));
+}
+```
+
+The qubit ID is a simple incrementing counter, and the `Qubit` object is just a pointer (as defined in [CoreTypes.hpp](https://github.com/microsoft/qsharp-runtime/blob/main/src/Qir/Runtime/public/CoreTypes.hpp)) with the qubit ID as its value:
```
swernli, line 69: Actually, per my other comment, this is pobably a better place to explicitly call out that ids are never reused. From a tracing perspective, this can have an effect on total qubit count, which makes sense but might not be obvious to a user.

examples/QIR/Simulation/SimulatorTemplate.cpp
swernli: Yes, the QIR runtime from qsharp-runtime is already using C++ 17:
https://github.com/microsoft/qsharp-runtime/blob/7bf27f0b3b1a1752c5789c9ad4c8f454b8c4e046/src/Qir/Runtime/CMakeLists.txt#L8-L10

examples/QIR/Simulation/StateSimulator/README.md
```diff
@@ -0,0 +1,375 @@
...
+    cp tmp/Microsoft.Quantum.Qir.Runtime.0.18.2106148911-alpha/runtimes/win-x64/native/* build
+    rm -r tmp
+    ```
+
+- **Linux** (installs mono for the NuGet CLI):
+
+    ```shell
+    mkdir build
+    sudo apt update && sudo apt install -y mono-complete
+    curl https://dist.nuget.org/win-x86-commandline/latest/nuget.exe --output build/nuget
```
swernli, line 345: The steps to download nuget and setup via mono can actually be simplified since mono-based nuget is it's own package. See for example:
https://github.com/microsoft/qsharp-compiler/blob/edd7caadca25f8deb556f33c80a6755c6f2a5808/.devcontainer/Dockerfile#L9

swernli (approved):
I think the sample looks good. Thanks for working on this! I've left a few comments regarding places that could use a little clarification, and the nuget package on Linux can simplify the setup instructions, but otherwise it should be ready to merge.

examples/QIR/Simulation/TraceSimulator/README.md
```diff
@@ -0,0 +1,285 @@
...
+Applying gate "Exp(0.100000, X Y)" on qubits 0 1
+Measuring qubits:
+    0 in base Z
+    1 in base X
+```
+
+## Compiling the simulator
+
+The simulator samples require a working [Clang](https://clang.llvm.org/) installation to compile.
+Refer to the [Optimization example](../Optimization#installing-clang) for instructions on setting up Clang and LLVM.
```
swernli, line 181: @dime10 This line seems to be the source of the broken link. Looks like it's missing an extra `..` and should be:
```suggestion
Refer to the [Optimization example](../../Optimization#installing-clang) for instructions on setting up Clang and LLVM.
```

examples/QIR/Simulation/StateSimulator/README.md
```diff
@@ -0,0 +1,378 @@
...
+}
+```
+
+The function `BuildPauliUnitary` simply generates the `Operator` "`P_1⊗P_2⊗..⊗P_n`" over the active qubit space.
+
+
+## Compiling the simulator
+
+The simulator samples require a working [Clang](https://clang.llvm.org/) installation to compile.
+Refer to the [Optimization example](../Optimization#installing-clang) for instructions on setting up Clang and LLVM.
```
swernli, line 325:
```suggestion
Refer to the [Optimization example](../../Optimization#installing-clang) for instructions on setting up Clang and LLVM.
```

PR #1121, opened 2021-08-17
swernli/version-update

swernli:
This switches to a newer QDK beta package that has name fixes in the QIR generation for target packages (see https://github.com/microsoft/qsharp-runtime/pull/789). It includes the corresponding updates to the README.md and .ll files.

PR #1122, opened 2021-08-18
CTest Infrastructure, documentation Docker image and improved docs

troelsfr:
# CTest Infrastructure, test coverage and documentation 

This PR adds:
- Configuration files for tests in CMake 
- A profile test framework that allows to test transformations
- Name testing of `__quantum_(qis|rt)` patterns through the passes directory to ensure consistency with 
- Documentation generation using Doxygen and markdown to produce a unified documentation server

## Dependencies
No dependencies.

## Documentation 
This PR improves the documentation, by generating API and developer documentation using Doxygen. This documentation is combined with custom written pages. The result is packaged up as a docker image that serves all documentation via nginx. To build and serve the documentation, run

```
make serve-docs
```

## Updated style checking

The style checker now sanitises all QIR qis symbols to make the development more error prune:
```
% ./manage stylecheck --fix-issues
2021-08-19 09:18:21,090 - FormatChecker - ERROR - Invalid string format for: '"__quantum__rt__qubit_allocate_array ' in Passes/Source/Rules/Factory.cpp
2021-08-19 09:18:21,911 - FormatChecker - ERROR - Invalid instruction 's_adj' in Passes/docs/src/base-profile-transformations.md
2021-08-19 09:18:21,912 - FormatChecker - ERROR - Missing specialisation for quantum symbol: __quantum__qis__h in Passes/docs/src/DeveloperGuide/QuickStart.md
```
This is still an experimental feature and despite reporting errors, the stylechecker will not fail at this point. The main reason is that #1121 should be merged before activating the enforcement of this.

## Test coverage reports

Additionally, this PR adds test coverage reports for the individual tests. These can be generated as follows:

```
./manage  coverage --output [output_dir]
```

which will store a coverage report for each test in the library.


src/Passes/cmake/Library.cmake
```diff
@@ -0,0 +1,38 @@
...
+      list(APPEND filelist ${child})
+    endif()
+  endforeach()  
+
+  set(${result} ${filelist})
+endmacro()
+
+function (microsoft_add_library          
+          library)
+  list(REMOVE_AT ARGV 0)
```
swernli, line 15: Why is the first element removed from `ARGV`? I don't see it used again anywhere, but might be missing it.

src/Passes/cmake/Library.cmake
```diff
@@ -0,0 +1,38 @@
...
+
+  target_include_directories(
+    ${library} 
+    PRIVATE
+    "${CMAKE_CURRENT_SOURCE_DIR}/include"
+  )
+
+  if(MICROSOFT_ENABLE_DYNAMIC_LOADING)
+    target_link_libraries(${library} 
+      "$<$<PLATFORM_ID:Darwin>:-undefined dynamic_lookup>")
```
swernli, line 33: Is this `PLATFORM_ID:Darwin` something that works on all platforms or is it Mac OS specific?

src/Passes/cmake/Testing.cmake
```diff
@@ -0,0 +1,46 @@
...
+             ${name}
+             --gtest_shuffle
+             --gtest_random_seed=1337)  
+  endif (MICROSOFT_ENABLE_TESTS)
+endfunction()
+
+function (microsoft_add_library_tests
+          library)
+  if (MICROSOFT_ENABLE_TESTS)
+    list(REMOVE_AT ARGV 0)
```
swernli, line 28: Here again, I'm don't see why removing the first item from `ARGV` is needed.

src/Passes/docs/src/DeveloperGuide/QuickStart.md
```diff
@@ -0,0 +1,150 @@
...
+
+The QAT tool will now attempt to map the QIR in `analysis-example.ll` into a QIR which is compatible with the base format. Removing type and function declarations, the correspoding code reads:
+
+```
+; ModuleID = './analysis-example.ll'
+source_filename = "./analysis-example.ll"
+
+define internal fastcc void @TeleportChain__DemonstrateTeleportationUsingPresharedEntanglement__body() unnamed_addr {
+entry:
+  call void @__quantum__qis__h(%Qubit* null)
```
swernli, line 89: Just FYI, with the coming update to the newer beta package, these now show up with `__body` at the end of the names. See the changes to the old readme here: https://github.com/microsoft/qsharp-compiler/pull/1121.
It probably makes sense to merge this PR first, since it includes build fixes, and then update 1121 to include changes to these files.

swernli (commented):
I have a few questions posted, but one thing I think we need follow up on is whether adding the submodule for googletest is something that requires an update to our NOTICE.txt that advertises the licenses for any external dependencies the repo has. @idavis for advice on handling the submodule.

src/Passes/cmake/Library.cmake
troelsfr: At this point, this is mostly just book keeping on what was already consumed and not. The reason why it is there is that at a later point, we may choose to extend the command such that it becomes `microsoft_add_library(name dep1 dep2 ... depN)` and by removing the arguments which were already consumed the arg list is ready to do this. However, there is no decision to do one or the other thing atm., so I have just prepared it for it.

swernli:
Regarding the test dependency on an external submodule, would it be possible to just use Catch2 like the qsharp-runtime repo? See https://github.com/microsoft/qsharp-runtime/tree/main/src/Qir/Common/Externals#catch2 for where we use the library, and https://github.com/microsoft/qsharp-runtime/blob/main/src/Qir/Tests/QIR-dynamic/qir-driver.cpp as an example of how we use it.

src/Passes/cmake/Library.cmake
troelsfr: Yes, that is OS X specific. I am not sure whether it is needed for other platforms, but a quick Google suggests that these look substantially different for Linux (and possibly Windows).

src/Passes/docs/src/DeveloperGuide/QuickStart.md
troelsfr: Thanks for the heads up. Do we have a full list of all symbol names somewhere? If so, I might just scan the entire Passes directory for invalid names.

src/Passes/cmake/Testing.cmake
troelsfr: > Regarding the test dependency on an external submodule, would it be possible to just use Catch2 like the qsharp-runtime repo? See https://github.com/microsoft/qsharp-runtime/tree/main/src/Qir/Common/Externals#catch2 for where we use the library, and https://github.com/microsoft/qsharp-runtime/blob/main/src/Qir/Tests/QIR-dynamic/qir-driver.cpp as an example of how we use it.

What I agreed with @bettinaheim is that we follow the standard used by LLVM and try to align with them as closely as possible. That said, I am happy for us to use Catch2 instead of gmock if the consensus is that this is a better choice.

troelsfr:
> Regarding the test dependency on an external submodule, would it be possible to just use Catch2 like the qsharp-runtime repo? See https://github.com/microsoft/qsharp-runtime/tree/main/src/Qir/Common/Externals#catch2 for where we use the library, and https://github.com/microsoft/qsharp-runtime/blob/main/src/Qir/Tests/QIR-dynamic/qir-driver.cpp as an example of how we use it.

What I agreed with @bettinaheim is that we follow the standard used by LLVM and try to align with them as closely as possible. That said, I am happy for us to use Catch2 instead of gmock if the consensus is that this is a better choice.

src/Passes/docs/src/DeveloperGuide/QuickStart.md
swernli: We don't have that full list in one place, no, but it comes from the change made to the runtime repo. In #1121 I went ahead and did what you suggest: searched the whole Passes directory and double checked every use of a `__quantum__qis__` function to make sure I either regenerated the corresponding file or manually updated the references. I can just do that again after this PR merges.

swernli:
> What I agreed with @bettinaheim is that we follow the standard used by LLVM and try to align with them as closely as possible. That said, I am happy for us to use Catch2 instead of gmock if the consensus is that this is a better choice.

Ah, that makes sense. I could see it going either way; aligning with the patterns and processes used by LLVM is a good idea, but bringing in a submodule comes with its own considerations while this is under the Microsoft org on GitHub (this would be the first submodule added to any of our repos, though likely not the last). Catch2 is an easy, one file dependency, but means that our CMakeLists and test files would look rather different than the ones from LLVM, which also brings overhead for developers. Unfortunately, there isn't an easy or quick answer here.

src/Passes/docs/src/DeveloperGuide/QuickStart.md
troelsfr: NIce, thanks!

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "src/Passes/vendors/googletest"]
```
idavis, line 1: Should this be shallow?

src/Passes/Makefile
```diff
@@ -1,6 +1,18 @@
...
-	
+
+documentation:
+	docker build -f Docker/Docs.dockerfile -t qir-passes-docs:latest .	
+
+documentation-shell: documentation
+	docker run --rm -i -t qir-passes-docs:latest sh
+
+serve-docs:  documentation
+	docker run -it --rm -p 8080:80 --name qir-documentation -t qir-passes-docs:latest
```
idavis, line 11: We should gen the docs into the project/docs folder as markdown instead of a container

idavis (changes requested):

.gitmodules
troelsfr: For me it is fine - I prefer to manually initialise submodules. However, ultimately, this is up to you.

PR #1124, opened 2021-08-24
Add QIR JIT compilation from Python example

dime10:
This example accesses the LLVM libraries from Python to JIT compile QIR programs.
It uses the llvmlite Python bindings for easier access to the LLVM API.

A small wrapper with C linkage around the QIR context and simulator instantiation needs to be pre-compiled before running the JIT, since you can't directly call into the C++ ABI from Python.

Preview sample [here](https://github.com/microsoft/qsharp-compiler/tree/dime-qir-jit/examples/QIR/JITCompilation).


swernli (approved):
Very cool sample, thanks for putting it together!

Issue #1128, opened 2021-09-01
Potential memory leak in generated QIR


bettinaheim:
The following code causes a potential memory leak:

```qsharp
    @EntryPoint()
    operation RunExample() : String {

        let model = SequentialModel([], [0.], 0.0);
        let _ = Enumerated([model]); // commenting out this line avoids the issue
        return "Executed successfully!";
    }
```

bettinaheim:
Possibly related: https://github.com/microsoft/qsharp-runtime/issues/757

bettinaheim:
More infos:
```qsharp
    internal newtype Foo = Unit;
    internal function Identity<'T>(input : 'T) : 'T { return input; }

    function MappedByIndex<'T, 'U> (mapper : ((Int, 'T) -> 'U), array : 'T[]) : 'U[] {
        mutable resultArray = new 'U[Length(array)];

        for idxElement in IndexRange(array) {
            set resultArray w/= idxElement <- mapper(idxElement, array[idxElement]);
        }

        return resultArray;
    }

    @EntryPoint()
    operation RunExample() : String {

        // FIXME: fails execution indicating a potential memory leak
        //let model = SequentialModel([], [0.], 0.0);
        //let _ = Enumerated([model]);

        // FIXME: throws an exception in QIR generation
        let item = Foo(); //"hello";
        let _ = MappedByIndex(Identity, [item]);

        // FIXME: fails execution without any info
        //let str = "hello";
        //let _ = MappedByIndex(Identity, [str]);

        return "Executed successfully!";
    }
```

bettinaheim:
Fixed the type issue for udts consisting of a single Unit-valued item, and the issue with default values for string.
Looking into the memory leak now, which seems occur when invoking a callable value (opposed to directly the declaration) of a type parameterized callable that takes a tuple as argument. 

This works fine:
```qsharp
    function MappedByIndex (mapper : SequentialModel -> SequentialModel) : SequentialModel[] {
        return [mapper(SequentialModel([], [0.], 0.0))];
    }

    @EntryPoint()
    operation RunExample() : String {

        // FIXME: fails execution indicating a potential memory leak
        let _ = MappedByIndex(Identity);
        return "Executed successfully!";
    }
```
This doesn't:
```qsharp
    function MappedByIndex (mapper : ((Int, SequentialModel) -> (Int, SequentialModel))) : (Int, SequentialModel)[] {
[...]

PR #1129, opened 2021-09-02
Fixing memory leak in generated QIR

bettinaheim:
Fixes all issues listed in https://github.com/microsoft/qsharp-compiler/issues/1128.


src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestReferenceCounts3.ll
```diff
@@ -0,0 +1,32 @@
+define internal void @Microsoft__Quantum__Testing__QIR__Main__body() {
+entry:
+  %0 = call { %String*, %Array* }* @Microsoft__Quantum__Testing__QIR__TestPendingRefCountIncreases__body(i1 true)
+  call void @Microsoft__Quantum__Testing__QIR__TestRefCountsForItemUpdate__body(i1 true)
+  %id = call %Callable* @__quantum__rt__callable_create([4 x void (%Tuple*, %Tuple*, %Tuple*)*]* @Microsoft__Quantum__Testing__QIR_____GUID___Identity, [2 x void (%Tuple*, i32)*]* null, %Tuple* null)
+  call void @__quantum__rt__capture_update_alias_count(%Callable* %id, i32 1)
```
swernli, line 6: One thing I'm still confused by here: why does the `%id` callable have it's capture alias and callable alias incremented here? Presumably it starts at one, then gets incremented here to two, then it is used, and then on lines 21-22 and 27-28 it is decremented twice to put it to zero and get released. It doesn't appear to be leaked at all, but it's not clear to me why it needed alias counts of 2 instead of just using the 1 it starts with and then decrementing once at the end of the scope.

bettinaheim: No, the alias always starts at 0. It gets incremented to 1 due to the assignment. This could potentially be optimized away, but I won't to any more involved optimizations until we have a better test coverage across the board.

swernli (approved):

PR #1134, opened 2021-09-07
QAT: Improving architecture and configuration management

troelsfr:
This PR improves the architecture and the command-line interface.

## Dependences

This PR has no formal dependencies. However, as it has been necessary to merge main to get CI to pass, it is suggested that we merge #1140 to reduce the size of the diff and make this PR more manageable.

## Change log

- Creation of component based configuration.
- Improved command line interface
- Refactor of naming and architecture
- Initial logging mechanism
- Elimination of several TODOs
- Allowing (almost) non-invasive 3rd party modules
- Much improved test coverage

## Base configuration

Each profile component/extension now requires its own configuration class. This forms the basis for modularising the framework. An example component which manages the LLVM extensions looks like:

```c++
    class LlvmPassesConfiguration
    {
      public:
        void setup(ConfigurationManager& config);

        static LlvmPassesConfiguration disable();
        bool                           alwaysInline() const;

        bool isDisabled() const;

        bool isDefault() const;

      private:
        bool always_inline_{false};
    };
```

where the parameters are explained in the configuration setup function:
```c++
    void LlvmPassesConfiguration::setup(ConfigurationManager& config)
    {
        config.setSectionName("LLVM Passes", "Configuration of LLVM passes.");
        config.addParameter(always_inline_, "always-inline", "Aggresively inline function calls.");
    }
```
Upon registering a new configuration, these parameters are propagated automatically to the commaline interface:

```sh
% ./qat
Usage: ./Source/Apps/qat [options] filename

Pass configuration - Configuration of the pass and its corresponding optimisations.

--[no-]delete-dead-code                           Deleted dead code. Default: true
--[no-]clone-functions                            Clone functions to ensure correct qubit allocation. Default: true
--max-recursion                                   max-recursion Default: 512
--[no-]reuse-qubits                               reuse-qubits Default: true
--entry-point-attr                                Specifies the attribute indicating the entry point. Default: EntryPoint
--group-measurements                              NOT IMPLEMENTED - group-measurements Default: false
--[no-]one-shot-measurement                       NOT IMPLEMENTED - one-shot-measurement Default: true

LLVM Passes - Configuration of LLVM passes.

--always-inline                                   Aggresively inline function calls. Default: false

Base configuration - Configuration of the quantum adoption tool to execute a specific behaviour.

--generate                                        Transforms the IR in correspondance with the specified transformation. Default: false
--validate                                        Executes the validation produre. Default: false
--profile                                         Sets the profile. Default: baseProfile
--S                                               Emits LLVM IR to the standard output. Default: false
--O0                                              Optimisation level 0. Default: false
--O1                                              Optimisation level 1. Default: false
--O2                                              Optimisation level 2. Default: false
--O3                                              Optimisation level 3. Default: false
--verify-module                                   Verifies the module after transformation. Default: false
--dump-config                                     Prints the configuration to the standard output. Default: false

Transformation rules - Rules used to transform instruction sequences in the QIR.

--[no-]disable-reference-counting                 Disables reference counting by instruction removal. Default: true
--[no-]disable-reference-counting                 Disables reference counting by instruction removal. Default: true
--[no-]disable-alias-counting                     Disables alias counting by instruction removal. Default: true
--[no-]disable-string-support                     Disables string support by instruction removal. Default: true
--[no-]optimise-branch-quatum-one                 Maps branching based on quantum measurements compared to one to base profile type measurement. Default: true
--[no-]optimise-branch-quatum-zero                Maps branching based on quantum measurements compared to zero to base profile type measurement. Default: true
--[no-]use-static-qubit-array-allocation          Maps allocation of qubit arrays to static array allocation. Default: true
--[no-]use-static-qubit-allocation                Maps qubit allocation to static allocation. Default: true
--[no-]use-static-result-allocation               Maps result allocation to static allocation. Default: true
```

## Allowing 3rd party registration

It is now possible to register profile extensions outside of the generator:

```c++
        generator.registerProfileComponent<RuleTransformationPassConfiguration>(
            "transformation-rules", [this](RuleTransformationPassConfiguration const& config, IProfileGenerator* ptr) {
                auto& ret = ptr->modulePassManager();

                // Defining the mapping
                RuleSet rule_set;
                auto    factory = RuleFactory(rule_set);
                factory.usingConfiguration(configurationManager().get<FactoryConfiguration>());

                // Creating profile pass
                ret.addPass(RuleTransformationPass(std::move(rule_set), config));
            });
```

This paves the way for allowing dynamic modules which can be created and loaded non-invasively by third party.

## Test coverage

This PR moves the average test coverage well-above 75% and mostly remaining parts of the code are utility parts such as configuration parsing etc. meaning that the core functionality is covered by tests at this point:

Filename | Function Coverage | Line Coverage | Region Coverage
-- | -- | -- | --
AllocationManager/AllocationManager.cpp | 100.00% (4/4) | 100.00% (61/61) | 100.00% (16/16)
AllocationManager/AllocationManager.hpp | 100.00% (1/1) | 100.00% (1/1) | 100.00% (1/1)
AllocationManager/IAllocationManager.cpp | 100.00% (1/1) | 100.00% (1/1) | 100.00% (1/1)
Commandline/ConfigurationManager.cpp | 100.00% (1/1) | 100.00% (1/1) | 100.00% (1/1)
Commandline/ConfigurationManager.hpp | 70.97% (22/31) | 65.90% (143/217) | 78.69% (48/61)
Commandline/ParameterParser.cpp | 66.67% (6/9) | 63.74% (58/91) | 73.81% (31/42)
Commandline/ParameterParser.hpp | 100.00% (2/2) | 100.00% (2/2) | 100.00% (2/2)
Generators/DefaultProfileGenerator.cpp | 88.89% (8/9) | 82.05% (96/117) | 76.92% (10/13)
Generators/IProfileGenerator.cpp | 55.56% (5/9) | 65.91% (29/44) | 58.33% (7/12)
Generators/IProfileGenerator.hpp | 100.00% (4/4) | 100.00% (16/16) | 100.00% (4/4)
Generators/LlvmPassesConfig.cpp | 100.00% (5/5) | 100.00% (19/19) | 100.00% (5/5)
RuleTransformationPass/Configuration.cpp | 75.00% (9/12) | 86.15% (56/65) | 90.62% (29/32)
RuleTransformationPass/Profile.cpp | 76.19% (16/21) | 52.24% (362/693) | 52.25% (93/178)
RuleTransformationPass/Profile.hpp | 100.00% (3/3) | 100.00% (4/4) | 100.00% (3/3)
Rules/Factory.cpp | 95.24% (20/21) | 93.50% (748/800) | 82.83% (82/99)
Rules/Factory.hpp | 100.00% (1/1) | 100.00% (1/1) | 100.00% (1/1)
Rules/FactoryConfig.hpp | 100.00% (13/13) | 100.00% (90/90) | 100.00% (41/41)
Rules/Notation/BitCast.cpp | 100.00% (1/1) | 100.00% (6/6) | 100.00% (1/1)
Rules/Notation/Branch.cpp | 100.00% (1/1) | 100.00% (9/9) | 100.00% (1/1)
Rules/Notation/Call.ipp | 100.00% (1/1) | 100.00% (15/15) | 100.00% (2/2)
Rules/Notation/CallByNameOnly.cpp | 100.00% (1/1) | 100.00% (4/4) | 100.00% (1/1)
Rules/Notation/Capture.cpp | 100.00% (3/3) | 100.00% (10/10) | 100.00% (3/3)
Rules/Notation/ConstInt.cpp | 100.00% (1/1) | 100.00% (5/5) | 100.00% (1/1)
Rules/Notation/IntToPtr.cpp | 100.00% (1/1) | 100.00% (6/6) | 100.00% (1/1)
Rules/Notation/Load.cpp | 100.00% (1/1) | 100.00% (6/6) | 100.00% (1/1)
Rules/Notation/Notation.cpp | 100.00% (2/2) | 100.00% (11/11) | 100.00% (2/2)
Rules/Notation/Phi.ipp | 100.00% (1/1) | 100.00% (12/12) | 100.00% (2/2)
Rules/OperandPrototype.cpp | 100.00% (8/8) | 90.77% (59/65) | 93.10% (27/29)
Rules/OperandPrototype.hpp | 100.00% (2/2) | 100.00% (5/5) | 100.00% (2/2)
Rules/Operands/Any.cpp | 100.00% (4/4) | 100.00% (8/8) | 100.00% (4/4)
Rules/Operands/Call.cpp | 100.00% (4/4) | 100.00% (25/25) | 100.00% (10/10)
Rules/Operands/Instruction.cpp | 100.00% (2/2) | 100.00% (14/14) | 100.00% (5/5)
Rules/Operands/Instruction.hpp | 100.00% (1/1) | 100.00% (1/1) | 100.00% (1/1)
Rules/Operands/Phi.cpp | 66.67% (2/3) | 46.67% (7/15) | 66.67% (4/6)
Rules/Operands/Phi.hpp | 100.00% (1/1) | 100.00% (1/1) | 100.00% (1/1)
Rules/ReplacementRule.cpp | 100.00% (5/5) | 100.00% (32/32) | 100.00% (13/13)
Rules/RuleSet.cpp | 100.00% (4/4) | 100.00% (31/31) | 100.00% (11/11)
Rules/RuleSet.hpp | 100.00% (3/3) | 100.00% (3/3) | 100.00% (3/3)
TestTools/IrManipulationTestHelper.cpp | 100.00% (18/18) | 93.85% (183/195) | 92.59% (50/54)
Totals | 87.44% (188/215) | 79.24% (2141/2702) | 78.23% (521/666)


bettinaheim:
@DmitryVasilevsky Could you please take a quick look at the AllocationManager logic and APIs?

src/Passes/Source/Generators/LlvmPassesConfig.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+    class LlvmPassesConfiguration
+    {
+      public:
+        void setup(ConfigurationManager& config);
+
+        static LlvmPassesConfiguration disable();
+        bool                           alwaysInline() const;
+
+        bool isDisabled() const;
```
bettinaheim, line 20: My understanding is that this is a shared configuration for all passes, rather than each pass having its config - is that correct? If so, what does it mean for the configuration to be disabled?

src/Passes/Source/Generators/LlvmPassesConfig.hpp
```diff
@@ -0,0 +1,29 @@
...
+// Licensed under the MIT License.
+
+#include "Commandline/ConfigurationManager.hpp"
+
+namespace microsoft
+{
+namespace quantum
+{
+
+    class LlvmPassesConfiguration
```
bettinaheim, line 12: It might be worthwhile to think about how defining a configuration for each pass would look like and whether a common interface for doing so would make sense. It might make sense to have a per-pass pieces for each pass as part of a common configuration for all passes.

src/Passes/Source/Logging/ILogger.hpp
```diff
@@ -0,0 +1,29 @@
...
+    class ILogger
+    {
+      public:
+        using String                                                             = std::string;
+        virtual ~ILogger()                                                       = default;
+        virtual void debug(String const& message)                                = 0;
+        virtual void info(String const& message)                                 = 0;
+        virtual void warning(String const& message)                              = 0;
+        virtual void error(String const& message)                                = 0;
+        virtual void internalError(String const& message)                        = 0;
```
bettinaheim, line 22: I like the distinction of an internalError here.

src/Passes/Source/RuleTransformationPass/Configuration.hpp
```diff
@@ -0,0 +1,58 @@
...
+// Licensed under the MIT License.
+
+#include "Commandline/ConfigurationManager.hpp"
+
+namespace microsoft
+{
+namespace quantum
+{
+
+    class RuleTransformationPassConfiguration
```
bettinaheim, line 12: At some point we will need more descriptive doc comments for this class as well.

src/Passes/Source/RuleTransformationPass/Tests/Unit/LoopUnrolling.cpp
```diff
@@ -0,0 +1,93 @@
...
+    }
+    return ir_manip;
+}
+
+} // namespace
+
+// Single allocation with action and then release
+TEST(RuleTransformationPass, LoopUnroll)
+{
+    auto ir_manip = newIrManip(R"script(
```
bettinaheim, line 50: I would prefer to split out larger snippets to compare against into separate files rather than inline them in the test code. It gets messy quickly, and anywhere where we had not split things such as this split out into separate files I believe we have had a desire to revise it later.

src/Passes/Source/Rules/Factory.cpp
```diff
@@ -375,6 +419,11 @@ namespace quantum
         addRule({call("__quantum__qis__m__body", "qubit"_cap = _), std::move(replace_measurement)});
     }
 
+    void RuleFactory::optimiseBranchQuatumZero()
```
bettinaheim, line 422: Could you please do a search for "Quatum" everywhere in the folder and replace it with "Quantum"? Also, I would recommend using a spell check extension (e.g. https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker).

src/Passes/Source/Rules/Factory.cpp
```diff
@@ -375,6 +419,11 @@ namespace quantum
...
+    {
+        // TODO(tfr): Not implemented
```
bettinaheim, line 424: Should this throw a not implemented exception for now?

src/Passes/Source/Rules/FactoryConfig.hpp
```diff
@@ -0,0 +1,153 @@
...
+        {
+            return disable_alias_counting_;
+        }
+
+        bool disableStringSupport() const
+        {
+            return disable_string_support_;
+        }
+
+        bool optimiseBranchQuatumOne() const
```
bettinaheim, line 77: These methods could also profit from documentation.

src/Passes/Source/Rules/Tests/Unit/RemoveFunctionCall.cpp
```diff
@@ -51,9 +51,10 @@ TEST(RuleSetTestSuite, RemovingFunctionCall)
         factory.removeFunctionCall("__quantum__qis__h__body");
     };
 
-    auto profile = std::make_shared<RuleSetProfile>(std::move(configure_profile));
-    ir_manip->applyProfile(profile);
+    auto profile = std::make_shared<DefaultProfileGenerator>(std::move(configure_profile));
 
+    ir_manip->applyProfile(profile);
+    llvm::errs() << *ir_manip->module() << "\n";
```
bettinaheim, line 57: Is there a check somewhere that there are no errors?

src/Passes/examples/TeleportChain/Makefile
```diff
@@ -15,3 +15,5 @@ QSharpVersion/qir/Example.ll:
 clean:
 	cd QSharpVersion && make clean
 
+
+# --no-disable-reference-counting --no-disable-reference-counting --no-disable-alias-counting --no-disable-string-support --no-optimise-branch-quatum-one --no-optimise-branch-quatum-zero --no-use-static-qubit-array-allocation --no-use-static-qubit-allocation --no-use-static-result-allocation
```
bettinaheim, line 19: What are these? This comment needs a description of what these are or should be removed.

bettinaheim (commented):
What are the two .profraw binary files that are checked in?

src/Passes/Source/Generators/LlvmPassesConfig.cpp
```diff
@@ -0,0 +1,42 @@
...
+namespace quantum
+{
+
+    void LlvmPassesConfiguration::setup(ConfigurationManager& config)
+    {
+        config.setSectionName("LLVM Passes", "Configuration of LLVM passes.");
+        config.addParameter(always_inline_, "always-inline", "Aggresively inline function calls.");
+    }
+
+    LlvmPassesConfiguration LlvmPassesConfiguration::disable()
```
DmitryVasilevsky, line 18: It's not clear why this function is called disable and why does it return and object.

src/Passes/Source/Generators/LlvmPassesConfig.hpp
troelsfr: This is not entirely correct. This configuration is only concerned with LLVM passes. LLVM passes has two options: Either they are on or they are off. While you could make a separate config class for each LLVM pass, it would always only have one boolean flag hence they are bundled into one configuration.

The static member function `disabled` creates a config where all LLVM passes are disabled and the corresponding member function `isDisabled` checks whether the configuration is an instance of completely disabling all classes.

For our own passes on the other hand, we can pass configurations to them and consequently each of these passes have their own config class and in some cases even subsystems of that pass can have its own config.

src/Passes/Source/Generators/LlvmPassesConfig.hpp
troelsfr: That is already done. This setup is a per-piece configuration and you will notice that each profile component (which may consist of one or more passes) have their own config. See for instance component registration in `DefaultProfileGenerator`:

```c++
        registerProfileComponent<RuleTransformationPassConfiguration>(
            "transformation-rules",
            [this](RuleTransformationPassConfiguration const& config, IProfileGenerator* ptr, Profile&) {
                auto& ret = ptr->modulePassManager();

                // Defining the mapping
                RuleSet rule_set;
                auto    factory = RuleFactory(rule_set);
                factory.usingConfiguration(configurationManager().get<FactoryConfiguration>());

                // Creating profile pass
                ret.addPass(RuleTransformationPass(std::move(rule_set), config));
            });

        registerProfileComponent<LlvmPassesConfiguration>(
            "llvm-passes", [](LlvmPassesConfiguration const& config, IProfileGenerator* ptr, Profile&) {
                // Configuring LLVM passes
                if (config.alwaysInline())
                {
                    auto& ret          = ptr->modulePassManager();
                    auto& pass_builder = ptr->passBuilder();
                    ret.addPass(llvm::AlwaysInlinerPass());

                    auto inliner_pass = pass_builder.buildInlinerPipeline(
                        ptr->optimisationLevel(), llvm::PassBuilder::ThinLTOPhase::None, ptr->debug());
                    ret.addPass(std::move(inliner_pass));
                }
            });
```
Whether or not we choose that a component _always_ must be comprised of exactly one pass is a different discussion.

src/Passes/Source/Rules/Factory.cpp
troelsfr: I am happy either way. As such it should just be implemented - I just did not find the time for it yet.

src/Passes/Source/Rules/Tests/Unit/RemoveFunctionCall.cpp
troelsfr: Yes,
```c++
    void IrManipulationTestHelper::applyProfile(
        GeneratorPtr const&      generator,
        OptimizationLevel const& optimisation_level,
        bool                     debug)
    {
        auto profile = generator->newProfile(optimisation_level, debug);
        profile.apply(*module_);

        // Verifying that the module is valid
        if (isModuleBroken())
        {
            throw std::runtime_error("Module was broken after applying result");
        }
    }
```

src/Passes/Source/Generators/LlvmPassesConfig.cpp
troelsfr: It is a static function that returns a instance of the configuration which disables all passes.

src/Passes/Source/Apps/Qat/Config.cpp
```diff
@@ -0,0 +1,87 @@
+// Copyright (c) Microsoft Corporation.
```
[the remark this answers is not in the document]
troelsfr: I am not sure I understand the comment. If you are referring to the setup function, then this is used to automatically configure the commandline parser such that whenever a configuration is attached to the ConfigurationManager the options in the configuration are automatically available from the CLI.

src/Passes/Source/Apps/Qat/Qat.cpp
```diff
@@ -31,35 +64,33 @@ int main(int argc, char** argv)
...
+        // picture of what is going.
+        if (config.dumpConfig())
+        {
+            configuration_manager.printConfiguration();
+        }
 
         if (parser.arguments().empty())
         {
             std::cerr << "Usage: " << argv[0] << " [options] filename" << std::endl;
+            configuration_manager.printHelp();
```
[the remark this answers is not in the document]
troelsfr: Arguably only errors should go to the error stream which is what is happening here. I think it would make more sense to change `llvm::errs()` to `llvm::outs()` the places where it is actually not errors. What are your thoughts?

src/Passes/Source/RuleTransformationPass/Tests/Unit/LoopUnrolling.cpp
troelsfr: In principle I agree with you, although I do not see it as a huge issue. Since we are using a combination of CMake and Google Test, I do not think that there is an elegant solution to this, but I am open to suggestions. Keep in mind that you would need the code snippet together with the expected instruction and unexpected instructions to be separated out.

How do you envision this will be structured? Where would these files live? And how would the program know where to find these files?

src/Passes/Source/Apps/Qat/Config.cpp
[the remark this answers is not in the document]
bettinaheim: I thought I had seen the option to generate a config file somewhere, but maybe I am misremembering. I'll keep my eyes out next time I take a look.

src/Passes/Source/Apps/Qat/Qat.cpp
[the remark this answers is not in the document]
bettinaheim: I agree to limit the error stream to be only errors. I think we need to carefully examine and probably replace all statements that write to one of the streams to accumulate and handle that properly in the service. @idavis If I remember correct we briefly touched on this today - you may have some suggestions off the top of your head (for a next PR)?

src/Passes/Source/RuleTransformationPass/Tests/Unit/LoopUnrolling.cpp
bettinaheim: I don't know about elegant, but the simple setup that has worked well is simply by having these files checked in in a subfolder and the tests have hardcoded paths to those files. Within dotnet, the files are considered to be content files that are part of the project and hence automatically copied to the build output folder. I don't know if there is a similar setup that can conveniently be used here.

src/Passes/Source/Rules/Factory.cpp
bettinaheim: I would suggest to throw a not-implemented-exception then if the intent is that it should have to have an implementation. 
Also it look like there is another type of Quatum opposed to Quantum here.

src/Passes/Source/Generators/LlvmPassesConfig.cpp
bettinaheim: This needs to be documented.

src/Passes/Source/Generators/LlvmPassesConfig.hpp
bettinaheim: The disabled property on each pass makes sense, but having it as a property to indicate if all llvm passes are disabled is confusing. I would suggest to drop the isDisabled, and instead of a static member disabled, have a either something like a .WithLlvmPassesDisabled member function, or a constructor that e.g. takes an enum to select predefined configurations via a descriptive name.

src/Passes/Source/Apps/Qat/Config.hpp
```diff
@@ -0,0 +1,44 @@
...
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#include "Commandline/ConfigurationManager.hpp"
+
+namespace microsoft
+{
+namespace quantum
+{
+    class QatConfig
```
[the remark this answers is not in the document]
troelsfr: Please see #1147.

src/Passes/Source/Apps/Qat/Qat.cpp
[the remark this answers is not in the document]
troelsfr: I am happy to discuss this, but think it is out of scope for this PR.

src/Passes/Source/Commandline/ConfigurationManager.hpp
```diff
@@ -0,0 +1,372 @@
...
+#include <sstream>
+#include <type_traits>
+#include <typeindex>
+
+namespace microsoft
+{
+namespace quantum
+{
+
+    class IConfigBind
```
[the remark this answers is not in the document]
troelsfr: Please see #1147.

src/Passes/Source/RuleTransformationPass/Configuration.hpp
troelsfr: Please see #1147

src/Passes/Source/RuleTransformationPass/Tests/Unit/LoopUnrolling.cpp
troelsfr: I am happy to have that discussion. I do think it is out of scope for the present PR as this would require a large change in how we deal with tests atm.

src/Passes/Source/Rules/FactoryConfig.hpp
troelsfr: Please see #1147.

src/Passes/Source/Generators/LlvmPassesConfig.cpp
troelsfr: Please see #1147.

src/Passes/Source/RuleTransformationPass/Tests/Unit/LoopUnrolling.cpp
bettinaheim: I agree to defer this conversation to a separate PR, given that we have enough dependencies between PRs as it is.

src/Passes/Source/Apps/Qat/Qat.cpp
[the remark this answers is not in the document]
bettinaheim: Agreed.

bettinaheim (approved):

src/Passes/Source/Generators/LlvmPassesConfig.hpp
troelsfr: First of all, notice that `isDisabled` is there because it is used. A disabled configuration for each of the passes does not have any meaning as this already covered by the passes flag by itself. For instance, `alwaysInline()` tells you whether or not the always inline pass will be added or not.

While I think it may make sense to choose a better name for the static function, the proposed change would change the interface when creating a generator from

`Generator(RuleTransformationPassConfiguration::disabled(), LlvmPassesConfiguration::disabled());`

to

`Generator(RuleTransformationPassConfiguration(RuleTransformationPassConfiguration::Config::DisableAll), 
LlvmPassesConfiguration(LlvmPassesConfiguration::Config::DisableAll));`

I am reluctant to make this change as it decreases code readibility substantially and does not provide any additional benefits in terms of readibility or code efficiency. Likewise, checking if a configuration disables a component altogether is currently

```c++
if(config.isDisabled())
{
  
}
```

vs

```c++
if(config.alwaysInline() == false && config.pass2() == false && /* ... */)
{
  
}
```

This will only look worse as more passes are added to the LLVM passes component. Also the readibility of the implementation would be affected by the proposed. If this is something you feel strongly about, lets have a discussion about it.

src/Passes/Source/Rules/Factory.cpp
troelsfr: That's fine. I will add a throw in #1142 for the sake of closing some of these PRs.

PR #1136, opened 2021-09-09
Remove For-Loop Parenthesis

ScottCarda-MS:
Adds data structures for for-loops and for-loop bindings.
Adds a rule that removes the deprecated parenthesis around the bindings of for-loop statements.

Addresses Issue #1144

### DIFF
--- a/src/QsFmt/Formatter/SyntaxTree/Statement.fs
+++ b/src/QsFmt/Formatter/SyntaxTree/Statement.fs
@@ -50,6 +50,17 @@ module QubitBinding =
     let mapPrefix mapper (binding: QubitBinding) =
         { binding with Name = SymbolBinding.mapPrefix mapper binding.Name }
 
+type ForBinding =
+    {
+        Name: SymbolBinding
+        In: Terminal
+        Value: Expression
+    }
+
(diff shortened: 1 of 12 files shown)


src/QsFmt/Formatter/SyntaxTree/Statement.fs
```diff
@@ -50,6 +50,17 @@ module QubitBinding =
...
+type ForBinding =
+    {
+        Name: SymbolBinding
+        In: Terminal
+        Value: Expression
+    }
+
+module ForBinding =
+    let mapPrefix mapper (binding: ForBinding) =
+        { binding with Name = SymbolBinding.mapPrefix mapper binding.Name }
```
swernli, line 62: This probably already handles `_` as the symbol binding, but I'm not sure how to check. Can you confirm if that case is handled properly?

ScottCarda-MS: Sorry for the late response. I've just double-checked. The `_` symbol is treated as any other symbol binding, so there is no special logic needed for handling that case.

swernli (approved):

PR #1138, opened 2021-09-10
Adding new examples

troelsfr:
This PR adds two examples of random walk phase estimation. This PR depends on #1140


src/Passes/examples/RandomWalkPE2/QSharpVersion/Example.qs
```diff
@@ -0,0 +1,67 @@
+namespace Microsoft.Quantum.Qir.Emission {
```
swernli, line 1: Do we really need all three of these variants of RWPE? I think only 1 and 3 are generally relevant, this one is overly specific to just floating point ranges and rotation units without changing anything about the format or flow compared to version 3.

src/Passes/examples/RandomWalkPE3/QSharpVersion/Example.qs
```diff
@@ -0,0 +1,78 @@
...
+
+    function ClassicalComputeInput1(mu : Double, sigma : Double) : Double {
+        return (PI() / 2.0) - (mu / sigma);
+    }
+
+    function ClassicalComputeInput2(mu : Double, sigma : Double) : Double {
+        return PI() * (mu - PI() * sigma / 2.0);
+    }
+
+    // function ClassicalCompute(mu : Double, sigma : Double) : (Double, Double) {
```
swernli, line 33: These commented out functions were for my personal reference and don't need to be kept. Likewise with the calls to them below.

src/Passes/examples/RandomWalkPE2/QSharpVersion/Example.qs
troelsfr: Until we have more examples, I think it is good to have some variations of the same example. The main reason for this is that different ways of constructing Q# code results in different QIRs which will expose different issues. I cannot remember whether it was example 1 or 2, but one of them revealed a segfault the other did not. Hence, I would leave it for now with an expectation that in time we will only need a single version of random walk phase estimation.

troelsfr: Hmmm... I see your point: It is basically the exact same code except that you have manually const folded. I revert my standpoint and have deleted 2nd example.

src/Passes/examples/RandomWalkPE3/QSharpVersion/Example.qs
troelsfr: Removed.

swernli (approved):

Issue #1089, opened 2021-07-21
Reference and/or alias counting bug


bettinaheim:
The following example currently fails with either a "Attempting to resurrect a previously released object!" upon evaluation of _PrepareClassification(prep, model, _), or with a "Alias count cannot be negative" somewhere after a call to EstimateClassificationProbability but before finishing EstimateClassificationProbabilities:
https://github.com/microsoft/Quantum/tree/anpaz/training/samples/machine-learning/half-moons

bettinaheim:
Minimal repro:
```qsharp
    function Mapped<'T, 'U> (mapper : ('T -> 'U), array : 'T[]) : 'U[] {
        mutable resultArray = new 'U[Length(array)];

        for idxElement in IndexRange(array) {
            set resultArray w/= idxElement <- mapper(array[idxElement]);
        }

        return resultArray;
    }

    function Id<'T>(a : 'T) : 'T { return a; }

    @EntryPoint()
    operation RunExample() : String {

        let _ = Mapped(Id, ["hello"]);
        return "Executed successfully!";
    }
```

bettinaheim:
More concisely: 
For 
```qsharp
    function Id<'T>(a : 'T) : 'T { return a; }

    @EntryPoint()
    operation RunExample() : String {

        let id = Id<String[]>;
        let _ = id(["hello"]);
        return "Executed successfully!";
    }
```
The Id_wrapper accidentally releases the array returned from the Id, despite that is is effectively a return value (which is returned by storing it in the output tuple of the wrapper)...

bettinaheim:
Narrowed it down further:
```qsharp
    function WithProductKernel(sample : Double[]) : Double[] {
        //return sample; // works
        return sample + [1.]; // doesn't work...
    }

    @EntryPoint()
    operation RunExample() : String {
        let feature = WithProductKernel((Features())[0]);
        let sample = LabeledSample(feature, 1); 
        let labeled = _Features(sample);

        // commenting out this line avoids the issue
        let _ = ApproximateInputEncoder(1e-6, labeled);
        return "Executed successfully!";
    }
```

PR #1141, opened 2021-09-12
Fixing memory management issue

bettinaheim:
This PR fixes https://github.com/microsoft/qsharp-compiler/issues/1089. 
See the code comments and the issue above for more detail.

To properly validate the change, I set up execution tests for QIR bitcode. These test work via JIT compilation and should not require llvm or clang to be installed on the system. They did require me to create and use a runtime package with the necessary extern C exposed.


src/QsCompiler/QirGeneration/ScopeManager.cs
```diff
@@ -126,28 +126,106 @@ private static IEnumerable<(IValue, bool)> Expand(Func<ITypeRef, string?> getFun
...
             /// </summary>
-            public void RegisterVariable(string varName, IValue value)
+            public void RegisterVariable(string varName, IValue value, IValue? fromLocalId)
             {
                 this.variables.Add(varName, value);
 
-                // Since the value is necessarily created in the current or a parent scope,
-                // it won't go out of scope before the variable does.
-                // There is hence no need to increase the reference count if the variable is never rebound.
                 var change = new[] { (value, true) }; // true refers to whether the change also applies to inner items
```
troelsfr, line 192: The comment above refers to whether or not it is a shallow assignment?

src/QsCompiler/QirGeneration/ScopeManager.cs
```diff
@@ -126,28 +126,106 @@ private static IEnumerable<(IValue, bool)> Expand(Func<ITypeRef, string?> getFun
...
+                // to increase the reference count immediately when
+                // a) binding a value accessed via a local variable to a mutable variable, or
+                // b) assigning a value accessed via a mutable variable.
                 if (value is PointerValue)
                 {
-                    // If the variable can be rebound, however, then updating an item via a copy-and-reassign statement
-                    // potentially leads to the updated item(s) being unreferenced in an inner scope, i.e. before the
-                    // pending reference count increases of this scope are applied.
-                    // We hence need to make sure to increase the reference count immediately when binding to a mutable variable.
+                    // Assignment to a mutable variable:
```
troelsfr, line 210: It seems that being a `PointerValue` is equivalent to being a mutable variable. I think it would make it easier to understand how the comments relate to the code by changing the code to `if(IsMutable(value))` as this makes the meaning of the statement explicit or start by explaining that we consider a variable mutable if it is an instance of `PointerValue`. Note it is non-obvious that PointerValue is equivalent to mutability as it could be a pointer to a const value.

troelsfr (approved):
Disclaimer: I do not feel that I am the best person to review this PR as it assumes deep knowledge on how Q# handles variables and values with respect to reference and alias counting.

With that in mind, overall it looks OK to me. As it seems that there are tests written which could be part of this PR, I think it would be great to include those. Another thing I personally really like, would be two diagrams that maps out all the different paths through which a variable a may increase or decrease its reference count. I will leave it up to you whether you wish to include that.

Approved, with a recommendation of getting a second pair of eyes to look at it.

bettinaheim:
> Disclaimer: I do not feel that I am the best person to review this PR as it assumes deep knowledge on how Q# handles variables and values with respect to reference and alias counting.
> 
> With that in mind, overall it looks OK to me. As it seems that there are tests written which could be part of this PR, I think it would be great to include those. Another thing I personally really like, would be two diagrams that maps out all the different paths through which a variable a may increase or decrease its reference count. I will leave it up to you whether you wish to include that.
> 
> Approved, with a recommendation of getting a second pair of eyes to look at it.

Yes, as far as the tests are concerned, that was the plan and I've added execution tests for them. I believe in this case the most reliable way to test them is to check the execution rather than checking the abstraction/compiled code (also, that is much more feasible to maintain for the large set of tests that is required). I just needed to first set it up. 

As far as the diagram is concerned, I will defer the challenge of drawing a diagram in this case. Rather than with a diagram, I would in this case rather keep a description of the logic and also that requires quite a bit of explanation. The good news is that all of that logic is contained in the scope manager, and looking at the code there in detail should give sufficient information to follow the logic.

src/QsCompiler/QirGeneration/ScopeManager.cs
bettinaheim: No, good point - this applies to the entirety of the logic. I'll make it more apparent.

src/QsCompiler/QirGeneration/ScopeManager.cs
bettinaheim: Hm, the explanation is largely valid even considering pointers more broadly, or put differently: I believe if this logic was invoked whenever a pointer is updated/defined it would be correct (albeit less efficient due to unnecessary alias count changes). The only reason we have special handling here is that we do some trickery with in-place modifications, and this logic is hence implementing optimizations specifically knowing which pointer values may be accessed in the source code later on. However, it would be fair to store other pointers in the scope manager, too; if that is done then we simply have some unnecessary overhead for those. 
Across other pieces of the code we do have PointerValues besides for things stored in mutable variables, and I hence believe that adding a function in the scope manager that equates a PointerValue with it being a mutable variable would be misleading.

PR #1142, opened 2021-09-13
Fixing segfault for phi node replacement

troelsfr:
This PR fixes a segfault that occurs during phi node replacement.

This PR depends on #1138 and #1134.


src/Passes/examples/TeleportChain/Makefile
```diff
@@ -15,3 +15,5 @@ QSharpVersion/qir/Example.ll:
 clean:
 	cd QSharpVersion && make clean
 
+
+# --no-disable-reference-counting --no-disable-reference-counting --no-disable-alias-counting --no-disable-string-support --no-optimise-branch-quatum-one --no-optimise-branch-quatum-zero --no-use-static-qubit-array-allocation --no-use-static-qubit-allocation --no-use-static-result-allocation
```
troelsfr, line 19:
```suggestion
```

src/Passes/Source/RuleTransformationPass/Profile.cpp
```diff
@@ -442,14 +440,32 @@ namespace quantum
...
+                {
+                    auto& instr = **it2;
+                    llvm::errs() << " -- " << instr << "\n";
                     instr.replaceAllUsesWith(llvm::UndefValue::get(instr.getType()));
-                    to_delete.push_back(&instr);
+                    instr.eraseFromParent();
+                    // to_delete.push_back(&instr);
                 }
 
                 // Removing all block references
```
swernli, line 471: The line wasn't near enough to the ones that have been changed for me to comment on it directly, but it looks like the `replaceAllUsesWith` function for the blocks gets called twice. Since it is called on line 472, shouldn't it not need to be called on line 482? It seems like the block removal could be rearranged a bit to match the same pattern as the new instruction removal pattern you introduce here, but it's possible I'm missing a subtlety as to why it can't. If it can be refactored to match, then the `blocks_to_delete_` variable can be eliminated and the block erased from parent as soon as all uses are replaced.

src/Passes/Source/RuleTransformationPass/Profile.cpp
```diff
@@ -442,14 +440,32 @@ namespace quantum
...
+
+                // Removing all instructions
+                for (auto it2 = instructions.rbegin(); it2 != instructions.rend(); ++it2)
+                {
+                    auto& instr = **it2;
+                    llvm::errs() << " -- " << instr << "\n";
                     instr.replaceAllUsesWith(llvm::UndefValue::get(instr.getType()));
-                    to_delete.push_back(&instr);
+                    instr.eraseFromParent();
+                    // to_delete.push_back(&instr);
```
swernli, line 468: I think this commented line can be removed now, as the `to_delete` variable is no longer used anywhere else.

src/Passes/examples/Makefile
```diff
@@ -1,7 +1,15 @@
...
+	cd RandomWalkPE2 && make all && make clean
+	cd RandomWalkPE3 && make all && make clean
 
 clean:
-	cd TeleportChain && make clean
+	cd SimpleLoop && make clean	
+	cd TeleportChain && make clean	
 	cd LoopRecursion && make clean
+	cd RandomWalkPE1 && make clean
+	cd RandomWalkPE2 && make clean
```
swernli, line 14: looks like this may need a merge from the feature branch after #1138 merges since some of the examples have changed.

swernli (commented):
The updates for instruction removal look good. I'd like to re-review once the examples PR is merged, and I think it's worth considering simplifying the block removal to match the pattern you use for instruction removal if possible.

src/Passes/Source/RuleTransformationPass/Profile.cpp
troelsfr: The one called on 472 is for the block whereas here we are replacing the instruction. I am not sure that I follow the suggestion - may it was a result of the original PR looking cluttered due to unmerged dependencies? If not, could you elaborate a bit?

src/Passes/Source/RuleTransformationPass/Profile.cpp
troelsfr: Remove the comment

troelsfr:
> The updates for instruction removal look good. I'd like to re-review once the examples PR is merged, and I think it's worth considering simplifying the block removal to match the pattern you use for instruction removal if possible.

I've updated the PR and we are - please re-review.

swernli (approved):
Changes look good! One last question: was the checking in of the two binary files named `default.profraw` intentional? They don't seem to have any references.

PR #1143, opened 2021-09-13
Refactor to prepare for qubit allocation counting

troelsfr:
This refactor prepares the code base for qubit allocation counting. The main new thing is that it creates a `Profile` class that which holds all the LLVM logic to apply a profile to a IR.

## Dependencies
This PR depends on #1142 .

## Change log
- Refactoring profile logic into its own class `Profile`
- Moving ownership of allocation managers to profile
- Moving the config functions into cpp files
- Adding the possibility of annotating the entry function with required number of qubits

## Example output

The new option gives the possibility for the code to produce an output like:

```sh
define void @TeleportChain__QuantumProgram() local_unnamed_addr #0 {
entry:
  tail call void @__quantum__qis__h__body(%Qubit* nonnull inttoptr (i64 1 to %Qubit*))
  ; ...
}

attributes #0 = { "EntryPoint" "requiredQubits"="6" }
```
We note that attribute set #0 is extended with an extra attribute. This example can be reproduced using the LoopRecursion example.


src/Passes/examples/RandomWalkPE2/QSharpVersion/Example.qs.orig
```diff
@@ -0,0 +1,146 @@
+<<<<<<< HEAD
```
bettinaheim, line 1: Was this file committed accidentally? It looks like it is not in use, given that it still has the diff markers in it. Also, there is a bug fix to this particular example (time and theta were accidentally swapped).

src/Passes/Source/AllocationManager/AllocationManager.cpp
```diff
@@ -61,9 +63,21 @@ namespace quantum
...
 
-        mappings_.erase(it);
-        if (reuse_qubits_)
+        if (!reuse_qubits_)
+        {
+            mappings_.erase(it);
+        }
+        else
         {
+            if (it->size > registersInUse())
```
bettinaheim, line 72: Does the allocation manager rely on consecutive ordering of the qubit indices for qubits that are currently in use?

src/Passes/Source/Profile/Profile.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include "Profile/Profile.hpp"
+
+#include "Llvm/Llvm.hpp"
+
+namespace microsoft
+{
+namespace quantum
+{
+
+    Profile::Profile(
```
bettinaheim, line 13: Nice. I'll need to read the docs in detail to fully understand how all components are structured precisely and where the extensibility points are - but at first glance this looks like a good setup.

bettinaheim (approved):

src/Passes/Source/AllocationManager/AllocationManager.cpp
troelsfr: It does not rely on consecutive, but it operates optimally with FILO/stack-based qubit allocation. That said, if you have another scheme that you prefer, you can use the interface to write a new manager to replace this manager. We just need to add the external interface to adding that but that is relatively easy,

PR #1145, opened 2021-09-14
Add generateazurepayload command to standalone executables

cesarzc:
This change generates C# code needed to handle the he `generateazurepayload` command for Q# standalone executables.


cesarzc:
This should be merged at the same time than this PR in qsharp-runtime: https://github.com/microsoft/qsharp-runtime/pull/812

kuzminrobin (approved):

PR #1148, opened 2021-09-15
Use only annotated type ranges in `IdentifierReferences` transformation

samarsha:
In this example,

```qsharp
newtype Foo = Unit;

function MappedByIndex(mapper : (Int, Foo) -> (Int, Foo)) : (Int, Foo)[] {
    return [mapper(0, Foo())];
}
```

clicking `Foo` in the editor highlights all occurrences of `Foo`, but also `mapper`, which has an inferred type that includes `Foo`. The `OnType` handler in `IdentifierReferences` should be changed to only look at annotated type ranges (not inferred type ranges).

### DIFF
--- a/src/QsCompiler/Transformations/SearchAndReplace.cs
+++ b/src/QsCompiler/Transformations/SearchAndReplace.cs
@@ -233,16 +233,17 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.SearchAndReplace
 
             public override ResolvedType OnType(ResolvedType type)
             {
-                switch (type.Resolution)
+                var id = type.Resolution switch
                 {
-                    case QsTypeKind.UserDefinedType { Item: var udt }:
-                        var id = Identifier.NewGlobalCallable(new QsQualifiedName(udt.Namespace, udt.Name));
-                        this.SharedState.LogIdentifierLocation(id, TypeRangeModule.TryRange(type.Range));
-                        break;
-                    case QsTypeKind.TypeParameter _:
-                        id = Identifier.NewLocalVariable(SyntaxTreeToQsharp.Default.ToCode(type) ?? "");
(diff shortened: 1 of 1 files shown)


src/QsCompiler/Transformations/SearchAndReplace.cs
```diff
@@ -233,16 +233,17 @@ public TypeTransformation(SyntaxTreeTransformation<TransformationState> parent)
...
-                        id = Identifier.NewLocalVariable(SyntaxTreeToQsharp.Default.ToCode(type) ?? "");
-                        this.SharedState.LogIdentifierLocation(id, TypeRangeModule.TryRange(type.Range));
-                        break;
+                    QsTypeKind.UserDefinedType udt =>
+                        Identifier.NewGlobalCallable(new QsQualifiedName(udt.Item.Namespace, udt.Item.Name)),
+                    QsTypeKind.TypeParameter _ => Identifier.NewLocalVariable(SyntaxTreeToQsharp.Default.ToCode(type)),
+                    _ => null,
+                };
+
+                if (!(id is null) && type.Range is TypeRange.Annotated annotatedRange)
```
idavis, line 244: Would `(id is object) && ...` or `(id is not null) && ...` work here to avoid the negation?

In C# 7 & 8 we can use `value is object` and in C# 9: `value is not null` to do not null checks.

idavis (approved):

src/QsCompiler/Transformations/SearchAndReplace.cs
samarsha: Yes, I'd use `id is not null` in C# 9, but the QDK can only build with C# 8 right now. I prefer `!(id is null)` over `id is object` because I feel that it expresses my intent better. (I care that it's not null, but I don't care that it's a subtype of `object`.)

Issue #1152, opened 2021-09-16
QIR generation can fail when there are blocks with pending reference counts that terminate in being unreachable


bettinaheim:
The following code causes a failure "cannot close scope that has pending calls to increase reference counts" upon QIR generation:

```qsharp
    @EntryPoint()
    operation TestInterpolatedStrings() : String {

        let arr1 = [1,2,3];
        let arr2 = ["1","2","3"];
        let arr3 = ["","2","","4"];
        let tuple1 = (1,(2,3));
        let tuple2 = ("1",("2","3"));
        let tuple3 = ("1",("","3"));
        let res = One;
        let pauli = PauliX;
        let pauliArr = [PauliY, PauliI];
        let range = 0..-1..0;
        let (fct, op) = (PrintSection, RunExample);
        let udt = Tuple("Hello", "World");
        
        Message("simple string");
        Message($"{"interpolated string"}");
        // Todo: bigint is not yet supported in the runtime
        Message($"{true} or {false}");
        Message($"{1}, {-1}, {1 - 1}");
        Message($"{1.}, {2.0}, {1e5}, {.1}, {-1.}, {1. - 1.}");
        Message($"{Zero}, {res}");
        Message($"{PauliZ}, {pauli}, {pauliArr[0]}, {pauliArr[1...]}");
        Message($"{1..3}, {3..-1..1}, {range}");
        Message($"{[1,2,3]}, {["1","2","3"]}, {arr1}, {arr2}, {arr3}");
        Message($"{()}, {(1,(2,3))}, {("1",("2","3"))}, {tuple1}, {tuple2}, {tuple3}");
        
        use (q, qs) = (Qubit(), Qubit[3]);
        Message($"{q}, {qs}");
        Message($"{PrintSection}, {RunExample}, {(fct, op)}");
        Message($"{udt::Item1}, {Foo(1)}, {udt}");
        return "All good!";
    }
```

Issue #1151, opened 2021-09-16
String arrays are not properly printed


bettinaheim:
The array `["", "hello"]` is printed as `[hello]`. 
The issue is likely caused by missing quotation marks around interpolation arguments if those arguments are of type string.

Issue #1096, opened 2021-07-26
Generate less redundant string declarations


bettinaheim:
The QIR generation currently adds a global constant for each used string, and in particular also the constants for empty string, command and parentheses repeatedly. It would be nicer to avoid this redundancy.

Issue #1086, opened 2021-07-21
Bool to string causes crash when conditional expression evaluates to true


bettinaheim:
Works:
```qsharp
        let res = One;
        let overflow = res == Zero ? true | false;
        Message($"{overflow}");
```
Doesn't work (crashes):
```qsharp
        let res = Zero;
        let overflow = res == Zero ? true | false;
        Message($"{overflow}");
```
Crash:
```
Unhandled exception at 0x00007FFB98E6FA80 (Microsoft.Quantum.Qir.Runtime.dll) in BoolToStrCrash.exe: 
0xC0000005: Access violation reading location 0xFFFFFFFFFFFFFFFF.
```

To reproduce:
Create a new folder, create a Test.csproj with the following content:
```
<Project Sdk="Microsoft.Quantum.Sdk/0.18.2109162713">

  <PropertyGroup>
    <QscVerbosity>Detailed</QscVerbosity>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp3.1</TargetFramework>
    <QirGeneration>true</QirGeneration>
  </PropertyGroup>

  <ItemGroup>
    <None Include="qir\**" />
  </ItemGroup>

  <Target Name="CleanAll" BeforeTargets="QSharpClean">
    <ItemGroup>
      <_FilesToClean Include="$(GeneratedFilesOutputPath)**" />
      <_FilesToClean Include="$(QirOutputPath)**/*.ll" />
      <Compile Remove="$(_FilesToClean)**/*.g.cs" />
    </ItemGroup>
    <Delete Files="@(_FilesToClean)" />
  </Target>

</Project>
```
and a Program.qs file with the following content:
```qsharp
namespace Microsoft.Quantum.Qir.Emission {
    open Microsoft.Quantum.Intrinsic;

    @EntryPoint()
    operation RunExample() : Unit {

        let res = Zero;
        let overflow = res == Zero ? true | false;
        Message($"{overflow}");
    }
}
```
From within that folder, run `dotnet build` to get the .ll in qir/Test.ll that will cause the crash if using the current QIR runtime version:
```
define internal void @Microsoft__Quantum__Qir__Emission__RunExample__body() {
entry:
  %res = call %Result* @__quantum__rt__result_get_zero()
  %0 = call %Result* @__quantum__rt__result_get_zero()
  %1 = call i1 @__quantum__rt__result_equal(%Result* %res, %Result* %0)
[...]

PR #1153, opened 2021-09-16
QIR generation bug fixes

bettinaheim:
This PR fixes https://github.com/microsoft/qsharp-compiler/issues/1152, https://github.com/microsoft/qsharp-compiler/issues/1151, https://github.com/microsoft/qsharp-compiler/issues/1086, https://github.com/microsoft/qsharp-compiler/issues/1159, https://github.com/microsoft/qsharp-compiler/issues/1160, https://github.com/microsoft/qsharp-compiler/issues/1161, https://github.com/microsoft/qsharp-compiler/issues/1162, https://github.com/microsoft/qsharp-compiler/issues/1164, and https://github.com/microsoft/qsharp-compiler/issues/1096.

This PR also removes the use of __quantum__rt__pauli_to_string, and adjust the signature of __quantum__rt__array_slice_1d to match the runtime implementation (which deviates from spec - this will be remedied shortly).


src/QsCompiler/QirGeneration/Interop.cs
```diff
@@ -539,13 +539,8 @@ private IrFunction GenerateEntryPoint(string entryPointName, ArgumentTuple argum
 
                 // print the return value
                 var message = this.sharedState.GetOrCreateRuntimeFunction(RuntimeLibrary.Message);
-                var outputStr = QirExpressionKindTransformation.CreateStringLiteral(
-                    this.sharedState,
-                    result.QSharpType.Resolution.IsString ? "\"{0}\"" : "{0}",
-                    result);
+                var outputStr = QirExpressionKindTransformation.CreateStringLiteral(this.sharedState, "{0}", result);
```
cesarzc, line 542: Before this change, if the type was string, it was wrapped around additional quotation marks. Why is it no longer the case?

src/QsCompiler/QirGeneration/QIR/RuntimeLibrary.cs
```diff
@@ -27,13 +27,11 @@ public static class RuntimeLibrary
...
         // to-string
         public const string BigIntToString = "bigint_to_string";
-        public const string BoolToString = "bool_to_string";
         public const string DoubleToString = "double_to_string";
         public const string IntToString = "int_to_string";
-        public const string PauliToString = "pauli_to_string";
         public const string QubitToString = "qubit_to_string";
-        public const string RangeToString = "range_to_string";
-        public const string ResultToString = "result_to_string";
+        public const string RangeToString = "range_to_string"; // todo: eliminate
```
cesarzc, line 33: Do these `todo`s need to be addressed before merging this case? If they will be kept for this change, it would be good to comment on when/why the strings need to be eliminated.

src/QsCompiler/QirGeneration/Interop.cs
bettinaheim: With the latest change, interpolation arguments of type string are wrapped inside quotation marks, hence there is no need to additionally wrap it once more.

src/QsCompiler/QirGeneration/QIR/RuntimeLibrary.cs
bettinaheim: No, this is a todo related to a spec update - or rather, it is possible (and likely reasonable) to not relay on the runtime for this functionality, but also lower priority compared to the general fixes in this PR.

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
```diff
@@ -262,55 +267,91 @@ void StoreElement(PointerValue pointer, IValue value, string? fromLocalId, bool
...
+
                     // do not increase the ref count here - we will increase the ref count of all new items at the end
                     IValue newItemValue = sharedState.EvaluateSubexpression(updated);
                     var index = sharedState.EvaluateSubexpression(accEx);
-                    UpdateElement(_ => newItemValue, index.Value);
+                    UpdateElement(() => newItemValue, array.GetArrayElementPointer(index.Value));
                 }
                 else if (accEx.ResolvedType.Resolution.IsRange)
                 {
+                    // fixme: if fromLocalId equals the lhs then we need to make a deep copy of updated ...
```
cesarzc, line 306: Should this fixme be tracked as an issue?

bettinaheim: This comment is just a remainder of my notes, and I have revised the comment here to be a description of what is done.

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
```diff
@@ -584,16 +658,21 @@ Value ArrayToString(ArrayValue array)
                 var ty = evaluated.QSharpType.Resolution;
                 if (ty.IsString)
                 {
-                    UpdateStringRefCount(evaluated.Value, 1);
-                    return evaluated.Value;
+                    Value quote = CreateConstantString("\"");
```
cesarzc, line 661: Nit: Is there a reason why `quote` was created using the `Value` type instead of `var`? I see that the `comma` variable a few lines above uses the same pattern.

bettinaheim: I think this might have been left from an earlier stage of the code where the explicit Value was needed for the nullcheck to be happy. I'll change it to var.

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
```diff
@@ -605,7 +684,29 @@ Value ArrayToString(ArrayValue array)
                 }
                 else if (ty.IsPauli)
                 {
-                    return SimpleToString(RuntimeLibrary.PauliToString);
+                    Value LoadPauli(QsPauli pauli)
```
cesarzc, line 687: It seems like a rather complicated change to avoid calling `PauliToString`. Why the `Pauli` type needs to be handled this way?

bettinaheim: There are a couple of general considerations:
1) Generally, I think it is good to trim down what the runtime needs to implement to keep things simple for backends. 
2) If the runtime implements it, then the compiled code will break down into something similar, so there is no advantage or disadvantage really one way or the other after all is compiled.
3) The disadvantage of implementing it in the compiler is a little bit more code complexity compared to implementing it in the runtime (the code for us to maintain, not the code that a Q# project compiles into).

What the reason was for the immediate change is that with the current implementation of the runtime, there can be issues with bool and Pauli, since these don't match the bitsize used in the IR. This is only relevant in edge cases, and the proper solution is to revise the API to avoid such incompatibilities. We will do this as part of moving to a C-API. In the meantime, avoiding to use things we know can cause issues (for edge cases, most of the time the runtime functions work fine as is) is a mitigation.

src/QsCompiler/Tests.Compiler/TestCases/QirTests/ExecutionTests.qs
```diff
@@ -1048,11 +1614,59 @@
...
+    operation ReturnsString() : String {
+        return "Success!";
+    }
+
     @EntryPoint()
     operation RunExample() : Unit {
 
         PrintSection(1, "");
-
+
```
cesarzc, line 1665: Nit: it seems like the editor might have added a few space characters.

cesarzc (approved):

src/QsCompiler/Tests.Compiler/TestCases/QirTests/ExecutionTests.qs
bettinaheim: I'll wait to have a formatter to remove these automatically. :) They occur when editing multiple lines, e.g. commenting them out.

Issue #1155, opened 2021-09-16
Formatter Takes Directories as Inputs


ScottCarda-MS:
The formatter should be able to take directories as input paths and process all .qs files in the given directories. Directories should be able to be given along side files as inputs (see multiple inputs #1154). A flag should be added to the CLI that allows directories to be processed recursively, processing all .qs files in subdirectories.

Issue #1154, opened 2021-09-16
Formatter Should Take Multiple Inputs


ScottCarda-MS:
The formatter should be able to process multiple file inputs with one use of the tool. The CLI should reflect this by allowing multiple files to be passed in as inputs.

PR #1156, opened 2021-09-16
Q# Command Line Process Multiple Files and Directories

ScottCarda-MS:
Allows multiple inputs to be passed to the CLI. Inputs can be any mix of files, folders, and the `-` for specifying standard input. Folders are resolved to all the `.qs` files  directly under them. The `--recurse` or `-r` flag can be specified to resolve folders recursively, processing all of their sub-directories and their sub-directories and so on.

Addresses #1154 and #1155.


swernli:
It looks like the build is failing due to an extra newline that snuck into the formatted help output.

Also, the original PR comment at the top referred to #155 but it looks like that was meant to be #1155, so I took the liberty of updating it!

src/QsFmt/App.Tests/Tests.fs
```diff
@@ -129,3 +144,176 @@ let ``Shows file not found error`` path =
...
+    function Bar() : Int {
+        return 0;
+    }
+}
+"
+
+    let result = run [| "format"; "Example1.qs"; "-"; "NotFound.qs"; "Example2.qs" |] (input |> standardizeNewLines)
+    Assert.Equal(3, result.Code)
+    Assert.Equal(output |> standardizeNewLines, result.Out)
+    Assert.NotEmpty(result.Error)
```
swernli, line 319: Is it worth asserting that a specific error is thrown here rather than just any error?

src/QsFmt/App/Program.fsi
```diff
@@ -7,3 +7,4 @@ module Microsoft.Quantum.QsFmt.App.Program
 /// Runs the Q# formatter.
 [<CompiledName "Main">]
 val main: string [] -> int
+
```
swernli, line 10: Is this newline necessary for correctness?

swernli (approved):
Signed off with minor comments

ScottCarda-MS:
> It looks like the build is failing due to an extra newline that snuck into the formatted help output.
> 
> Also, the original PR comment at the top referred to #155 but it looks like that was meant to be #1155, so I took the liberty of updating it!

Thanks, that's what I get for trying to do a last minute change, haha. It is fixed now.

src/QsFmt/App.Tests/Tests.fs
ScottCarda-MS: This would be a 'file not found' error, and those aren't controlled or maintained by our code. The specific wording might even be dependent on the OS you are running. For that reason, trying to verify that a 'file not found' error is thrown would create a dependency on the test being run in an environment we predict. Just didn't seem that worth it to me to go through that trouble.

src/QsFmt/App/Program.fsi
ScottCarda-MS: Nope, not sure how that got in there. It is removed now.

src/QsFmt/App.Tests/Tests.fs
swernli: Makes sense!

PR #1165, opened 2021-09-22
Formatter - Output to Files

ScottCarda-MS:
Input files are now overwritten instead of having the output go to the standard out. Backup files can be made if the `--backup` flag is used.


src/QsFmt/App.Tests/Tests.fs
```diff
@@ -148,173 +176,126 @@ let ``Shows file not found error`` path =
...
+            {
+                Code = 0
+                Out = StandardInputTest.Updated
+                Error = ""
+            },
+            result
+        )
+
+        for file in files do
+            Assert.True(File.Exists(file.Path + "~"))
```
swernli, line 269: This pattern of adding "~" to the path repeats enough that it might be worth adding that as a function, which will also have the advantage that if we update the backup file naming pattern later there will be one place to change. It's only in this one test so far, so it's not a big deal at the moment and it's up to you if you want to refactor that now or leave as-is for the time being.

src/QsFmt/App/Program.fsi
```diff
@@ -7,3 +7,4 @@ module Microsoft.Quantum.QsFmt.App.Program
 /// Runs the Q# formatter.
 [<CompiledName "Main">]
 val main: string [] -> int
+
```
swernli, line 10: Huh... this extra newline snuck back in again. I wonder if it's a formatting tool making this change?

src/QsFmt/Formatter.Tests/Formatter.Tests.fsproj
```diff
@@ -10,6 +10,7 @@
       <Link>TestCases\%(FileName)%(Extension)</Link>
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+    <Compile Include="Tests.fs" />
```
swernli, line 13: I'm a little surprised this is necessary... doesn't an fsproj automatically find any .fs files in the same directory tree?

swernli (approved):
Looks good, a few minor comments!

src/QsFmt/Formatter.Tests/Formatter.Tests.fsproj
ScottCarda-MS: It is necessary, and F# projects don't automatically include files. F# files are compiled in the order they appear here in the project file (usually reflected in the solution explorer in the IDE). If you change the precedence of the files in an F# project, you can get different behavior, although usually you will just get errors. Since this order should be based on the logic and not on the alphabetical ordering of the file names, the order is not assumed from the directory structure, and so files can't be automatically included as it wouldn't know where to put the files in the order.

src/QsFmt/App/Program.fsi
ScottCarda-MS: Weirdly persistent little new line. I've removed it again.

src/QsFmt/App.Tests/Tests.fs
ScottCarda-MS: Hmm, it is hard to consolidate logic that is referenced so sparsely. Outside of this test, the tilde only comes up once in the Program file. I've consolidated the logic a bit in the test, but I'll leave the rest alone. Future use cases would better inform how that bridge should be crossed.

PR #1166, opened 2021-09-23
Allow ProjectReference as Target Package

swernli:
Previously only a `PackageReference` would be recognized as a target package. This allows `ProjectReference` to be used as a target package, and still enforces that only one such reference exists.

### DIFF
--- a/src/QuantumSdk/DefaultItems/DefaultItems.targets
+++ b/src/QuantumSdk/DefaultItems/DefaultItems.targets
@@ -97,14 +97,17 @@
   <!-- Computes the ResolvedTargetSpecificDecompositions property. Generates a build error if more than one target package has been specified. -->
   <Target Name="ResolveTargetPackage" DependsOnTargets="ResolveProjectReferences;ResolveAssemblyReferences" BeforeTargets="BeforeQSharpCompile">
     <ItemGroup>
+      <ResolvedTargetSpecificDecompositions Include="@(_ResolvedProjectReferencePaths)" Condition="@(_ResolvedProjectReferencePaths->Count()) &gt; 0 And %(_ResolvedProjectReferencePaths.IsTargetPackage)" />
+    </ItemGroup>
+    <ItemGroup>
       <_TargetPackageReference Include="@(PackageReference)" Condition="@(PackageReference->Count()) &gt; 0 And %(PackageReference.IsTargetPackage)" />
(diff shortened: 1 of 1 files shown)


src/QuantumSdk/DefaultItems/DefaultItems.targets
```diff
@@ -96,15 +96,18 @@
...
+      <ResolvedTargetSpecificDecompositions Include="@(_ResolvedProjectReferencePaths)" Condition="@(_ResolvedProjectReferencePaths->Count()) &gt; 0 And %(_ResolvedProjectReferencePaths.IsTargetPackage)" />
+    </ItemGroup>
     <ItemGroup>
       <_TargetPackageReference Include="@(PackageReference)" Condition="@(PackageReference->Count()) &gt; 0 And %(PackageReference.IsTargetPackage)" />
       <_TargetPackageReferencePathProperty Include="@(_TargetPackageReference->'QscRef_$([System.String]::Copy('%(_TargetPackageReference.Identity)').Replace('.','_'))')" />
       <_ResolvedTargetPackageReference Include="$(%(_TargetPackageReferencePathProperty.Identity))" />
       <ResolvedTargetSpecificDecompositions Include="@(_ResolvedTargetPackageReference->Split(`;`))" />
     </ItemGroup>
-    <Error Condition="@(_TargetPackageReference->Count()) &gt; 1" Text="More than one target package was defined. The defined target package needs to be unique." />
+    <Error Condition="@(ResolvedTargetSpecificDecompositions->Count()) &gt; 1" Text="More than one target package was defined. The defined target package needs to be unique." />
```
ScottCarda-MS, line 108: Before, it seems like it was counting all target package references. After this change, it looks like it will only be counting reference paths that are target packages. This would seem to exclude any target package reference that is not a resolved project reference path, whereas before those would have been included.

swernli: `ResolvedTargetSpecificDecompositions` gets the union of any package references that have `IsTargetPackage` set to true and now any project references that have it set to true. So for example, I tried this kind of set up:
```xml
  <ItemGroup>
    <ProjectReference Include="..\..\MyCustomTarget\CustomTargetPackage.csproj" IsTargetPackage="true" />
    <PackageReference Include="Microsoft.Quantum.Type2.Core" Version="0.18.2109164716-beta" IsTargetPackage="true" />
  </ItemGroup>
```
And it got correctly rejected by this check. Since lines 104-106 don't have any conditionals, it will include any package that was linked this way that defines the `QsRef` property correctly.

ScottCarda-MS (approved):

PR #1171, opened 2021-09-27
added MSBuild_Logs to gitignore

filipw:
There is a possible new `MSBuild_Logs` generated under the working directory (see https://github.com/dotnet/msbuild/commit/cdb5077c451180ab38161e0b5e70f5448e70355b) when MSBuild crashes.

This should not happen often but it showed up for me in three places when working with this repo, so I thought it would make sense to extend the default gitignore

Bildschirmfoto 2021-09-27 um 21 22 57


bettinaheim (approved):
Thanks!

PR #1174, opened 2021-09-28
Only re-indent if closing brace occurs in indentation region

### LINKED ISSUE #1163: Behavior for closing parenthesis in VS is incorrect
The behavior for closing parenthesis in general does no work properly when closing parenthesis on the same line. 
In particular, the example
```qsharp
        set arr = cond
            ? [[0], size = 5] w/ 4..-2..-1 <- [[] // close the missing parenthesis here
            | value w/ 1 <- [];
```
results in the code being replaced by
```qsharp
        set arr = cond
        0], size = 5] w/ 4..-2..-1 <- [[]] // after closing the parenthesis
            | value w/ 1 <- [];
```
Worse, the edit cannot even be undone without also removing the closing parenthesis.

samarsha:
Closes #1163.


src/VisualStudioExtension/QSharpVsix/QsSmartIndent.cs
```diff
@@ -125,7 +126,13 @@ private string CreateIndentation(int indent)
...
         /// Returns true if the character is an indentation character (a space or a tab).
         /// </summary>
-        private static bool IsIndentation(char c) => c == ' ' || c == '\t';
+        private static bool IsIndentationChar(char c) => c == ' ' || c == '\t';
+
+        /// <summary>
+        /// Returns true if the column occurs within the indentation region at the beginning of the line.
+        /// </summary>
+        private static bool IsInIndentation(ITextSnapshotLine line, int column) =>
+            line.GetText().Substring(0, column).All(IsIndentationChar);
```
idavis, line 135: Do `ITextSnapshotLine`s have carriage returns or line feeds? If so `char.IsWhiteSpace` might be a better fit for `IsIndentationChar`

samarsha: `GetText` excludes line breaks. I didn't use `char.IsWhiteSpace` originally since I wasn't sure if I really wanted all Unicode whitespace characters to be counted as indentation... but I think it's probably fine, so I removed `IsInIndentation`.

PR #1176, opened 2021-09-29
Minor changes for passes to build on Linux

bettinaheim:


swernli:
Is this still needed after #1173? After that PR, the pipeline correctly builds and tests on Linux, and while this specific change was not included it doesn't seem to be needed for the compilation to succeed.

bettinaheim:
Ah - great; I overlooked that one. Looks like both change are included in that PR.

PR #1179, opened 2021-10-01
Support `select` measurement transformation

swernli:
This adds simple support for LLVM `select` instruction measurement use transformation. LLVM inlining can at times take a branch instruction and turn it into a select instruction, which means it would no longer match the branch rule and end up leaving behind calls to `__quantum__rt__result_equal`. This treats select like branch such that select on result value `One` will be turned into an appropriate use of `__quantum__rt__read_result`.


src/Passes/Source/Rules/FactoryConfig.hpp
```diff
@@ -138,8 +143,9 @@ namespace quantum
 
         /// Optimisations
         /// @{
-        bool optimise_branch_quatum_one_{true};
-        bool optimise_branch_quatum_zero_{true};
+        bool optimise_branch_quantum_one_{true};
+        bool optimise_branch_quantum_zero_{true};
+        bool optimise_select_quatnum_zero_{true};
```
ScottCarda-MS, line 148: quatum

ScottCarda-MS: should be `optimise_select_quantum_zero_` instead of `optimise_select_quatum_zero_`.

src/Passes/Source/Rules/Notation/Select.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include "Llvm/Llvm.hpp"
+
+#include <unordered_map>
+#include <vector>
+
+namespace microsoft
+{
+namespace quantum
+{
+    namespace notation
```
kuzminrobin, line 18: So far (in qsharp-runtime) all the namespace names were capitalized (**M**icrosoft, **Q**uantum).

If this is C++17 or newer then I'd suggest
```suggestion
namespace Microsoft::Quantum::Notation
{
```

src/Passes/Source/Rules/Tests/Unit/SelectOnOne.cpp
```diff
@@ -0,0 +1,86 @@
...
+    ir_manip->declareFunction("void @__quantum__qis__reset__body(%Qubit*)");
+    ir_manip->declareFunction("%Result* @__quantum__rt__result_get_one()");
+    ir_manip->declareFunction("i1 @__quantum__rt__result_equal(%Result*, %Result*)");
+    ir_manip->declareFunction("void @__quantum__qis__mz__body(%Qubit*, %Result*)");
+
+    if (!ir_manip->fromBodyString(script))
+    {
+        llvm::outs() << ir_manip->getErrorMessage() << "\n";
+        exit(-1);
+    }
```
kuzminrobin, line 35: Looks like `ir_manip` is filled in with hard-coded data. The arg `script` is too. The failure of the call `ir_manip->fromBodyString(script)` feels like a detection of the internal bug in the code (not a detection of the error in the user-provided data). If that's always the case then I'd suggest:
```suggestion
    assert(ir_manip->fromBodyString(script));
```

src/Passes/Source/Rules/Tests/Unit/SelectOnOne.cpp
```diff
@@ -0,0 +1,86 @@
...
+        auto factory = RuleFactory(rule_set, BasicAllocationManager::createNew(), BasicAllocationManager::createNew());
+        // factory.useStaticResultAllocation();
+
+        factory.optimiseBranchQuantumOne();
+    };
+
+    auto profile = std::make_shared<DefaultProfileGenerator>(std::move(configure_profile));
+    ir_manip->applyProfile(profile);
+
+    // This optimistation is specific to the the __quantum__qir__read_result which
```
kuzminrobin, line 65:
```suggestion
    // This optimization is specific to the the __quantum__qir__read_result which
```

src/Passes/Source/Rules/Factory.cpp
```diff
@@ -502,6 +507,82 @@ namespace quantum
...
+            auto                      module   = orig_instr->getModule();
+            auto                      function = module->getFunction("__quantum__qir__read_result");
+            std::vector<llvm::Value*> arguments;
+            arguments.push_back(result);
+
+            if (!function)
+            {
+                std::vector<llvm::Type*> types;
+                types.resize(arguments.size());
+                for (uint64_t i = 0; i < types.size(); ++i)
```
kuzminrobin, line 531: Nit: I'd suggest
```suggestion
                for (auto i = 0; i < types.size(); ++i)
```
or
```suggestion
                for (size_t i = 0; i < types.size(); ++i)
```

src/Passes/Source/Rules/Factory.cpp
```diff
@@ -502,6 +507,82 @@ namespace quantum
...
+                function                     = llvm::Function::Create(
+                    fnc_type, llvm::Function::ExternalLinkage, "__quantum__qir__read_result", module);
+            }
+
+            builder.SetInsertPoint(llvm::dyn_cast<llvm::Instruction>(val));
+            auto new_call = builder.CreateCall(function, arguments);
+            if (cond == nullptr)
+            {
+                return false;
+            }
```
kuzminrobin, line 548: Nit: 
Feels like this `if` can be moved all the way up, and placed right after the line 514. In that case we can avoid calculating `new_call`, `function`, `arguments`, and may be even `orig_instr`.

If line 514 and this fragment are placed right after the line 512, then we can avoid calculating `result` too.

src/Passes/Source/Rules/Factory.cpp
```diff
@@ -502,6 +507,82 @@ namespace quantum
...
+            auto cond   = llvm::dyn_cast<llvm::Instruction>(cap["cond"]);
+            // Replacing result
+            auto orig_instr = llvm::dyn_cast<llvm::Instruction>(val);
+            if (orig_instr == nullptr)
+            {
+                return false;
+            }
+
+            auto                      module   = orig_instr->getModule();
+            auto                      function = module->getFunction("__quantum__qir__read_result");
```
kuzminrobin, line 523: After this line there is a risk (in the future) of name conflict between variable `function` and [`std::function`](https://cplusplus.com/reference/functional/function/).

src/Passes/Source/Rules/Factory.cpp
```diff
@@ -502,6 +507,82 @@ namespace quantum
...
+                user->setOperand(use.getOperandNo(), new_call);
+            }
+            cond->replaceAllUsesWith(new_call);
+
+            // Deleting the previous condition and function to fetch one
+            replacements.push_back({cond, nullptr});
+            replacements.push_back({cap["one"], nullptr});
+
+            return false;
+        };
```
kuzminrobin, line 564: In all the cases the lambda returns `false`?

src/Passes/Source/Rules/FactoryConfig.hpp
```diff
@@ -27,11 +27,11 @@ namespace quantum
             config.addParameter(
                 disable_string_support_, "disable-string-support", "Disables string support by instruction removal.");
             config.addParameter(
-                optimise_branch_quatum_one_, "optimise-branch-quatum-one",
+                optimise_branch_quantum_one_, "optimise-branch-quatum-one",
```
kuzminrobin, line 30:
```suggestion
                optimise_branch_quantum_one_, "optimise-branch-quantum-one",
```

src/Passes/Source/Rules/FactoryConfig.hpp
```diff
@@ -27,11 +27,11 @@ namespace quantum
...
                 "Maps branching based on quantum measurements compared to one to base profile "
                 "type measurement.");
             config.addParameter(
-                optimise_branch_quatum_zero_, "optimise-branch-quatum-zero",
+                optimise_branch_quantum_zero_, "optimise-branch-quatum-zero",
```
kuzminrobin, line 34:
```suggestion
                optimise_branch_quantum_zero_, "optimise-branch-quantum-zero",
```

src/Passes/Source/Rules/Factory.cpp
```diff
@@ -502,6 +507,82 @@ namespace quantum
...
+          %1 = call %Result* @__quantum__rt__result_get_one()
+          %2 = call i1 @__quantum__rt__result_equal(%Result* %0, %Result* %1)
+          ...
+          %5 = select i1 %2, <type> %3, <type> %4
+        */
+
+        // Variations of get_one
+        auto get_one = call("__quantum__rt__result_get_one");
+        addRule(
+            {select("cond"_cap = call("__quantum__rt__result_equal", "result"_cap = _, "one"_cap = get_one), _, _),
```
kuzminrobin, line 578: Just curiosity (feel free to ignore):
What is `"cond"_cap`, `"result"_cap`?  (I have never met such C++ constructs before)
And what is `_` in the fragment `"result"_cap = _, "one"_cap = get_one), _, _`?

src/Passes/Source/Rules/Factory.hpp
```diff
@@ -69,6 +69,7 @@ namespace quantum
         /// @{
         void optimiseBranchQuantumOne();
         void optimiseBranchQuantumZero();
+        void optimiseSelectQuantumOne();
         /// @}
```
kuzminrobin, line 73: Nit:
The mismatch between the file name (Factory.hpp) and the class name (**Rule**Factory), defined in the that file, is suspicious.

src/Passes/Source/Rules/FactoryConfig.hpp
```diff
@@ -76,12 +76,17 @@ namespace quantum
...
         bool optimiseBranchQuantumZero() const
         {
-            return optimise_branch_quatum_zero_;
+            return optimise_branch_quantum_zero_;
+        }
+
+        bool optimiseSelectQuantumOne() const
+        {
+            return optimise_select_quantum_zero_;
         }
```
kuzminrobin, line 90: Suspicious function:
The function name ends with `One` but the returned value name ends with `zero`. Did you mean what is below?
```suggestion
        bool optimiseSelectQuantumOne() const
        {
            return optimise_select_quantum_one_;
        }
```

kuzminrobin:
I'm done with the review round.

swernli:
Thanks @kuzminrobin! I'll work through the feedback later; for now I'm just trying to make sure the new unit test actually passes...

swernli:
Finally got the passes CI working, but the component governance check is still failing. I think the merge from main in #1181 should resolve that.

src/Passes/Source/Rules/Factory.cpp
troelsfr: No that should not be an issue given that the cpp file does not use namespace std.

src/Passes/Source/Rules/Factory.cpp
troelsfr: That is a good point. The function should return `true` here to indicate that the replacement was successful.

src/Passes/Source/Rules/Factory.cpp
troelsfr: These are user literals to allow for a custom pattern matching language using native C++. All `"xxx"_cap` states is: If there is a match of the right hand side, store the match in the variable name `xxx`. The variable `_` is a shorthand notation for pattern.

src/Passes/Source/Rules/Factory.hpp
```diff
@@ -69,6 +69,7 @@ namespace quantum
         /// @{
         void optimiseBranchQuantumOne();
         void optimiseBranchQuantumZero();
+        void optimiseSelectQuantumOne();
```
troelsfr, line 72: Is it worthwhile to implement `selectQauntumZero` as well?

troelsfr (approved):
Overall, this PR looks good to me. There is one minor detail detected by @kuzminrobin with a function that returns false where it should return true, but other than that. I've left a few comments for you to optionally consider.

PR #1184, opened 2021-10-05
QAT interface update

troelsfr:
This PR increases the customisability of QAT allowing users to deactivate various instances of the standard LLVM pipelines.

## Dependencies
This PR depends on #1175 .

## Changelog
- Refactor of how LLVM passes are loaded.
- Added interface support for `--disable-default-pipeline`
- Added interface support for `--no-simplify-prior-transform`


troelsfr:
> In the ConstArray example it seems a few empty .ll files may have been included in the commit by mistake, but perhaps they are placeholders. I have one other question below about the repeated pattern of setting up the default passes, but otherwise this seems ready to merge to the feature branch. Disabling the auto-merge just to make sure you have the chance to see and respond, but feel free to merge if you'd rather address this in the next PR.

Thanks - I've deleted ll files.

PR #1193, opened 2021-10-11
Adding parser support to PyQir

swernli:
This adds to the PyQir wheel a function and set of classes to support parsing QIR. This initial version focuses on the subset of LLVM we commonly use in certain examples, but sets the stage for further extension and general support. It introduces classes like `QirParameter` or `QirType` based on the underlying Rust crate `llvm_ir` but with some extra helpers that make parsing QIR in particular easier. For example, a `QirConstant` exposes a `is_qubit` property along with a `qubit_static_id` that make it easy to identify arguments to a call that are statically allocated qubits and find what their equivalent integer is. Likewise some additional helpers for traversing program control flow are added, such as `get_block_by_name` and `get_instruction_by_output_name` on `QirFunction` and `get_phi_pairs_by_source_name` for `QirBasicBlock`. The test for this uses a teleport chain sample which has some of these elements, but further test programs will need to be added that allow for verifying more complex features like phi node handling.

### DIFF
--- a/src/QirTools/pyqir/pyqir/parser.py
+++ b/src/QirTools/pyqir/pyqir/parser.py
@@ -89,7 +89,7 @@ class QirArrayType(QirType):
         """
         Gets the ordered list of QirTypes representing the underlying array types.
         """
-        return list(map(QirType, self.ty.array_element_type))
+        return [QirType(i) for i in self.ty.array_element_type]
 
     @property
     def element_count(self) -> int:
@@ -108,7 +108,7 @@ class QirStructType(QirType):
         """
         Gets the ordered list of QirTypes representing the underlying struct types.
         """
-        return list(map(QirType, self.ty.struct_element_types))
+        return [QirType(i) for i in self.ty.struct_element_types]
 
 class QirNamedStructType(QirType):
     """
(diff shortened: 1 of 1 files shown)


swernli:
Thanks for the feedback @cgranade, in particular the ideas about how to make it more Python friendly (I'll admit that's not my strength so I could use all the advice you've got there). I'll work on naming updates and usability improvements over the next few days and update the PR as I make progress. My hope was to get the initial functionality up and running first and get that into this PR, and now I can use your feedback and any other I receive to refine the structure and API.

Case in point, I left a comment at the top of one of the files describing how I imagine this will need to evolve, and it lines up with some of your feedback. Namely, the few places I've added extensions/helper functions should get pushed own into the qirlib rust module so it can be used from other rust clients, and this layer I've worked in here can focus on just exposing that into a convenient pyo3 form. Then I would wrap it with a thin layer in the python module itself so I can provide doc comments specific to python consumers. Given that idea and the feedback you've provided on the design, I'll work on that next and incorporate your feedback as part of that, committing it into this same working branch and updating this PR. Given that, I'll change this PR to a draft for now!

swernli:
@crazy4pi314 Thanks for the info and interest! I think your scenario sounds cool and could definitely make use of this new functionality. If you'd like to try the wheels from the build and see how those work so far, I'd appreciate any feedback here or in issues about the Python usability and API. For now, I've definitely focused on a subset of QIR based on our current needs, but left room to do more general support going forward. To that end, I'd be happy to review any PRs or feature requests as we continue building out the python support for QIR!

One thing that immediately comes to mind is that while PyQir contains both support for generating QIR and this new support for parsing it, it doesn't actually connect the two yet. So if you use the parser to get an in-memory representation of the program, there isn't a good way to modify it and output that modified program as QIR. It seems like we should have the pieces needed, and that would go a long way to enabling things like optimizations and transformations of programs, though it is unfortunately out of scope for this specific PR. If that is an area of interest for you, would you mind filing an issue describing that use case and tagging myself, @idavis and @bettinaheim in it so we can discuss there? Thanks!

PR #1199, opened 2021-10-15
Quantum Development Kit Telemetry Library

vxfield:
This PR introduces a common Telemetry Library (.NET class library) to be used by the Quantum Development Kit.

This library provides a single place where we can put all of our core telemetry logic, reducing code duplication across multiple QDK products and bringing the following benefits:

1. Makes it easier to comply to telemetry, data collection and privacy policies in a standard/consistent way, such as opt-in and opt-out controls
2. Comprehensive unit test (and, in the future, integration test) coverage
3. Ability to share common event properties and schemas across different QDK products
4. Makes it easier for QDK developers to add instrumentation to the products (with included C# and F# samples)
5. Support for out-of-process telemetry upload, to reduce the run time of short-lived console applications


vxfield:
@anpaz, @bettinaheim, @ScottCarda-MS:

This is the .NET class library idea that I mentioned to you.
I only worked on it for a couple hours and it still need some clean-up (specially in the TypeConversionHelper), and documentation.
I'll also work on a F# sample on how to use the library.

The main motivation for it is make it easy to add telemetry to the Q# Formatter, and later standardize our telemetry on all of our .NET-based tools (IQ#, part of Visual Studio Extension, Q# compiler, etc...)

Let me know what you think.

vxfield:
Just as a heads-up, these are the next things I'm planning to work on this PR:
- Reorganize the classes in their own files and create some folder structures to group the files together (still keeping all in the same namespace).
- Finishing adding unit tests for the out of process scenario. That will increase the code coverage to almost 100%.
- Add more asserts in the unit tests, particularly to validate things are getting logged correctly.
- Implement simple serialization protocol for the out of process scenario instead of using json (I'll make it easy to change by using the Strategy Pattern).

vxfield:
@samarcha, could you take a look at the F# sample (in the Samples folder) to see if it looks good?
That is my first F# program ever, so forgive me if I'm murdering the language :-)

src/Telemetry/Samples/FSharp/Program.fs
```diff
@@ -0,0 +1,158 @@
...
+        SampleArray: string []
+        SampleTimeSpan: TimeSpan
+        SampleInt: int
+        SampleGuid: Guid
+        [<SerializeJson>]
+        SampleObjectToBeSerialized: Map<string, string>
+        SampleObjectToBeIgnored: obj
+        SampleException: Exception
+        SampleNullableWithValue: Nullable<int>
+        SampleNullableWithNull: Nullable<int>
```
samarsha, line 29: System.Nullable generally isn't used in F# unless it's specifically for C# interop. More natural is to use `int option`. It looks like the telemetry library has special support for ignoring nulls; does it also support F# options?

src/Telemetry/Library/Telemetry.csproj
```diff
@@ -0,0 +1,10 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
```
ScottCarda-MS, line 3: I'm not certain of how these frameworks work, but most everything else in the compiler repo is netcoreapp3.1 or netstandard2.1. Is it going to create compatibility issues for this to be targeting net5.0?

cgranade: net5.0 supercedes netcoreapp3.1 and netstandard2.1 both; in particular, a net5.0 exe or dll can reference a netstandard2.1 project. The bad case is referring to a net5.0 dll from a netcoreapp3.1 exe. That said, we're in the process of upgrading to net6.0 for exes across the board, such that this should work well once bumped to net6.0.

src/Telemetry/Samples/FSharp/Program.fs
vxfield: Thanks, Sarah. I'll change it to use F# options and make sure it works well.

src/Telemetry/Library/Telemetry.csproj
ScottCarda-MS: That sounds especially problematic for what the intended use of this library is. This library needs to be referenced by other projects in the repo. It sounds like this needs to be lowered to at least netcoreapp3.1, if not netstandard2.1.

vxfield: @cgranade, the initial use case for this library is for the Q# formatter (which targets netstandard2.1) to use it.
Do you foresee a problem that would require us to temporarily target netstandard2.1, maybe by having different build profiles?

vxfield: just saw your comment, @ScottCarda-MS. I'll work on an option to target the older frameworks.

cgranade: @ScottCarda-MS: Given that those exe projects will be upgraded to net6.0 ASAP (.NET 6.0 launches and is the default early this month), I don't think it's problematic at all for this library to target net6.0. There's not much reason to upgrade existing libraries from netstandard2.1 to net6.0, but I also think there's also not a lot of reason to develop _new_ libraries against netstandard2.1 instead of net6.0. In particular, the net6.0 TFM brings with it support for C# 10 (netstandard2.1 targets 7.0, IIRC); that in turn makes it easier to provide null-safe code, given improvements to nullable reference types in more recent versions of C#.

+@ricardo-espinoza for .NET 6.0 discussion/impact.

ScottCarda-MS: @cgranade If the update to net6.0 is, say this week, than we can hold off on this PR till then. Otherwise we are in a bit of a time crunch to get this in, and we need to have this PR free of dependencies on code changes that will happen later in the month. In general code should be written to be compatible with the codebase as it is, not as it will be. Based on that, I think it will be more appropriate to updated this library to net6.0 as well as the Q# formatter along with everything else later in the month.

cgranade: I'll let @ricardo-espinoza speak in more detail to specific timelines, but given that .NET 6.0 goes live in a few days, I do believe the update to the net6.0 TFM is intended to be somewhat timely.

> In general code should be written to be compatible with the codebase as it is, not as it will be.

It may be helpful to assume good faith on the part of other developers! In particular, I think it's relevant in this case that good code should also anticipate future needs, especially when those needs are planned on a short time frame. Considering how code is affected by other features and maintenance can be a good way of preventing churn.

For example, one solution to the concern that you raised would be to temporarily enable `netstandard2.1` support as per @vxfield's comment, but to do so with multi-framework targeting:

```xml
<TargetFrameworks>net6.0;netstandard2.1</TargetFrameworks>
```

That ensures that dependencies can be taken now as-is, and also that code builds correctly against changes currently in progress.

vxfield: @ScottCarda-MS, any chance you can make the Q# Formatter target .NET 5 as of now?
Downgrading the Telemetry library back to netstandard2.1 will require a lot of (silly) code changes from C# 9 to C# 8 that I would like to avoid if possible.
Let me know.

src/Telemetry/Samples/FSharp/Program.fs
vxfield: Added support to use F# Options and updated the F# Sample.

src/Telemetry/Library/Telemetry.csproj
ScottCarda-MS: Yah I can look into that if it will save you a lot of changes like that. I'll need to double check with Bettina if that will work with the SDK integration she is working on, but I expect that it will be okay to go forward with that. For now let's assume we can, and I'll update the Q# Formatter.

src/Telemetry/Samples/CSharp/Program.cs
```diff
@@ -0,0 +1,209 @@
...
+                {
+                    CollectTargetSite = true,
+                    CollectSanitizedStackTrace = true,
+                },
+                SendTelemetryInitializedEvent = true,
+                SendTelemetryTearDownEvent = true,
+            };
+
+            // REQUIRED: Initialize
+            // Initialize the TelemetryManager right at the beggining of your program
```
ScottCarda-MS, line 33:
```suggestion
            // Initialize the TelemetryManager right at the beginning of your program

```

src/Telemetry/Samples/FSharp/Program.fs
```diff
@@ -0,0 +1,158 @@
...
+            MaxTeardownUploadTime = TimeSpan.FromSeconds(2.0),
+            OutOfProcessUpload = true,
+            ExceptionLoggingOptions =
+                ExceptionLoggingOptions(CollectTargetSite = true, CollectSanitizedStackTrace = true),
+            SendTelemetryInitializedEvent = true,
+            SendTelemetryTearDownEvent = true
+        )
+
+    // REQUIRED: Initialize
+    // Initialize the TelemetryManager right at the beggining of your program
```
ScottCarda-MS, line 63:
```suggestion
    // Initialize the TelemetryManager right at the beginning of your program
```

src/Telemetry/Library/InternalsVisibleToTests.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Microsoft.Quantum.Telemetry.Tests")]
```
anpaz:
```suggestion
[assembly: InternalsVisibleTo("Microsoft.Quantum.Telemetry.Tests" + SigningConstants.PUBLIC_KEY)]
```
needs to change for signing.

src/Telemetry/Library/Telemetry.csproj
```diff
@@ -0,0 +1,10 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
anpaz, line 1: My understanding is that this is going to be distributed as a nuget package?
If so I believe it's missing a lot of properties (like Author, Copyright, PackageIcon, etc).
The project is also missing the AssemblyConstants needed for DelaySigning.

src/Telemetry/Library/TelemetryManager.cs
```diff
@@ -0,0 +1,444 @@
...
+                // We don't log exception fields as they can potentially contain customer data
+                if (obj is Exception exception)
+                {
+                    obj = exception.ToExceptionLogRecord(Configuration.ExceptionLoggingOptions);
+                    eventName = $"{Configuration.EventNamePrefix}_{Configuration.AppId}_Exception";
+                }
+
+                var type = obj.GetType();
+                eventProperties.Name = eventName ?? $"{Configuration.EventNamePrefix}_{Configuration.AppId}_{type.Name}";
+                var properties = ReflectionCache.GetProperties(type);
```
anpaz, line 306: Are we worried that reflection is very slow? Since one of the goals was to make this fast so it doesn't add a significant overhead to short processes like qsfmt, I'm wondering if this defeats the purpose.

src/Telemetry/Library/TelemetryManager.cs
```diff
@@ -0,0 +1,444 @@
...
+            if (TelemetryOptOut)
+            {
+                return;
+            }
+
+            try
+            {
+                if (telemetryLogger == null && !initializingOrTearingDown)
+                {
+                    throw new InvalidOperationException("TelemetryManager has not been initialized. Please call TelemetryManager.Initialize before attempting to log anything.");
```
anpaz, line 390: not very safe if it thrown an exception, is it? :)
in this particular case it might be ok to throw one, but in general we should avoid anything that causes telemetry to create any blockers from the actual app.

src/Telemetry/Nuget.config
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
```
anpaz: we should just use the nuget.config in the source of the project (it will be automatically used if this is missing).

src/Telemetry/Samples/CSharp/Program.cs
```diff
@@ -0,0 +1,209 @@
...
+
+namespace Microsoft.Quantum.Telemetry.Samples.CSharp
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            var stopwatch = Stopwatch.StartNew();
+
+            var telemetryConfig = new TelemetryManagerConfig()
```
anpaz, line 16: initializing and enabling telemetry should be only be enabled the the `TELEMETRY` compiler constant is on.

src/Telemetry/Samples/CSharp/Program.cs
```diff
@@ -0,0 +1,209 @@
...
+        public string? SampleString { get; set; }
+
+        public bool SampleBool { get; set; }
+
+        public SampleEnumType SampleEnum { get; set; }
+
+        [PiiData]
+        public string? SamplePII { get; set; }
+
+        [SerializeJson]
```
anpaz, line 193: is this still needed?

src/Telemetry/Samples/FSharp/Program.fs
```diff
@@ -0,0 +1,158 @@
...
+    try
+        throwANestedException throw
+        null
+    with
+    | ex -> ex
+
+[<EntryPoint>]
+let main args =
+    // Initialize the TelemetryManager
+    let telemetryConfig =
```
anpaz, line 51: telemetry should only be enabled and initialized if the `TELEMETRY` compiler flag is set.

src/Telemetry/Tests/OutOfProcess/SimpleYamlSerializerTests.cs
```diff
@@ -0,0 +1,136 @@
...
+            var commandResults = (await AsyncEnumerableToEnumerable(deserializedResults)).ToList();
+
+            Assert.AreEqual(commands.Count, commandResults.Count);
+            for (int i = 0; i < commands.Count; i++)
+            {
+                Assert.AreEqual(commands[i], commandResults[i]);
+            }
+        }
+    }
+}
```
anpaz, line 137: We should add more tests for invalid cases/usage. For example:
* when we pass multiple commands not seperated by blank lines
* when the client tries to log an event but the logmanager process is not there (for example, it crashed)
* when the client sends an invalid payload

src/Telemetry/readme.md
```diff
@@ -0,0 +1,36 @@
...
+https://github.com/microsoft/artifacts-credprovider
+
+
+Run:
+iex "& { $(irm https://aka.ms/install-artifacts-credprovider.ps1) }"
+
+Run:
+dotnet restore --interactive   
+
+Env vars:
```
anpaz: this is a bit terse, could you explain when/how each one is used.

anpaz (dismissed):
Looks great overall, but I do have several things that should be revised.

src/Telemetry/Build/Test.ps1
```diff
@@ -0,0 +1,15 @@
...
+
+Push-Location $PSScriptRoot/../Tests/
+dotnet reportgenerator "-reports:$coverageFile" "-targetdir:TestResults/html" -reporttypes:HTML;
+Pop-Location
+
+$reportFile = Resolve-Path "$PSScriptRoot/../Tests/TestResults/html/index.htm"
+
+Write-Output "Attempting to open: file:///$reportFile"
+    
+Start-Process "file:///$reportFile"
```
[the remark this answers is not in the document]
vxfield: Oh, we don't call these scripts from the build agents, but I'll move this test coverage to a separate script just in case.

src/Telemetry/Build/pack.ps1
```diff
@@ -0,0 +1,3 @@
+dotnet publish -c release --self-contained --runtime linux-x64 --framework net5.0
```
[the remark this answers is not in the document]
vxfield: this was just an ad hoc test for the out of process sample.
I'll remove it

src/Telemetry/Samples/CSharp/Program.cs
vxfield: Do we really need this `TELEMETRY` compiler constant?
Why not always build with telemetry included and opt-out during runtime if needed?

src/Telemetry/Samples/CSharp/Program.cs
vxfield: If you mean the `[SerializeJson]` attribute, yes, it's needed to optionally serialize a whole object to json and send it as a telemetry payload to be consumed in Kusto.

src/Telemetry/Samples/FSharp/Program.fs
vxfield: Do we really need this TELEMETRY compiler constant?
Why not always build with telemetry included and opt-out during runtime if needed?

src/Telemetry/readme.md
vxfield: I've included a comprehensive readme last night

src/Telemetry/Library/OutOfProcess/OutOfProcessLogger.cs
```diff
@@ -0,0 +1,220 @@
...
+        }
+
+        private EVTStatus SendCommand(OutOfProcessCommand command)
+        {
+            // let's not start a new process just to ask it to quit
+            if (!(command is OutOfProcessQuitCommand))
+            {
+                this.CreateExternalProcessIfNeeded();
+
+                if (!this.externalProcess.IsRunning)
```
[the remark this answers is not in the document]
vxfield: The `externalProcess` is a non-null property of type IExternalProcessConnector which is never null.
The ExternalProcessConnector is basically a wrapper around the actual .NET `Process` or, in the case of a unit test, a Mock to simulate an external process.

src/Telemetry/Library/OutOfProcess/OutOfProcessLogger.cs
```diff
@@ -0,0 +1,220 @@
...
+                {
+                    throw new InvalidOperationException("OutOfProcessUpload external process is not running");
```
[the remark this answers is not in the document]
vxfield: The design that I chose is for the Telemetry Library to internally throw exceptions as needed to surface unexpected errors.

But then, there is another option for exceptions to be suppressed, including this one, or other unexpected exceptions. The ReadMe provides an explanation on how it works.

src/Telemetry/Library/OutOfProcess/SimpleYamlSerializer.cs
```diff
@@ -0,0 +1,232 @@
...
+                case TelemetryPropertyType.Double:
+                    return double.Parse(value);
+                case TelemetryPropertyType.Guid:
+                    return Guid.Parse(value);
+                case TelemetryPropertyType.Long:
+                    return long.Parse(value);
+                case TelemetryPropertyType.String:
+                    return ReplaceLineBreaks(value);
+                default:
+                    throw new ArgumentOutOfRangeException(message: $"{propertyType} conversion not implemented", innerException: null);
```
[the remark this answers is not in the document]
vxfield: The design that I chose is for the Telemetry Library to internally throw exceptions as needed to surface unexpected errors.

But then, there is another option for exceptions to be suppressed, including this one, or other unexpected exceptions. The ReadMe provides an explanation on how it works.

src/Telemetry/Library/TelemetryManager.cs
```diff
@@ -0,0 +1,444 @@
...
+using System.Diagnostics;
+using System.Linq;
+using System.Net.NetworkInformation;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Text.RegularExpressions;
+using Microsoft.Applications.Events;
+using Microsoft.Quantum.Telemetry.OutOfProcess;
+
+[assembly: InternalsVisibleTo("Microsoft.Quantum.Telemetry.Tests")]
```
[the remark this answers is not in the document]
vxfield: Yes, I removed it.

src/Telemetry/Library/InternalsVisibleToTests.cs
vxfield: I removed this attribute.

src/Telemetry/Library/Telemetry.csproj
vxfield: As of now, this will be distributed together with the Q# Formatter via project reference.

But I'll add the required metadata.

src/Telemetry/Library/OutOfProcess/SimpleYamlSerializer.cs
```diff
@@ -0,0 +1,232 @@
...
+using System.Text.RegularExpressions;
+using Microsoft.Applications.Events;
+
+namespace Microsoft.Quantum.Telemetry.OutOfProcess
+{
+    internal class SimpleYamlSerializer : IOutOfProcessSerializer
+    {
+        private static readonly string LineBreak = @"__\r\n__";
+        private static readonly string EventNamePropertyName = "__name__";
+        private static readonly Regex CommandRegex = new(@"^((?<command>- command:\s+!(?<commandType>[^\s$]+).*$)|(?<property>\s{4}(?<key>[^\s:]+):\s*(!(?<type>[^\s:\+]+)(?<pii>\+Pii)?\s+)?(?<value>.+)))$", RegexOptions.Compiled);
```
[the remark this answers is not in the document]
vxfield: A compiled Regex should perform as good as if we were to build our own state machine parser.
The slow down could come from looking for the capture groups by name in the Match object, but since this is running out of process anyway, I didn't want to over optimize it.

src/Telemetry/Library/OutOfProcess/SimpleYamlSerializer.cs
```diff
@@ -0,0 +1,232 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text.RegularExpressions;
+using Microsoft.Applications.Events;
+
+namespace Microsoft.Quantum.Telemetry.OutOfProcess
+{
+    internal class SimpleYamlSerializer : IOutOfProcessSerializer
```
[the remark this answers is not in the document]
vxfield: Will do!

src/Telemetry/Library/TelemetryManager.cs
vxfield: Yes, in general reflection is slow, but not that slow if only used once, hence why I created a cache.
IMO, the convenience of using reflection outweighs the verbose code of logging each property line by line.
This makes it a lot easier for developers to instrument applications by definition the events schemas as strongly typed records.
With everything else that goes on with the instrumentation (like uploading data, out-of-process communication, etc.) the cost of a one-time reflection should be neglectable. 
If we really want to optimize this, I would rather use a code generation tool to "reflect" at compile time rather than asking the developer to log property by property.

src/Telemetry/Library/TelemetryManager.cs
vxfield: The design that I chose is for the Telemetry Library to internally throw exceptions as needed to surface unexpected errors.

But then, there is another option for exceptions to be suppressed, including this one, or other unexpected exceptions. The ReadMe provides an explanation on how it works.

src/Telemetry/Library/OutOfProcess/SimpleYamlSerializer.cs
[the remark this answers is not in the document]
vxfield: Done.

src/Telemetry/Library/Telemetry.csproj
vxfield: The common assembly metadata and delay signing info is automatically added via [/src/Common/AssemblyCommon.props](https://github.com/microsoft/qsharp-compiler/blob/main/src/Common/AssemblyCommon.props) and [/src/Common/DelaySign.props](https://github.com/microsoft/qsharp-compiler/blob/main/src/Common/DelaySign.props) at the root of the repo.

src/Telemetry/Nuget.config
vxfield: Sure. Removed.

src/Telemetry/Tests/OutOfProcess/SimpleYamlSerializerTests.cs
vxfield: The TestSimpleYamlSerializerSingleCommand test method tests for invalid payloads, commands, indentation, etc.
The behavior of the parser, by design, is to ignore anything that is invalid and print the errors to the debug console without throwing exceptions.

We actually allow multiple commands to not be separated by blank lines, but I added a case for that anyway.

The serializer is just a serializer and is unaware of an external process.

The case of handling an external process crash is handled by the `OutOfProcessClient` class and will be tested in by the `OutOfProcessTests` class.
I'll add more test cases there, including with a real external process, but I want to do it in another PR.

src/Telemetry/Samples/CSharp/Program.cs
vxfield: Accepted you suggestion as discussed.

src/Telemetry/Samples/FSharp/Program.fs
vxfield: Accepted you suggestion as discussed.

ScottCarda-MS (approved):

PR #1212, opened 2021-10-21
Statement Kinds Support

ScottCarda-MS:
Adds support for various statement kinds in the syntax tree.
Statement kinds that need support are:
- [x] Expression Statements
- [x] Fail Statements
- [x] Mutable Declarations
- [x] Set Statements
- [x] Update Statements
- [x] Set-With Statements
- [x] Elif Statements
- [x] While Statements
- [x] Repeat Statements
- [x] Until Statements
- [x] Within Statements
- [x] Apply Statements


src/QsFmt/Formatter/SyntaxTree/Statement.fsi
```diff
@@ -141,75 +175,65 @@ type internal Let =
     }
 
 /// <summary>
-/// A <c>return</c> statement.
+/// An <c>update</c> statement.
```
bettinaheim, line 178: This could use a bit more detail; the language docs do not have a section about an update statement. It would be helpful if the doc comments in the code here use terminology that facilitates finding the corresponding section in the language docs. If we wanted to *really* nice, we could even link to the language docs...
Could you please clarify how an update statement differs from a SetWith?

To align with language docs, both SetWith and UpdateStatement are referred to as "evaluate-and-reassign statements" in the language docs: https://github.com/microsoft/qsharp-language/blob/main/Specifications/Language/2_Statements/VariableDeclarationsAndReassignments.md#evaluate-and-reassign-statements.

src/QsFmt/Formatter/SyntaxTree/Statement.fsi
```diff
@@ -141,75 +175,65 @@ type internal Let =
...
-
-    /// The block of statements after the declaration.
-    | Block of Statement Block
-
-/// A qubit declaration statement.
-and internal QubitDeclaration =
+/// <summary>
+/// A <c>set w/=</c> statement.
+/// </summary>
+type internal SetWith =
```
bettinaheim, line 203: This naming to me just feels a bit inconsistent. Aligning closer with the UpdateStatement, I would consider calling it UpdateWithStatement.

src/QsFmt/Formatter/SyntaxTree/Statement.fsi
```diff
@@ -141,75 +175,65 @@ type internal Let =
...
         /// </summary>
-        ReturnKeyword: Terminal
+        SetKeyword: Terminal
 
-        /// The returned expression.
-        Expression: Expression
+        /// The identifier being updated.
+        Name: Terminal
+
+        /// The update operator.
```
bettinaheim, line 190: I feel it is worth calling out that the equal sign is included here:
```suggestion
        /// The update operator, e.g. <c>+=</c>.
```

src/QsFmt/Formatter/SyntaxTree/Statement.fsi
```diff
@@ -141,75 +175,65 @@ type internal Let =
     }
 
 /// <summary>
-/// A <c>return</c> statement.
+/// An <c>update</c> statement.
 /// </summary>
-type internal Return =
+type internal UpdateStatement =
```
bettinaheim, line 180: Is this used for both 'set a = 5;' and for 'set a += 5;'? 
If it is, then the structure of this statement doesn't cover the 'set (a,b) = (1,2);'. If it is not, then where is the SetStatement? 
Edit: Looks like it isn't, and I've found the SetStatement below.

src/QsFmt/Formatter/SyntaxTree/Statement.fsi
```diff
@@ -255,36 +275,160 @@ and internal For =
         Block: Statement Block
     }
 
+/// <summary>
+/// The concluding section of an <c>until</c> statement.
+/// </summary>
+and internal UntilCoda =
```
bettinaheim, line 281: What's a coda? I don't think I've heard that before.

src/QsFmt/Formatter/SyntaxTree/Statement.fsi
```diff
@@ -255,36 +275,160 @@ and internal For =
...
     /// </summary>
-    | Return of Return
+    | Mutable of BindingStatement
 
-    /// A qubit declaration statement.
-    | QubitDeclaration of QubitDeclaration
+    /// <summary>
+    /// A <c>set</c> statement.
+    /// </summary>
+    | SetStatement of BindingStatement
```
bettinaheim, line 372: Set and update statement seem to be the only thing called statement; aligning closer with Let and Mutable, it would make sense to call them Set, Update, and UpdateWith.

src/QsFmt/Formatter/ParseTree/Statement.fs
```diff
@@ -104,51 +104,85 @@ type StatementVisitor(tokens) =
...
+        {
+            Keyword = context.set |> Node.toTerminal tokens
+            Binding = symbolBindingVisitor.Visit context.binding
+            Equals = context.equals |> Node.toTerminal tokens
+            Value = expressionVisitor.Visit context.value
+            Semicolon = context.semicolon |> Node.toTerminal tokens
+        }
+        |> SetStatement
+
+    override _.VisitSetUpdateStatement context =
```
bettinaheim, line 160: Why the discrepancy in naming between the parse tree and the rewriter, the reducer and the syntax tree?

bettinaheim (commented):
I don't have much to add - all looks reasonable to me and I like the tests. I just have a couple of comments on what caused some naming confusion for me.

src/QsFmt/Formatter/SyntaxTree/Statement.fsi
ScottCarda-MS: The definition for coda is "a concluding event, remark, or section". I found it was a good description to describe the end-part of the statement.

src/QsFmt/Formatter/SyntaxTree/Statement.fsi
ScottCarda-MS: I mentioned the evaluate-and-reassign name and gave an example for the Update statements.

src/QsFmt/Formatter/SyntaxTree/Statement.fsi
ScottCarda-MS: done

src/QsFmt/Formatter/SyntaxTree/Statement.fsi
ScottCarda-MS: done

src/QsFmt/Formatter/SyntaxTree/Statement.fsi
ScottCarda-MS: Added Statement to the end of all the statement data structures.

src/QsFmt/Formatter/ParseTree/Statement.fs
ScottCarda-MS: Hopefully I've resolved the discrepancies here.

bettinaheim (approved):

src/QsFmt/Formatter/SyntaxTree/Rewriter.fs
```diff
@@ -5,7 +5,41 @@ namespace Microsoft.Quantum.QsFmt.Formatter.SyntaxTree
 
 open Microsoft.Quantum.QsFmt.Formatter.Utils
 
-type 'context Rewriter() =
+type 'context Rewriter() as rewriter =
```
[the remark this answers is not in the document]
ScottCarda-MS: I didn't find a difference between these two forms of the `self` reference. To make this file consistent I've gone ahead and removed the member-specific `self` references, just as the Reducer has.

PR #1215, opened 2021-10-23
Fix language server symbol info for lambda parameters

samarsha:
There's a bug (not specific to lambdas) that prevents hover information from showing if the same variable occurs more than once in a statement (except for the first occurrence). Also another bug where the lambda's parameter tuple doesn't say "declaration of" in the hover, like it does for bindings in statements. I'll fix those in a separate PR.


src/QsCompiler/CompilationManager/EditorSupport/SymbolInfo.cs
```diff
@@ -277,13 +342,168 @@ bool OverlapsWithPosition(Range symRange)
...
+        /// <summary>
+        /// Finds the innermost scope that contains <paramref name="position"/> and returns the scope, statements in the
+        /// innermost scope before and including <paramref name="position"/>, and statements after
+        /// <paramref name="position"/>. If <paramref name="position"/> points to a statement that declares variables
+        /// and contains a child scope, the statements after <paramref name="position"/> are from the child scope.
+        /// </summary>
+        /// <param name="scope">A scope that contains <paramref name="position"/>.</param>
+        /// <param name="position">The position to split by.</param>
+        /// <returns>The innermost scope and statements before and after <paramref name="position"/>.</returns>
+        private static (QsScope, ImmutableList<QsStatement>, ImmutableList<QsStatement>) SplitStatementsByPosition(
```
bettinaheim, line 363: How does it deal with statements whose position is Null? That would be good to document.

src/QsCompiler/CompilationManager/EditorSupport/SymbolInfo.cs
```diff
@@ -122,6 +127,66 @@ bool OverlapsWithPosition(Range symRange)
...
+                QsStatementKind.QsExpressionStatement expression => ExpressionDeclarations(expression.Item),
+                QsStatementKind.QsReturnStatement @return => ExpressionDeclarations(@return.Item),
+                QsStatementKind.QsFailStatement fail => ExpressionDeclarations(fail.Item),
+                QsStatementKind.QsVariableDeclaration declaration => ExpressionDeclarations(declaration.Item.Rhs),
+                QsStatementKind.QsValueUpdate update => ExpressionDeclarations(update.Item.Rhs),
+                QsStatementKind.QsConditionalStatement cond => CondExpressionDeclarations(cond),
+                QsStatementKind.QsForStatement @for => ExpressionDeclarations(@for.Item.IterationValues),
+                QsStatementKind.QsWhileStatement @while => ExpressionDeclarations(@while.Item.Condition),
+                QsStatementKind.QsRepeatStatement repeat => ExpressionDeclarations(repeat.Item.SuccessCondition),
+                _ => Enumerable.Empty<LocalVariableDeclaration<string>>(),
```
bettinaheim, line 164: While unlikely, a qubit scope could have a variable declaration.

src/QsCompiler/CompilationManager/EditorSupport/SymbolInfo.cs
```diff
@@ -277,13 +342,168 @@ bool OverlapsWithPosition(Range symRange)
...
+            IEnumerable<LocalVariableDeclaration<string>> Binary(TypedExpression e1, TypedExpression e2) =>
+                DeclarationsInExpression(e1, position).Concat(DeclarationsInExpression(e2, position));
+
+            IEnumerable<LocalVariableDeclaration<string>> Many(IEnumerable<TypedExpression> es) =>
+                es.SelectMany(e => DeclarationsInExpression(e, position));
+
+            IEnumerable<LocalVariableDeclaration<string>> Lambda(Lambda<TypedExpression> lambda)
+            {
+                var inputType = CallableInputType(expression.ResolvedType);
+                var inferred = new InferredExpressionInformation(false, true);
```
bettinaheim, line 470: I am curious about your thinking on always setting hasLocalQuantumDependency to true. It seems reasonable as a pessimistic assumption, given that hasLocalQuantumDependency always has to assume that there is a dependency when there could be. I think it merits a comment in the code calling this out.

src/QsCompiler/CompilationManager/EditorSupport/SymbolInfo.cs
```diff
@@ -277,13 +342,168 @@ bool OverlapsWithPosition(Range symRange)
...
+            };
+        }
+
+        private static IEnumerable<LocalVariableDeclaration<string>> DeclarationsInTypedSymbol(
+            QsSymbol symbol, ResolvedType type, InferredExpressionInformation inferred)
+        {
+            switch (symbol.Symbol, type.Resolution)
+            {
+                case (QsSymbolKind.Symbol name, _):
+                    var range = symbol.Range.ValueOr(Range.Zero);
```
bettinaheim, line 489: Hm, I don't see how this would result in correct behavior is the range is not set. Is this ever supposed to be Null? If not, then in the spirit of failing early, it would make more sense to throw. If there is reason to expect that the range is null, why is the declaration return with incorrect information, rather than ignored?

bettinaheim (commented):
All in all, looks correct and reasonable to me. I've added some questions and comments.

src/QsCompiler/CompilationManager/EditorSupport/SymbolInfo.cs
samarsha: I don't think so; a qubit initializer is not an expression and cannot contain a lambda.

src/QsCompiler/CompilationManager/EditorSupport/SymbolInfo.cs
samarsha: Throwing feels too strict because there could be other symbols with ranges in a tuple; if it returned an empty enumerable rather than throwing, it could extract as much useful information as it could. If LocalVariableDeclaration had an optional range, I would just return one without a range if the source symbol was missing a range, but it's required.

src/QsCompiler/CompilationManager/EditorSupport/SymbolInfo.cs
samarsha: I think it has the same behavior as the original `StatementsAfterAndLocalDeclarationsAt` that it's replacing: it ignores any statement without a location, *and* any statements after it (because of the `TakeWhile`). So it basically treats statements without a location as if it occurred after the given position.

Added comment.

src/QsCompiler/CompilationManager/EditorSupport/SymbolInfo.cs
samarsha: Added comment.

src/QsCompiler/CompilationManager/EditorSupport/SymbolInfo.cs
samarsha: I guess you can argue that the function can't do what it's claiming to do (return LocalVariableDeclarations for every symbol) if any of the symbols is missing a range, because LocalVariableDeclaration requires a range, so throwing makes sense.

samarsha: Updated code to throw if the symbol range is null.

src/QsCompiler/CompilationManager/EditorSupport/SymbolInfo.cs
bettinaheim: The qubit array initializer takes an expression, and that expression could contain a lambda. Think of some function that takes a function as argument an returns the number of qubits. While not very common, 
a) the compiler needs to cover all cases not just the common ones
b) it is not unreasonable to have something like the following e.g. in a Chemistry application:
```
let qs = Qubit[RequiredNumberOfQubits(x => x *x, ...)];
```

bettinaheim (approved):
Please take a look at the qubit array initializer. Other than that, looks good.

src/QsCompiler/CompilationManager/EditorSupport/SymbolInfo.cs
samarsha: I don't mean that the compiler shouldn't handle uncommon cases; I just forgot about the qubit array length expression. Thanks for explaining.

PR #1207, opened 2021-10-21
Adding Sdk integration, and formatting support + code action for syntax updates to the language server

bettinaheim:
This PR does three things:
- It includes the published version of the formatter in the Microsoft.Quantum.Sdk, and adds a build target for updating the syntax in the project (no formatting).
- It adds support for formatting files in the language server. This currently as written includes both formatting and updating syntax.
- It adds a code action for the things that can be fixed by the syntax updater. If that action is selected, the deprecated syntax in the whole file is updates. No formatting is performed. 

There is a test for the language server functionality, which tests two scenarios:
1) The first part checks formatting works when using the formatter packaged in the Sdk, and relies on a fixed version of the Sdk that has the formatter included. 
2) The second part tests formatting works when using the (published) source code version of the formatter, and should catch any breaking changes to the formatter.

Additionally, we will set up a sanity test in the qdk repo to ensure that the formatter is correctly packaged in all future versions of the Sdk and the defined build target works as well. 
Telemetry will follow in a separate PR.

PR #1220, opened 2021-10-25
Refactor language server code for extracting symbols from Q# fragments

samarsha:
* Significantly simplifies the `SymbolInformation` type and function and makes it easier to extract symbol declarations from an expression (useful for lambdas).
* Improves type safety for code that doesn't support overlapping occurrences.
* Improves type safety by allowing code to exhaustively match over the occurrence kinds.


src/QsCompiler/SyntaxProcessor/SyntaxExtensions.fs
```diff
@@ -18,222 +18,55 @@ open Microsoft.Quantum.QsCompiler.SyntaxTokens
...
-        ]
-
-    symDecl,
-    (vars,
-     List.concat [ typeParams
-                   types
-                   signature.ReturnType |> TypeNameSymbols ],
-     exs)
-
+// TODO: Remove after 2022-05 release.
```
bettinaheim, line 28: Could you please follow the patter used for other such todos, so that they are easier to find in the code:
TODO: RELEASE 2022-05: Remove SymbolInformation.

src/QsCompiler/SyntaxProcessor/SyntaxExtensions.fs
```diff
@@ -18,222 +18,55 @@ open Microsoft.Quantum.QsCompiler.SyntaxTokens
...
     }
-    static member internal New(decl: QsSymbol list, (vars: QsSymbol list, ts: QsType list, ex: QsExpression list)) =
-        {
-            DeclaredSymbols = decl.ToImmutableHashSet()
-            UsedVariables = vars.ToImmutableHashSet()
-            UsedTypes = ts.ToImmutableHashSet()
-            UsedLiterals = ex.ToImmutableHashSet()
-        }
 
+// TODO: Remove after 2022-05 release.
```
bettinaheim, line 38: Same here.

src/QsCompiler/SyntaxProcessor/SymbolOccurrence.fs
```diff
@@ -0,0 +1,194 @@
...
+        let typeParams = Seq.map Declaration callable.Signature.TypeParameters |> Seq.toList
+        let args = inNamedTuple callable.Signature.Argument
+        let returnType = inType callable.Signature.ReturnType
+        Declaration callable.Name :: typeParams @ args @ returnType
+
+    [<CompiledName "InFragment">]
+    let inFragment fragment =
+        match fragment with
+        | NamespaceDeclaration s -> [ Declaration s ]
+        | OpenDirective (_, s) -> QsNullable<_>.Choose id [ s ] |> Seq.toList |> List.map Declaration
```
bettinaheim, line 129: The namespace name of an open directive was previously considered to be a used variable - that is lost here.

src/QsCompiler/SyntaxProcessor/SymbolOccurrence.fs
```diff
@@ -0,0 +1,194 @@
...
+
+module SymbolOccurrence =
+    let rec flattenSymbol symbol =
+        match symbol.Symbol with
+        | SymbolTuple items -> Seq.collect flattenSymbol items |> Seq.toList
+        | _ -> [ symbol ]
+
+    let symbolDeclarations = flattenSymbol >> List.map Declaration
+
+    let rec inType type_ =
```
bettinaheim, line 25: Can we please still use the private keyword for the corresponding functions also when a .fsi is present? It is somewhat cumbersome to have to open the fsi to figure out which things are public and which things are not.

src/QsCompiler/SyntaxProcessor/SymbolOccurrence.fs
```diff
@@ -0,0 +1,194 @@
...
+    | UsedVariable of QsSymbol
+    | UsedLiteral of QsExpression
+
+module SymbolOccurrence =
+    let rec flattenSymbol symbol =
+        match symbol.Symbol with
+        | SymbolTuple items -> Seq.collect flattenSymbol items |> Seq.toList
+        | _ -> [ symbol ]
+
+    let symbolDeclarations = flattenSymbol >> List.map Declaration
```
bettinaheim, line 23: This used to not return anything for missing and invalid symbols (see also the comment on the type declaration).

src/QsCompiler/SyntaxProcessor/SymbolOccurrence.fs
```diff
@@ -0,0 +1,194 @@
...
+        | BodyDeclaration g
+        | AdjointDeclaration g
+        | ControlledDeclaration g
+        | ControlledAdjointDeclaration g -> inGenerator g
+        | ImmutableBinding (s, e)
+        | MutableBinding (s, e)
+        | ForLoopIntro (s, e) -> symbolDeclarations s @ inExpression e
+        | UsingBlockIntro (s, i)
+        | BorrowingBlockIntro (s, i) -> symbolDeclarations s @ inInitializer i
+        | TypeDefinition t -> Declaration t.Name :: inNamedTuple t.UnderlyingType
```
bettinaheim, line 142: This might run into trouble when items are unnamed; before Invalid and Missing symbols were not collected, now they are.

src/QsCompiler/SyntaxProcessor/SymbolOccurrence.fs
```diff
@@ -0,0 +1,194 @@
...
+        | Declaration s -> declaration.Invoke s
+        | UsedType t -> usedType.Invoke t
+        | UsedVariable s -> usedVariable.Invoke s
+        | UsedLiteral e -> usedLiteral.Invoke e
+
+    [<MaybeNull>]
+    member occurrence.AsDeclaration =
+        match occurrence with
+        | Declaration s -> s
+        | _ -> Unchecked.defaultof<QsSymbol>
```
bettinaheim, line 176: I would suggest to stick with returning a QsNullable instead. See also here, and the tldr: https://stackoverflow.com/questions/13520720/unchecked-defaultof-on-a-list.

src/QsCompiler/CompilationManager/ProjectManager.cs
```diff
@@ -1111,7 +1111,9 @@ editCollection switch
         /// </remarks>
         public Hover? HoverInformation(TextDocumentPositionParams? param, MarkupKind format = MarkupKind.PlainText) =>
             this.Manager(param?.TextDocument?.Uri)?.FileQuery(
-                param?.TextDocument, (file, c) => file.HoverInformation(c, param?.Position?.ToQSharp(), format), suppressExceptionLogging: true);
+                param?.TextDocument,
```
idavis, line 1114: It feels like there a lot of changes without adding tests. Is the code covered in this PR already under test?

src/QsCompiler/SyntaxProcessor/SymbolOccurrence.fs
samarsha: IMO, this is an improvement. For example, previous behavior is to show nothing when hovering over `_` in `let _ = ...`. Now it shows the hover text "Discarded symbol assignment."

src/QsCompiler/CompilationManager/EditorSupport/SymbolInfo.cs
```diff
@@ -191,21 +175,20 @@ bool OverlapsWithPosition(Range symRange)
...
+            if (fragment?.Kind is QsFragmentKind.NamespaceDeclaration)
             {
                 return false;
             }
 
-            var sym = symbolInfo.UsedTypes.Any()
-                && symbolInfo.UsedTypes.Single().Type is QsTypeKind<QsType, QsSymbol, QsSymbol, Characteristics>.UserDefinedType udt ? udt.Item
-                : symbolInfo.UsedVariables.Any() ? symbolInfo.UsedVariables.Single()
-                : symbolInfo.DeclaredSymbols.Any() ? symbolInfo.DeclaredSymbols.Single() : null;
+            var occurrence = SymbolOccurrence(file, position, true);
```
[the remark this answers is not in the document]
samarsha: The only thing SymbolOccurrence used the file for was to get the fragment, so I just replaced the file parameter with a fragment parameter instead. The fragment can then be reused after it's given to SymbolOccurrence.

src/QsCompiler/SyntaxProcessor/SymbolOccurrence.fs
samarsha: I guess this is a matter of style. The F# style guide doesn't explicitly discourage `private` if you use a signature file, but it gets close with [this sentence](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/component-design-guidelines#hide-the-representation-of-f-union-types-in-vanilla-net-apis):

> When designing a vanilla .NET API, consider hiding the representation of a union type by using either a private declaration or a signature file.

"Either ... or ..." is exclusive, so it seems to suggest that you should use `private` in the .fs file, or omit it from an .fsi file, but not both.

My opinion is: the .fsi file *defines* the public interface for the .fs file, so I think it is a good thing that you need to open it to figure out what is public.

src/QsCompiler/SyntaxProcessor/SymbolOccurrence.fs
samarsha: From testing, I couldn't find any issues with unnamed items in type declarations. Note that the omitted symbol in this case is missing a range, so the language server should ignore it.

src/QsCompiler/SyntaxProcessor/SymbolOccurrence.fs
samarsha: This is part of the C# interop API for SymbolOccurrence, so it should be as idiomatic to use from C# as possible. That's why I used a nullable type. The IL for the default case here is:

```
    IL_001f: ldnull
    IL_0020: ret
```

That should be equivalent to a C# method with a `SymbolOccurrence?` return type returning null, which is what I intended.

I don't think this will cause issues like that StackOverflow question, because in that question they are assigning `null` to a field of an F# list, which has a special interpretation. Here, `null` is only used as a return type for a method that is called from C#. One snag is that it does assume that `SymbolOccurrence` is a reference type; I don't think this will work if it has the `[<Struct>]` attribute.

src/QsCompiler/SyntaxProcessor/SymbolOccurrence.fs
```diff
@@ -0,0 +1,194 @@
...
+        | DoubleLiteral _
+        | BoolLiteral _
+        | ResultLiteral _
+        | PauliLiteral _
+        | MissingExpr -> [ UsedLiteral expression ]
+        // TODO: Remove special case for un-interpolated strings once overlapping occurrences are handled.
+        | StringLiteral (_, es) when es.IsEmpty -> [ UsedLiteral expression ]
+        | Identifier (s, ts) ->
+            let ts' = QsNullable.defaultValue ImmutableArray.Empty ts |> Seq.collect inType |> Seq.toList
+            UsedVariable s :: ts'
```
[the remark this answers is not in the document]
samarsha: I'm not sure - these are pretty old todos that I moved over here.

src/QsCompiler/CompilationManager/EditorSupport/SymbolInfo.cs
```diff
@@ -191,21 +175,20 @@ bool OverlapsWithPosition(Range symRange)
...
         {
             (referenceLocations, declarationLocation) = (null, null);
-            if (file == null || compilation == null)
-            {
-                return false;
-            }
 
-            var symbolInfo = file.TryGetQsSymbolInfo(position, true, out var fragment); // includes the end position
-            if (symbolInfo == null || fragment?.Kind is QsFragmentKind.NamespaceDeclaration)
+            var fragment = file.TryGetFragmentAt(position, out _, true);
```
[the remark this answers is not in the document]
samarsha: Do you mean the `symbolInfo == null` check? I just coalesced it with the `sym == null` check that's later down, by using a null-guarded `occurrence?.Match`.

src/QsCompiler/CompilationManager/ProjectManager.cs
samarsha: The language server has a very limited test suite - mostly, it tests that it can load projects and track edits correctly. There is no test infrastructure for hover information, finding references, etc.

src/QsCompiler/SyntaxProcessor/SymbolOccurrence.fs
bettinaheim: Yes, my main concern was that it may not work as expected in some of the places where it is used (e.g. type declarations came to mind). That there is no range and it doesn't show up makes sense, though.

src/QsCompiler/CompilationManager/ProjectManager.cs
bettinaheim: While there is definitively a testing gap that we know of and are tracking (within the QDK team), it would be good to add a basic test case specifically for scenarios we are adding. See e.g. the test added for the new formatting support here: https://github.com/microsoft/qsharp-compiler/pull/1207/files#diff-b160aef5299bccdebd1263eafd99d9d7811162a4ac2bc45cc007d318afdb5c02
The coverage won't be perfect, but some coverage is better than none.

src/QsCompiler/SyntaxProcessor/SymbolOccurrence.fs
bettinaheim: There are a couple Unchecked.defaultof in this file, and I am just a bit worried that ultimately, we might change one of the types or one of these methods will be used from F#, and this will lead to odd issues that go unnoticed. It just feels a bit unsafe, and I think the "use with caution" disclaimer (Unchecked) overall has good reason. I understand the desire for keeping it idiomatic, but this feels like one check too little especially for a public API. Plus, it doesn't align well with the rest of the code base, and I would have concerns using it broadly across the code base.

src/QsCompiler/SyntaxProcessor/SymbolOccurrence.fs
bettinaheim: The style guide here I believe is vague. As such, I have no objections keeping it as is, merely a preference as stated.

src/QsCompiler/SyntaxProcessor/SymbolOccurrence.fs
samarsha: It looks like the [F# component design guidelines recommend the TryGet pattern](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/component-design-guidelines#use-the-trygetvalue-pattern-instead-of-returning-f-option-values-and-prefer-method-overloading-to-taking-f-option-values-as-arguments) for C# interop, as the replacement for returning an option type.

src/QsCompiler/CompilationManager/ProjectManager.cs
samarsha: OK, so you mean add tests to the feature/lambda branch before it's merged to main, but tests are not necessary for this PR?

src/QsCompiler/SyntaxProcessor/SymbolOccurrence.fs
bettinaheim: I'd prefer to go with that then. That seems reasonable to me - what do you think?

src/QsCompiler/CompilationManager/ProjectManager.cs
bettinaheim: Yes, I was suggesting to add a test case for the hover and find all references for lambdas before merging to main, but not necessarily as part of this PR (in the spirit of testing the functionality rather than the implementation, even though some of these are technically part of a public API).

src/QsCompiler/SyntaxProcessor/SymbolOccurrence.fs
samarsha: I updated this to use the TryGet pattern.

bettinaheim (approved):

PR #1222, opened 2021-10-26
Adding python gate jit api

idavis:


src/QirTools/qirlib/src/interop/pyjit/gates.rs
```diff
@@ -0,0 +1,216 @@
+#![allow(non_upper_case_globals)]
+#![allow(non_camel_case_types)]
+#![allow(non_snake_case)]
+#![allow(dead_code)]
```
cgranade, line 4: The other `allow` attributes are fairly straightforward, but it may help to add a comment explaining why allowing dead code is helpful.

src/QirTools/qirlib/src/interop/pyjit/intrinsics.rs
```diff
@@ -0,0 +1,237 @@
+#![allow(non_upper_case_globals)]
+#![allow(non_camel_case_types)]
+#![allow(non_snake_case)]
+#![allow(dead_code)]
+#![allow(unused_variables)]
+#![allow(unused_imports)]
```
cgranade, line 6: Similarly here.

src/QirTools/qirlib/src/interop/pyjit/intrinsics.rs
```diff
@@ -0,0 +1,237 @@
...
+    //gs.h_ctl(control, get_cubit_string(qubit));
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn __quantum__qis__measure__body(
+    qubits: *mut QirArray,
+    registers: *mut QirArray,
+) {
+    log::debug!("/__quantum__qis__measure__body/");
+
```
cgranade, line 41: Is it intended for this to be blank? Should there be a `todo!` macro here?

src/QirTools/qirlib/src/interop/pyjit/intrinsics.rs
```diff
@@ -0,0 +1,237 @@
...
+#[no_mangle]
+pub unsafe extern "C" fn __quantum__qis__r__body(pauli: PauliId, theta: f64, qubit: QUBIT) {
+    log::debug!("/__quantum__qis__r__body/");
+    let mut gs = get_current_gate_processor();
+    match pauli {
+        1 => gs.rx(theta, qubit),
+        3 => gs.ry(theta, qubit),
+        2 => gs.rz(theta, qubit),
+        _ => panic!("Unsupported Pauli value: {}", pauli),
```
cgranade, line 58: Technically, rotating about `PauliI` is valid. It's a no-op in hardware, but still is valid.

cgranade (commented):
Added a first round of comments. I also wanted to ask about including *.so/*.dylib/*.dll files directly in the repo, as that effectively creates a circular dependency with the runtime repo. Perhaps the same question, but the runtime repo also contains a Rust crate implementing CHP and density matrix simulation that would be good to include as an `impl GateSet` if possible without running into the same circular dependency issue.

bettinaheim:
@idavis Is this PR obsolete?

Issue #1202, opened 2021-10-18
Formatter Updates Deprecated Specialization Declaration Parameters


ScottCarda-MS:
The formatter should have a rule for updating the specialization declaration syntax, replacing the deprecated syntax for parameters with the supported syntax.
- `body ()` should be replaced with `body (...)`.
- `adjoint ()` should be replaced with `adjoint (...)`.
- `controlled (cs)` should be replaced with `controlled (cs, ...)`
- `adjoint controlled (cs)` should be replaced with `adjoint controlled (cs, ...)`.

For example:
```qsharp
operation Foo () : Unit is Ctl + Adj {
    body () {
        // ...
    }
    adjoint () {
        // ...
    }
    controlled (q) {
        // ...
    }
    controlled adjoint (q) {
        // ...
    }
}
```
should be updated to
```qsharp
operation Foo () : Unit is Ctl + Adj {
    body (...) {
        // ...
    }
    adjoint (...) {
        // ...
    }
    controlled (q, ...) {
        // ...
    }
    controlled adjoint (q, ...) {
        // ...
    }
}
```

yjt98765:
Then it should be easy to implement. I will create a pull request soon...

ScottCarda-MS:
Resolved with #1226.

PR #1226, opened 2021-10-27
Formatter Updates Deprecated Specialization Declaration Parameters

yjt98765:
This PR implements issue #1202, i.e., added an update rule that appends deprecated specialization declaration with a `...` parameter.

To do so, I first modified the syntax tree definition of `SpecializationGenerator.Provider`. Previously, all the parameters are regarded as a `Terminal` without a detailed structure. Therefore, updating and formatting rules cannot be applied. For example, the extra space in `controlled(q,` `...)` cannot be removed. Instead, I define `Provider.parameters` as [`specializationParameter tuple`](https://github.com/microsoft/qsharp-compiler/compare/main...yjt98765:specialization?expand=1#diff-6660bf1f7c254fee4924922540ad22a55df67965b62c9d806fae6ee10f803aeeL17-R19), which solves the problem.

Based on the previous definition, I implemented the [rule](https://github.com/microsoft/qsharp-compiler/compare/main...yjt98765:specialization?expand=1#diff-0aa06a0f8f64b964b555407d003971f9273a8243d632ee750a3509ce8358d902R163-R208). It updates the parameters based on the number and the text of existing parameters.

The code example used in the issue is added to `Examples.fs` as a test case.


src/QsFmt/Formatter/SyntaxTree/Namespace.fs
```diff
@@ -12,9 +12,11 @@ type TypeParameterBinding =
         CloseBracket: Terminal
     }
 
+type internal SpecializationParameter = Terminal
```
ScottCarda-MS, line 15: I'm not sure I see the need for this wrapper type. If we keep the `Provided` case of type `Terminal Tuple option * Statement Block` would that break anything?

src/QsFmt/Formatter/SyntaxTree/Namespace.fs
```diff
@@ -12,9 +12,11 @@ type TypeParameterBinding =
...
+
 type SpecializationGenerator =
     | BuiltIn of name: Terminal * semicolon: Terminal
-    | Provided of parameters: Terminal option * statements: Statement Block
+    | Provided of parameters: SpecializationParameter Tuple option * statements: Statement Block
```
ScottCarda-MS, line 19: Since this is a Tuple type now, I don't think there needs to be an option type here.

src/QsFmt/Formatter/SyntaxTree/Namespace.fs
yjt98765: No, everything will be fine. Actually, I just used `Terminal` for rewriting and reducing. I have no problem with removing it.

src/QsFmt/Formatter/SyntaxTree/Namespace.fs
yjt98765: But the ANTLR rule is [`providedSpecialization : parameters=specializationParameterTuple? block=scope;`](https://github.com/microsoft/qsharp-compiler/blob/main/src/QsFmt/Parser/QSharpParser.g4#L110). Note the `?` after `specializationParameterTuple`. It means that the entire tuple can be missing, including the parentheses. I have also found some code examples that do not have `()`, e.g., `body { }`([CodegenTests.qs](https://github.com/microsoft/qsharp-compiler/blob/main/src/QsCompiler/Tests.CSharpGeneration/Circuits/CodegenTests.qs#L59)) and
[InternalRenaming.qs](https://github.com/microsoft/qsharp-compiler/blob/main/src/QsCompiler/Tests.Compiler/TestCases/LinkingTests/InternalRenaming.qs#L103):
```
        body {
        }
```
Maybe this style is also deprecated? In that case, we do not only need to remove the `option`, but also modify the parser and the examples. Maybe even add move update rules?

ScottCarda-MS: Oh I didn't know we could just _not_ have the parameter tuple. That is also deprecated syntax, but to be able to handle it, we would need to keep the `option` type as you have it here. Could you add the mapping of `body { }` -> `body (...) { }` syntax to this rule, or is it a significant enough change that it needs to be a separate rule?

yjt98765: I have added the supporting for `body { }` in recent commits. To add the `()`, I hard-coded an empty `Tuple` object,  which I hope is not a problem.

src/QsFmt/Formatter/ParseTree/Namespace.fs
```diff
@@ -33,8 +33,21 @@ type SpecializationGeneratorVisitor(tokens) =
         toBuiltIn context.intrinsic context.semicolon
 
     override _.VisitProvidedGenerator context =
+        let tospecializationParameters tokens (context: QSharpParser.SpecializationParameterTupleContext) =
+            let toSpecializationParameter (context: QSharpParser.SpecializationParameterContext) =
+                { Prefix = Node.prefix tokens context.Start.TokenIndex; Text = context.GetText() }
+
+            let parameters = context._parameters |> Seq.map toSpecializationParameter
```
ScottCarda-MS, line 40:
```suggestion
            let parameters = context._parameters |> Seq.map (Node.toUnknown tokens)
```

src/QsFmt/Formatter/SyntaxTree/Namespace.fs
ScottCarda-MS: Thanks!

src/QsFmt/Formatter/SyntaxTree/Namespace.fs
ScottCarda-MS: Thanks!

src/QsFmt/Formatter/Rules.fs
```diff
@@ -160,6 +160,61 @@ let forParensUpdate =
...
+            | [] -> [ ellipsisItem 0 ]
+            // Replace, e.g., `controlled (q)` with `controlled (q, ...)`
+            | [ x ] ->
+                match Option.get(x.Item).Text with
+                | "..." -> [ x ]
+                | _ -> [ ensureComma x; ellipsisItem 1 ]
+            | _ -> parameters.Items
+    }
+
+let specializationUpdate =
```
ScottCarda-MS, line 194: The documentation for the Formatter has recently been updated to include the rule set that is run. Could you also add this rule to the QsFmt's README.md file, under the section "Rules that are run"? [Link to Section](https://github.com/yjt98765/qsharp-compiler/blob/specialization/src/QsFmt/README.md?plain=1#L52)

yjt98765: I have added the first version to that section. It is challenging to summarize this update rule in a sentence, though. Please modify it if you have a better idea!

ScottCarda-MS: Looks good, thanks!

ScottCarda-MS (approved):
This looks great! Thanks for your contribution!

Issue #1224, opened 2021-10-27
Quantum Development Kit should be updated to .NET 6


ricardo-espinoza:
With the upcoming release of .NET6 on November 9th, 2021 we'll have a new LTS version which will be supported until the end 2024. See here the official [.NET support policy](https://dotnet.microsoft.com/platform/support/policy).

Currently, the QDK is built on, and requires .NET 3.1, which is the current LTS version. However, this version is already almost two years old and approaching the end of its supported lifetime.

We're proposing to update the target SDK of the QDK binaries to be based on .NET6 and make this version the new minimum required one.

This migration will bring the following benefits:

- Based on a newer version with a long support lifetime ahead.
- Enable C# 10 features in the QDK binaries.
- Align with the recommended and default version that users will get when downloading .NET.

PR #1232, opened 2021-11-03
Update .NET version as target framework for qsharp-compiler binaries

ricardo-espinoza:
This pull request is updating binaries of the Q# Compiler to target .NET6 as part of a larger effort tracked by issue #1224 .

This pull request is partial work into a feature branch, so not all checks are expected to pass.

### DIFF
--- a/src/QsCompiler/Tests.LanguageServer/ProjectLoaderTests.cs
+++ b/src/QsCompiler/Tests.LanguageServer/ProjectLoaderTests.cs
@@ -68,7 +68,7 @@ namespace Microsoft.Quantum.QsLanguageServer.Testing
             Assert.IsTrue(loader.IsSupportedQsFramework("netcoreapp2.1"));
             Assert.IsTrue(loader.IsSupportedQsFramework("netcoreapp2.2"));
             Assert.IsTrue(loader.IsSupportedQsFramework("netcoreapp3.0"));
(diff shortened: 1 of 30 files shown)


src/QsCompiler/Tests.LanguageServer/ProjectLoaderTests.cs
```diff
@@ -68,7 +68,7 @@ public void SupportedTargetFrameworks()
             Assert.IsTrue(loader.IsSupportedQsFramework("netcoreapp2.1"));
             Assert.IsTrue(loader.IsSupportedQsFramework("netcoreapp2.2"));
             Assert.IsTrue(loader.IsSupportedQsFramework("netcoreapp3.0"));
-            Assert.IsTrue(loader.IsSupportedQsFramework("netcoreapp3.1"));
+            Assert.IsTrue(loader.IsSupportedQsFramework("net6.0"));
```
cgranade, line 71: It looks from the above as though the project loader should still support older versions, even if they can't be built?

cgranade (approved):
LGTM, but may be worthwhile to get a second opinion as well.

src/QsCompiler/Tests.LanguageServer/ProjectLoaderTests.cs
ricardo-espinoza: Thanks for the catch, Chris! You're entirely right.

PR #1238, opened 2021-11-06
Sdk changes to support Qir targets

bettinaheim:


src/QuantumSdk/DefaultItems/DefaultItems.targets
```diff
@@ -107,17 +115,24 @@
       <_ResolvedTargetPackageReference Include="$(%(_TargetPackageReferencePathProperty.Identity))" />
       <ResolvedTargetSpecificDecompositions Include="@(_ResolvedTargetPackageReference->Split(`;`))" />
     </ItemGroup>
-    <Error Condition="@(ResolvedTargetSpecificDecompositions->Count()) &gt; 1" Text="More than one target package was defined. The defined target package needs to be unique." />
```
bettinaheim, line 110: I randomly came across that the Sdk prevents that more than more than one dll can contain target specific decompositions. The compiler supports multiple dlls - I believe I merely forgot to remove this here.

src/QuantumSdk/DefaultItems/DefaultItems.targets
```diff
@@ -107,17 +115,24 @@
...
+      Text="Resolved target package assemblies: @(ResolvedTargetSpecificDecompositions)"
       Importance="High" />
   </Target>
 
-  <!-- Computes ResolvedQscReferences, ResolvedQscReferencesAndPriorities, QscProjectReference and QscPackageReference. -->
+  <!-- Populates the ResolvedQSharpReferences item group. -->
+  <Target Name="ResolveQSharpReferences" DependsOnTargets="ResolveTargetPackage;ResolveAssemblyReferences;_CopyFilesMarkedCopyLocal" BeforeTargets="BeforeQSharpCompile">
+    <ItemGroup>
+      <ResolvedQSharpReferences Include="@(ReferencePath)" Condition="$([System.Text.RegularExpressions.Regex]::IsMatch(%(FullPath), '(?i)system.|mscorlib|netstandard.library|microsoft.netcore.app|csharp|fsharp|microsoft.visualstudio|microsoft.testplatform|microsoft.codeanalysis|fparsec|newtonsoft|roslynwrapper|yamldotnet|markdig|serilog')) == false" />
+      <ResolvedQSharpReferences Exclude="@(ResolvedTargetSpecificDecompositions)" />
```
bettinaheim, line 131: I opted that both the core package to support compilation in the first place as well as the target specific decompositions should be referenced by the project. That means that the target package can contain multiple dlls, and some or all of these dlls may be declared as the actual target decompositions. The content of dlls that are not explicitly declared as target specific decompositions in the target package are exposed in exactly the same way that other references are and will hence be visible to the user, while the dll content of target specific decompositions is only loaded after the initial compilation such that it does not conflict with other declarations and instead replaces them. Declarations in those dlls are not visible to the user.

src/QuantumSdk/Sdk/Sdk.props
```diff
@@ -23,43 +23,42 @@
...
   <!-- Include the qir generation and docs generation compiler extensions if needed. -->
   <ItemGroup>
     <PackageReference Condition="$(QSharpDocsGeneration)"
-                      Include="Microsoft.Quantum.DocumentationGenerator" Version="0.18.2107151063-beta"
+                      Include="Microsoft.Quantum.DocumentationGenerator" Version="0.20.2110171573"
                       IsImplicitlyDefined="true" 
                       IsQscReference="true" ExecutionTarget="Any" Priority="-1" PrivateAssets="All" />
-    <PackageReference Condition="$(QirGeneration) Or '$(ResolvedProcessorArchitecture)' == 'MicrosoftSimulator'"
-                      Include="Microsoft.Quantum.QirGeneration" Version="0.18.2107151063-beta"
+    <PackageReference Condition="$(QirGeneration) Or $(EnableQirSubmission)"
```
cesarzc, line 45: It seems like this condition is failing during builds because when this is evaluated, `$(QirGeneration)` is still a string (`default`).

swernli: I think you are right, the value for `QirGeneration` in src/QuantumSdk/DefaultItems/DefaultItems.props.v.template needs to be changed from `default` to `false`.

swernli (changes requested):

src/QuantumSdk/DefaultItems/DefaultItems.targets
bettinaheim: The compiler should actually generate an error in this case; i.e. the ordering is not hierarchical and the expectation is that there are declarations within all target package dlls are unique.

swernli (approved):
Approving based on our investigation.

src/QuantumSdk/Sdk/Sdk.props
bettinaheim: The QirGeneration property is set to either true or false in the DefaultItems.targets file. The issue turned out to be that the .targets file wasn't value due to using Exclude without an Include.

Issue #1216, opened 2021-10-23
NamespaceItem items are incomplete in formatter


bettinaheim:
Since namespace items are incomplete, e.g. type declarations are not currently not processed by qsfmt. 

Namespace kinds that need support are:
- [x] Open Directives
- [x] Type Declarations

yjt98765:
I have started to implement them. However, the ANTLR rule [`attribute`](https://github.com/microsoft/qsharp-compiler/blob/main/src/QsFmt/Parser/QSharpParser.g4#L35) confuses me:
```
attribute : at='@' expr=expression;

access : 'internal';

declarationPrefix : attributes+=attribute* access?;

// Type Declaration

typeDeclaration : declarationPrefix 'newtype' Identifier '=' underlyingType ';';
```
I cannot find related information in the Q# [user guide](https://docs.microsoft.com/en-us/azure/quantum/user-guide/language/programstructure/typedeclarations) or code examples in test files. Maybe it is a deprecated syntax and does not need to be implemented?

yjt98765:
I see... It refers to the `@Attribute()`

ScottCarda-MS:
Completed with #1239

PR #1239, opened 2021-11-06
Complete NamespaceItem in qsfmt

yjt98765:
Issue #1216 points out that the implementation of `NamespaceItem` is incomplete and hence some code cannot be formatted. This PR implements the open directive and type declaration. In addition to the callable, all the three types of `NamespaceItem` are implemented now.

To demonstrate the added implementation, I modified several lines of existing test cases. It shows that formatting rules can be applied to the open directive and type declaration.


src/QsFmt/Parser/QSharpParser.g4
```diff
@@ -21,14 +21,14 @@ namespace
 qualifiedName : Identifier ('.' Identifier)*;
 
 namespaceElement
-    : openDirective # OpenElement
-    | typeDeclaration # TypeElement
+    : directive=openDirective # OpenElement
+    | typeDecl=typeDeclaration # TypeElement
     | callable=callableDeclaration # CallableElement
```
ScottCarda-MS, line 26:
```suggestion
    : openDirective # OpenElement
    | typeDeclaration # TypeElement
    | callableDeclaration # CallableElement
```
I suggest we drop the inner labels here, and reorganize the NamespaceItemVisitor a bit. See my comment there.

ScottCarda-MS: I realize the `callable` label was already there, but now that the other cases are getting filled out, it could be removed.

src/QsFmt/Formatter/ParseTree/Namespace.fs
```diff
@@ -146,6 +187,29 @@ type NamespaceItemVisitor(tokens) =
 
     override _.VisitChildren node = Node.toUnknown tokens node |> Unknown
 
+    override _.VisitOpenElement context =
```
ScottCarda-MS, line 190: It would be better if we didn't use these inner labels for the namespace elements. We should organize this NamespaceItemsVisitor similar to how the ParameterVisitor is structured for consistency.

src/QsFmt/Formatter/ParseTree/Namespace.fs
```diff
@@ -146,6 +187,29 @@ type NamespaceItemVisitor(tokens) =
...
+
+    override _.VisitTypeElement context =
+        {
+            Attributes =
+                context.typeDecl.prefix._attributes |> Seq.map (NamespaceContext.toAttribute tokens) |> Seq.toList
+            Access = context.typeDecl.prefix.access () |> Option.ofObj |> Option.map (Node.toUnknown tokens)
+            NewtypeKeyword = context.typeDecl.keyword |> Node.toTerminal tokens
+            DeclaredType = context.typeDecl.declared |> Node.toTerminal tokens
+            Equals = context.typeDecl.equals |> Node.toTerminal tokens
+            UnderlyingType = context.typeDecl.underlying |> (UnderlyingTypeVistor tokens).Visit
```
ScottCarda-MS, line 208: nit: When a visitor needs to use another visitor, even if it is just once, it is preferred if the used visitor is bound to a variable at the top of the visitor definition. This just makes it a bit easier to see at a glance what other visitors a given visitor relies on.

src/QsFmt/Parser/QSharpParser.g4
```diff
@@ -40,18 +40,20 @@ declarationPrefix : attributes+=attribute* access?;
...
-typeDeclaration : declarationPrefix 'newtype' Identifier '=' underlyingType ';';
+typeDeclaration
+    : prefix=declarationPrefix keyword='newtype' declared=Identifier equals='=' underlying=underlyingType semicolon=';'
+    ;
 
 underlyingType
-    : typeDeclarationTuple
-    | type
+    : typeDeclarationTuple # TupleUnderlyingType
+    | type # OtherUnderlyingType
```
ScottCarda-MS, line 49: nit: I'd like to find a more descriptive name for this other than `OtherUnderlyingType`, though I can see it is a bit tricky with this rules scheme. Maybe `UnnamedUnderlyingType` or `UnlabeledUnderlyingType` since this is the only way to get a type into the definition without an accompanying label. Though that also kinda sounds like an anonymous type, which is not what this is. Feel free to come up with a different name too; I just would rather find something other than `OtherUnderlyingType`.

ScottCarda-MS (commented):
Just found a few things to improve on.

src/QsFmt/Formatter/ParseTree/Namespace.fs
yjt98765: I have modified it. The repetitive `.directive` etc is gone, so the code has become clearer.

src/QsFmt/Formatter/ParseTree/Namespace.fs
yjt98765: Thanks. I have modified them.

yjt98765: However, I got a problem. When I add a `TypeTupleItemVistor` to `UnderlyingTypeVistor` using the following line:
```
    let typeTupleItemVistor = TypeTupleItemVistor tokens
```
The formatted can be built but reported a stack overflow error. It would be fine if I use `(TypeTupleItemVistor tokens).Visit`, though.

I doubt that this problem is related to the fact that `UnderlyingTypeVistor` and `TypeTupleItemVistor` are defined together with the `and` keyword, and they have to visit each other. There is already an `underlyingTypeVistor` defined in `TypeTupleItemVistor`. Somehow, it is not allowed to define the other visitor in the other class at the same time. I looked up a little bit and find that a let binding in a class [creates a private function](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/members/let-bindings-in-classes), which is different from the let bindings in functions. Maybe this is related to the problem.

src/QsFmt/Parser/QSharpParser.g4
yjt98765: Thank you for seeing my struggle... The purpose of `typeDeclarationTuple` is to define types that [are similar to record types in F#](https://docs.microsoft.com/en-us/azure/quantum/user-guide/language/programstructure/typedeclarations). In the record type, there are named and unnamed items. So, I agree that `UnnamedUnderlyingType` or `UnlabeledUnderlyingType` would be confusing here. I have found the term "scala type" that is used as the opposite of "aggregate types" (maybe where "record type" belongs?). However, "scala type" mainly describes a scala value, which is also different from the "single-item type" semantics here. I was considering `NonRecordType`, but there is no context related to `record` here. Finally, I replace "record" with "tuple", and that is the current version.

Please replace it with a better name if you get one.

src/QsFmt/Formatter/ParseTree/Namespace.fs
ScottCarda-MS: Oh that makes sense. The top-level let bindings in a class like that get evaluated each time the constructor for that type is called, and if a let binding in `UnderlyingTypeVistor` is creating a `TypeTupleItemVistor`, which in turn has a let binding that is creating an `UnderlyingTypeVistor` object, the constructors will recursively call each other infinitely, causing the stack overflow issue you saw. Looking at how other tuple-like grammar rules are handled, I think it is fine if you leave the `(TupleTypeItemVisitor tokens).Visit` here.

src/QsFmt/Parser/QSharpParser.g4
ScottCarda-MS: Let's go with UnnamedTypeItem.

src/QsFmt/Formatter/ParseTree/Namespace.fs
ScottCarda-MS: Thanks that looks good now!

ScottCarda-MS (approved):
This looks good! If you could just updated the label on the UnderlyingTyple case one more time, then I think it will be good to merge in. Thanks for another great contribution!

src/QsFmt/Parser/QSharpParser.g4
yjt98765: No problem. It is done.

PR #1230, opened 2021-11-01
QAT v. 0.1

troelsfr:
This PR introduces the QIR adaptor tool (QAT) to the main repository. This tool allows the user to take a generic QIR and adapt to a specific profile. Through these transformations, the use is able to adapt a generic QIR to a constrained QIR that is more suitable for a specific architecture.

PR #1240, opened 2021-11-08
Addressing last comments - QAT v0.1

troelsfr:
Updating comments which were updated in the original PR #1230 .


src/Passes/Source/Rules/Notation/Call.hpp
```diff
@@ -0,0 +1,69 @@
+#pragma once
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+/// @defgroup shorthandNotation Shorthand Notation
+
+#include "Rules/Notation/Call.ipp"
```
kuzminrobin, line 7: (I looked at your links about .ipp files. Thank you for those links)

FYI: The Call.ipp file provides the _definition_ (of the template function) first, and then the remaining part of this header (Call.hpp) provides the _declaration_. So the order of declaration and definition seems to me _reversed_ in all the translation units that include this Call.hpp. Such an order makes it questionable the break-up into separate .hpp (with declaration) and .ipp (with definition). I personally would consider merging these two files. But up to you.

kuzminrobin (approved):

DmitryVasilevsky (approved):
My comments are addressed.

PR #1242, opened 2021-11-09
[DRAFT] Attempt of instrumenting QsFmt

vxfield:


vxfield:
@ScottCarda-MS, @samarsha 

This is my humble attempt to instrument the Q# Formatter.
It's very verbose and I had to introduce a lot of data types to return the data that we want to log.

Let me know what you think and if there is a better/simpler way to do it.

src/QsFmt/App/Program.fs
```diff
@@ -106,20 +129,174 @@ let runUpdateAndFormat (arguments: UpdateAndFormatArguments) =
...
-    | Error errorCode -> errorCode
+    | Error errorCode -> { RunResult.Default with ExitCode = errorCode }
+
+type ParseArgsResult =
+    | Success of ParserResult<obj>
+    | Error of Exception
+
+type ExecutionResult =
+    | Success of RunResult
+    | Error of Exception
```
samarsha, line 140: You could use `Result<ParserResult<obj>, Exception>` and `Result<RunResult, Exception>` instead (`Result` is a built-in F# type with constructors `Ok` and `Error`).

src/QsFmt/App/Program.fs
```diff
@@ -106,20 +129,174 @@ let runUpdateAndFormat (arguments: UpdateAndFormatArguments) =
...
+        {
+            StartTime = startTime
+            Command = argumentOptions.Command
+            RecurseFlag = argumentOptions.RecurseFlag
+            BackupFlag = argumentOptions.BackupFlag
+            InputType = argumentOptions.InputType
+            QSharpVersion = argumentOptions.QSharpVersion
+            UnhandledException = unhandledException
+            SyntaxErrors = syntaxErrors
+            ExecutionTime = DateTime.Now - startTime
```
ScottCarda-MS, line 295: Since the execution time here is measured on a very small scale, I found that it was highly recommended to use the System.Diagnostics.Stopwatch class to measure the execution time, as it is much more accurate than subtracting DateTime objects. You could still use DateTime.Now to get the start time.

src/QsFmt/App/Program.fs
```diff
@@ -56,29 +71,37 @@ let run arguments inputs =
...
+
+    and foldResults (previousRunResult: RunResult) filePath =
+        let newRunResult = (filePath |> doOne arguments)
+        {
+            FilesProcessed = previousRunResult.FilesProcessed + newRunResult.FilesProcessed
+            ExitCode = max previousRunResult.ExitCode newRunResult.ExitCode
+            SyntaxErrors = previousRunResult.SyntaxErrors @ newRunResult.SyntaxErrors
+        }
 
     and doMany arguments inputs =
```
ScottCarda-MS, line 96: Looks like the `arguments` parameter is not used here anymore. Actually, since `arguments` isn't modified anywhere under `run`, I think you could remove the `arguments` parameters from all the nested functions. All the references to `arguments` can just refer directly to the `run` function's `arguments` parameter.

src/QsFmt/App/Program.fs
vxfield: Suggestion implemented.

src/QsFmt/App/Program.fs
vxfield: Suggestion implemented.

src/QsFmt/App/Program.fs
vxfield: Suggestion accepted.

src/QsFmt/App/Program.fs
```diff
@@ -14,11 +14,11 @@ open Microsoft.Quantum.QsFmt.Formatter
 let makeFullPath input =
     if input = "-" then input else Path.GetFullPath input
 
-let run (commandWithOptions: CommandWithOptions) inputs =
+let runCommand (commandWithOptions: CommandWithOptions) inputs =
 
     let mutable paths = Set.empty
 
-    let rec doOne input =
+    let rec runOneCommand input =
```
ScottCarda-MS, line 21: nit: The nested functions are really operating on inputs rather than commands. It is not as though `runManyCommands` actually runs multiple _commands_. It processes multiple inputs, so maybe the names could be something like `doOneInput` and `doManyInputs`?

src/QsFmt/App/CommandLineArguments.fs
```diff
@@ -178,22 +243,18 @@ type UpdateAndFormatArguments =
...
-    | Update
-    | Format
-    | UpdateAndFormat
-
-type Arguments =
+type CommandWithOptions =
     {
-        CommandKind: CommandKind
+        CommandKind: CommandKind option
+        InputKind: InputKind option
```
ScottCarda-MS, line 249: I'm not quite seeing the need for these to be option types. Both of these need to have values and have default values when none are given.

vxfield: This is for the case in which we didn't have a chance to actually get a valid Command or Input type due to bad arguments

vxfield:
Replaced by #1249

PR #1243, opened 2021-11-09
Adds breakpoints at statements, debug info for functions, and partial debug info for variables

ryan-moreno:
This PR adds breakpoints at all Q# statements, adds debug information for functions that allows stepping through scopes, and adds partial debug info for variables of basic types. The next PR will take care of variables of all types and specifically will handle variables that are callables.


src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -170,46 +393,138 @@ private DebugInfoBuilder GetOrCreateDIBuilder(string sourcePath)
...
-        /// Note: because this is called from within the constructor of the GenerationContext,
-        /// we cannot access this.Module or anything else that uses this.sharedState
+        /// Creates a debug type for a particular Q# type. The debug type connects a debug info type with its
+        /// corresponding LLVM Native type.
         /// </summary>
-        internal void AddTopLevelDebugInfo(BitcodeModule owningModule, ImmutableArray<QsQualifiedName> entryPoints)
+        /// <param name="resolvedType">The Q# type to create a DebugType from.</param>
+        /// <param name="dIBuilder">The <see cref="DebugInfoBuilder"/> to use to create the debug type.</param>
+        /// <returns>The Debug Type corresponding to the given Q# type. Returns null if the debug type is not implemented.</returns>
+        private IDebugType<ITypeRef, DIType>? GetDebugTypeFor(ResolvedType resolvedType, DebugInfoBuilder dIBuilder)
```
ryan-moreno, line 402: @bettinaheim we'd discussed using a type transformation, but I made this helper function for the following reasons. Let me know if you think I should still set it up as a type transformation.

- I tried to have an abstract types class that NativeLlvmType (the existing Type class) and DebugType both inherit from, but C# doesn't support covariant return types in overrides, so I can't specify that we need outputs that implement IDebugType (rather than ITypeRef which it implements).
- If NativeType and DebugType don't both inherit from some Type, then can't reuse TypeTransformation. It seems like quite the overhead to implement another TypeTransformation when we're not actually transforming anything (the existing TypeTransformation just uses a builtType side effect to get the LLVM Native Type)
- We need a dIBuilder to create a DebugType, so we would need a different TypeTransformation for each source file so that it can have its own dIBuilder.

src/QsCompiler/LlvmBindings/BitcodeModule.cs
```diff
@@ -380,6 +380,66 @@ public IrFunction CreateFunction(string name, IFunctionType signature)
...
+            this.ThrowIfDisposed();
+
+            if (string.IsNullOrWhiteSpace(mangledName))
+            {
+                mangledName = name;
+            }
+
+            if (signature.DIType == null)
+            {
+                throw new ArgumentException();
```
bettinaheim, line 420:
```suggestion
                throw new ArgumentException("DIType is null");
```

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -132,6 +158,211 @@ internal void FinalizeDebugInfo()
...
+                string shortName = spec.Parent.Name; // We want to show the user the short name in the debug info instead of the mangled name
+
+                // get the location info
+                QsNullable<QsLocation> debugLoc = spec.Location;
+                if (debugLoc.IsNull)
+                {
+                    return this.CreateFunctionNoDebug(mangledName, signature);
+                }
+
+                Position absolutePosition = debugLoc.Item.Offset; // Because spec is its own namespace, the position is already an absolute position
```
bettinaheim, line 274: I don't understand what you mean by "spec is its own namespace" (same for a similar comment below).

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -132,6 +158,211 @@ internal void FinalizeDebugInfo()
...
+        /// <param name="value">The value of the function's argument.</param>
+        /// <param name="argNo">The index of the argument (expected to be 1-indexed).</param>
+        internal void CreateFunctionArgument(QsSpecialization spec, string name, IValue value, int argNo)
+        {
+            if (this.DebugFlag)
+            {
+                // get the DIBuilder
+                string sourcePath = spec.Source.CodeFile;
+                DebugInfoBuilder curDIBuilder = this.GetOrCreateDIBuilder(sourcePath);
+                DIFile debugFile = curDIBuilder.CreateFile(Path.ChangeExtension(sourcePath, ".c")); // TODO: change this now that we have a language solution
```
bettinaheim, line 322: I'd suggest to address this todo. As far as I know, things work correctly on Linux, and the remaining Windows issues are also reproduceable when debugging C++, correct?

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -155,13 +381,10 @@ private DebugInfoBuilder GetOrCreateDIBuilder(string sourcePath)
...
 
-                // TODO: If we need compilation flags (an optional argument to CreateCompileUnit) in the future we will need
-                // to figure out how to get the compile options in the CompilationLoader.Configuration
-                // and turn them into a string (although the compilation flags don't seem to be emitted to the IR anyways)
                 di = this.Module.CreateDIBuilder();
                 di.CreateCompileUnit(
-                    QSharpLanguage, // Note that to debug the source file, you'll have to copy the content of the .qs file into a .c file with the same name
-                    cSourcePath,
+                    QSharpLanguage,
+                    cSourcePath, // TODO: change this now that we have a solution to language/extension issue
```
bettinaheim, line 387: here's another todo related to the comment above

src/QsCompiler/QirGeneration/DebugInfoManager.cs
bettinaheim: I think if you populate it here to include all types, you will effectively just end up reimplementing the type transformation without using a type transformation. I am hence not sure I understand your second bullet point, and also the third bullet is not quite clear to me. I'd be happy to chat to better understand what you mean.

src/QsCompiler/QirGeneration/Context.cs
```diff
@@ -776,9 +781,9 @@ internal void GenerateConstructor(QsSpecialization spec, ArgumentTuple argTuple)
         /// as arguments and returns void.
         /// </summary>
         internal IrFunction GeneratePartialApplication(string name, QsSpecializationKind kind, Action<IReadOnlyList<Argument>> body)
-        {
+        { // RyanQuestion: What is this for? Does it need debug info?
```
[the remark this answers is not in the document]
ryan-moreno: It looks like I forgot to hit save after I got rid of the some of the random comments to myself in this file, sorry about that. Thanks for the info. I don't have access to the backlog, but I'll touch bases with you when we sync about how you'd like me to track these.

src/QsCompiler/QirGeneration/DebugInfoManager.cs
ryan-moreno: The Windows issues aren't reproduceable when debugging C++ because it only comes up when we use a language enum from outside the C family. Things work correctly on Linux if I hand-edit the .c extension in the LLVM IR to the .qs extension. However, I discovered yesterday that when I automate outputting the .qs extension into the LLVM IR file, I'm hitting an LLVM bug on both Windows and Linux (because of some side effects in the DebugInfoBuilders). I filed a bug with LLVM and think I have a workaround. I'll update with that once I verify the workaround.

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
```diff
@@ -1586,6 +1597,7 @@ public override ResolvedExpressionKind OnIdentifier(Identifier sym, QsNullable<I
             }
 
             this.SharedState.ValueStack.Push(value);
+            this.SharedState.DIManager.EmitLocation(Position.Zero);
```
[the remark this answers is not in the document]
ryan-moreno: This Function takes in a relative position and then automatically adds in the offset from the statement and expression stacks. I'll rename the function to make it clearer

src/QsCompiler/QirGeneration/Subtransformations/StatementTransformation.cs
```diff
@@ -33,10 +33,10 @@ public QirStatementTransformation(GenerationContext sharedState, TransformationO
 
         public override QsStatement OnStatement(QsStatement stm)
         {
-            QsNullable<QsLocation> loc = stm.Location;
-            this.SharedState.DIManager.LocationStack.Push(loc);
+            this.SharedState.DIManager.StatementLocationStack.Push(stm.Location.Item);
+            this.SharedState.DIManager.EmitLocation(Position.Zero);
```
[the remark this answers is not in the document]
ryan-moreno: The Position.Zero was supposed to be the offset relative to the current namespace and statement stack. I renamed the functions for clarity. Having two statements on the same line isn't an issue, there just are two emitted locations with the same line. Ultimately it only matters if there is at least one emitted location with whatever line you want to set the breakpoint on. I am emitting column information as well, so technically this would allow you to set two breakpoints on the same line if you were debugging with column specificity.

src/QsCompiler/QirGeneration/Subtransformations/StatementKindTransformation.cs
```diff
@@ -99,7 +100,7 @@ void DestructTuple(ImmutableArray<SymbolTuple> symbols, IValue value)
 
             if (symbols is SymbolTuple.VariableName varName)
             {
-                bindVariable(varName.Item, value);
+                bindVariable(varName.Item, value); // RyanNOte: probably better not to do it in here since we don't know what bindVariable does necessarily
```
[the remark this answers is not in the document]
ryan-moreno: Sorry, none of these comments were meant to be in the PR or for other people to understand. The comment is not relevant; it was just an old note I left for myself when I was looking through where to emit locations from. I went through and deleted my random comments but clearly didn't save the changes properly. Next time I'll make sure to skim through the PR within GitHub.

src/QsCompiler/QirGeneration/DebugInfoManager.cs
ryan-moreno: For almost all of the syntax tree nodes, except for Namespaces of course, the position that is stored is relative to the position of the NamespaceElement ancestor as well as the stack of ancestor statements. However, with Specializations specifically, they are absolute positions instead of relative positions. I'm not sure why it was done this way, but I assume the position is made absolute for these because specializations are callables, which each become their own namespace. That's what I meant by the comment, but I can see why it's not clear -- I'll rewrite that

src/QsCompiler/QirGeneration/Subtransformations/StatementTransformation.cs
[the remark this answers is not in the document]
ryan-moreno: The Position.Zero was supposed to be the offset relative to the current namespace and statement stack. I renamed the functions for clarity. Having two statements on the same line isn't an issue, there just are two emitted locations with the same line. Ultimately it only matters if there is at least one emitted location with whatever line you want to set the breakpoint on. I am emitting column information as well, so technically this would allow you to set two breakpoints on the same line if you were debugging with column specificity.

src/QsCompiler/LlvmBindings/DebugInfo/DILocation.cs
```diff
@@ -15,24 +15,22 @@ public class DILocation
     {
         /// <summary>Initializes a new instance of the <see cref="DILocation"/> class.</summary>
         /// <param name="context">Context that owns this location</param>
-        /// <param name="line">line number for the location</param>
-        /// <param name="column">Column number for the location</param>
+        /// <param name="position">1-based line/col number for the location</param>
         /// <param name="scope">Containing scope for the location</param>
-        public DILocation(Context context, uint line, uint column, DILocalScope scope)
-            : this(context, line, column, scope, null)
+        public DILocation(Context context, DebugPosition position, DILocalScope scope)
```
ryan-moreno, line 20: Before this goes to main, I should probably change the DebugPosition back to the line, column format since this is a public API. I can move DebugPosition information to within the QIR Generation only.

src/QsCompiler/LlvmBindings/BitcodeModule.cs
```diff
@@ -22,34 +22,34 @@ namespace Ubiquity.NET.Llvm
...
     [SuppressMessage("Design", "CA1027:Mark enums with FlagsAttribute", Justification = "It isn't a flags enum")]
-    public enum ModuleFlagBehavior
+    public enum ModuleFlagBehavior // There is currently a bug in LLVMSharp such that all values in the LLVMModuleFlagBehavior enum are off by one.
     {
         /// <summary>Invalid value (default value for this enumeration).</summary>
         Invalid = 0,
 
         /// <summary>Emits an error if two values disagree, otherwise the resulting value is that of the operands.</summary>
-        Error = LLVMModuleFlagBehavior.LLVMModuleFlagBehaviorError,
+        Error = 1, // LLVMModuleFlagBehavior.LLVMModuleFlagBehaviorError,
```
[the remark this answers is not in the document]
ryan-moreno: I can't have a function here because the assignment to ModuleFlagBehavior.Error or any others must be a constant since it's directly in a namespace. However, I could have it be LLVMModuleFlagBehavior.LLVMModuleFlagBehaviorError + 1 instead of just hard-coding 1. Is that an improvement?

src/QsCompiler/LlvmBindings/BitcodeModule.cs
```diff
@@ -22,34 +22,34 @@ namespace Ubiquity.NET.Llvm
 {
     /// <summary>Enumeration to indicate the behavior of module level flags metadata sharing the same name in a <see cref="BitcodeModule"/>.</summary>
     [SuppressMessage("Design", "CA1027:Mark enums with FlagsAttribute", Justification = "It isn't a flags enum")]
-    public enum ModuleFlagBehavior
+    public enum ModuleFlagBehavior // There is currently a bug in LLVMSharp such that all values in the LLVMModuleFlagBehavior enum are off by one.
```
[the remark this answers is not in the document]
ryan-moreno: Issue linked. I also created a PR to fix the issue in LLVMSharp and add a unit test (I didn't spend a bunch of time on this part, I had already created the test when I was trying to make sure there was a bug and just had to clean it up).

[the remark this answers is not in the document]
ryan-moreno: Resolving this conversation because I found an appropriate way to add the module flags correctly

src/QsCompiler/LlvmBindings/BitcodeModule.cs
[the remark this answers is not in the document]
ryan-moreno: Resolving this conversation because I found an appropriate way to add the module flags correctly.

src/QsCompiler/LlvmBindings/BitcodeModule.cs
```diff
@@ -380,6 +380,66 @@ public IrFunction CreateFunction(string name, IFunctionType signature)
...
+        /// <param name="linePosition">1-based <see cref="DebugPosition"/> with the line number of the function definition</param>
+        /// <param name="signature">LLVM Function type for the signature of the function</param>
+        /// <param name="isLocalToUnit">Flag to indicate if this function is local to the compilation unit</param>
+        /// <param name="isDefinition">Flag to indicate if this is a definition</param>
+        /// <param name="scopeLinePosition">1-based <see cref="DebugPosition"/> with the first line of the function's outermost scope, this may not be the same as the first line of the function definition due to source formatting</param>
+        /// <param name="debugFlags">Additional flags describing this function</param>
+        /// <param name="isOptimized">Flag to indicate if this function is optimized</param>
+        /// <param name="dIBuilder"><see cref="DebugInfoBuilder"/>to use when creating debug info</param>
+        /// <returns>Function described by the arguments</returns>
+        public IrFunction CreateFunction(
```
cesarzc, line 397: Not blocking for this PR, but for clarity this function could be named `CreateFunctionWithDebugInfo`.

src/QsCompiler/LlvmBindings/DebugInfo/DILocation.cs
cesarzc: Can you please add a TODO comment about this in the code so we don't forget?

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -5,22 +5,38 @@
...
+    using ResolvedTypeKind = QsTypeKind<ResolvedType, UserDefinedType, QsTypeParameter, CallableInformation>;
+
     /// <summary>
     /// This class holds shared utility routines for generating QIR debug information.
     /// It contains a reference to the GenerationContext that owns it because it needs
     /// access to the same shared state represented by the GenerationContext.
     /// </summary>
     internal sealed class DebugInfoManager
     {
+// SECTION: Config values
```
cesarzc, line 33: For these configuration options, it might make sense to encapsulate them in a private nested class called `Configuration` or `Options`.

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -52,16 +68,24 @@ private static string GetQirProducerIdent()
             return compilationInfo.Name + " with " + qirGenerationInfo.Name + " V " + qirGenerationInfo.Version;
         }
 
+// SECTION: Exposed member variables
```
cesarzc, line 71: In general the code in this repo does not explicitly marks sections. When this is done, it is usually in large files and using `#region` and `#endregion`. Since this is still a small file, I would recommend not explicitly marking sections but if you do, I would recommend using `#region` and `#endregion`.

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -132,6 +158,225 @@ internal void FinalizeDebugInfo()
...
+        /// which is converted to a 1-based <see cref="DebugPosition"/>.
+        /// </summary>
+        /// <param name="relativePosition">The 0-based <see cref="Position"/> at which to emit the location
+        /// relative to the namespace and stack of parent <see cref="QsStatement"/>s.</param>
+        internal void EmitLocationWithOffset(Position? relativePosition)
+        {
+            if (this.DebugFlag)
+            {
+                DISubProgram? sp = this.sharedState.CurrentFunction?.DISubProgram;
+                QsLocation? namespaceLoc = this.sharedState.DIManager.CurrentNamespaceElementLocation;
```
cesarzc, line 173: Variable names that are not abbreviations are usually more readable. In this case I would suggest changing `sp` to `subProgram` and `namespaceLoc` to `namespaceLocation`.

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -132,6 +158,225 @@ internal void FinalizeDebugInfo()
...
+        /// <summary>
+        /// If the debug flag is set to true, this creates the debug information for a local variable.
+        /// </summary>
+        /// <param name="name">The name of the local variable.</param>
+        /// <param name="value">The value of the variable.</param>
+        internal void CreateLocalVariable(string name, IValue value)
+        {
+            if (this.DebugFlag)
+            {
+                // get the DebugInfoBuilder for this variable
```
cesarzc, line 203: Nitpick (not blocking): In this codebase in general, the first letter in a comment is capitalized.

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -132,6 +158,225 @@ internal void FinalizeDebugInfo()
...
+        /// with the global function.
+        /// </summary>
+        /// <param name="spec">The <see cref="QsSpecialization"/> for the function.</param>
+        /// <param name="mangledName">The mangled name for the function.</param>
+        /// <param name="signature">The signature for the function.</param>
+        /// <param name="isDefinition">Whether this is a function definition (rather than a declaration).</param>
+        /// <returns>The <see cref="IrFunction"/> corresponding to the given Q# type.</returns>
+        internal IrFunction CreateGlobalFunction(QsSpecialization spec, string mangledName, IFunctionType signature, bool isDefinition)
+        {
+            if (this.DebugFlag && spec.Kind == QsSpecializationKind.QsBody)
```
cesarzc, line 266: Since many of these functions do nothing when `this.DebugFlag` is `false`, to separate this mechanism from other conditions, I would suggest that `this.DebugFlag` is evaluated individually at the beginning of these methods and that the method returns if it is set to false:
```
if (!this.DebugFlag)
{
    return;
}

// Rest of the logic.
```

This is just a suggestion so it is not blocking.

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -132,6 +158,225 @@ internal void FinalizeDebugInfo()
...
+                string sourcePath = spec.Source.CodeFile;
+                DebugInfoBuilder curDIBuilder = this.GetOrCreateDIBuilder(sourcePath);
+                DIFile debugFile = curDIBuilder.CreateFile(sourcePath);
+
+                // set up debug info for the function
+                IDebugType<ITypeRef, DIType>? retDebugType;
+                IDebugType<ITypeRef, DIType>?[] argDebugTypes;
+                retDebugType = this.DebugTypeFromQsharpType(spec.Signature.ReturnType, curDIBuilder);
+                argDebugTypes = spec.Signature.ArgumentType.Resolution is ResolvedTypeKind.TupleType ts ? ts.Item.Select(t => this.DebugTypeFromQsharpType(t, curDIBuilder)).ToArray() :
+                    new IDebugType<ITypeRef, DIType>?[] { this.DebugTypeFromQsharpType(spec.Signature.ArgumentType, curDIBuilder) };
```
cesarzc, line 278:
```suggestion
                argDebugTypes = spec.Signature.ArgumentType.Resolution is ResolvedTypeKind.TupleType ts ?
                    ts.Item.Select(t => this.DebugTypeFromQsharpType(t, curDIBuilder)).ToArray() :
                    new IDebugType<ITypeRef, DIType>?[] { this.DebugTypeFromQsharpType(spec.Signature.ArgumentType, curDIBuilder) };
```

cesarzc: Just a legibility change.

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -170,46 +404,132 @@ private DebugInfoBuilder GetOrCreateDIBuilder(string sourcePath)
...
+                    offset += range.Start;
                 }
-
-                this.GetOrCreateDIBuilder(sourcePath);
             }
+
+            return offset;
         }
+
+// SECTION: Private getters
```
cesarzc, line 501: I think this private getters are not needed.

ryan-moreno: I included them in case we ended up needing to perform some logic to when deciding the versions (like check the target architecture etc.) But changing it if/when we need to do that shouldn't be difficult, so should I remove them and return the values directly?

src/QsCompiler/QirGeneration/DebugInfoManager.cs
ryan-moreno: The sections are mostly just for me while I'm developing in this file. I added a TODO note to remove the sections before merging to main. Is that ok to do?

src/QsCompiler/QirGeneration/DebugInfoManager.cs
ryan-moreno: If I have a private nested class, then I don't think I can access the member variables in side the class directly (I can't make them public since the class is private) without also creating getters for each config value. This seems like it would make it a bit messy (similar to the getters I had for the CodeviewVersion etc that were really just returning a private string. What do you think?

cesarzc: I believe you would be able to do it if the class member variables are public. The class wouldn't be accessible from outside the `DebugInfoManager` class but it would be accessible from inside which is the intent.

I don't think this is blocking so we can go ahead without making this change.

src/QsCompiler/QirGeneration/DebugInfoManager.cs
cesarzc: Yes, that's ok for this PR.

cesarzc (approved):

src/Passes/Source/Validator/Validator.cpp.orig
```diff
@@ -0,0 +1,167 @@
+// Copyright (c) Microsoft Corporation.
```
cesarzc, line 1: Did you mean to include file in source control?

cesarzc (changes requested):
It seems like many more (possibly unrelated) files are now appearing in this PR (maybe due to a merge). Can you please check why this happened and revert the merge such that only relevant changes appear?

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -5,63 +5,94 @@
...
-        /// </summary>
-        private static readonly uint DwarfVersion = 4;
+// TODO: remove SECTIONS before merging to main
+// SECTION: Config values
+        private static class DebugInfoConfig
+        {
+            /// <summary>
+            /// Whether or not to emit debug information during QIR generation
+            /// </summary>
+            private static bool debugFlag = false;
```
cesarzc, line 40: Since this is a static class, `DebugFlag` can be public and not readonly. That way you don't need the getter and setter either.

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -5,63 +5,94 @@
...
     /// </summary>
     internal sealed class DebugInfoManager
     {
-        /// <summary>
-        /// Dwarf version we are using for the debug info in the QIR generation
-        /// </summary>
-        private static readonly uint DwarfVersion = 4;
+// TODO: remove SECTIONS before merging to main
+// SECTION: Config values
+        private static class DebugInfoConfig
```
cesarzc, line 35: Since this is a private nested class, it could just be named `Config`,

src/QsCompiler/QirGeneration/DebugInfoManager.cs
ryan-moreno: If I make it public and not readonly I get the following error: [repo path]\src\QsCompiler\QirGeneration\DebugInfoManager.cs(40,32): warning SA1401: Field should be private [[repo path]\src\QsCompiler\QirGeneration\QirGeneration.csproj]. If I make it public and readonly the error goes away (but it shouldn't be readonly).

cesarzc: In this case using the following syntax should get rid of the warning:
```csharp
public bool DebugSymbolsEnabled { get; set; } = false;
```

I changed the name to `DebugSymbolsEnabled` to make clear the intent of the property.

cesarzc: Actually since this is a static class, the rest of the properties should follow the same pattern and not use `readonly`. Rather just having a `get` but not a `set`.

cesarzc (approved):

PR #1248, opened 2021-11-10
Allowing to use the LLVM intrinsic readcyclecounter from Q#

bettinaheim:


src/QsCompiler/Core/Dependencies.fs
```diff
@@ -238,6 +239,12 @@ type BuiltIn =
     static member EnableTestingViaName =
         { FullName = { Name = "EnableTestingViaName"; Namespace = BuiltIn.DiagnosticsNamespace }; Kind = Attribute }
 
+    static member ReadCycleCounter =
+        {
+            FullName = { Name = "ReadCycleCounter"; Namespace = BuiltIn.DiagnosticsNamespace }
```
swernli, line 244: I think this should be in the Core namespace or the intrinsic namespace. Diagnostic feels odd to me, though perhaps there is a logic I'm missing.

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestBuiltInIntrinsics1.ll
```diff
@@ -50,6 +50,7 @@ entry:
   store %String* %24, %String** %27, align 8
   call void @__quantum__rt__callable_invoke(%Callable* %19, %Tuple* %25, %Tuple* null)
   call void @__quantum__rt__callable_invoke(%Callable* %14, %Tuple* null, %Tuple* null)
+  %28 = call i64 @llvm.readcyclecounter()
```
swernli, line 53: Awesome! So glad you got this working, I was super stuck.

swernli (approved):
One minor thing about namespace, I think I'd prefer Core, but I defer to you to decide what you think is best for this one.

src/QsCompiler/Core/Dependencies.fs
bettinaheim: I suggest to put it into a new namespace Microsoft.Quantum.LlvmIntrinsics.

PR #1249, opened 2021-11-10
Q# Formatter Telemetry

vxfield:
Adds telemetry information to the Q# formatter tool. Data reported in telemetry include: execution start time, execution duration, which command-line options were used, number of source files processed, transformation rules warnings, exit code, and unhandled exception class name if one occurred.


vxfield:
@samarsha, Fantomas is crashing when trying to format or check [this file](https://github.com/microsoft/qsharp-compiler/blob/03415aadf4ffb994874428034b07f2b85bd9255b/src/QsFmt/App/Telemetry.fs).

Do you have an idea why?
I was able to run it before, it updated the file but now it's in a way that is breaking the tool with this error message:

> Fantomas is trying to format the input multiple times due to the detect of multiple defines.
> There is a problem with merging all the code back together. Please raise an issue at 

For now I just put the Telemetry.fs file in the .fantomasignore to unblock the CI build

vxfield:
> @samarsha, Fantomas is crashing when trying to format or check [this file](https://github.com/microsoft/qsharp-compiler/blob/03415aadf4ffb994874428034b07f2b85bd9255b/src/QsFmt/App/Telemetry.fs).
> 
> Do you have an idea why? I was able to run it before, it updated the file but now it's in a way that is breaking the tool with this error message:
> 
> > Fantomas is trying to format the input multiple times due to the detect of multiple defines.
> > There is a problem with merging all the code back together. Please raise an issue at
> 
> For now I just put the Telemetry.fs file in the .fantomasignore to unblock the CI build

I found the issue.
I had a compiler directive in the middle of some constructor parameters.

```f#
let telemetryConfig =
    TelemetryManagerConfig(
        SendTelemetryInitializedEvent = false,
        SendTelemetryTearDownEvent = false,
        #if TELEMETRY
        TestMode = false
        #else
        TestMode = true
        #endif
    )
```

I moved them out of there and it's working now.

src/QsFmt/App/Program.fs
```diff
@@ -33,93 +34,107 @@ let run arguments inputs =
...
-                3
+                { RunResult.Default with ExitCode = ExitCode.IOError }
             | :? UnauthorizedAccessException as ex ->
                 eprintfn "%s" ex.Message
-                4
+                { RunResult.Default with ExitCode = ExitCode.UnauthorizedAccess }
 
-    and doMany arguments inputs =
-        inputs |> Seq.fold (fun (rtrnCode: int) filePath -> max rtrnCode (filePath |> doOne arguments)) 0
+    and foldResults (previousRunResult: RunResult) filePath =
```
ScottCarda-MS, line 74: `doOneInput` and `doManyInputs` are `and`-ed together because they are mutually recursive, but `foldResults` can be independent. If you move its definition to above `doOneInput` then it doesn't need to be `and`-ed with the others. I'm not entirely sure, but I _think_ it is more efficient in F# to avoid `and`-ing when possible, but also it just disentangles otherwise independent pieces.

src/QsFmt/App/Program.fs
```diff
@@ -33,93 +34,107 @@ let run arguments inputs =
...
+    use _telemetryManagerHandle = Telemetry.initializeTelemetry args
+
+    let parseArgsResult =
+        try
+            assemblyLoadContextSetup ()
+
+            Ok(
+                CommandLine.Parser.Default.ParseArguments<FormatArguments, UpdateArguments, UpdateAndFormatArguments>
+                    args
+            )
```
ScottCarda-MS, line 103:
```suggestion
                CommandLine.Parser.Default.ParseArguments<FormatArguments, UpdateArguments, UpdateAndFormatArguments> args |> Ok
```

src/QsFmt/App/Program.fs
```diff
@@ -33,93 +34,103 @@ let run arguments inputs =
...
+            | Result.Error ex -> ex |> Result.Error
+        with
+        | ex -> ex |> Result.Error
+
+    Telemetry.logExecutionCompleted commandWithOptions runResult startTime executionTime.Elapsed
+
+    match runResult with
+    | Ok runResult -> runResult.ExitCode |> int
+    | Error ex ->
+        eprintf "Unexpected Error: %s" (ex.ToString())
```
vxfield, line 135: @ScottCarda-MS , this is the place that you might want to throw the exception that was previously capture if you want the exception to crash the program execution at the end (after logging it).

ScottCarda-MS (approved):

PR #1250, opened 2021-11-10
Updates for QIR validation

troelsfr:
This PR makes minor updates to make QAT more robust and provide better QIR validation.


src/Passes/Source/Rules/Factory.cpp
```diff
@@ -632,11 +632,16 @@ namespace quantum
 
     void RuleFactory::disableStringSupport()
     {
-        removeFunctionCall("__quantum__rt__string_create");
-        removeFunctionCall("__quantum__rt__string_update_reference_count");
-        removeFunctionCall("__quantum__rt__string_update_alias_count");
-        removeFunctionCall("__quantum__rt__message");
         removeFunctionCall("__quantum__rt__fail");
+        removeFunctionCall("__quantum__rt__message");
+        removeFunctionCall("__quantum__rt__bool_to_string");
```
bettinaheim, line 637: There are a couple missing compared to what is in the spec. See here for the full list of string-related runtime functions: https://github.com/microsoft/qsharp-language/blob/main/Specifications/QIR/Data-Types.md#strings

src/Passes/Source/Validator/Validator.cpp
```diff
@@ -48,37 +71,94 @@ namespace quantum
...
     {
         return gscc_analysis_manager_;
     }
+
     llvm::ModuleAnalysisManager& Validator::moduleAnalysisManager()
     {
         return module_analysis_manager_;
     }
 
+    void Validator::saveLogsToFileIfNeeded()
```
bettinaheim, line 112: I would expect such a helper to be in the logger rather than part of the validator.

bettinaheim (approved):
lgtm

src/Passes/Source/Rules/Factory.cpp
troelsfr: I've updated adding all functions. It feels like we should consider what our string elimination strategy should be. Regardless, we probably should emit warnings for every function removed without further processing.

src/Passes/Source/Validator/Validator.cpp
troelsfr: The idea is that this will write a validation report even if it just writing a log to JSON format atm. Hence it is meant to be beyond just writing a log. I've updated the function name to reflect this.

PR #1255, opened 2021-11-12
Structure for QAT + stdlib + examples

troelsfr:


src/Passes/qir/qsharp/analysis.py
```diff
@@ -0,0 +1,52 @@
+import os
```
bettinaheim, line 1: Is this file documented in a readme somewhere?

src/Passes/qir/stdlib/lib/stdlib.ll
```diff
@@ -0,0 +1,507 @@
+; ModuleID = 'src/stdlib.c'
+source_filename = "src/stdlib.c"
+target datalayout = "e-m:o-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
+target triple = "x86_64-apple-macosx11.0.0"
```
bettinaheim, line 4: My expectation is that nothing depending on this target triple for the code in question. Would it make sense to remove it and let llvm default to the current machine as needed?

src/QsCompiler/LanguageServer/TemporaryProject.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation.
```
bettinaheim, line 1: This file was accidentally committed and should be removed.

bettinaheim (approved):
I haven't looked very closely at the stdlib.c, but am happy to approve so we can move ahead with structural change/migration. Before it is more broadly used, we should make another pass to carefully look over it.

src/Passes/qir/qsharp/analysis.py
troelsfr: No, this is just a copy from the examples repo and it is WiP

src/Passes/qir/stdlib/lib/stdlib.ll
troelsfr: It is auto-generated when running the `clang`. The only way to remove it consistently is post-processing. We probably should decide on a triplet for quantum machines as it _does_ impact how optimisations are performed by LLVM.

src/QsCompiler/LanguageServer/TemporaryProject.cs
troelsfr: Removed.

PR #1257, opened 2021-11-12
Fix hover info for declarations of `for` and `use` statements

samarsha:
* Fix a regression introduced on the feature/lambda branch that didn't include the `for` loop variable, or the declared qubits from `use`, in `LocalsInScope` when `inclusive: true`.
* Add `seq` interface to `QsNullable` for convenience.
* Remove deprecated `QsNullable<'T>.FromOption`.

### DIFF
--- a/src/QsCompiler/CompilationManager/EditorSupport/SymbolInfo.cs
+++ b/src/QsCompiler/CompilationManager/EditorSupport/SymbolInfo.cs
@@ -124,58 +124,65 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
         internal static LocalDeclarations LocalsInScope(QsScope scope, Position position, bool inclusive)
         {
             var (parent, statementsBefore, _) = SplitStatementsByPosition(scope, position);
-            var declarationsBefore = parent.KnownSymbols.Variables
-                .Concat(statementsBefore.SkipLast(1).SelectMany(s => s.SymbolDeclarations.Variables));
             var currentStatement = statementsBefore.LastOrDefault();
+            var varsBefore = parent.KnownSymbols.Variables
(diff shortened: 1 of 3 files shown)


src/QsCompiler/CompilationManager/EditorSupport/SymbolInfo.cs
```diff
@@ -124,58 +124,70 @@ bool RangeOverlaps(QsNullable<Range> range)
...
+            var statementVars = currentStatement.Statement switch
+            {
+                QsStatementKind.QsForStatement f => f.Item.Body.KnownSymbols.Variables,
+                QsStatementKind.QsQubitScope q => q.Item.Body.KnownSymbols.Variables,
+                _ => currentStatement.SymbolDeclarations.Variables.Concat(varsBefore),
+            };
 
-            IEnumerable<LocalVariableDeclaration<string>> ExpressionDeclarations(TypedExpression expr) =>
-                DeclarationsInExpression(expr, position - currentStatement.Location.Item.Offset);
+            return new LocalDeclarations((expressionVars ?? statementVars).ToImmutableArray());
```
samarsha, line 159: This is the key part that fixes the bug - the previous version of LocalsInScope didn't include the child scope's environment of QsForStatement and QsQubitScope. However, we only want that if the position *isn't* in the statement's expression, so expressionVars is now null in that case. @ScottCarda-MS

ScottCarda-MS: I see, that makes sense. Thanks!

ScottCarda-MS (approved):

src/QsCompiler/CompilationManager/EditorSupport/SymbolInfo.cs
ScottCarda-MS: I see, that makes sense. Thanks!

PR #1259, opened 2021-11-16
Adding language server telemetry events

bettinaheim:


src/QsCompiler/CompilationManager/ProjectManager.cs
```diff
@@ -798,20 +798,33 @@ public bool ManagerTask(Uri file, Action<CompilationUnitManager> executeTask, ID
...
         /// it is called whenever diagnostics for the project have changed and are ready for publishing.
         /// <para/>
         /// Any exceptions caught during processing are logged using <paramref name="exceptionLogger"/>.
         /// </remarks>
-        public ProjectManager(Action<Exception>? exceptionLogger, Action<string, MessageType>? log = null, Action<PublishDiagnosticParams>? publishDiagnostics = null)
+        public ProjectManager(
+            Action<Exception>? exceptionLogger,
+            Action<string, MessageType>? log = null,
+            Action<PublishDiagnosticParams>? publishDiagnostics = null,
+            Action<string, Dictionary<string, string?>, Dictionary<string, int>>? sendTelemetry = null)
```
vxfield, line 819: I suggest using a well defined delegate here to make it easier to understand the purpose of each input parameter of the sendTelemetry action.
And maybe make this an event instead of a standard field.

src/QsCompiler/CompilationManager/ProjectManager.cs
```diff
@@ -1291,6 +1310,14 @@ editCollection switch
                             codeActionSuggestions = codeActionSuggestions.Append(
                                 ("Update deprecated syntax in file.", file.GetWorkspaceEdit(formattingEdits)));
                         }
+
+                        // send telemetry if telemetry is enabled
+                        var telemetryProps = new Dictionary<string, string?>();
+                        telemetryProps["quantumSdkVersion"] = c.BuildProperties.SdkVersion?.ToString();
```
vxfield, line 1316: This might not be needed. If the Extension version is supposed to be the same as QDK version, then we already get the version for every event emitted (every event includes the Extension version).

src/QsCompiler/LanguageServer/EditorState.cs
```diff
@@ -29,7 +28,7 @@ internal class EditorState : IDisposable
         public void Dispose() => this.projects.Dispose();
 
         private readonly Action<PublishDiagnosticParams> publish;
-        private readonly Action<string, Dictionary<string, string?>, Dictionary<string, int>> sendTelemetry;
+        private readonly Action<string, Dictionary<string, string?>, Dictionary<string, int>>? sendTelemetry;
```
vxfield, line 31: As before, I suggest using a well defined delegate here to make it clear what each input parameter of the action mean.
And maybe make this an event instead of a standard field.

src/QsCompiler/LanguageServer/LanguageServer.cs
```diff
@@ -185,6 +185,12 @@ internal async Task CheckDotNetSdkVersionAsync()
         /// </summary>
         internal void OnInternalError(Exception ex)
         {
+            var telemetryProps = new Dictionary<string, string?>();
```
vxfield, line 188: Thanks for finding this! I assumed we were capturing all errors/exceptions in the previous telemetry implementation!

src/QsCompiler/CompilationManager/ProjectManager.cs
```diff
@@ -1162,6 +1175,12 @@ editCollection switch
                 this.log?.Invoke("Failed to format document. Formatter may be unavailable.", MessageType.Info);
             }
 
+            // send telemetry if telemetry is enabled
+            var telemetryProps = new Dictionary<string, string?>();
+            telemetryProps["quantumSdkVersion"] = manager?.BuildProperties.SdkVersion?.ToString();
```
vxfield, line 1180: This might not be needed. If the Extension version is supposed to be the same as QDK version, then we already get the version for every event emitted (every event includes the Extension version).

src/QsCompiler/CompilationManager/ProjectManager.cs
```diff
@@ -1162,6 +1175,12 @@ editCollection switch
...
+            var telemetryMeas = new Dictionary<string, int>();
+            telemetryMeas["totalEdits"] = edits?.Count() ?? 0;
```
vxfield, line 1182: Nit
```suggestion
            var telemetryMeas = new Dictionary<string, int>()
            {
                { "totalEdits", edits?.Count() ?? 0 },
            };
```

src/QsCompiler/CompilationManager/ProjectManager.cs
```diff
@@ -1291,6 +1310,14 @@ editCollection switch
                             codeActionSuggestions = codeActionSuggestions.Append(
                                 ("Update deprecated syntax in file.", file.GetWorkspaceEdit(formattingEdits)));
                         }
+
+                        // send telemetry if telemetry is enabled
+                        var telemetryProps = new Dictionary<string, string?>();
+                        telemetryProps["quantumSdkVersion"] = c.BuildProperties.SdkVersion?.ToString();
+                        var telemetryMeas = new Dictionary<string, int>();
+                        telemetryMeas["qsfmtUpdateEdits"] = formattingEdits?.Count() ?? 0;
+                        telemetryMeas["totalEdits"] = codeActionSuggestions.Count();
```
vxfield, line 1319: Nit
```suggestion
                        var telemetryMeas = new Dictionary<string, int>()
                        {
                            { "qsfmtUpdateEdits", formattingEdits?.Count() ?? 0 },
                            { "totalEdits", codeActionSuggestions.Count() },
                        };
```

src/QsCompiler/LanguageServer/LanguageServer.cs
```diff
@@ -185,6 +185,12 @@ internal async Task CheckDotNetSdkVersionAsync()
...
+            telemetryProps["exceptionType"] = ex.GetType().ToString();
+            telemetryProps["message"] = ex.Message;
```
vxfield, line 190: Nit
```suggestion
            var telemetryProps = new Dictionary<string, string?>()
            {
                { "exceptionType" } = ex.GetType().FullName;
            }
```

vxfield: It would be nice to capture the Exception message for easier investigations, but it can contain customer data / PII and therefore we should avoid collecting it.

vxfield (changes requested):
Looks good overall, but I have not tested it to make sure it's sending the telemetry that it's supposed to send.

I have a few minor suggestions that it's up to you to accept.

My main request for change is to not log the Exception message as it can contain customer data or PII.

src/QsCompiler/CompilationManager/ProjectManager.cs
bettinaheim: The sdk version is the sdk version of the project that is being edited, not the version of the extension.

src/QsCompiler/LanguageServer/LanguageServer.cs
bettinaheim: No, this was long overdue. It would be great if we could actually have some tracking for this number, and raise some flags if the number goes up.

src/QsCompiler/CompilationManager/ProjectManager.cs
bettinaheim: Same command as above; this is the sdk version of the project rather than the extension.

src/QsCompiler/LanguageServer/LanguageServer.cs
bettinaheim: That's a good point, thanks. It is a bit unfortunate since the stack trace is what we would need to fix the issue, but I don't see a good way to get more information. I'll remove it.

src/QsCompiler/LanguageServer/LanguageServer.cs
cgranade: > ...and raise some flags if the number goes up.

We currently have similar telemetry on the VS Code extension and IQ#, but no flags if they go up; perhaps it would be good to include those as well?

vxfield: I'll create a work item for us to add monitoring and alerts for these events.

src/QsCompiler/LanguageServer/LanguageServer.cs
vxfield: I wrote a [(private static) method](https://github.com/microsoft/qsharp-compiler/blob/8da3079da3c5387f12e0dfd059d4cb4ab6cf0ca2/src/Telemetry/Library/TypeConversionHelper.cs#L197) in a Telemetry Library helper class to sanitize the StackTrace of an exception.
If you want to take a dependency on the Telemetry Library, you could just make that method public and use it.
Another option would be to copy that code, but I would prefer to have it in a single place to make it easier to fix bugs in the future.

vxfield (approved):

PR #1260, opened 2021-11-16
Add aggressive inline pass to QAT

swernli:
This adds an aggresive inline pass to QAT that will ensure functions remaining at the end of transformations get inlined even beyond the default limit of 250 lines.


troelsfr (approved):
LGTM

PR #1265, opened 2021-11-18
Removing the qat tool due to migration

bettinaheim:
The qat tool has been migrated here: https://github.com/qir-alliance/qat


swernli (approved):

cesarzc (approved):

bettinaheim:
It looks like also after removing the repo build is failing due to a security alert for json-schema 0.2.3. I believe this will need to be resolved separately and suggest to go ahead with the merge.

PR #1267, opened 2021-11-23
Add testing for existing debug info generation

ryan-moreno:
This PR adds tests for the debug info generation already in the feature branch. It also makes 2 bug fixes related to debug info generation.


src/QsCompiler/CommandLineTool/Commands/Build.cs
```diff
@@ -79,14 +79,6 @@ public static IEnumerable<Example> UsageExamples
                 HelpText = "Destination folder where the output of the performance assessment will be generated.")]
             public string PerfOutputFolder { get; set; }
 
-            [Option(
```
cesarzc, line 82: Why was this option removed?

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -318,13 +317,12 @@ internal IrFunction CreateGlobalFunction(QsSpecialization spec, string mangledNa
...
                         signature: debugSignature,
                         isLocalToUnit: true, // We're using the compile unit from the source file this was declared in
-                        isDefinition: isDefinition,
+                        isDefinition: true, // if this isn't set to true, it results in temporary metadata nodes that don't get resolved.
                         scopeLinePosition: ConvertToDebugPosition(absolutePosition), // TODO: Could make more exact bc of formatting and white space (see lastParamLocation in Kaleidescope tutorial)
                         debugFlags: DebugInfoFlags.None,
                         isOptimized: false,
                         curDIBuilder);
 
-                    this.EmitLocation(absolutePosition, func.DISubProgram);
```
cesarzc, line 327: Why invoking `EmitLocation` is no longer needed?

src/QsCompiler/QirGeneration/RewriteSteps.cs
```diff
@@ -74,8 +74,17 @@ public bool PreconditionVerification(QsCompilation compilation)
         /// <inheritdoc/>
         public bool Transformation(QsCompilation compilation, out QsCompilation transformed)
         {
+            bool debugSymbolsEnabled = false;
+            if (this.AssemblyConstants.TryGetValue(ReservedKeywords.AssemblyConstants.DebugSymbolsEnabled, out string? debugEnabledString))
```
cesarzc, line 78: Is this a workaround to pass whether debug symbols are enabled to the QirGeneration rewrite step?

cesarzc:
Was this file: src/QsCompiler/Tests.Compiler/TestCases/DebugInfoTests/TestFunctionReturnsUnit8.ll meant to be empty?

src/QsCompiler/CommandLineTool/Commands/Build.cs
ryan-moreno: I had added the option a while back expecting to use this method as the gating. Instead, I made it so that you specify whether you want debug information in the same way that you specify if you want qir to be emitted & where it should be emitted to (within the assembly properties)

src/QsCompiler/QirGeneration/DebugInfoManager.cs
ryan-moreno: It was redundant with other debug locations, and the scoping was slightly off. This emitted a debug location wherever the function is declared, yet set the scope to the function itself. The debug locations should be set within the scope of the function only when steps are executed within the function. And then they should be set within the scope of the calling function when this function is called. Those situations are taken care of elswhere.

src/QsCompiler/QirGeneration/RewriteSteps.cs
ryan-moreno: That's what it's for, but I don't think it's a workaround. I think this is the way the assembly constants are meant to be used. Shortly after this the Qir Output Path is received using the same method. This just needs to occur earlier because it is relevant whether debug symbols are enabled before the generator is applied whereas the output path is only relevant after the generator is applied and we have the content of the QIR.

ryan-moreno:
> Was this file: src/QsCompiler/Tests.Compiler/TestCases/DebugInfoTests/TestFunctionReturnsUnit8.ll meant to be empty?

Not sure how to reply normally, but now I must have accidentally deleted the content. I'll fix that.

src/QsCompiler/QirGeneration/RewriteSteps.cs
cesarzc: Sounds good, thanks for the clarification Ryan!

cesarzc (approved):

PR #1275, opened 2021-12-04
Replace illegal characters in namespace name

warren-jones:
A Q# project's Namespace name must be alphanumeric, except for dots and underscores.  When using “Q#: Create new project…” from the VS Code Command Palette, a user may specify a valid project name containing hyphens, but the namespace name in the automatically generated Program.qs will contain underscores in place of the hyphens.

Dots are left in place where valid, but other special characters are replaced with underscores.  Other character combination restrictions are described in https://docs.microsoft.com/en-us/azure/quantum/user-guide/language/programstructure/namespaces


warren-jones:
Found a bug in the bug fix while adding test cases to the unit test -- changed the PR back to "Draft"

warren-jones:
Fixed the bug I mentioned yesterday.  Working on unit tests.

ricardo-espinoza (approved):

Issue #1115, opened 2021-08-11
Nested Conditional Blocks are Lifted when they Shouldn't Be


ScottCarda-MS:
**Describe the bug**

When a conditional includes nested conditionals that are invalid for lifting, the outer conditional may incorrectly be lifted into a generated callable by the ConditionallyControlled rewrite step. See discussion with concrete examples in comments here #767 (comment).

**To Reproduce**

Remove the Skip condition from [this test](https://github.com/microsoft/qsharp-compiler/blob/main/src/QsCompiler/Tests.Compiler/ClassicalControlTests.fs#L1631) in the ClassicallyControlledSupportTests.qs and run it. The test contains a nested condition with a return statement in it. Neither of the conditionals should be lifted into their own callables because of the return statement. However, the outer condition does get lifted, resulting in incorrect behavior.

**Expected behavior**

The test should pass. Because the inner conditional contains a return statement, neither the inner conditional nor the outer conditional should be lifted into their own callables.

bettinaheim:
@ScottCarda-MS Based on the description, I thought this should be resolved with https://github.com/microsoft/qsharp-compiler/pull/1114 - is that not correct? The test case in the description is still commented out. If the issue is resolved with #1114, please enable the test case and close this issue.

ScottCarda-MS:
> @ScottCarda-MS Based on the description, I thought this should be resolved with #1114 - is that not correct? The test case in the description is still commented out. If the issue is resolved with #1114, please enable the test case and close this issue.

This is not resolved with #1114. This is the more fundamental issue with lifting that has yet to be addressed. #1114 just prevented this issue from surfacing to the level of mattering for hardware jobs.

ScottCarda-MS:
Resolved with #1277

PR #1277, opened 2021-12-10
Base Lifter Cleanup

ScottCarda-MS:
This PR makes some structural changes to the Base Lifter to allow it to be more flexible and easier to work with. A functional change is that the generated callable will now only have the minimal parameter set, rather than taking all the known symbols as parameters. This PR also addresses #1115.

### DIFF
--- a/src/QsCompiler/Tests.Compiler/QirTests.fs
+++ b/src/QsCompiler/Tests.Compiler/QirTests.fs
@@ -8,9 +8,16 @@ open Microsoft.Quantum.QsCompiler.CommandLineCompiler
 open Xunit
 open System.Reflection
 open System.Text.RegularExpressions
+open System
 
 let private GUID =
     new Regex(@"[({]?[a-fA-F0-9]{8}[-]?([a-fA-F0-9]{4}[-]?){3}[a-fA-F0-9]{12}[})]?", RegexOptions.IgnoreCase)
+
+/// <summary>
+/// Ensures that the new line characters will conform to the standard of the environment's new line character.
+/// </summary>
+let private standardizeNewLines (s: string) =
+    s.Replace("\r", "").Replace("\n", Environment.NewLine)
 
 let private testOne expected args =
     let result = Program.Main args
@@ -21,7 +28,8 @@ let private clearOutput name =
 
 let private checkAltOutput name actualText =
     let expectedText = ("TestCases", "QirTests", name) |> Path.Combine |> File.ReadAllText
-    Assert.Contains(expectedText, GUID.Replace(actualText, "__GUID__"))
+    let replacedGUID = GUID.Replace(actualText, "__GUID__")
+    Assert.Contains(standardizeNewLines expectedText, standardizeNewLines replacedGUID)
 
 let private compilerArgs target (name: string) =
     seq {
--- a/src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestTargeting.ll
+++ b/src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestTargeting.ll
@@ -3,13 +3,13 @@ entry:
   %q = call %Qubit* @__quantum__rt__qubit_allocate()
   %0 = call %Result* @__quantum__qis__mz(%Qubit* %q)
   %1 = call %Tuple* @__quantum__rt__tuple_create(i64 mul nuw (i64 ptrtoint (i1** getelementptr (i1*, i1** null, i32 1) to i64), i64 2))
-  %2 = bitcast %Tuple* %1 to { %Callable*, %Qubit* }*
-  %3 = getelementptr inbounds { %Callable*, %Qubit* }, { %Callable*, %Qubit* }* %2, i32 0, i32 0
-  %4 = getelementptr inbounds { %Callable*, %Qubit* }, { %Callable*, %Qubit* }* %2, i32 0, i32 1
+  %2 = bitcast %Tuple* %1 to { %Callable*, %Tuple* }*
+  %3 = getelementptr inbounds { %Callable*, %Tuple* }, { %Callable*, %Tuple* }* %2, i32 0, i32 0
+  %4 = getelementptr inbounds { %Callable*, %Tuple* }, { %Callable*, %Tuple* }* %2, i32 0, i32 1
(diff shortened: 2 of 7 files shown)


kuzminrobin (approved):
The `.ll`, `.qs`, and `.cs` files look good to me. 
Approving to unblock. But I would recommend an approval from someone more knowledgeable than me in this topic.

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestTargeting.ll
```diff
@@ -3,13 +3,13 @@ entry:
...
-  %4 = getelementptr inbounds { %Callable*, %Qubit* }, { %Callable*, %Qubit* }* %2, i32 0, i32 1
+  %2 = bitcast %Tuple* %1 to { %Callable*, %Tuple* }*
+  %3 = getelementptr inbounds { %Callable*, %Tuple* }, { %Callable*, %Tuple* }* %2, i32 0, i32 0
+  %4 = getelementptr inbounds { %Callable*, %Tuple* }, { %Callable*, %Tuple* }* %2, i32 0, i32 1
   %5 = call %Callable* @__quantum__rt__callable_create([4 x void (%Tuple*, %Tuple*, %Tuple*)*]* @Microsoft__Quantum__Testing__QIR_____GUID___NoArgs__FunctionTable, [2 x void (%Tuple*, i32)*]* null, %Tuple* null)
   store %Callable* %5, %Callable** %3, align 8
-  store %Qubit* %q, %Qubit** %4, align 8
-  call void @Microsoft__Quantum__ClassicalControl_____GUID___ApplyIfOne__body(%Result* %0, { %Callable*, %Qubit* }* %2)
+  store %Tuple* null, %Tuple** %4, align 8
+  call void @Microsoft__Quantum__ClassicalControl_____GUID___ApplyIfOne__body(%Result* %0, { %Callable*, %Tuple* }* %2)
```
samarsha, line 12: Why did this change?

src/QsCompiler/Transformations/ClassicallyControlled.cs
```diff
@@ -997,14 +1015,11 @@ public override QsStatementKind OnConditionalStatement(QsConditionalStatement st
...
                         }
-
-                        this.SharedState.GeneratedOpParams = contextParams;
-                        this.SharedState.IsValidScope = contextValidScope;
                     }
 
                     if (this.SharedState.IsConditionLiftable)
                     {
-                        this.SharedState.GeneratedOperations?.AddRange(generatedOperations);
+                        this.SharedState.GeneratedCallables?.AddRange(generatedOperations);
```
samarsha, line 1022: Is it valid for `this.SharedState.GeneratedCallables` to be null here?

ScottCarda-MS: It should not be null here, as the GeneratedCallables should get a new List at the start of every namespace.

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestTargeting.ll
ScottCarda-MS: The behavior for lifted callables changed to take only a minimal parameter set based on the variables from the calling context that are actually used in the lifted body, whereas before it would just take all known variables as parameters. This change in parameter behavior is reflected here by removing the expected qubit parameter to the lifted callable and replacing it with an empty tuple, as the qubit was not used in the lifted body.

src/QsCompiler/Tests.Compiler/QirTests.fs
```diff
@@ -21,7 +28,8 @@ let private clearOutput name =
 
 let private checkAltOutput name actualText =
     let expectedText = ("TestCases", "QirTests", name) |> Path.Combine |> File.ReadAllText
-    Assert.Contains(expectedText, GUID.Replace(actualText, "__GUID__"))
+    let replacedGUID = GUID.Replace(actualText, "__GUID__")
+    Assert.Contains(standardizeNewLines expectedText, standardizeNewLines replacedGUID)
```
swernli, line 32: I'm glad this ended up working, but looking again now I can't help but wonder why it is needed for only this test. I think this may have something to do with the git settings on your machine, but for now we can leave it as-is since it's technically safer this way anyway.

src/QsCompiler/Transformations/ContentLifting.cs
```diff
@@ -456,12 +465,225 @@ public TypeTransformation(SyntaxTreeTransformation<TransformationState> parent)
...
+                    {
+                        if (this.SharedState.ReturnType == null)
+                        {
+                            this.SharedState.ReturnType = ex.ResolvedType;
+                        }
+                        else
+                        {
+                            this.SharedState.ReturnType = this.ValidateAndCombineReturnTypes(this.SharedState.ReturnType, ex.ResolvedType);
+                        }
+                    }
```
swernli, line 642: Do we have any tests for lifting when return is allowed, or is that always disabled in our currently supported scenarios?

swernli: Looks like we always pass false here... I'm curious when we anticipate we would start allowing returns from lifted functions, but in the meantime this is essentially untested dead code. Might be worth a comment calling it out as speculative implementation and highlighting when it might need to be used?

swernli (approved):

src/QsCompiler/Tests.Compiler/QirTests.fs
ScottCarda-MS: On my local machine this was causing all of the QIR tests to fail. This fix is just to allow easier use of running the tests locally.

src/QsCompiler/Transformations/ContentLifting.cs
ScottCarda-MS: There is not currently a way (and I can't think of any) to test the features of the base lifter directly. We can only test the features of the base lifter as they are used in derived lifters like that of the classical control lifter, which we do test. Since a lot of this PR was a refactoring of the base lifter to 1. clean it up and 2. prepare it for the derived lambda lifter, there were features added here that can't be tested directly until used by the lambda lifter. Return values are one and lifting in functions is another that I can think of.

ScottCarda-MS: I've added a comment per your suggestion.

src/QsCompiler/Transformations/ClassicallyControlled.cs
ScottCarda-MS: Changed to use the null-forgiveness operator instead.

PR #1278, opened 2021-12-10
Handle display of variable values in debug info

ryan-moreno:
This PR makes it so that values are displayed for variables of simple non-quantum types, and for other types the variable is displayed with a note that the type is not yet supported for value display.

It also sets up a method for testing the debug info feature within the development project and updates and adds to the unit tests.


src/QsCompiler/QirGeneration/QIR/DataStructures.cs
```diff
@@ -140,7 +140,7 @@ internal SimpleValue(Value value, ResolvedType type)
     internal class PointerValue : IValue
     {
         private readonly IValue.Cached<IValue> cachedValue;
-        private readonly Value pointer;
+        internal readonly Value Pointer;
```
ryan-moreno, line 143: This causes a warning that this field should be private. What is the proper way to access this field?

examples/QIR/Development/DebugInfo/CoreTypes.hpp
```diff
@@ -0,0 +1,58 @@
+// Copyright (c) Microsoft Corporation.
```
ryan-moreno, line 1: These files are here so that the development QIR can easily be compiled to an executable. However, they are library files that would need updating when they change. Is there a better way to make the development process easier for the user without including all these files? I did at least put a note in the README that these files may need to be updated.

cesarzc: These files are included as part of the QIR runtime package. In order to compile the executable, the library files should have been pulled from somewhere and it is likely these headers can be pulled from the same place.

cesarzc:
Was this [README](https://github.com/microsoft/qsharp-compiler/pull/1278/files#diff-34ee18d211c850c3022a6f1dfc552c0bb545babab84dd46c61c66a0fa69c6769) meant to be empty?

cesarzc:
Similar to QIR runtime headers, QIR runtime binaries ([ibMicrosoft.Quantum.Qir.QSharp.Core.so](https://github.com/microsoft/qsharp-compiler/pull/1278/files#diff-ac4a94500ed5c9665b11048d954eccb0c93324219edcc691ddef27574e2efb75), [libMicrosoft.Quantum.Qir.Runtime.so](https://github.com/microsoft/qsharp-compiler/pull/1278/files#diff-ce7bfb2cdd3d5803ac775ee0b4b8201d374d6b61724d368e1367dc17ce0bcdd3), [libMicrosoft.Quantum.Simulator.Runtime.so](https://github.com/microsoft/qsharp-compiler/pull/1278/files#diff-48b99615843e120eb164721627ad2190db35c32bfaedf019161f96e9ffb3d322)) shouldn't be in source control.

cesarzc (changes requested):
Given that there's some files that are originally from the qsharp-runtime repo and some binaries that are built for a specific platform, we should address fixing their usage here such that they are not in source control.

examples/QIR/Development/DebugInfo/CoreTypes.hpp
ryan-moreno: You're right, these are all in the build folder so I'll adjust accordingly

ryan-moreno:
Not sure how to reply to this without a quote reply, but I removed all these and changed the process for running the debugger on the Q# file in the development folder

ryan-moreno:
Fixed :)

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -228,14 +230,18 @@ internal void CreateLocalVariable(string name, IValue value)
             DIType? dIType = resType == null ? null : this.DebugTypeFromQsharpType(resType, dIBuilder)?.DIType;
 
             // Get the location information for the variable declaration
-            QsLocation? namespaceLocation = this.sharedState.DIManager.CurrentNamespaceElementLocation;
+            QsLocation? namespaceLocation = this.CurrentNamespaceElementLocation;
             Position namespaceOffset = namespaceLocation?.Offset ?? Position.Zero;
             Position absolutePosition = namespaceOffset + this.TotalOffsetFromStatements() + this.TotalOffsetFromExpressions();
 
+            // LLVM Native API has a bug and will crash if we pass in a null dIType
+                // for either CreateLocalVariable or InsertDeclare (https://bugs.llvm.org/show_bug.cgi?id=52459)
```
cesarzc, line 238:
```suggestion
            // for either CreateLocalVariable or InsertDeclare (https://bugs.llvm.org/show_bug.cgi?id=52459)
```

cesarzc: nit: indentation

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -254,12 +260,24 @@ internal void CreateLocalVariable(string name, IValue value)
 
                 if (this.sharedState.CurrentBlock != null)
                 {
-                    // Create the debug info for the local variable declaration
-                    dIBuilder.InsertDeclare(
-                        storage: value.Value,
+                    if (isMutableBinding)
+                    { // variable is represented as a pointer to the value
```
cesarzc, line 264: nit: It would be good if the comment was in its own line. Same for the other similar comment after the `else` statement.

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -302,8 +320,9 @@ internal IrFunction CreateGlobalFunction(QsSpecialization spec, string mangledNa
                     return this.CreateFunctionNoDebug(mangledName, signature);
                 }
 
-                Position absolutePosition = debugLocation.Item.Offset; // The position stored in a QsSpecialization is absolute, not relative to the ancestor namespace and statements
+                Position absolutePosition = debugLocation.Item.Offset; // the position stored in a QsSpecialization is absolute, not relative to the ancestor namespace and statements
```
cesarzc, line 323: nit: comments usually start with an uppercase character.

cesarzc: Same for other comments in the source code.

cesarzc (approved):

PR #1285, opened 2021-12-15
Fix exponent and '+' unary operator bugs in QsFmt

### LINKED ISSUE #1283: Q# Formatter fails on valid Q# code
**Describe the bug**

I used the Q# Formatter to update the deprecated code in the Katas, and on several of the projects it failed to parse the code. The same code builds successfully with Q# compiler, so it is not the case of invalid Q# code.

**To Reproduce**

1. Clone the [katas repo](https://github.com/microsoft/QuantumKatas/) and go to the KeyDistribution_BB84 folder.
2. Run `dotnet msbuild -t:UpdateSyntax`.
3. See error
    c:\GitHub\QuantumKatas\KeyDistribution_BB84\Tests.qs, Line 36, Character 121: extraneous input ')' expecting ';'
    c:\GitHub\QuantumKatas\KeyDistribution_BB84\Tests.qs, Line 36, Character 118: extraneous input 'e' expecting {',', ')'}

Projects MagicSquareGame and PhaseEstimation fail as well on multiple lines, and are worth looking into.

**Expected behavior**

Q# formatter correctly processes any valid Q# code.

**System information**

- QDK 0.21.2111177148
- .NET Core 3.1.415

yjt98765:
This PR solves the problems reported in #1283.

1. The exponent symbols are not included in the `DoubleLiteral` rules with `.`, so tokens such as `1.2e3` cannot be lexed.
2. The `+` is not considered as a unary operator, making expressions like `+1` illegal.

This PR modifies the ANTLR grammar to fix these problems. Some test cases are added to FixedPoints.fs as well.

After this modification, all the .ps files in the [QuantumKatas repository](https://github.com/microsoft/QuantumKatas/) can be updated without errors.

close #1283


ScottCarda-MS (approved):
I have confirmed that this fixes the issue. Thanks @yjt98765 for another great contribution! As always, much appreciated!

samarsha (approved):

PR #1288, opened 2021-12-15
Emit debug info within QIR when requested

ryan-moreno:
This PR bundles all of the work on emitting debug info within QIR from my internship and merges it to main. The feature is gated, and debug info will only be emitted when requested (DebugSymbolsEnabled can be set to "true" in the same way that QirOutputPath is set). The feature is tested within the Tests.Compiler project.


ryan-moreno:
@bettinaheim you had some thoughts about my changes to the LLVM Bindings' public interfaces. When I merged the change to the feature branch that allowed for multiple DebugInfoBuilders per module, I believe that we talked about reverting the changes where I switched a module input to a DebugInfoBuilder (using a default selection if there is only one DebugInfoBuilder) and then adding another function that accepts the specific DebugInfoBuilder instead of the module. Just double checking that's how you'd like that to be formatted before I go through and make the changes. (This is referring to changes in the files like DebugArrayType, DebugBasicType, etc.)

ryan-moreno:
It looks like I can't request specific reviewers now that I'm not internal, but I think this would be relevant to @bettinaheim @cesarzc @swernli and @idavis. Thanks for taking a look!

README.md
```diff
@@ -35,7 +35,7 @@ Before you can build the source code on this repository and start contributing t
...
+The Q# compiler and language server in this repository are built using [.NET Core](https://docs.microsoft.com/en-us/dotnet/core/introduction#net-core-and-net-5). Building the [QsCompiler.sln](./QsCompiler.sln) builds the Q# compiler and language server. To test your changes to the compiler, open the project file of a Q# project that uses the latest version of the [Microsoft.Quantum.Sdk](https://www.nuget.org/packages/Microsoft.Quantum.Sdk/) in a text editor. You can confirm the Sdk version that the project is using by looking at the first line in the project file. You may need to edit that line to update to the latest version, and adjust your project as needed. Confirm that the project is building correctly using that version by executing the command
```
cesarzc, line 38:
```suggestion
The Q# compiler and language server in this repository are built using [.NET Core](https://docs.microsoft.com/dotnet/core/). Building the [QsCompiler.sln](./QsCompiler.sln) builds the Q# compiler and language server. To test your changes to the compiler, open the project file of a Q# project that uses the latest version of the [Microsoft.Quantum.Sdk](https://www.nuget.org/packages/Microsoft.Quantum.Sdk/) in a text editor. You can confirm the Sdk version that the project is using by looking at the first line in the project file. You may need to edit that line to update to the latest version, and adjust your project as needed. Confirm that the project is building correctly using that version by executing the command
```

cesarzc: This is just reverting the change since the link that used to return 404 error is now working and no longer failing validation pipelines.

examples/QIR/Development/README.md
```diff
@@ -8,3 +8,31 @@ Note: If you are using the Development project on Linux, you may run into the fo
 ```
 export LD_LIBRARY_PATH=/usr/lib/llvm-11/lib
 ```
+
+# Development of the QIR Debug Info Generation
+NOTE: This debug feature is currently incompatible with Windows due to some LLVM tool limitations. If you request debug symbols within Development.csproj while using the Windows platform, you will run into errors when you build the Development project. These errors come from compiling the emitted QIR (with debug information) into an executable, meaning that the QIR emission itself will still work on Windows, but you won't be able to actually test the Q# debugging experience.
```
cesarzc, line 13: Is this due to the problem with the .qs extension on Windows? If so, can you please expand this note adding a brief couple of sentences explaining what problem would someone run into when trying to test the new Q# debugging experience?

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -1256,5 +1256,13 @@ public static string GeneratedFile(string fileId, string outputFolder, string fi
             File.WriteAllText(targetFile, content);
             return targetFile;
         }
+
+        /// <summary>
+        /// Gets the executing assembly (the Qsharp Compiler) AssemblyName
+        /// </summary>
+        public static AssemblyName GetQSharpCompilerAssemblyName()
```
cesarzc, line 1263:
```suggestion
        public static AssemblyName QSharpCompilerAssemblyName =>
            Assembly.GetExecutingAssembly().GetName();
```

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -0,0 +1,575 @@
...
+            /// Returns a string representing the producer information for the QIR
+            /// </summary>
+            public static string GetQirProducerIdent()
+            {
+                AssemblyName compilationInfo = CompilationLoader.GetQSharpCompilerAssemblyName();
+                AssemblyName qirGenerationInfo = Assembly.GetExecutingAssembly().GetName();
+                return compilationInfo.Name + " with " + qirGenerationInfo.Name + " V " + qirGenerationInfo.Version;
+            }
+        }
+
```
cesarzc, line 69:
```suggestion
            public static string QirProducerIdentifier =>
                $"{CompilationLoader.QSharpCompilerAssemblyName.Name} with {DebugInfoProducerAssemblyName.Name} V {DebugInfoProducerAssemblyName.Version}";
        }

        private static AssemblyName DebugInfoProducerAssemblyName => Assembly.GetExecutingAssembly().GetName();

```

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -0,0 +1,575 @@
...
+        /// <param name="owningModule">The <see cref="BitcodeModule"/> that this DebugInfoManager is related to</param>
+        internal void AddTopLevelDebugInfo(BitcodeModule owningModule)
+        {
+            if (!Config.DebugSymbolsEnabled)
+            {
+                return;
+            }
+
+            // Add Module identity and Module Flags
+            owningModule.AddProducerIdentMetadata(Config.GetQirProducerIdent());
```
cesarzc, line 169:
```suggestion
            owningModule.AddProducerIdentMetadata(Config.QirProducerIdentifier);
```

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -0,0 +1,575 @@
...
+            {
+                return di;
+            }
+            else
+            {
+                di = this.Module.CreateDIBuilder();
+                di.CreateCompileUnit(
+                    Config.QSharpLanguage,
+                    sourcePath,
+                    Config.GetQirProducerIdent(),
```
cesarzc, line 465:
```suggestion
                    Config.QirProducerIdentifier,
```

src/QsCompiler/LlvmBindings/BitcodeModule.cs
```diff
@@ -120,19 +118,60 @@ private BitcodeModule(LLVMModuleRef handle)
             }
         }
 
-        /// <summary>Gets the <see cref="DebugInfoBuilder"/> used to create debug information for this module</summary>
-        /// <remarks>The builder returned from this property is lazy constructed on first access so doesn't consume resources unless used.</remarks>
+        /// <summary>Creates and returns a new <see cref="DebugInfoBuilder"/> used to create debug information for this module</summary>
+        public DebugInfoBuilder CreateDIBuilder()
+        {
+            this.ThrowIfDisposed();
```
cesarzc, line 124: Was `this.ThrowIfDisposed()` taken from the Llvm.NET repo?

src/QsCompiler/LlvmBindings/BitcodeModule.cs
```diff
@@ -383,6 +422,77 @@ public IrFunction CreateFunction(string name, IFunctionType signature)
...
+            uint line,
+            DebugFunctionType signature,
+            bool isLocalToUnit,
+            bool isDefinition,
+            uint scopeLine,
+            DebugInfoFlags debugFlags,
+            bool isOptimized,
+            DebugInfoBuilder dIBuilder)
+        {
+            this.ThrowIfDisposed();
```
cesarzc, line 460: I see `this.ThrowIfDisposed()` in multiple files. What is the purpose of this call?

src/QsCompiler/LlvmBindings/DebugInfo/DebugBasicType.cs
```diff
@@ -28,20 +28,19 @@ public class DebugBasicType
...
         /// <param name="name">Source language name of the type</param>
         /// <param name="encoding">Encoding for the type</param>
-        public DebugBasicType(ITypeRef llvmType, BitcodeModule module, string name, DiTypeKind encoding)
+        public DebugBasicType(ITypeRef llvmType, DebugInfoBuilder dIBuilder, string name, DiTypeKind encoding)
             : base(
                 llvmType,
-                module
-                          .DIBuilder
+                dIBuilder
                           .CreateBasicType(
```
cesarzc, line 38: Nit: this indentation is not consistent with the rest of the code now that the `module` part is no longer there.

src/QsCompiler/LlvmBindings/DebugInfo/DebugPointerType.cs
```diff
@@ -51,18 +51,17 @@ public DebugPointerType(ITypeRef llvmElementType, BitcodeModule module, DIType?
...
             : base(
                 llvmPtrType,
-                module.DIBuilder
-                          .CreatePointerType(
+                dIBuilder.CreatePointerType(
                               elementType,
                               name,
-                              module.Layout.BitSizeOf(llvmPtrType),
+                              dIBuilder.OwningModule.Layout.BitSizeOf(llvmPtrType),
                               alignment))
```
cesarzc, line 65: Nit: the indentation here is also a bit off now that `module` has been removed.

src/QsCompiler/LlvmBindings/LLVMSharpExtensions/LLVMDIBuilderRefExtensions.cs
```diff
@@ -4,6 +4,7 @@
 // </copyright>
 
 using System;
+using Ubiquity.NET.Llvm.DebugInfo;
```
cesarzc, line 7:
```suggestion
```

cesarzc: This doesn't seem to be needed.

src/QsCompiler/QirGeneration/Context.cs
```diff
@@ -661,6 +670,7 @@ internal void GenerateFunctionHeader(QsSpecialization spec, ArgumentTuple argTup
             var innerTuples = new Queue<(string?, ArgumentTuple)>();
             var outerArgItems = ArgTupleToArgItems(argTuple, innerTuples);
             var innerTupleValues = new Queue<TupleValue>();
+            int argNo = 1; // arg numbers are 1-indexed
```
cesarzc, line 673: Nit: consider moving this to just before the `for` loop where it is used.

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -0,0 +1,575 @@
...
+            /// <summary>
+            /// CodeView version we are using for the debug info in the QIR generation
+            /// </summary>
+            public static uint CodeviewVersion { get; } = 1;
+
+            /// <summary>
+            /// The source language information for Dwarf.
+            /// For now, we are using the C interface. Ideally this would be a user defined language for Q#.
+            /// </summary>
+            public static SourceLanguage QSharpLanguage { get; } = SourceLanguage.C99;
```
cesarzc, line 57: Since this is defined in DwarfEnumerations.cs, it could be expanded to include Q# just as it was expanded to include C#.

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -0,0 +1,575 @@
...
+        /// <param name="signature">The signature for the function.</param>
+        /// <returns>The <see cref="IrFunction"/> corresponding to the given Q# type.</returns>
+        internal IrFunction CreateGlobalFunction(QsSpecialization spec, string mangledName, IFunctionType signature)
+        {
+            if (!Config.DebugSymbolsEnabled)
+            {
+                return this.CreateFunctionNoDebug(mangledName, signature);
+            }
+
+            if (spec.Kind == QsSpecializationKind.QsBody)
```
cesarzc, line 327: For adjoint and controlled operations, this just creates a function with no debug info but it is not clear in the code that it is doing that. I would suggest encapsulating the contents of this `if` statement in a method and using a `switch` statement here to make clear that `QsSpecializationKind.QsBody` is the only kind for which debug symbols are currently supported and that for the other ones debug symbols are explicitly not supported:
```csharp
var func = spec.Kind switch
    {
        QsSpecializationKind.QsAdjoint => this.CreateFunctionNoDebug(mangledName, signature),
        QsSpecializationKind.QsBody => this.CreateQsBodyFunction(spec, mangledName, signature),
        QsSpecializationKind.QsControlled => this.CreateFunctionNoDebug(mangledName, signature),
        QsSpecializationKind.QsControlledAdjoint => this.CreateFunctionNoDebug(mangledName, signature),
        _ => throw new ArgumentException($"Unsupported QsSpecializationKind '{spec.Kind}'"),
    }
```

src/QsCompiler/QirGeneration/Subtransformations/NamespaceTransformation.cs
```diff
@@ -55,6 +56,24 @@ public QirNamespaceTransformation(GenerationContext sharedState)
...
         /* public overrides */
 
+        public override QsNamespaceElement OnNamespaceElement(QsNamespaceElement nse)
+        {
+            // keep track of the location of the current namespace so that we can calculate the offset of locations accurately
+            QsNullable<QsLocation> locNullable = nse switch
+            {
+                QsNamespaceElement.QsCallable callable => callable.Item.Location,
+                QsNamespaceElement.QsCustomType type => type.Item.Location,
+                _ => throw new NotImplementedException("unknown namespace element"),
```
cesarzc, line 66: I believe a more appropriate exception here would be `ArgumentException`.

src/QuantumSdk/README.md
```diff
@@ -3,7 +3,7 @@
 ## Content ##
 
 
-The NuGet package Microsoft.Quantum.Sdk serves a .NET Core Sdk for developing quantum applications, meaning it acts as [shared SDK Component](https://docs.microsoft.com/en-us/dotnet/core/tools/cli-msbuild-architecture#the-tooling-layers).
+The NuGet package Microsoft.Quantum.Sdk serves a .NET Core Sdk for developing quantum applications, meaning it acts as [shared SDK Component](https://docs.microsoft.com/en-us/dotnet/core/project-sdk/overview).
```
cesarzc, line 6:
```suggestion
The NuGet package Microsoft.Quantum.Sdk serves a .NET Core Sdk for developing quantum applications, meaning it acts as [shared SDK Component](https://docs.microsoft.com/en-us/dotnet/core/tools/cli-msbuild-architecture#the-tooling-layers).
```

cesarzc: Reverting since the 404 error with the link seems resolved now.

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -0,0 +1,575 @@
...
+        /// Converts from a 0-based <see cref="Position"/> to a 1-based <see cref="DebugPosition"/>.
+        /// </summary>
+        /// <param name="position">The 0-based <see cref="Position"/></param>
+        /// <return>The 1-based <see cref="DebugPosition"/></return>
+        private static DebugPosition ConvertToDebugPosition(Position position)
+        {
+            return DebugPosition.FromZeroBased(position.Line, position.Column);
+        }
+
+        public static string DebugTypeNotSupportedMessage { get; } = "This debug type is not yet supported";
```
samarsha, line 115: nit:
```suggestion
        public static string DebugTypeNotSupportedMessage => "This debug type is not yet supported";
```

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -0,0 +1,575 @@
...
+        private readonly GenerationContext sharedState;
+
+        /// <summary>
+        /// Contains the DIBuilders included in the module related to this DebugInfoManager.
+        /// The key is the source file path of the DebugInfoBuilder's compile unit
+        /// </summary>
+        private readonly Dictionary<string, DebugInfoBuilder> dIBuilders;
+
+        /// <summary>
+        /// Constructs a DebugInfoManater with access to the <see cref="GenerationContext"/> as the shared state.
```
samarsha, line 144:
```suggestion
        /// Constructs a DebugInfoManager with access to the <see cref="GenerationContext"/> as the shared state.
```

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -0,0 +1,575 @@
...
+        internal void FinalizeDebugInfo()
+        {
+            if (!Config.DebugSymbolsEnabled)
+            {
+                return;
+            }
+
+            foreach (KeyValuePair<string, DebugInfoBuilder> entry in this.dIBuilders)
+            {
+                entry.Value.Finish(); // must be called for every DIBuilder after all QIR generation
```
samarsha, line 187: If this clean up has to happen when DebugInfoManager is done being used, should it be part of an IDisposable implementation for DebugInfoManager instead?

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -0,0 +1,575 @@
...
+                    else
+                    {
+                        // variable is represented as the value itself
+                        Value variable = value.Value;
+                        dIBuilder.InsertValue(
+                            value: variable,
+                            varInfo: dIVar,
+                            location: dILocation,
+                            insertAtEnd: this.sharedState.CurrentBlock);
+                    }
```
samarsha, line 307:
```suggestion
                    // The variable is represented as a pointer if it's mutable, or its value if it's immutable.
                    var variable = isMutableBinding ? ((PointerValue)value).Pointer : value.Value;
                    dIBuilder.InsertValue(
                        value: variable,
                        varInfo: dIVar,
                        location: dILocation,
                        insertAtEnd: this.sharedState.CurrentBlock);
```

src/QsCompiler/QirGeneration/DebugInfoManager.cs
ryan-moreno: You should be able to use the space reserved for user-defined languages in DwarfEnumeration.cs. However, there seems to be some bug in the LLVM Binding APIs we use such that if you define a value in the user-defined section and pass it in to the creation of a DICompileUnit, it gets translated into some absurdly large and out-of-range integer by the time it's emitted to the IR. Ian helped me take a brief look into this and we didn't figure out the answer

ryan-moreno: Ultimately, we will want to ask DWARF to add Q# to their language enumeration, so it seemed most reasonable to just keep using C99 in the meantime. I wrote aboout this in my internship notes, but I will create work requests in GitHub for things like this so that I can link them into the code and they're easy to find

src/QsCompiler/LlvmBindings/BitcodeModule.cs
ryan-moreno: No, I added it in when I created this function because it's used throughout our BitcodeModule.cs

src/QsCompiler/LlvmBindings/BitcodeModule.cs
```diff
@@ -120,19 +118,60 @@ private BitcodeModule(LLVMModuleRef handle)
...
+            return dIBuilder;
+        }
+
+        /// <summary>Gets the list of <see cref="DebugInfoBuilder"/>s used to create debug information for this module</summary>
+        /// <remarks>Each DebugInfoBuilder can own one paired compile unit.</remarks>
+        public List<DebugInfoBuilder> DIBuilders { get; }
+
+        /// <summary>
+        /// If there is exactly one <see cref="DebugInfoBuilder"/> used to create debug info for this module,
+        /// we return it through a get. This field is retained for backward compatibility.
```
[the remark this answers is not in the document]
ryan-moreno: It's because DIBuilder was already exposed as public. Bettina and I had talked about everything that was exposed as public needing to remain exposed. @bettinaheim any thoughts?

src/QsCompiler/LlvmBindings/BitcodeModule.cs
ryan-moreno: BitcodeModule has a member called isDisposed that can be set to true to "dispose" of the module. This is a safety check that was already used throughout our codebase in the places where LLVM.NET had different safety checks.

src/QsCompiler/QirGeneration/DebugInfoManager.cs
ryan-moreno: I couldn't use the switch statement like this because it expects a constant instead of something like QsSpecializationKind.QsAdjoint. There's probably a workaround I could figure out, but I did something else to make it clearer. Could you let me know if this fix is what you were looking for?

examples/QIR/Development/README.md
ryan-moreno: Yes, that's what it's related to. I wrote a bit more here.

src/QsCompiler/LlvmBindings/BitcodeModule.cs
```diff
@@ -120,19 +118,60 @@ private BitcodeModule(LLVMModuleRef handle)
...
+                    throw new Exception("Reference to the DIBuilder is ambiguous because this module does not contain exactly one DebugInfoBuilder.");
+                }
             }
         }
 
-        /// <summary>Gets the Debug Compile unit for this module</summary>
-        public DICompileUnit? DICompileUnit { get; internal set; }
+        /// <summary>
+        /// If there is exactly one <see cref="DebugInfoBuilder"/> used to create debug info for this module,
+        /// we return its <see cref="DICompileUnit"/> through a get. This field is retained for backward compatibility.
```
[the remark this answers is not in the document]
ryan-moreno: Same as above (just commenting to keep track of what I've addressed)

src/QsCompiler/QirGeneration/DebugInfoManager.cs
ryan-moreno: I tried this, but the problem is that the only place it would make sense (to me) to call Dispose would be within the GenerationContext when it is being disposed. (Since the GenerationContext is paired 1-1 with a DebugInfoManager and the GenerationContext is the shared state for the transformation. However, FinalizeDebugInfo needs to be called before we emit the QIR which we do before we dispose of the GenerationContext. 

Calling Finish() on each DIBuilder is so that any unresolved dependencies in the debug info tags are resolved instead of being left as "temporary" (this is all related to the LLVM Bindings). This feels less about managed resources, and more about signaling to the LLVM Bindings that we are finished parsing the functions in the source file and ready to do the final wrap up.

Let me know if you think it should be handled differently.

src/QsCompiler/QirGeneration/DebugInfoManager.cs
cesarzc: Thanks for the detailed answer, Ryan!

src/QsCompiler/QirGeneration/DebugInfoManager.cs
cesarzc: That's interesting, `QsSpecializationKind.QsBody` is a constant and it is used in a switch statement [here](https://github.com/microsoft/qsharp-compiler/blob/a4ef30735d856069f92ef33e00e92adcf433f49b/src/QsCompiler/BondSchemas/CompilerObjectTranslator.cs#L380). I'm ok with the way it currently is but I would prefer the `switch` expression since makes it clear what are all the possible specialization kinds.

cesarzc:
It seems like the only thing left is to merge this branch with the latest changes in `main` and we should be good to go :)

src/QsCompiler/QirGeneration/RewriteSteps.cs
```diff
@@ -74,8 +74,17 @@ public bool PreconditionVerification(QsCompilation compilation)
         /// <inheritdoc/>
         public bool Transformation(QsCompilation compilation, out QsCompilation transformed)
         {
+            bool enableDebugSymbols = false;
+            if (this.AssemblyConstants.TryGetValue(ReservedKeywords.AssemblyConstants.EnableDebugSymbols, out string? debugEnabledString))
+            {
+                if (debugEnabledString != null && string.Equals(debugEnabledString.Trim(), "true", StringComparison.OrdinalIgnoreCase))
+                {
+                    enableDebugSymbols = true;
+                }
```
samarsha, line 83: nit:
```suggestion
                enableDebugSymbols = string.Equals(debugEnabledString?.Trim(), "true", StringComparison.OrdinalIgnoreCase);
```

src/QsCompiler/QirGeneration/DebugInfoManager.cs
samarsha: Thanks for the explanation. That sounds fine, it can stay the way it is.

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -0,0 +1,569 @@
...
+            /// <summary>
+            /// CodeView version we are using for the debug info in the QIR generation
+            /// </summary>
+            public static uint CodeviewVersion { get; } = 1;
+
+            /// <summary>
+            /// The source language information for Dwarf.
+            /// For now, we are using the C interface. Ideally this would be a user defined language for Q#.
+            /// </summary>
+            public static SourceLanguage QSharpLanguage { get; } = SourceLanguage.C99;
```
samarsha, line 57: nit: For read-only properties that return a constant, I think it's better to use `=>` - the syntax is shorter and it avoids increasing the size of the class with constant fields.

```suggestion
            /// <summary>
            /// Dwarf version we are using for the debug info in the QIR generation
            /// </summary>
            public static uint DwarfVersion => 4;

            /// <summary>
            /// Title of the CodeView module flag for debug info
            /// </summary>
            public static string CodeviewName => "CodeView";

            /// <summary>
            /// CodeView version we are using for the debug info in the QIR generation
            /// </summary>
            public static uint CodeviewVersion => 1;

            /// <summary>
            /// The source language information for Dwarf.
            /// For now, we are using the C interface. Ideally this would be a user defined language for Q#.
            /// </summary>
            public static SourceLanguage QSharpLanguage => SourceLanguage.C99;
```

src/QsCompiler/QirGeneration/DebugInfoManager.cs
ryan-moreno: I'm changing this back because I didn't look closely enough when I committed this and it the distinction between InsertValue and InsertDeclare is important

bettinaheim (changes requested):
Blocking the PR for now to clarify the use of expression.Range.Item.

bettinaheim:
I am pulling this into the compiler repo on a separate branch so we can wrap things up there.

src/QsCompiler/QirGeneration/QIR/DataStructures.cs
```diff
@@ -140,7 +140,8 @@ internal SimpleValue(Value value, ResolvedType type)
     internal class PointerValue : IValue
     {
         private readonly IValue.Cached<IValue> cachedValue;
-        private readonly Value pointer;
+
+        public Value Pointer { get; }
```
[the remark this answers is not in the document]
bettinaheim: Ah, the e2e pipelines didn't run; this explains why this didn't cause any failure - there is a test gap for the debug infos themselves, and none of the other QIR test cases here would have run into this.

src/QsCompiler/QirGeneration/Generator.cs
```diff
@@ -36,12 +36,13 @@ public class Generator : SyntaxTreeTransformation<GenerationContext>, IDisposabl
         /// Instantiates a transformation capable of emitting QIR for the given compilation.
         /// </summary>
         /// <param name="compilation">The compilation for which to generate QIR</param>
-        public Generator(QsCompilation compilation)
-        : base(new GenerationContext(compilation.Namespaces, compilation.EntryPoints.Length == 0), TransformationOptions.NoRebuild)
+        public Generator(QsCompilation compilation, bool enableDebugSymbols)
```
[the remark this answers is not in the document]
bettinaheim: addressed in https://github.com/microsoft/qsharp-compiler/pull/1349.

src/QsCompiler/QirGeneration/DebugInfoManager.cs
```diff
@@ -0,0 +1,585 @@
...
+        /// </summary>
+        private readonly Dictionary<string, DebugInfoBuilder> dIBuilders;
+
+        /// <summary>
+        /// Constructs a DebugInfoManager with access to the <see cref="GenerationContext"/> as the shared state.
+        /// </summary>
+        internal DebugInfoManager(GenerationContext generationContext, bool enableDebugSymbols)
+        {
+            this.sharedState = generationContext;
+            Config.EnableDebugSymbols = enableDebugSymbols;
```
bettinaheim, line 147: It is odd that an argument to a class instance construction is used to set a property in the Config that is no limited to this instance.

src/QsCompiler/QirGeneration/Subtransformations/StatementKindTransformation.cs
```diff
@@ -485,6 +485,7 @@ public override QsStatementKind OnVariableDeclaration(QsBinding<TypedExpression>
 
                     QirExpressionKindTransformation.AccessViaLocalId(ex, out var localId);
                     this.SharedState.ScopeMgr.RegisterVariable(varName, value, fromLocalId: localId);
+                    this.SharedState.DIManager.CreateLocalVariable(varName, value, stm.Kind.IsMutableBinding);
```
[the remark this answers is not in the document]
bettinaheim: There is a call to this.DIManager.CreateFunctionArgument in the Context.

PR #1290, opened 2021-12-16
Syntax tree transformation improvements - part 1

bettinaheim:
The transformations now use OnLocalNameDeclaration and OnLocalName for all locally declared symbols. 
The change also introduces OnItemNameDeclaration and OnItemName to easily modify item named in user defined types. These methods along with OnAbsoluteLocation and OnRelativeLocation are virtual methods in the SyntaxTreeTransformation itself; i.e. this PR moves the common handles that are used by several kinds of nodes into the syntax tree transformation, rather than having them live in node specific transformations.

Here is the second half of the changes, that updates the data structures used to represent lambdas: https://github.com/microsoft/qsharp-compiler/pull/1332


src/QsCompiler/Core/Transformations/ExpressionTransformation.fs
```diff
@@ -461,6 +471,12 @@ and ExpressionTransformationBase internal (options: TransformationOptions, _inte
 
     // supplementary expression information
 
+    abstract OnLocalNameDeclaration : string -> string
+    default this.OnLocalNameDeclaration name = name
+
+    abstract OnLocalName : string -> string
+    default this.OnLocalName name = name
```
samarsha, line 478: It seems like if I want to visit callable parameters, or let/for statement bindings, I now need to override a handler in ExpressionTransformation, even if the declarations don't occur in an expression. Is that correct? I think that is confusing.

bettinaheim: I extracted these, and also all other shared pieces into the syntax tree transformation itself.

src/QsCompiler/Core/Transformations/SyntaxTreeTransformation.fs
```diff
@@ -3,52 +3,37 @@
...
 
 type SyntaxTreeTransformation<'T> private (state: 'T, options: TransformationOptions, _internal_: string) =
-
-    let mutable _Types = new TypeTransformation<'T>(TransformationOptions.Default, _internal_)
-    let mutable _ExpressionKinds = new ExpressionKindTransformation<'T>(TransformationOptions.Default, _internal_)
-    let mutable _Expressions = new ExpressionTransformation<'T>(TransformationOptions.Default, _internal_)
-    let mutable _StatementKinds = new StatementKindTransformation<'T>(TransformationOptions.Default, _internal_)
-    let mutable _Statements = new StatementTransformation<'T>(TransformationOptions.Default, _internal_)
-    let mutable _Namespaces = new NamespaceTransformation<'T>(TransformationOptions.Default, _internal_)
+    inherit CommonTransformationItems()
```
samarsha, line 16: Is it intended for users to override these methods only from their class that extends `SyntaxTreeTransformation`, rather than creating a separate class to extend `CommonTransformationItems`?

src/QsCompiler/Core/Transformations/TypeTransformation.fs
```diff
@@ -17,22 +17,46 @@ open Microsoft.Quantum.QsCompiler.Transformations.Core.Utils
...
+        new TypeTransformationBase(options, "_internal_")
+        then this.CommonTransformationItemsHandle <- getCommonItems
 
+    new(options: TransformationOptions) as this =
+        new TypeTransformationBase(options, "_internal_")
+        then
+            let commonItems = new CommonTransformationItems()
+            this.CommonTransformationItemsHandle <- fun _ -> commonItems
+
+    new() = TypeTransformationBase TransformationOptions.Default
```
samarsha, line 41: I think this can be simplified a lot. You don't need `missingTransformation` or a constructor with an unused `_internal_` parameter if you make the primary constructor take in a `getCommonItems`.

```fsharp
type TypeTransformationBase internal (options, getCommonItems) =
    static let createCommonGetter () =
        let common = CommonTransformationItems()
        fun () -> common

    let Node = if options.Rebuild then Fold else Walk

    member val internal CommonTransformationItemsHandle = getCommonItems with get, set

    member this.Common = this.CommonTransformationItemsHandle()

    new(options) = TypeTransformationBase(options, createCommonGetter ())

    new() = TypeTransformationBase TransformationOptions.Default
```

src/QsCompiler/Tests.LanguageServer/Tests.LanguageServer.csproj
```diff
@@ -26,6 +26,9 @@
     <ProjectReference Include="..\LanguageServer\LanguageServer.csproj">
       <Private>true</Private>
     </ProjectReference>
+    <ProjectReference Include="..\..\QsFmt\App\App.fsproj">
+      <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
+    </ProjectReference>
```
samarsha, line 31: Is this supposed to be part of this PR?

src/QsCompiler/Core/Transformations/CommonTransformationItems.fs
```diff
@@ -0,0 +1,38 @@
...
+
+open Microsoft.Quantum.QsCompiler.DataTypes
+open Microsoft.Quantum.QsCompiler.SyntaxTree
+
+
+/// <summary>
+/// Contains virtual methods that are invoked
+/// by several different kinds of nodes in the syntax tree.
+/// </summary>
+type CommonTransformationItems internal () =
```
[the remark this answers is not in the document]
bettinaheim: Yes. The syntax tree transformation inherits from it, so overrides are done in the syntax tree transformation. For all practical purposes, this class is not usable outside from using/overriding the handles in the syntax tree transformation.

src/QsCompiler/Core/Transformations/CommonTransformationItems.fs
```diff
@@ -0,0 +1,38 @@
...
+    default this.OnAbsoluteLocation loc = loc
+
+    abstract OnRelativeLocation : QsNullable<QsLocation> -> QsNullable<QsLocation>
+    default this.OnRelativeLocation loc = loc
+
+    abstract OnTypeRange : TypeRange -> TypeRange
+    default this.OnTypeRange range = range
+
+    abstract OnExpressionRange : QsNullable<Range> -> QsNullable<Range>
+    default this.OnExpressionRange range = range
```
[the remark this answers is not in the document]
bettinaheim: There is a third handle coming that is on SymbolRange. It seems most consistent to have them all in the same location, along with the other location information (absolute and relative location).

src/QsCompiler/Core/Transformations/SyntaxTreeTransformation.fs
bettinaheim: Yes. We could seal it more explicitly, but I believe the internal constructor effectively takes care of that.

src/QsCompiler/Core/Transformations/TypeTransformation.fs
bettinaheim: It just seemed so inconsistent, and I got rid of having independent initialization for all subtransformations. This seems all in all much simpler, and I was just bothered by one constructor looking different than for the ones for the rest of the transformations.

src/QsCompiler/Tests.LanguageServer/Tests.LanguageServer.csproj
bettinaheim: Kind of - I was noticed and was bothered by the solution not building properly, and it seemed silly to create a separate PR.

.fantomasignore
```diff
@@ -1,2 +1,6 @@
 ﻿# Ignore signature files because of formatting bugs.
 *.fsi
+
+# Disabling the check for the transformations due to formatting bugs.
```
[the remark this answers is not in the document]
bettinaheim: I added a comment.

src/QsCompiler/Core/Transformations/CommonTransformationItems.fs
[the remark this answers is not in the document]
bettinaheim: I renamed it to CommonTransformationNodes. CommonTransformation seems odd to me since it is not fully self contained transformation but rather a collection of nodes shared by all transformations.

src/QsCompiler/Core/Transformations/TypeTransformation.fs
samarsha: Sorry, I'm not sure what you mean? In terms of consistency, I think that all of the transformation constructors could be cleaned up in the same way to avoid the `_internal_` parameter hack.

src/QsCompiler/Core/Transformations/CommonTransformationNodes.fs
```diff
@@ -0,0 +1,38 @@
...
+
+open Microsoft.Quantum.QsCompiler.DataTypes
+open Microsoft.Quantum.QsCompiler.SyntaxTree
+
+
+/// <summary>
+/// Contains virtual methods that are invoked
+/// by several different kinds of nodes in the syntax tree.
+/// </summary>
+type CommonTransformationNodes internal () =
```
samarsha, line 14: Since this class is only an implementation detail, I don't think it's right to make it a public type (even if the constructor is internal). Instead, I would recommend making it an internal interface. That makes it possible for the transformations to treat `SyntaxTreeTransformation` as implementing `CommonTransformationNodes` but outside code wouldn't know that it existed.

```suggestion
type internal CommonTransformationNodes =
```

src/QsCompiler/Core/Transformations/TypeTransformation.fs
ScottCarda-MS: Looking through the transformations as they are before this PR, I am also not seeing the need for the "_internal_" constructors when they are just called by, what I would consider, the proper constructors. Though I am also not sure what the purpose of the `getCommonItems` parameter is either.

ScottCarda-MS: For example with the TypeTrasformation:
```fsharp
and TypeTransformation<'T> internal (options, _internal_) =
    inherit TypeTransformationBase(options)
    let mutable _Transformation : SyntaxTreeTransformation<'T> option = None // will be set to a suitable Some value once construction is complete

    /// Handle to the parent SyntaxTreeTransformation.
    /// This handle is always safe to access and will be set to a suitable value
    /// even if no parent transformation has been specified upon construction.
    member this.Transformation
        with get () = _Transformation.Value
        and private set value = _Transformation <- Some value

    member this.SharedState = this.Transformation.SharedState

    new(parentTransformation: SyntaxTreeTransformation<'T>, options: TransformationOptions) as this =
        new TypeTransformation<'T>(options, "_internal_")
        then this.Transformation <- parentTransformation

    new(sharedState: 'T, options: TransformationOptions) as this =
        TypeTransformation<'T>(options, "_internal_")
        then
            this.Transformation <- new SyntaxTreeTransformation<'T>(sharedState, options)
            this.Transformation.Types <- this

    new(parentTransformation: SyntaxTreeTransformation<'T>) =
        new TypeTransformation<'T>(parentTransformation, TransformationOptions.Default)

    new(sharedState: 'T) = new TypeTransformation<'T>(sharedState, TransformationOptions.Default)
```
I believe could be rewritten as:
```fsharp
and TypeTransformation<'T> internal (parentTransformation: SyntaxTreeTransformation<'T>, options: TransformationOptions) =
    inherit TypeTransformationBase(options)
    let mutable _Transformation : SyntaxTreeTransformation<'T> = parentTransformation

    /// Handle to the parent SyntaxTreeTransformation.
    /// This handle is always safe to access and will be set to a suitable value
    /// even if no parent transformation has been specified upon construction.
    member this.Transformation
        with get () = _Transformation
        and private set value = _Transformation <- value

    member this.SharedState = this.Transformation.SharedState

    new(sharedState: 'T, options: TransformationOptions) as this =
        TypeTransformation<'T>(new SyntaxTreeTransformation<'T>(sharedState, options), options)
        then
            this.Transformation.Types <- this

    new(parentTransformation: SyntaxTreeTransformation<'T>) =
        new TypeTransformation<'T>(parentTransformation, TransformationOptions.Default)

    new(sharedState: 'T) = new TypeTransformation<'T>(sharedState, TransformationOptions.Default)
```
eliminating the `_internal_` constructor. Though, here again I'm not sure where the new `getCommonItems` fits into this formulation. This is just my take on the pre-existing framework.

src/QsCompiler/Core/Transformations/CommonTransformationNodes.fs
bettinaheim: Agreed in principle. I may be missing something here in terms of F# features, but I don't see how the `inherit ...` would work with this - the edit as suggested doesn't seem to define or allow to call a constructor as part of the SyntaxTreeTransformation class, and `type internal CommonTransformationNodes() = ...` results in an error that the type of the base class is less accessible that the SyntaxTreeTransformation that inherits from it.

bettinaheim:
The build seems to be failing a) due to a security alert that is unrelated to this PR, b) some tests are indicated as failing - that is odd since locally they pass fine, and they also passed fine in CI a couple of days ago when I put up the PR. I'll double check that shortly.

src/QsCompiler/Core/Transformations/TypeTransformation.fs
bettinaheim: Sarah's comment applied to a different file, where part of the reason for the setup is that I wanted to avoid having to put all transformations into the same file. @ScottCarda-MS Your edit I believe is intended for the SyntaxTreeTransformation.fs file where the recursive type definition is not a problem (here we assemble it all, after all). Try to apply the same change to the rest of the transformations in that file (along with the corresponding edit to the constructor calls in the SyntaxTreeTransformation), and you should see the unit test failures - unless I am misinterpreting what you had in mind?

ScottCarda-MS: Alright, I made a branch where I removed the `_internal_` constructors from the transformation framework: [here](https://github.com/microsoft/qsharp-compiler/compare/sccarda/RemoveInternalConstructors)
@bettinaheim you are right that a lot of the unit tests fail after this, although none of the failures really tell me why. I'm not sure what the recursive problem is, but there is clearly an issue with doing this. Since it is not the focus of this PR, maybe you could catch me up on what the root issue is offline.

bettinaheim:
Closing this in favor of https://github.com/microsoft/qsharp-compiler/pull/1332, that contains this change as well as the second half to update the lambda data structures. Thanks for the reviews so far - the feedback is incorporated in https://github.com/microsoft/qsharp-compiler/pull/1332 and please continue there.

PR #1291, opened 2021-12-20
Update QDK to .NET6.0 (qsharp-compiler)

ricardo-espinoza:
We're migrating the QDK to the most recent Long Time Support version of the .NET framework. For details about this change, refer to the original issue [Q#C:1224](https://github.com/microsoft/qsharp-compiler/issues/1224).

As part of this change, we're:

- Re-targeting all .NetCoreApp3.1 binaries to .NET6.0
- Updating Docker images, samples and templates.
- Libraries using .NetStandard2.1 are not affected by this change.
- The minimum supported .NET version in the QDK will also be updated from 3.1 to 6.0

This is a multi-repo change, and all changes will need to be merged concurrently.

### DIFF
--- a/build/test.ps1
+++ b/build/test.ps1
@@ -7,6 +7,9 @@ $ErrorActionPreference = 'Stop'
 $all_ok = $True
 Write-Host "Assembly version: $Env:ASSEMBLY_VERSION"
 
+Write-Host "##vso[task.logissue type=warning;]Forcing all tests to be skipped."
+return
+
 if ($Env:ENABLE_TESTS -eq "false") {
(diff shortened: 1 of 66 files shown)


build/test.ps1
```diff
@@ -7,6 +7,9 @@ $ErrorActionPreference = 'Stop'
 $all_ok = $True
 Write-Host "Assembly version: $Env:ASSEMBLY_VERSION"
 
+Write-Host "##vso[task.logissue type=warning;]Forcing all tests to be skipped."
```
ricardo-espinoza, line 10: This part of the change will need to be removed before this can be merged. We need to fix the Project Loader tests for this.

ricardo-espinoza: Removed.

PR #1299, opened 2022-01-10
Add llvm-build logic for pipeline to produce LLVM binaries and generated C#

swernli:
This change adds all the files and build logic needed to create a private build pipeline to produce a package with generated C# and binaries for our LlvmBindings to depend on. The resulting .cs files and multi-platform binaries are packed via nuspec and uploaded to the qdk-alpha in a package like `Microsoft.Quantum.LlvmBindings.Native.13.0.0-CI-20220110-201043.nupkg`. This is the first step of removing the dependency on the LLVMSharp.Interop package and upgrading to LLVM/Clang 13. See the build pipeline [here](https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=36731&view=results). Note that this pipeline should only be triggered manually, and none of this build logic will run during normal qsharp-compiler or e2e builds.


swernli:
Whoops... looks like the LLVM pipeline did get triggered by this PR. Of note, the builds should only take 20 min and run in parallel with the other builds, but this is not what I expected. @anpaz, can we change the rules for that pipeline such that it only gets run if contents of the `llvm-build` folder or `build/llvm.yml` file change?

kuzminrobin:
In the description of this PR, the presence of the space (` `) in the fragment `Microsoft.Quantum.LlvmBindings.Native 13.0.0-CI-20220110-201043.nupkg` looks suspicious. So far I don't remember spaces used in package names.

build/llvm.yml
```diff
@@ -0,0 +1,231 @@
...
+  AQ_LLVM_PACKAGE_GIT_VERSION: ""
+  ENABLE_COMPILER: false
+  ENABLE_VSIX: false
+  ENABLE_LLVM_BUILDS: true
+  AQ_DOWNLOAD_LLVM: false
+
+jobs:
+- job: GetLlvmTag
+  pool:
+    vmImage: 'ubuntu-20.04'
```
kuzminrobin, line 20: Strictly 'ubuntu-**20.04**'? (Not 'ubuntu-**latest**'?)
If yes then a code comment answering "why?" would not harm.

build/llvm.yml
```diff
@@ -0,0 +1,231 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+trigger: none
+
+variables:
+  CCACHE_DIR: $(Pipeline.Workspace)/ccache
```
kuzminrobin, line 7: Just curiosity. 
Is `CCACHE_DIR` related to https://ccache.dev/?
What is '**S**CCACHE' below? (in `SCCACHE_DIR`)

build/llvm.yml
```diff
@@ -0,0 +1,231 @@
...
+      choco install --accept-license -y sccache ninja
+    displayName: Install build dependencies, sccache, and update PATH to use linked versions of gcc, cc, etc
+
+  - task: Cache@2
+    inputs:
+      key: 'version7 | sccache | "$(Agent.OS)"'
+      path: $(SCCACHE_DIR)
+      restoreKeys: |
+        version7 | sccache | "$(Agent.OS)"
+    displayName: sccache
```
kuzminrobin, line 95: (I don't know what 'CCACHE', '**S**CCACHE', and 'Cache@2' are in this context)

I see that 
`CCACHE_DIR` is also used outside of `task: Cache@2`, but
`SCCACHE_DIR` is only used in `task: Cache@2`.
That looks suspicious.

build/llvm.yml
```diff
@@ -0,0 +1,231 @@
...
+    vmImage: 'ubuntu-20.04'
+  steps:
+  - checkout: self
+  - pwsh:  Write-Host "##vso[task.setvariable variable=AQ_LLVM_PACKAGE_GIT_VERSION;isoutput=true]$(Get-Content ./llvm.tag.txt)"
+    workingDirectory: $(System.DefaultWorkingDirectory)/llvm-build
+    displayName: "Get LLVM Tag"
+    name: Tag
+
+- job: LLVMLinuxBuild
+  timeoutInMinutes: 300
```
kuzminrobin, line 29: 5-hour time-out? If the [pipeline runs for 20 minutes](https://github.com/microsoft/qsharp-compiler/pull/1299#issuecomment-1009438656), then 5-hour-long time-out seems redundantly long. I would consider 40 - 60 minutes.

build/llvm.yml
swernli: I think it should work with ubuntu-latest, since on Linux we build with a container using manylinux anyway. So the underlying OS doesn't actually matter.

build/llvm.yml
swernli: This one is a bit tricky... the pipeline runs for 20 minutes if the cache is fully populated, but if not, or if the LLVM version has been updated to one where a cached build does not exist, the build takes about 1.5 hours. In testing, I believe @idavis saw builds that took upwards of 3 hours if additional features were enabled. So under normal circumstances, 20 minutes is the expected build time, but when intentionally doing an update or running after cache expiration, then it could be longer.

Incidentally, this is why the intent is to not build this pipeline unless manually invoked; regular PRs should not be producing any changes to the binary dependencies we have on LLVM.

build/llvm.yml
swernli: sccache comes from https://github.com/mozilla/sccache. Note that Windows uses sccache while Linux and MacOS use ccache. I believe this was an intentional choice for compatibility, but I'm building off of infrastructure @idavis introduced for pyqir. That logic lives on here: https://github.com/qir-alliance/pyqir/blob/main/.github/workflows/ci.yml#L45. In that repo it also uses ccache for Linux and MacOS while Windows uses sccache.

build/llvm.yml
swernli: `Cache@2` is the caching task for Azure DevOps (see https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/cache?view=azure-devops). The environment variables ensure that this tasks loads the cache from the right backend storage.

swernli:
> In the description of this PR, the presence of the space (` `) in the fragment `Microsoft.Quantum.LlvmBindings.Native 13.0.0-CI-20220110-201043.nupkg` looks suspicious. So far I don't remember spaces used in package names.

Yup, that space was a copy paste error on my part. The actual package does not have a space there.

llvm-build/Microsoft.Quantum.LlvmBindings.Native.nuspec
```diff
@@ -0,0 +1,18 @@
...
+    <authors>Microsoft</authors>
+    <owners>QuantumEngineering, Microsoft</owners>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <description>LLVM native support for LLVM Bindings.</description>
+    <copyright>© Microsoft Corporation. All rights reserved.</copyright>
+  </metadata>
+  <files>
+    <file src=".\drops\libLLVM.dll" target="runtimes\win-x64\native\libLLVM.dll" />
+    <file src=".\drops\libLLVM.dylib" target="runtimes\osx-x64\native\libLLVM.dylib" />
+    <file src=".\drops\libLLVM.so" target="runtimes\linux-x64\native\libLLVM.so" />
```
kuzminrobin, line 15: For other dynamic libraries the Win version typically does not have the `lib` prefix (in `libLLVM.dll`).

swernli: A notable exception would be `libomp` ;). The naming here is intentional, and matches what was done by LLVMSharp in their interop package (see for example https://github.com/dotnet/LLVMSharp/blob/ad498be4815dc4699981ba61c9ade30a740d5904/sources/LLVMSharp.Interop/Extensions/LLVM.Manual.cs#L12 which explicitly uses `"libLLVM"` for all platforms).

llvm-build/libllvm.ps1
```diff
@@ -0,0 +1,131 @@
...
+        exec { 7z x $file }
+    }
+    elseif (".gz" -eq $extension) {
+        $tarGz = $file
+        Write-Vso "Processing $tarGz"
+        exec { 7z x $tarGz }
+        $tar = $tarGz.Name.TrimEnd($extension)
+        Write-Vso "Processing $tar"
+        exec { 7z x -aoa -ttar $tar }
+    }
```
kuzminrobin, line 49: I assume there is no single `7z` command to extract `.tar.gz` (two separate commands are _required_).

build/llvm.yml
swernli: I updated to using "latest" build images for both ubuntu and windows vms.

llvm-build/libllvm.ps1
swernli: Yes, unfortunately it's an archive in an archive, so the two calls are required.

llvm-build/libllvm.ps1
```diff
@@ -0,0 +1,131 @@
...
+Write-Vso "Test Linux Path: $linuxPath"
+Test-Path $linuxPath
+
+$windowsPath = Resolve-Path (Join-Path  '.' "$($env:WINDOWS_PKG_NAME)" "bin" "LLVM-C.dll")
+Write-Vso "Test Windows Path: $windowsPath"
+Test-Path $windowsPath
+
+$darwinPath = Resolve-Path (Join-Path  '.' "$($env:DARWIN_PKG_NAME)" "lib" "libLLVM.dylib")
+Write-Vso "Test Darwin Path: $darwinPath"
+Test-Path $darwinPath
```
kuzminrobin, line 66: (I'm not very familiar with pwsh)
What does change if any of these `Test-Path` commands returns `$False`? (What do these lines 56 - 66 do?)

swernli: This won't fail the script execution, but it does output `true` or `false` into the build log. See for example here:
https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=36741&view=logs&j=b1b4ecf1-bb33-53e2-fb11-822bd4878653&t=0625e23a-d1dd-514b-4717-f10b81a8af4e&l=119

llvm-build/libllvm.ps1
```diff
@@ -0,0 +1,131 @@
...
+
+Write-Vso "Copy-Item -Recurse -Verbose -Path $generatedDir -Destination $(Join-Path $PSScriptRoot drops)"
+Copy-Item -Recurse -Verbose -Path $generatedDir -Destination $(Join-Path $PSScriptRoot drops)
+
+foreach ($file in $llvmArchiveFiles) {
+    Move-Item -Path $file -Destination $artifactsDir
+}
+
+# Add it the job's artifacts
+Copy-Item -Path $libLlvmPackage -Destination $artifactsDir
```
kuzminrobin, line 119: Is something missing in the comment? 
"# Add it **to** the job's artifacts"?

(Also, I personally prefer when the use of pronouns (like "it", "there", "him") is _minimized_, and actual nouns are used instead (e.g. "# Add **the package** the job's artifacts" ). Because with pronouns the multiple interpretations are possible quite often)

llvm-build/llvm.cmake
```diff
@@ -0,0 +1,76 @@
...
+  set(LLVM_USE_CRT_MINSIZEREL "MT" CACHE STRING "")
+  set(LLVM_BUILD_LLVM_C_DYLIB ON CACHE BOOL "")
+  set(CMAKE_INSTALL_UCRT_LIBRARIES ON CACHE BOOL "")
+  set(CPACK_BINARY_ZIP ON CACHE BOOL "")
+  set(LLVM_DYLIB_COMPONENTS "core;debuginfodwarf;linker;support;target;bitwriter" CACHE STRING "")
+else()
+  set(LLVM_BUILD_LLVM_DYLIB ON CACHE BOOL "")
+  set(CPACK_BINARY_TGZ ON CACHE BOOL "")
+  set(LLVM_DYLIB_COMPONENTS "core;debuginfodwarf;linker;support;target;bitwriter" CACHE STRING "")
+endif()
```
kuzminrobin, line 76: Both `if` and `else` parts end with command `set(LLVM_DYLIB_COMPONENTS "core;debuginfodwarf;linker;support;target;bitwriter" CACHE STRING "")`. Looks like the command can be extracted and placed after the line 76.

It is suspicious that the Win-specific part (`if (${CMAKE_HOST_SYSTEM_NAME} MATCHES "Windows")`) contains the fragments `DYLIB` (in LLVM_BUILD_LLVM_C_**DYLIB**, LLVM_**DYLIB**_COMPONENTS ).

llvm-build/libllvm.ps1
swernli: Yeah, that's good advice. The "it" here is indeed the package.

llvm-build/llvm.cmake
swernli: In this case, `DYLIB` refers to dynamic library as a general term, and it is an option exposed by the CMakeLists for LLVM. See https://github.com/llvm/llvm-project/blob/ebd8eee62a431a6744c3f187fcda58e5dea08499/llvm/CMakeLists.txt#L641.

llvm-build/llvm.ps1
```diff
@@ -0,0 +1,150 @@
...
+$llvmDir = Join-Path $llvmRootDir llvm
+
+if (Test-CommandExists sccache) {
+    Write-Vso "Found sccache command"
+    # Set cap and make sure dir is created
+    $Env:SCCACHE_CACHE_SIZE = "2G"
+    Write-Vso "Starting sccache server"
+    & { sccache --start-server } -ErrorAction SilentlyContinue
+    Write-Vso "Started sccache server"
+}
```
kuzminrobin, line 22: Looks like the fragment "and make sure dir is created" is out-of-date.

llvm-build/llvm.ps1
```diff
@@ -0,0 +1,150 @@
...
+    if (-not (Test-Path env:\CCACHE_DIR)) {
+        # CCACHE_DIR needs to be set, get the value from ccache
+        $Env:CCACHE_DIR = exec { ccache -k cache_dir }
+    }
+    Assert (![string]::IsNullOrWhiteSpace($Env:CCACHE_DIR)) "CCACHE_DIR is not set"
+
+    # Set cap and make sure dir is created
+    if (!(Test-Path $Env:CCACHE_DIR)) {
+        mkdir $Env:CCACHE_DIR | Out-Null
+    }
```
kuzminrobin, line 35: Looks like [the `-p`](https://man7.org/linux/man-pages/man1/mkdir.1.html) will not harm right after `mkdir`: "mkdir **-p** $Env:CCACHE_DIR".

cgranade: That's only true when `mkdir` resolves to the native command on a system, but on some systems and platforms `mkdir` is a function that wraps `New-Item -ItemType Directory`. The most portable option would be as follows:
```suggestion
    if (!(Test-Path $Env:CCACHE_DIR)) {
        New-Item -ItemType Directory -Force $Env:CCACHE_DIR | Out-Null
    }
```

swernli: Cass is right, the explicit powershell way is the safest in this case. I'll update that in this file and in libllvm.ps1

llvm-build/llvm.ps1
```diff
@@ -0,0 +1,150 @@
...
+
+if ($IsLinux) {
+    function Build-ContainerImage {
+
+        Write-Vso "Building container image manylinux-llvm-builder"
+        exec -wd $buildDir {
+            $userName = [Environment]::UserName
+            $userId = $(id -u)
+            $groupId = $(id -g)
+            Write-Vso "Get-Content manylinux.Dockerfile | docker build -t manylinux-llvm-builder --build-arg USERNAME=$userName --build-arg USER_UID=$userId --build-arg USER_GID=$groupId --build-arg LLVM_BUILD_DIR=""$llvmBuildDir"" -" "command"
```
kuzminrobin, line 80: I have an impression that the fragment `" -" "command"` will print `-command` (no space between the dash and `command`). But the line below (line 81) makes me think that there should be some separator between the dash and `command`.

kuzminrobin, line 80: It is suspicious that the `$userName`, `$userId`, etc. are not surrounded with extra quotes (`"`), but `"$llvmBuildDir"` _is_.

llvm-build/manylinux.Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+    echo "echo \"cmake -G Ninja -C \${LLVM_CMAKEFILE} \${CMAKE_FLAGS} \${LLVM_DIR}\"" >> /tmp/entrypoint.sh && \
+    echo "cmake -G Ninja -C \${LLVM_CMAKEFILE} \${CMAKE_FLAGS} \${LLVM_DIR}" >> /tmp/entrypoint.sh && \
+    echo "echo \"ninja package\"" >> /tmp/entrypoint.sh && \
+    echo "ninja package" >> /tmp/entrypoint.sh && \
+    echo "echo \"([ ! -z \"\${LLVM_INSTALL_DIR}\" ] && ninja install) || [ -z \"\${LLVM_INSTALL_DIR}\" ]\"" >> /tmp/entrypoint.sh && \
+    echo "([ ! -z \"\${LLVM_INSTALL_DIR}\" ] && ninja install) || [ -z \"\${LLVM_INSTALL_DIR}\" ]" >> /tmp/entrypoint.sh && \
+    chmod +x /tmp/entrypoint.sh && \
+    chown ${USER_UID}:${USER_GID} /tmp/entrypoint.sh
+
+ENTRYPOINT ["bash", "-c", "/tmp/entrypoint.sh $*", "--"]
```
kuzminrobin, line 63: [FYI](https://stackoverflow.com/a/33979066/6362941). (I've never written Dockerfiles, have never checked if it works)

llvm-build/utils.ps1
```diff
@@ -0,0 +1,278 @@
...
+
+if (!(Test-Path function:\Get-RepoRoot)) {
+    # git revparse uses cwd. E2E builds use a different
+    # working dir, so we pin it to out repo (submodule in E2E)
+    function Get-RepoRoot {
+        exec -wd $PSScriptRoot {
+            git rev-parse --show-toplevel
+        }
+    }
+}
```
kuzminrobin, line 17: I have an impression that `exec` is not defined by the moment of line 13.
Some time later someone can add a change that calls `Get-RepoRoot` (line 12) earlier than `exec` is defined (line 55), and get something unexpected.

llvm-build/utils.ps1
```diff
@@ -0,0 +1,278 @@
...
+    # If no env var is defined we default to true
+    if (!(Test-Path env:\ENABLE_COMPILER)) {
+        $true
+    }
+    else {
+        # The env var exists, we use its value
+        # Coerce falsy values and negate to keep the defalt $true
+        $env:ENABLE_COMPILER -ne $false
+    }
+}
```
kuzminrobin, line 36: (I don't know powershell)
I have an impression that the lines 29 and 34 return a value. And if someone adds some code between lines 35 and 36 then the value returned can become unexpected. 
If supported, I would consider an explicit `return` inside of `if` and skipping `else` after `return`:
```suggestion
function Test-BuildCompiler {
    # By default we want to build the compiler
    # If no env var is defined we default to true
    if (!(Test-Path env:\ENABLE_COMPILER)) {
        return $true
    }
    # The env var exists, we use its value
    # Coerce falsy values and negate to keep the defalt $true
    return ( $env:ENABLE_COMPILER -ne $false )
}
```

llvm-build/utils.ps1
```diff
@@ -0,0 +1,278 @@
...
+    if (Test-Path env:\TF_BUILD) {
+        $true
+    }
+    elseif ((Test-Path env:\CI)) {
+        $env:CI -eq $true
+    }
+    else {
+        $false
+    }
+}
```
kuzminrobin, line 111: Nit: I personally prefer the functions returning Boolean to be named such that from the name it is immediately clear
* that the function returns Boolean,
* what it means if `true` is returned and what if `false` is returned.

Currently the names `Test-BuildVsix`, `Test-BuildLlvmComponents`, `Test-CI` do not tell me that. These names tell me that some sets of tests are executed for the Vsix Build, for Llvm Components Build, and for CI.

You may want to consider changing to something like '**Should**BuildVsix', '**Must**BuildLlvmComponents, '**Running**OnCI'.

llvm-build/utils.ps1
```diff
@@ -0,0 +1,278 @@
...
+
+# Writes an Azure DevOps message with default debug severity
+function Write-Vso {
```
kuzminrobin, line 114: Why `Vso`? What is the relation between the Azure DevOps and the `Vso`?

llvm-build/utils.ps1
cgranade: `Test-` is the idiomatic prefix for PowerShell commands that return streams of boolean outputs: 

image

https://docs.microsoft.com/en-us/powershell/scripting/developer/cmdlet/approved-verbs-for-windows-powershell-commands?view=powershell-7.2

kuzminrobin: Thanks!

kuzminrobin (approved):

llvm-build/manylinux.Dockerfile
swernli: I'll wait to defer to @idavis on this part, rather than try to clean up the dockerfile in this PR.

llvm-build/utils.ps1
swernli: Actually, between this and the comment below, I think I need to review the whole utils file and prune it. I don't think some of these are used anywhere...

llvm-build/utils.ps1
swernli: PowerShell allows for multiple returns from a function, so if any code was added after this it might cause the function to return a list of values instead of just one. That would still break callers, as you point out, and potentially be confusing to root cause. All that said, as I mentioned above it seems several of these utils aren't actually used, so I think this whole function can just be removed.

llvm-build/utils.ps1
swernli: Actually, until you called attention to it I hadn't even noticed it because some part of me is used to the old name... this terminology is a holdover from the days when Azure DevOps was called "Visual Studio Online." When it was first announced, git-compatible services via Azure was named [Visual Studio Online (VSO)](https://docs.microsoft.com/en-us/archive/msdn-magazine/2014/january/visual-studio-2013-introducing-visual-studio-online), then later rebranded to Visual Studio Team Services (VSTS), and finally in 2018 renamed to Azure DevOps. Interestingly, you can still see this legacy in the [default format for logging commands](https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash#logging-command-format):
`##vso[area.action property1=value;property2=value;...]message`
However, DevOps added an alternate logging format that doesn't use the `vso` anymore, and that happens to be what this utility uses:
`##[command]message`.

Still, I'm not opposed to renaming it for clarity since not everyone comes to DevOps with that naming baggage.

PR #1301, opened 2022-01-12
Rework LlvmBindings to use custom interop, LLVM 13

swernli:
This change updates the LlvmBindings package to use the custom LLVM binaries and generated C# from the LLVM pipeline. This allows us to eliminate the dependency on LLVMSharp and control the surface area and LLVM version we use. It also includes all test updates to use LLVM 13 compatible output.


src/QsCompiler/TestTargets/Simulation/Target/QirExecution.cs
```diff
@@ -28,7 +27,42 @@ public static class JitCompilation
...
+                var span = new ReadOnlySpan<byte>(msg, int.MaxValue);
+                var errTxt = span.Slice(0, span.IndexOf((byte)'\0')).AsString();
+                LLVMSharp.Interop.LLVM.DisposeMessage(msg);
+                throw new InternalCodeGeneratorException(errTxt);
+            }
+
+            fixed (LLVMSharp.Interop.LLVMModuleRef* pOutModule = &outModule)
+            {
+                sbyte* pMessage = null;
+                var result = LLVMSharp.Interop.LLVM.ParseBitcodeInContext(
```
alan-geller, line 45: So you still need LLVMSharp for testing?

I think I see why -- it looks like you'd have to add a lot of functionality to the custom interop to avoid it. Just making sure that testing is the only place it's used.

alan-geller (approved):

alan-geller (approved):

src/QsCompiler/TestTargets/Simulation/Target/QirExecution.cs
swernli: Yes, that's correct. Right now we only use LLVM JIT support in the tests, so it made sense not to include that in the built LLVM binaries. To avoid breaking the tests, I convert them to using LLVMSharp for JIT. If/when we decide to add JIT into the LLVM binaries we produce to line up with a production feature, we can switch this code to using that approach instead.

src/QsCompiler/LlvmBindings/DebugInfo/DebugInfoBuilder.cs
```diff
@@ -226,7 +224,7 @@ public DIFile CreateFile(string? path)
         /// </returns>
         public DIFile CreateFile(string? fileName, string? directory)
         {
-            var handle = this.BuilderHandle.CreateFile(fileName, directory);
+            var handle = this.BuilderHandle.CreateFile(fileName!, directory!);
```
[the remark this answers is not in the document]
swernli: That's a good point, I can add an explicit check for `null` to avoid mysterious failures later on.

src/QsCompiler/LlvmBindings/DebugInfo/DebugInfoBuilder.cs
```diff
@@ -226,7 +224,17 @@ public DIFile CreateFile(string? path)
...
-            var handle = this.BuilderHandle.CreateFile(fileName, directory);
+            if (fileName == null)
+            {
+                throw new ArgumentNullException(nameof(fileName));
+            }
+
+            if (directory == null)
+            {
+                throw new ArgumentNullException(nameof(directory));
+            }
```
samarsha, line 235: Why not make `CreateFile` take in non-nullable strings as parameters? This looks like a source of bugs especially since the documentation for `CreateFile` explicitly says `fileName` and `directory` are allowed to be null.

swernli: Ah, you're right, I missed the documentation there and the automerge kicked in before I saw your comment. Looking at the undelrying implementation, it's probably the intent to treat null as an empty string, so that's what I should be doing instead. `BuilderHandle.CreateFile` handles empty strings but does not check for null. This probably an oversight in the original code.

Issue #1294, opened 2022-01-04
Implement Lambda Lifter for Abstracting Lambda Expression in the AST


ScottCarda-MS:
We need to implement a content lifter transformation for lifting the bodies of lambda expressions into generated callables and replacing the lambda expression with a call to that generated callable.

- [X] Rework the base lifter to prepare for adding additional functionality.
- [X] Add functionality to base lifter to work with functions.
- [X] Update base lifter to generate callables with minimal parameter set.
- [X] Generate a callable for the body of lambdas.
- [x] Generate a call expression to the generated callable for the lambdas.
- [x] Change generated callable characteristic determination from context-based to content-based.
- [x] Add lots of tests.

ScottCarda-MS:
Closed with #1302

PR #1302, opened 2022-01-12
Lambda Lifter

ScottCarda-MS:
This PR redefines the lifter API and separates lifters into two kinds: those that use contextual information to generate lifted callables, and those that use the content to generate the lifted content. The lambda lifter is defined as a content-based lifter, while the classical control lifter is reworked in the new system to be a context-based lifter.
With the lambda lifter, lambdas will be removed from the AST through the Lift Lambdas rewrite step.

#1294


src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -213,6 +213,11 @@ public struct Configuration
             internal bool ConvertClassicalControl =>
                 this.RuntimeCapability != null && this.RuntimeCapability == RuntimeCapability.BasicMeasurementFeedback;
 
+            /// <summary>
+            /// Indicates whether the compiler will remove lambda expressions and replace them with calls to generated callables.
+            /// </summary>
+            internal bool LiftLambdaExpressions => true; // ToDo: replace this with the actual condition.
```
ScottCarda-MS, line 219: I'm not actually sure what the condition should be for the Lift Lambda Expressions rewrite step should be. @bettinaheim

src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs
```diff
@@ -0,0 +1,573 @@
...
+             })
+            result
+
+    [<Fact>]
+    [<Trait("Category", "Return Values")>]
+    member this.``With Nested Lambda Call``() = CompileLambdaLiftingTest 12
+
+    [<Fact>]
+    [<Trait("Category", "Return Values")>]
+    member this.``With Nested Lambda``() = CompileLambdaLiftingTest 13
```
samarsha, line 475: Where are the assertions for these cases?

src/QsCompiler/Tests.Compiler/TestUtils/Signatures.fs
```diff
@@ -609,6 +621,195 @@ let public ClassicalControlSignatures =
...
+         [|
+             LambdaLiftingNS, "Foo", [||], "Unit" // The original operation
+             LambdaLiftingNS, "_Foo", [| "Int" |], "Unit"
+             LambdaLiftingNS, "_Foo", [| "Int" |], "Unit"
+             LambdaLiftingNS, "_Foo", [| "Int"; "Double" |], "Unit"
+             LambdaLiftingNS, "_Foo", [| "Int"; "Double"; "String" |], "Unit"
+             LambdaLiftingNS, "_Foo", [| "Int"; "(Double, String)" |], "Unit"
+             LambdaLiftingNS, "_Foo", [| "Int"; "Double"; "String" |], "Unit"
+             LambdaLiftingNS, "_Foo", [| "(Int, Double)"; "String" |], "Unit"
+             LambdaLiftingNS, "_Foo", [| "Int"; "Double"; "String" |], "Unit"
```
samarsha, line 701: I don't think I understand the naming here. The lifted lambdas all have the same name `_Foo`?

samarsha (commented):
Just reviewed the tests so far.

src/QsCompiler/Tests.Compiler/TestUtils/Signatures.fs
ScottCarda-MS: They all end with "_Foo". Since I can't know the generated names of the lifted operations here, this field is for how the name of the operation ends.

src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs
```diff
@@ -0,0 +1,573 @@
...
+        let expected = sprintf "let lambda = %O(x, y, z, _);" generated.Parent
+        Assert.True(lines.[4] = expected, "The generated call expression did not have the correct arguments.")
+
+    [<Fact>]
+    [<Trait("Category", "Parameters")>]
+    member this.``Use Lots of Params``() = CompileLambdaLiftingTest 8 |> ignore
+
+    [<Fact>]
+    [<Trait("Category", "Parameters")>]
+    member this.``Use Closure With Params``() = CompileLambdaLiftingTest 9 |> ignore
```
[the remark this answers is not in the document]
ScottCarda-MS: That the result from compilation has the appropriate signatures is enough for this test case. That is handled in the signature check.

src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs
ScottCarda-MS: That the result from compilation has the appropriate signatures is enough for this test case. That is handled in the signature check.

src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs
```diff
@@ -0,0 +1,573 @@
...
+
+    let getManager uri content =
+        CompilationUnitManager.InitializeFileManager(
+            uri,
+            content,
+            compilationManager.PublishDiagnostics,
+            compilationManager.LogException
+        )
+
+    let ReadAndChunkSourceFile fileName =
```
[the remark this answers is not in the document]
ScottCarda-MS: These should somehow be private static member functions, which should have PascalCase. How do I specify them as static private members in F#?

src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs
```diff
@@ -0,0 +1,573 @@
...
+                Seq.item 0 x |> GetBodyFromCallable)
+
+        let lines = generated |> GetLinesFromSpecialization
+
+        Assert.True(
+            1 = Seq.length lines,
+            sprintf "Callable %O(%A) did not have the expected number of statements." generated.Parent generated.Kind
+        )
+
+        Assert.True(lines.[0] = "return 0;", "The generated callable did not have the expected content.")
```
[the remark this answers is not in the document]
ScottCarda-MS: Hmm, the suggestion doesn't seem to work. It is telling me that Assert.Equal has a limited set of types that it works on, and a string list isn't one.

src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs
```diff
@@ -0,0 +1,573 @@
...
+    member this.``With Return Value``() =
+        let result = CompileLambdaLiftingTest 1
+
+        let original = GetCallableWithName result Signatures.LambdaLiftingNS "Foo" |> GetBodyFromCallable
+
+        let generated =
+            GetCallablesWithSuffix result Signatures.LambdaLiftingNS "_Foo"
+            |> (fun x ->
+                Assert.True(1 = Seq.length x)
+                Seq.item 0 x |> GetBodyFromCallable)
```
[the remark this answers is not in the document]
ScottCarda-MS: replaced

src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs
```diff
@@ -0,0 +1,573 @@
...
+
+        charMatch.Value && adjMatch.Value && ctlMatch.Value
+
+    let AssertCallSupportsFunctors expectedFunctors call =
+        Assert.True(
+            DoesCallSupportFunctors expectedFunctors call,
+            sprintf "Callable %O did not support the expected functors" call.FullName
+        )
+
+    let AsserLambdaFunctorsByLine result line parentName expectedFunctors =
```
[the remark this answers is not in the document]
ScottCarda-MS: fixed

src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs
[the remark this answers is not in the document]
samarsha: > These should somehow be private static member functions

I don't think they should - they don't need to access any other instance member, so they work fine as `let`s.

src/QsCompiler/Tests.Compiler/TestUtils/Signatures.fs
samarsha: > Since I can't know the generated names of the lifted operations here

You could use a deterministic naming scheme, like numbering them by the order they appear. That would make it easier to write tests with asserts for specific lambdas.

src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs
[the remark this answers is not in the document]
samarsha: I guess Xunit doesn't support an assert message for Assert.Equal with IEnumerable. You might also need to turn the list into a seq. One of these should work:

```fsharp
Assert.Equal([ "return 0;" ], lines)
Assert.Equal(seq { "return 0;" }, lines)
```

src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs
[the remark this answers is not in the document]
samarsha: I'd be interested in some additional assertions, like

- What are the names of the parameters of the lifted lambda?
- What's the body of the lifted lambdas? Does it preserve the names of the arguments and the enclosed variables?
- Does the original lambda location get replaced with the right form of partial application? Are the arguments/closure tuple in the right order?

src/QsCompiler/Tests.Compiler/TestCases/LambdaLifting.qs
```diff
@@ -0,0 +1,219 @@
+﻿// Copyright (c) Microsoft Corporation.
```
samarsha, line 1: Some ideas for additional test cases:

- Lambda that uses its arguments, e.g. `x -> x` or `(x, y) -> (y, x)`
- Lambda that uses both its arguments and enclosed variables, e.g. `let a = 0; let f = x -> (a, x);`
- Lambda that discards parameters, e.g. `(x, _, _) -> x`.

src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs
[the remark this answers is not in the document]
ScottCarda-MS: I've settled for using the F# equality comparison `=` and wrapping that with the XUnit's Assert.True.

src/QsCompiler/Tests.Compiler/TestCases/LambdaLifting.qs
ScottCarda-MS: I can make tests for the first two suggestions. But for the third bullet point, lambda's can't be defined with missing parameters.

src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs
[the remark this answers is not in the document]
ScottCarda-MS: I guess they don't really work well as `let` bindings if you are asserting that they would then need to be camelCase, which isn't desirable, and they being let bindings doesn't accurately my intention for them. My suggestion for them to be static members is a compromise.

src/QsCompiler/Tests.Compiler/TestUtils/Signatures.fs
```diff
@@ -609,6 +621,195 @@ let public ClassicalControlSignatures =
...
+                     })
+                    |> QsTypeKind.Operation |]
+
+/// Expected callable signatures to be found when running Lambda Lifting tests
+let public LambdaLiftingSignatures =
+    [| // Basic Lift
+        (_DefaultTypes,
+         [|
+             LambdaLiftingNS, "Foo", [||], "Unit" // The original operation
+             LambdaLiftingNS, "_Foo", [||], "Int" // The generated operation
```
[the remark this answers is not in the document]
ScottCarda-MS: If this there is a small fix for what your asking, then please point it out and I can look into it. But I'm concerned that what you're suggesting would be a major refactor of the testing infrastructure, and if that's the case, it needs to be addressed in a separate PR as a separate work item.

src/QsCompiler/Tests.Compiler/TestUtils/Signatures.fs
ScottCarda-MS: I'm not interested in changing the naming scheme that's used throughout the compiler. I'm not sure there would be a significant benefit to that, and it would take a bit of looking into to make it work right.

src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs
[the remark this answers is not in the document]
samarsha: I don't understand why `camelCase` is not desirable? That is the standard F# naming convention.

samarsha (changes requested):

src/QsCompiler/Tests.Compiler/TestUtils/Signatures.fs
samarsha: In terms of benefits from testing, I might be missing something but I'm not sure how you are testing that the lambdas are substituted in the right places. For example, how do you assert that:

```qsharp
F(x -> x + 1, x -> x + 2);
```

transforms into:

```qsharp
function Lambda1(x : Int) : Int {
    return x + 1;
}

function Lambda2(x : Int) : Int {
    return x + 2;
}

F(Lambda1, Lambda2);
```

and not (incorrectly):

```qsharp
F(Lambda2, Lambda1);
```

Would this be caught by the existing test suite?

src/QsCompiler/Transformations/LiftLambdas.cs
```diff
@@ -0,0 +1,243 @@
...
+                                    .ToImmutableArray());
+                            }
+                            else
+                            {
+                                throw new ArgumentException("Lambda parameter type mismatch");
+                            }
+                        }
+                        else
+                        {
+                            throw new ArgumentException("Lambda parameter type mismatch");
```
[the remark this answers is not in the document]
ScottCarda-MS: Callables in Q# can't be defined with missing parameters, so lambdas shouldn't be defined with missing parameters either.

src/QsCompiler/Tests.Compiler/TestUtils/Signatures.fs
samarsha: > There are several tests so far that check that the original location of the lambda is replaced with the call to the lifted operation.

Yes, I saw the tests that check that the lambda is replaced by a reference to a lifted lambda, but it didn't look like they distinguish between which lifted lambda, if there is more than one. If the testing infrastructure can distinguish between the two cases above that would be good to know.

src/QsCompiler/Transformations/LiftLambdas.cs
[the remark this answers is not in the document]
samarsha: You can generate names for the parameters when lifting to work around that. Lambda parameters are designed to be like `let` bindings (no type annotations, and support for discards), rather than global callable declarations. It's convenient to be able to discard parameters since brevity is more important for lambdas than global callable declarations.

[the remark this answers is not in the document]
samarsha: (Although it might be worth checking if having a discard as a callable parameter actually causes issues. I think it's prevented by the parser, but I'm not sure if any later step currently depends on that.)

src/QsCompiler/Tests.Compiler/TestUtils/Signatures.fs
samarsha: So in terms of the test infrastructure, I think I'm looking for a test case where you have something like

```fsharp
GetCallablesWithSuffix result Signatures.LambdaLiftingNS "_Foo"
```

but *without* expecting there to be one element in the returned sequence. Right now all the tests that use `GetCallablesWithSuffix` also call `Seq.exactlyOne`. Then the test would determine which lambda has which body, and assert that the original lambdas were replaced with the correct corresponding lifted ones.

src/QsCompiler/Tests.Compiler/TestUtils/Signatures.fs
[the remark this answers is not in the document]
samarsha: My suggestion is: In the lambda lifting tests, you look up the `QsCallable` for the original callable and the lifted lambdas. I think it would be clearer if you asserted that the `QsCallable.Signature` matched the expected type in LambdaLiftingTests.fs, rather than in Signatures.fs.

src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs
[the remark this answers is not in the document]
ScottCarda-MS: `camelCase` is something that should be used to distinguish local variables. `PascalCase` is something that should be used to distinguish functions. It is very useful to be able to distinguish between things that are scoped to the immediate function versus things that are available to a wider scope.

src/QsCompiler/Transformations/LiftLambdas.cs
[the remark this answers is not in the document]
ScottCarda-MS: > Lambda parameters are designed to be like `let` bindings (no type annotations, and support for discards), rather than global callable declarations. It's convenient to be able to discard parameters since brevity is more important for lambdas than global callable declarations.

Okay, thanks for the clarity on the difference. That makes sense.

> You can generate names for the parameters when lifting to work around that. (Although it might be worth checking if having a discard as a callable parameter actually causes issues. I think it's prevented by the parser, but I'm not sure if any later step currently depends on that.)

My preference is that, if it is not OK with the parser, we shouldn't lean into it, even if nothing depends on it. Ideally, we would be able to take the code generated at the end of the transformation steps and be able to feed it back into the compiler from the start, although there are a few caveats for that like the use of the reserved double-underscore identifiers. I will look into generating parameters for these missing parameters, probably using the double-underscore identifiers to avoid colliding with existing symbols.

src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs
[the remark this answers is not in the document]
samarsha: Note that that's a C# naming convention - since this is F# code, it's preferrable to follow the conventions of the language being used. In this case, the `let`s are local variables (they are scoped to the constructor - in F# all members are also in the constructor's scope). IMO it's somewhat less idiomatic to make them private static members but that would also work.

> How do I specify them as static private members in F#?

Here's the documentation on defining methods and access control:

* https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/members/methods
* https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/access-control

src/QsCompiler/Tests.Compiler/TestUtils/Signatures.fs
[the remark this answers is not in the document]
ScottCarda-MS: Okay thank you for the suggestion, but unfortunately I'm going to have to reject that suggestion for this PR, as that would be too big a change to the testing infrastructure, and not particularly relevant to the purpose of this PR. If you'd like, you can file a GitHub issue for this refactor, and we can take it from there.

src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs
```diff
@@ -0,0 +1,573 @@
...
+                 else
+                     true)
+
+        charMatch.Value && adjMatch.Value && ctlMatch.Value
+
+    let AssertCallSupportsFunctors expectedFunctors call =
+        Assert.True(
+            DoesCallSupportFunctors expectedFunctors call,
+            sprintf "Callable %O did not support the expected functors" call.FullName
+        )
```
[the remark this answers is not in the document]
ScottCarda-MS: Common functionality has been moved to the TestUtils module.

src/QsCompiler/Tests.Compiler/TestUtils/Signatures.fs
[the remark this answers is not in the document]
samarsha: Sorry, but to clarify, I think it is relevant to this PR because it seems like it prevents the lambda lifting tests from having adequate test coverage of the lifted lambda types. For example, in the "Functor Support Basic Return" test:

```qsharp
// Functor Support Basic Return
namespace Microsoft.Quantum.Testing.LambdaLifting {
    operation Foo() : Unit {
        let lambda1 = () => 0;
        let lambda2 = () => ();
    }
}
```

```fsharp
// Functor Support Basic Return
(_DefaultTypes,
 [|
     LambdaLiftingNS, "Foo", [||], "Unit" // The original operation
     LambdaLiftingNS, "_Foo", [||], "Int"
     LambdaLiftingNS, "_Foo", [||], "Unit"
 |])
```

This asserts that there exists a lifted lambda of type `Unit => Int` and a lifted lambda of type `Unit => Unit`. But I couldn't  find any assertion that `lambda1` was specifically substituted with a lifted lambda of type `Unit => Int`, and `lambda2` was specifically substituted with a lifted lambda of type `Unit => Unit` (and not vice versa). Is this correct?

src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs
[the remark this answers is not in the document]
ScottCarda-MS: I had hoped this would become a moot discussion once I moved these items to the TestUtils, but the flaw in this F# standard seems to be even more egregious now. These functions are now defined in a totally different module as publicly available members of that module, and act as public static members (there should be some way of specifying that they're static). They are even more removed from the places in which they are called now, and are not in any way local. This should be made clear with their name.
I should also qualify them with the module name. Let me make that change quick.

src/QsCompiler/Tests.Compiler/TestCases/LambdaLifting.qs
ScottCarda-MS: As per [this comment](https://github.com/microsoft/qsharp-compiler/pull/1302#discussion_r791036804) I will look into adding tests that cover all three of these bullet points.

src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs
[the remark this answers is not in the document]
samarsha: Hmm, I'm not sure I'm following exactly... even in C#, camelCase and PascalCase doesn't distinguish local/non-local bindings. For example, private fields in C# classes are `camelCase` even though they are class-scoped. Also, local functions in C# are `PascalCase`, even though they are scoped to a single method.

In any case it's generally a good idea to follow the conventions of the language being used, even if it's different from what you're used to in other languages.

src/QsCompiler/Tests.Compiler/TestUtils/Signatures.fs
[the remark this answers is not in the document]
samarsha: OK, so ignoring the file organization for now, I'm just asking for this kind of assertion to be made somewhere in the test suite. I couldn't find a test case that does that, but I might have missed one.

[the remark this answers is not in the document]
ScottCarda-MS: This should be satisfied once I've added a test for the case suggested [here](https://github.com/microsoft/qsharp-compiler/pull/1302#discussion_r791074688).

src/QsCompiler/Transformations/ClassicallyControlled.cs
```diff
@@ -996,8 +996,16 @@ public override QsStatementKind OnConditionalStatement(QsConditionalStatement st
...
                             // Lift the scope to its own operation
-                            if (this.SharedState.LiftBody(block.Body, false, out var call, out var callable))
+                            if (this.SharedState.LiftBody(block.Body, null, false, out var call, out var callable))
                             {
+                                // Make sure the inferred information for the callable has local quantum dependency
+                                call = new TypedExpression(
+                                    call.Expression,
+                                    call.TypeArguments,
+                                    call.ResolvedType,
+                                    new InferredExpressionInformation(false, true),
```
swernli, line 1006: To make sure I understand, does adding the quantum dependency here apply to the call expression or to the whole lifted function? If the former then I think this makes sense, but I'm a little confused if it means the whole lifted function would inherit a dependency.

ScottCarda-MS: This just sets the inferred information for the call expression, so it won't affect the lifted callable. The calls for the classically controlled are set this way to reflect the fact that these calls are dependent on the results of quantum measurements.

src/QsCompiler/Transformations/LiftLambdas.cs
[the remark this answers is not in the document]
ScottCarda-MS: This should be supported now.

src/QsCompiler/Tests.Compiler/TestCases/LambdaLifting.qs
ScottCarda-MS: added

src/QsCompiler/Tests.Compiler/TestUtils/Signatures.fs
ScottCarda-MS: I was thinking of the `identifyGeneratedByCalls` function, but that is limited to looking at lines that are just call expressions. I've added the tests case you suggested above, and just used logic written in that test case. If it is something I use more as I continue to refactor code, I will look into making that logic a TestUtil function as well.

src/QsCompiler/Tests.Compiler/TestUtils/Signatures.fs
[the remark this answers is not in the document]
ScottCarda-MS: I've added a test case that checks for this now.

src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs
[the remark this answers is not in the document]
ScottCarda-MS: I don't think I can reasonable accomplish those assertions with these test case ("With Lots of Params" and "Use Closure With Params"). I've added some additional tests since this comment was posted, and I believe the second bullet point is a covered scenario now. If you ignore these two tests and look at the rest of the tests, what scenarios would you say still need coverage?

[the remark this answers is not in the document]
ScottCarda-MS: I've updated the tests ("With Lots of Params" and "Use Closure With Params") to check the parameter names of the lambdas, that the correct lambdas are called in the right places, and that the argument tuples are as expected. I've added some additional tests since this comment was posted, and I believe the second bullet point is a covered scenario now as well, but by other tests since these tests are meant to focus on the parameters.

src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs
ScottCarda-MS: resolved, check [this comment](https://github.com/microsoft/qsharp-compiler/pull/1302#discussion_r792346904)

swernli:
I've finished reviewing the non-test code and left some minor feedback. I'll continue reviewing the test code and following along with the ongoing discussion there.

src/QsCompiler/Transformations/ContentLifting.cs
```diff
@@ -16,341 +16,429 @@
...
+            {
+                parameters = ConcatParams(parameters, additionalParameters);
+            }
 
-                    arguments = new TypedExpression(
-                        ExpressionKind.NewValueTuple(argumentArray),
+            var (generatedCallable, generatedCallableCallType) = GenerateCallable(callingCallable, body, parameters, callableInformation, isFunction, returnType);
+
+            // Forward the type parameters of the parent callable to the type arguments of the call to the generated callable.
+            // ToDo: this may need to be per-specialization
```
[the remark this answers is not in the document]
ScottCarda-MS: I'd like to get some clarity on type specializations from @bettinaheim.

src/QsCompiler/Transformations/LiftLambdas.cs
```diff
@@ -0,0 +1,261 @@
...
+                                    throw new ArgumentException("Lambda parameter type length mismatch");
+                                }
+
+                                return ParameterTuple.NewQsTuple(subSymbols
+                                    .Select((symbol, i) => MatchNameWithType(subSybmolTypes[i], symbol))
+                                    .ToImmutableArray());
+                            }
+                            else
+                            {
+                                throw new ArgumentException("Lambda parameter type mismatch");
```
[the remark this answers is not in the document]
ScottCarda-MS: I've updated these messages a bit now, and they are all unique and, I think, a bit more informative.

src/QsCompiler/Transformations/ContentLifting.cs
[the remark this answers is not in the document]
bettinaheim: Replacing it with a specific tag seems reasonable to me. While the data structures in principle make room for it, it is not clear we will or should pursue this. I hence wouldn't file it just yet.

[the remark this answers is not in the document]
ScottCarda-MS: I've changed the tag to 'Future', as discussed with @bettinaheim.

src/QsCompiler/Transformations/ContentLifting.cs
```diff
@@ -689,22 +778,184 @@ public override ExpressionKind OnIdentifier(Identifier sym, QsNullable<Immutable
...
+
+            internal QsSpecialization? ControlledAdjoint { get; }
+
+            internal QsNullable<ImmutableArray<ResolvedType>> TypeParameters { get; }
+
+            internal CallableDetails(QsCallable callable)
+            {
+                this.Callable = callable;
+
+                // ToDo: this may need to be adapted once we support type specializations
```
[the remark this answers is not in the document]
ScottCarda-MS: I've changed the tag to 'Future'. See [this comment](https://github.com/microsoft/qsharp-compiler/pull/1302#discussion_r793154998).

src/QsCompiler/Transformations/LiftLambdas.cs
```diff
@@ -0,0 +1,261 @@
...
+                        ? op.Item2
+                        : ex.ResolvedType.Resolution is ResolvedTypeKind.Function
+                        ? new CallableInformation(ResolvedCharacteristics.Empty, InferredCallableInformation.NoInformation)
+                        : throw new ArgumentException("Lambda with non-callable type");
+
+                    // Returns are allowed only if we are not returning Unit from an operation.
+                    var isReturnStatement =
+                        !(ex.ResolvedType.Resolution is ResolvedTypeKind.Operation opType) || !opType.Item1.Item2.Resolution.IsUnitType;
+
+                    // Alternatively:
```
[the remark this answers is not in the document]
ScottCarda-MS: I'd like to wait and play around with both strategies after the lambda characteristics bug is addressed. We can leave this open until then.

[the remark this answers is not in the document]
ScottCarda-MS: I've switched to using the other 'Alternative' strategy now. After the characteristics bug was fixed I was playing around with these two and noticed some edge cases that were not well-handled by the first strategy. I've added some test cases for those edge cases now as well. See [Function Without Return Value](https://github.com/microsoft/qsharp-compiler/blob/sccarda/LambdaLifter/src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs#L687) and [Return Unit-Typed Expression](https://github.com/microsoft/qsharp-compiler/blob/sccarda/LambdaLifter/src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs#L713) tests.

src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs
```diff
@@ -0,0 +1,733 @@
...
+            |> TestUtils.getBodyFromCallable
+
+        let generated =
+            TestUtils.getCallablesWithSuffix result Signatures.LambdaLiftingNS "_Foo"
+            |> Seq.exactlyOne
+            |> TestUtils.getBodyFromCallable
+
+        let lines = generated |> TestUtils.getLinesFromSpecialization
+
+        let expectedContent = [| "return (y, x);" |]
```
samarsha, line 561: For these tests (Use Parameter Tuple, Use Parameter Tuple, Use Parameter and Closure, Use Parameter with Missing Params), I was hoping to have assertions about the order of names in the lifted function's parameter tuple. Otherwise, we don't know that `y` and `x` are referring to the correct parameters. (The idea being to catch bugs in how the lifter copies the lambda parameters and combines them with the closure parameters.)

src/QsCompiler/Tests.Compiler/TestUtils/TestUtils.fs
```diff
@@ -276,3 +280,206 @@ let testExpr (str, succExp, resExp, diagsExp) =
...
+            Assert.True(x <> None, "Did not find expected generated content")
+            rtrn <- Seq.append rtrn [ Seq.item x.Value callables |> fst ]
+            callables <- removeAt x.Value callables)
+
+    rtrn
+
+
+// utils for testing checks and assertions
+
+let checkIfStringIsCall = _checkIfStringIsCall
```
samarsha, line 424: nit: Can you update the original `let private _checkIfStringIsCall` instead?

```suggestion
```

src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs
ScottCarda-MS: The tests `With Lots of Params` and `Use Closure With Params` already test that the parameters for lambdas are generated correctly in a variety of cases. I don't see how that applies to these tests you mention as these are for testing cases where the parameters are used, and don't in general have a variety of parameter configurations to test against.

src/QsCompiler/Tests.Compiler/TestUtils/TestUtils.fs
ScottCarda-MS: It is done this way because let bindings are processed in order from top to bottom in F# and I needed the `_checkIfStringIsCall` for the `identifyGeneratedByCalls` function. However the proper `checkIfStringIsCall` function should be in the last grouping of functions. The requirement that functions be defined in a specific order is causing this awkward solution.

ScottCarda-MS: Alternatively, I have considered that maybe both the `checkIfStringIsCall` and the `identifyGeneratedByCalls` are not that generally useful and should be moved back to the ClassicallyControlled test file.

src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs
samarsha: "With Lots of Params" and "Use Closure With Params" don't make assertions about the content of the lambdas. Because the assertions about the parameter tuples use a different test case from the assertions about the contents, the tests don't establish that a lambda is lifted with both the right variable names in the body and the right parameter names in the declaration.

src/QsCompiler/Tests.Compiler/TestUtils/TestUtils.fs
samarsha: OK. I would personally go with a slightly wonky function grouping that reflects the dependencies accurately, instead of declaring the same function twice with two different names, but it's up to you.

src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs
ScottCarda-MS: It seems unreasonable to have unit tests for cross products of independent components of a feature. We have tests for checking that parameter lists are generated correctly, and we have tests that check that the content is generated correctly. We even have tests for check that the parameters are generated correctly when those parameters are expected to be influenced by the content. Beyond what we have, I don't see a need to further complicate unit tests that already accomplish what they are meant to check, and I don't see a gap in test coverage.

src/QsCompiler/Tests.Compiler/TestUtils/TestUtils.fs
ScottCarda-MS: I've moved the functions dealing with checking if lines are calls back to the Classically Controlled test module.

src/QsCompiler/Tests.Compiler/LambdaLiftingTests.fs
samarsha: No, I agree, I didn't mean that there needs to be a cross product of testing properties. I had a simple additional test case in mind, but I think on second thought the current tests probably cover it.

src/QsCompiler/Transformations/ContentLifting.cs
```diff
@@ -16,341 +16,429 @@
...
+            else if (parameters is ParameterTuple.QsTuple tuple)
+            {
+                return tuple.Item.SelectMany(FlattenParamTuple);
+            }
+
+            return ImmutableArray<LocalVariableDeclaration<QsLocalSymbol>>.Empty;
         }
 
-        public class TransformationState
+        private static ResolvedType ExractParamType(ParameterTuple parameters)
```
samarsha, line 42:
```suggestion
        private static ResolvedType ExtractParamType(ParameterTuple parameters)
```

src/QsCompiler/Transformations/LiftLambdas.cs
```diff
@@ -0,0 +1,257 @@
...
+                                    : paramName.Range.Item);
+                            return ParameterTuple.NewQsTupleItem(localVar);
+                        }
+                        else if (paramName.Symbol is QsSymbolKind<QsSymbol>.SymbolTuple tup)
+                        {
+                            if (tup.Item.Length == 0 && paramType.Resolution.IsUnitType)
+                            {
+                                // Need an artificial Unit parameter here
+                                var localVar = new LocalVariableDeclaration<QsLocalSymbol>(
+                                    QsLocalSymbol.NewValidName("__lambdaUnitParam__"),
```
samarsha, line 156: Is this to handle `() -> ...` lambdas? The declarations `function F() : Unit {}` and `function F(x : Unit) : Unit {}` should be equivalent in Q#, is the tuple item needed?

src/QsCompiler/Transformations/ContentLifting.cs
ScottCarda-MS: done

src/QsCompiler/Transformations/LiftLambdas.cs
```diff
@@ -0,0 +1,257 @@
...
+                                var subSymbols = tup.Item;
+                                var subSybmolTypes = tupType.Item;
+
+                                if (subSymbols.Length != subSybmolTypes.Length)
+                                {
+                                    throw new ArgumentException("Lambda parameter tuple type length mismatch.");
+                                }
+
+                                return ParameterTuple.NewQsTuple(subSymbols
+                                    .Select((symbol, i) => MatchNameWithType(subSybmolTypes[i], symbol))
```
samarsha, line 180:
```suggestion
                                var subSymbolTypes = tupType.Item;

                                if (subSymbols.Length != subSymbolTypes.Length)
                                {
                                    throw new ArgumentException("Lambda parameter tuple type length mismatch.");
                                }

                                return ParameterTuple.NewQsTuple(subSymbols
                                    .Select((symbol, i) => MatchNameWithType(subSymbolTypes[i], symbol))
```

src/QsCompiler/Transformations/LiftLambdas.cs
```diff
@@ -0,0 +1,257 @@
...
+                        QsComments.Empty);
+                    var generatedContent = new QsScope(
+                        lambdaBody.Expression.IsUnitValue
+                            ? ImmutableArray<QsStatement>.Empty // if it is just a single unit literal, there should be an empty body
+                            : new[] { bodyStatment }.ToImmutableArray(),
+                        new LocalDeclarations(this.SharedState.KnownVariables));
+
+                    var success = processedLambda.Kind.IsFunction
+                        ? ContentLifting.LiftContent.LiftFunctionBody(this.SharedState.CurrentCallable!, generatedContent, lambdaParams, isReturnStatement, out var call, out var callable)
+                        : ContentLifting.LiftContent.LiftOperationBody(this.SharedState.CurrentCallable!, generatedContent, lambdaParams, callableInfo, isReturnStatement, out call, out callable);
```
samarsha, line 237: In what situations can lifting fail?

src/QsCompiler/Transformations/LiftLambdas.cs
ScottCarda-MS: Consider the case:
```qsharp
let x = 3;
let lambda = () => x;
```
Here the parameter tuple needs to be mapped to a parameter `__lambdaUnitParam__` so that it can appear in the generated parameter list as `<lambda name>(x : Int, __lambdaUnitParam__ : Unit) : Int { return x; }` so that the original position of the lambda can be a partial application: `let lembda = <lambda name>(x, _);` Without this parameter the original expressions would be a completed call rather than the partial it needs to be.

ScottCarda-MS: It _does_ also mean that `() => ...` gets mapped to `<lambda name>(__lambdaUnitParam__ : Unit)` and the call `<lambda name>(_)` when there is no closure, which is equivalent, as you point out, to the lambda if it were defined and called with no parameters, but this is a relatively harmless side effect, I think.

src/QsCompiler/Transformations/LiftLambdas.cs
ScottCarda-MS: done

src/QsCompiler/Transformations/LiftLambdas.cs
ScottCarda-MS: For lifting in general, a block of code will fail to be lifted (or in other words is not liftable) if it contains a set statement on a mutable defined outside of the block of code, or if it either contains a return statement when not expecting a return or does not contain a return statement when expecting a return, as determined by the parameter `isReturnsAllowed`. It will also fail if multiple return statements are found that disagree on the return type. Most of this is beyond the scope of the lambda lifter as it only lifts a single expression.

ScottCarda-MS: In that sense, maybe the compiler should throw an error if `success` is false, as I don't think it should be possible for a Q# writer to write a lambda that can't be lifted. Failure here should be an indicator of the compiler failing on some level.

samarsha: Yeah, I think `success` being false is an internal compiler error, so throwing seems better.

src/QsCompiler/Transformations/ContentLifting.cs
```diff
@@ -16,341 +16,429 @@
...
-                    ExpressionKind.NewIdentifier(
-                        Identifier.NewGlobalCallable(generatedCallable.FullName),
-                        typeArguments),
-                    typeArguments.IsNull
-                        ? TypeArgsResolution.Empty
-                        : typeArguments.Item
-                            .Select(type => Tuple.Create(generatedCallable.FullName, ((ResolvedTypeKind.TypeParameter)type.Resolution).Item.TypeName, type))
-                            .ToImmutableArray(),
-                    generatedCallableCallType,
+            var isAdditionalParams = true;
```
samarsha, line 241: nit: Maybe `hasAdditionalParams`?

src/QsCompiler/Transformations/LiftLambdas.cs
ScottCarda-MS: I've changed this to throw an error on failure.

samarsha (approved):

src/QsCompiler/Transformations/ContentLifting.cs
ScottCarda-MS: sure

PR #1303, opened 2022-01-13
Fixing an issue where the runtime expects callable to be non-null

bettinaheim:
The QIR generation used null values to represent default values for callables when they were created as part of a (deprecated) new array expression. At the same time, the runtime functions for references count changes of the capture expects the callable to be non-null. Given that we want to eliminate such invalid (null) default values, it seems more reasonable to fix this compiler side such that the runtime can omit all null-checks.


src/QsCompiler/Tests.Compiler/TestCases/QirTests/ExecutionTests.qs
```diff
@@ -1651,6 +1651,26 @@
...
+        Message($"{new Int[3]}");
+        Message($"{new Double[3]}");
+        Message($"{new Bool[3]}");
+        Message($"{new Pauli[3]}");
+        Message($"{new String[3]}");
+        Message($"{new Range[3]}");
+        Message($"{new Qubit[3]}");
+        Message($"{new Result[3]}");
+        Message($"{new Unit[3]}");
+        // Todo: bigint is not yet supported in the runtime
```
kuzminrobin, line 1665: FYI: One of the best practices recommends for every "Todo" to provide either the alias or the bug number, e.g.
"Todo(**bettinaheim**): ..." or
"Todo(**\<bug number\>**): ...".

kuzminrobin (approved):

ltalirz:
@bettinaheim Can this be merged?

Would be great to have this available to test next week

cesarzc (approved):

bettinaheim:
> @bettinaheim Can this be merged?
> 
> Would be great to have this available to test next week

I just didn't get around addressing Robin's comment. I'll do that and merge it later today.

PR #1304, opened 2022-01-14
Add multiplatform testing to feature/llvm13

swernli:
This adds multiplatform compilation and testing to the compiler pipeline. Some tests are still skipped for non-Windows environments. QIR execution tests are temporarily skipped until better JIT support can be brought online for them.


src/QsCompiler/Tests.Compiler/QirTests.fs
```diff
@@ -271,3 +271,6 @@ let ``QIR Library generation`` () =
         "TestLibraryGeneration"
         compilerArgs
         [ "TestLibraryGeneration1"; "TestLibraryGeneration2"; "TestLibraryGeneration3" ]
+
+[<Fact(Skip = "Fails on Mac")>]
```
swernli, line 275: I need to add details of the Issue once I file one.

swernli:
@bettinaheim, I've seen intermittent failures with the QIR execution tests on Windows (they are still suppressed for other platforms), and the weird inconsistencies in the failures makes me think that the JIT hack of using LLVMSharp v12 to run QIR from LLVM 13 is not actually working. As previously discussed, we want to get rid of this hack anyway by trying out producing JIT capable binaries ourselves, and this just shows that we need that for Windows as well.

build/build.ps1
```diff
@@ -7,11 +7,37 @@ $ErrorActionPreference = 'Stop'
 $all_ok = $True
 Write-Host "Assembly version: $Env:ASSEMBLY_VERSION"
 
-choco install llvm --version=11.1.0 --allow-downgrade
-if (!(Get-Command clang -ErrorAction SilentlyContinue) -and (choco find --idonly -l llvm) -contains "llvm") {
-    # For some reason, adding to the path does not work on our build servers, even after calling refreshenv.
-    # LLVM was installed by Chocolatey, so add the install location to the path.
-    $env:PATH += ";$($env:SystemDrive)\Program Files\LLVM\bin"
+if (($IsWindows) -or ((Test-Path Env:/AGENT_OS) -and ($Env:AGENT_OS.StartsWith("Win")))) {
+    if (!(Get-Command clang        -ErrorAction SilentlyContinue) -or `
```
idavis, line 11: Why do we need clang and clang-format? Is it just for the development project's sample?

build/build.ps1
```diff
@@ -7,11 +7,37 @@ $ErrorActionPreference = 'Stop'
...
+        !(Get-Command clang-format -ErrorAction SilentlyContinue) -or `
+        (Test-Path Env:/AGENT_OS)) {
+        choco install llvm --version=13.0.0 --allow-downgrade
```
idavis, line 14: I'm worried that prepending LLVM to the path effects the private lookup path, but I'm not 100% sure.

build/test.ps1
```diff
@@ -41,7 +43,9 @@ function Test-One {
 
 Test-One '../QsCompiler.sln'
 Test-One '../src/Telemetry/Telemetry.sln'
-Test-One '../QsFmt.sln'
+if ($IsWindows) {
```
idavis, line 46: Does the Q# formatter not work in VSCode?

examples/QIR/Development/Development.csproj
```diff
@@ -133,7 +123,7 @@
       <SimulatorRuntime Condition="$([MSBuild]::IsOsPlatform('Linux'))">$(PkgMicrosoft_Quantum_Simulators)/runtimes/linux-x64/native/libMicrosoft.Quantum.Simulator.Runtime.so</SimulatorRuntime>
       <ClangName Condition="$([MSBuild]::IsOsPlatform('Windows')) Or $([MSBuild]::IsOsPlatform('OSX'))">clang++</ClangName>
       <ClangName Condition="$([MSBuild]::IsOsPlatform('Linux'))">clang++-11</ClangName>
```
idavis, line 125: Should this be `clang++-13`?

build/test.ps1
swernli: It is supposed to, but the tests fail on non-Windows platforms. Could just be an issue of line endings, which would hopefully be a simple fix. But for now I'm skipping them just to try and get the infrastructure stood up. I'll file an issue for follow up!

examples/QIR/Development/Development.csproj
swernli: Yup, you are totally right. That's what I get for copy pasting too quickly ;)

build/build.ps1
swernli: It's an interesting question... but I think it's mitigated somewhat by how we are trying to ensure the compiler QIR Generation step explicitly loads the binaries that we build, even if the environment has another LLVM in it. So it shouldn't interfere with the generation step, and if it does then that would be something we should fix in the compiler itself.

build/test.ps1
swernli: Added link to issue.

build/build.ps1
swernli: I'm going to try an approach that removes the need for installing these. Here's hoping it works...

swernli: I think I have a viable approach here, let me know what you think. I'll double check with @bettinaheim to make sure this makes sense too.

build/build.ps1
swernli: I think we can avoid installing clang entirely in the script, so this has been removed.

swernli:
I've simplified the commit history. This should now build clean across all platforms, including QIR execution tests.

examples/QIR/Development/Development.csproj
```diff
@@ -131,9 +121,8 @@
...
       <SimulatorRuntime Condition="$([MSBuild]::IsOsPlatform('Windows'))">$(PkgMicrosoft_Quantum_Simulators)/runtimes/win-x64/native/Microsoft.Quantum.Simulator.Runtime.dll</SimulatorRuntime>
       <SimulatorRuntime Condition="$([MSBuild]::IsOsPlatform('Linux'))">$(PkgMicrosoft_Quantum_Simulators)/runtimes/linux-x64/native/libMicrosoft.Quantum.Simulator.Runtime.so</SimulatorRuntime>
-      <ClangName Condition="$([MSBuild]::IsOsPlatform('Windows')) Or $([MSBuild]::IsOsPlatform('OSX'))">clang++</ClangName>
-      <ClangName Condition="$([MSBuild]::IsOsPlatform('Linux'))">clang++-11</ClangName>
-      <ClangCommand>$(ClangName) -o $(ExecutablePath) $(QirOutputPath)$(PathCompatibleAssemblyName).ll $(BuildOutputPath)/Main.cpp -I$(BuildOutputPath) -L$(BuildOutputPath) -lMicrosoft.Quantum.Qir.Runtime -lMicrosoft.Quantum.Qir.QSharp.Core -lMicrosoft.Quantum.Qir.QSharp.Foundation</ClangCommand>
+      <ClangOptions Condition="$([MSBuild]::IsOsPlatform('Linux')) Or $([MSBuild]::IsOsPlatform('OSX'))">-fseh-exceptions -lstdc++</ClangOptions>
```
idavis, line 124: Why do we want `-fseh-exceptions` ?

llvm-build/llvm.cmake
```diff
@@ -72,4 +72,4 @@ else()
   set(LLVM_BUILD_LLVM_DYLIB ON CACHE BOOL "")
   set(CPACK_BINARY_TGZ ON CACHE BOOL "")
 endif()
-set(LLVM_DYLIB_COMPONENTS "core;debuginfodwarf;linker;support;target;bitwriter;analysis" CACHE STRING "")
+set(LLVM_DYLIB_COMPONENTS "core;debuginfodwarf;linker;support;target;bitwriter;analysis;executionengine;runtimedyld;mcjit;bitstreamreader;bitreader;native" CACHE STRING "")
```
idavis, line 75: @swernli Might be good to add a permalink to the docs page that lists theses dep chains.

src/QsCompiler/CommandLineTool/CommandLineTool.csproj
```diff
@@ -32,6 +32,14 @@
     <ProjectReference Include="..\Compiler\Compiler.csproj" />
   </ItemGroup>
 
+  <ItemGroup Condition="!$([MSBuild]::IsOsPlatform('Windows'))">
```
idavis, line 35: Why aren't we doing this on Windows?

src/QsCompiler/LlvmBindings/Extensions/MarshaledStringArray.cs
```diff
@@ -0,0 +1,57 @@
...
+                    this.Values[i] = new MarshaledString(inputs[i].AsSpan());
+                }
+            }
+        }
+
+        public int Count { get; private set; }
+
+        public MarshaledString[]? Values { get; private set; }
+
+        public void Dispose()
```
idavis, line 32: Temp value to preclude NRE unless this is a covered idiom nowadays.
```csharp
        public void Dispose()
        {
            var values = this.Values;
            if (values is object)
            {
                for (int i = 0; i < values.Length; i++)
                {
                    values[i].Dispose();
                }

                this.Values = null;
                this.Count = 0;
            }
        }
```

src/QsCompiler/LlvmBindings/Interop/LLVMModuleRef.cs
```diff
@@ -271,6 +291,36 @@ public void SetModuleInlineAsm(ReadOnlySpan<char> asm)
...
+        public bool TryCreateMCJITCompiler(out LLVMExecutionEngineRef outJIT, ref LLVMMCJITCompilerOptions options, out string outError)
+        {
+            fixed (LLVMExecutionEngineRef* pOutJIT = &outJIT)
+            {
+                fixed (LLVMMCJITCompilerOptions* pOptions = &options)
+                {
+                    sbyte* pError = null;
+                    var result = LLVM.CreateMCJITCompilerForModule((LLVMOpaqueExecutionEngine**)pOutJIT, this, pOptions, (UIntPtr)Marshal.SizeOf<LLVMMCJITCompilerOptions>(), &pError);
+
+                    if (pError == null)
```
idavis, line 309: nit
`if (pError is null)`

src/QsCompiler/LlvmBindings/Extensions/MarshaledStringArray.cs
```diff
@@ -0,0 +1,57 @@
...
+                }
+
+                this.Values = null;
+                this.Count = 0;
+            }
+        }
+
+        public void Fill(sbyte** pDestination)
+        {
+            if (this.Values != null)
```
idavis, line 48: nit
`if (this.Values is object)`

src/QsCompiler/LlvmBindings/Interop/LLVMExecutionEngineRef.cs
```diff
@@ -0,0 +1,162 @@
...
+        }
+
+        public bool TryRemoveModule(LLVMModuleRef m, out LLVMModuleRef outMod, out string outError)
+        {
+            fixed (LLVMModuleRef* pOutMod = &outMod)
+            {
+                sbyte* pError = null;
+                var result = LLVM.RemoveModule(this, m, (LLVMOpaqueModule**)pOutMod, &pError);
+
+                if (pError == null)
```
idavis, line 148: nit
`if (pError is null)`

src/QsCompiler/TestTargets/Simulation/Target/QirExecution.cs
```diff
@@ -62,31 +62,54 @@ private static unsafe bool TryParseBitcode(string path, out LLVMSharp.Interop.LL
...
 
+        private static unsafe void ExplicitLibraryLoad(string libraryName)
+        {
+            if (LLVM.LoadLibraryPermanently(libraryName.AsMarshaledString()) != 0)
+            {
+                throw new ExternalException($"Failed explicit load of library '{libraryName}'");
+            }
+        }
+
         public static unsafe void BuildAndRun(string pathToBitcode, params string[] functionNames)
```
idavis, line 73: @swernli Probably worth documenting why the private lookup paths don't automatically work as this shouldn't be needed ordinarily. We see them loaded and then loaded again. There is a lot of magic knowledge here.

src/QsCompiler/LlvmBindings/Extensions/MarshaledStringArray.cs
```diff
@@ -0,0 +1,57 @@
...
+                this.Values = new MarshaledString[this.Count];
+
+                for (int i = 0; i < this.Count; i++)
+                {
+                    this.Values[i] = new MarshaledString(inputs[i].AsSpan());
+                }
+            }
+        }
+
+        public int Count { get; private set; }
```
DmitryVasilevsky, line 28: Count seems to always reflect the number of elements in Values. Do you need a separate field?

src/QsCompiler/TestTargets/Simulation/Target/Simulation.csproj
```diff
@@ -18,29 +18,13 @@
...
-      <SimulatorRuntime Condition="$([MSBuild]::IsOsPlatform('OSX'))">$(PkgMicrosoft_Quantum_Simulators)/runtimes/osx-x64/native/libMicrosoft.Quantum.Simulator.Runtime.dylib</SimulatorRuntime>
-      <SimulatorRuntime Condition="$([MSBuild]::IsOsPlatform('Windows'))">$(PkgMicrosoft_Quantum_Simulators)/runtimes/win-x64/native/Microsoft.Quantum.Simulator.Runtime.dll</SimulatorRuntime>
-      <SimulatorRuntime Condition="$([MSBuild]::IsOsPlatform('Linux'))">$(PkgMicrosoft_Quantum_Simulators)/runtimes/linux-x64/native/libMicrosoft.Quantum.Simulator.Runtime.so</SimulatorRuntime>
-    </PropertyGroup>
-    <ItemGroup>
-      <_QirRuntimeLibFiles Include="$(QirRuntimeLibs)/**/*.*" Exclude="$(QirRuntimeLibs)/**/*.exe" />
-    </ItemGroup>
-    <Copy SourceFiles="$(SimulatorRuntime)" DestinationFolder="$(OutputPath)" SkipUnchangedFiles="true" />
-    <Copy SourceFiles="@(_QirRuntimeLibFiles)" DestinationFolder="$(OutputPath)\%(RecursiveDir)" SkipUnchangedFiles="true" />
+  <Target Name="PrepareReferenceTests" Condition="'$(DesignTimeBuild)' != 'true' And $([MSBuild]::IsOsPlatform('OSX'))" BeforeTargets="CoreCompile">
```
idavis, line 25: Do we still need these conditionals?

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestDeepNesting.qs
```diff
@@ -0,0 +1,26 @@
...
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        }
+        return 0;
+    }
+}
```
idavis, line 26: nit
new line.

src/QsCompiler/Tests.Compiler/Tests.Compiler.fsproj
```diff
@@ -598,19 +599,8 @@
...
-      <SimulatorRuntime Condition="$([MSBuild]::IsOsPlatform('Windows'))">$(PkgMicrosoft_Quantum_Simulators)/runtimes/win-x64/native/Microsoft.Quantum.Simulator.Runtime.dll</SimulatorRuntime>
-      <SimulatorRuntime Condition="$([MSBuild]::IsOsPlatform('Linux'))">$(PkgMicrosoft_Quantum_Simulators)/runtimes/linux-x64/native/libMicrosoft.Quantum.Simulator.Runtime.so</SimulatorRuntime>
-    </PropertyGroup>
-    <ItemGroup>
-      <_QirRuntimeLibFiles Include="$(QirRuntimeLibs)/**/*.*" Exclude="$(QirRuntimeLibs)/**/*.exe" />
-    </ItemGroup>
-    <Copy SourceFiles="$(SimulatorRuntime)" DestinationFolder="$(OutputPath)" SkipUnchangedFiles="true" />
-    <Copy SourceFiles="@(_QirRuntimeLibFiles)" DestinationFolder="$(OutputPath)\%(RecursiveDir)" SkipUnchangedFiles="true" />
+    <Copy Condition="$([MSBuild]::IsOsPlatform('OSX'))" SourceFiles="$(PkgMicrosoft_Quantum_Simulators)/runtimes/osx-x64/native/libMicrosoft.Quantum.Simulator.Runtime.dylib" DestinationFolder="$(OutputPath)" SkipUnchangedFiles="true" />
```
idavis, line 602: Do we need arch specific here?

src/QsCompiler/LlvmBindings/Interop/LLVMExecutionEngineRef.cs
```diff
@@ -0,0 +1,162 @@
...
+
+        public static bool operator ==(LLVMExecutionEngineRef left, LLVMExecutionEngineRef right) => left.Handle == right.Handle;
+
+        public static bool operator !=(LLVMExecutionEngineRef left, LLVMExecutionEngineRef right) => !(left == right);
+
+        public void AddGlobalMapping(LLVMValueRef global, IntPtr addr) => LLVM.AddGlobalMapping(this, global, (void*)addr);
+
+        public void AddModule(LLVMModuleRef m) => LLVM.AddModule(this, m);
+
+        public void Dispose()
```
DmitryVasilevsky, line 33: Keeping unmanaged handles may require a more careful disposing and destructor. https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose

src/QsCompiler/LlvmBindings/Extensions/LLVMMCJITCompilerOptionsExtensions.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) .NET Foundation and Contributors. All Rights Reserved. Licensed under the MIT License (MIT). See License.md in the repository root for more information.
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace LlvmBindings.Interop
```
cesarzc, line 6: If the package is named `Microsoft.Quantum.LlvmBindings.Native`, why isn't it the root namespace instead of just `LlvmBindings`? I believe that is the pattern for most of the packages that we build and distribute.

src/QsCompiler/LlvmBindings/Extensions/MarshaledStringArray.cs
swernli: Perhaps not. This code was copied almost verbatim from LLVMSharp (I only fixed warnings due to our strictly repo settings). The original is at https://github.com/dotnet/LLVMSharp/blob/ad498be4815dc4699981ba61c9ade30a740d5904/sources/LLVMSharp.Interop/Internals/MarshaledStringArray.cs. We could do more general cleanup across these files, but for now my focus has been on copying functionality rather than on doing any refactoring.

src/QsCompiler/LlvmBindings/Extensions/MarshaledStringArray.cs
swernli: What is NRE?

examples/QIR/Development/Development.csproj
```diff
@@ -8,18 +8,7 @@
...
-    <PackageReference Condition="$([MSBuild]::IsOsPlatform('OSX'))"
-                      Include="libLLVM.runtime.osx-x64" Version="11.0.0"
-                      PrivateAssets="All" GeneratePathProperty="true" />
-    <PackageReference Condition="$([MSBuild]::IsOsPlatform('Linux')) And '$(UbuntuVersion)' == '18.04'"
-                      Include="libLLVM.runtime.ubuntu.18.04-x64" Version="11.0.0"
-                      PrivateAssets="All" GeneratePathProperty="true" />
-    <PackageReference Condition="$([MSBuild]::IsOsPlatform('Linux')) And '$(UbuntuVersion)' != '18.04'"
-                      Include="libLLVM.runtime.ubuntu.20.04-x64" Version="11.0.0"
-                      PrivateAssets="All" GeneratePathProperty="true" />
+    <PackageReference Include="Microsoft.Quantum.LlvmBindings.Native" Version="13.0.0-CI-20220129-005156" PrivateAssets="All" GeneratePathProperty="true"/>
```
cesarzc, line 11: Just curious, why does the `Microsoft.Quantum.LlvmBindings.Native` package has a `Native` suffix?

examples/QIR/Development/Development.csproj
swernli: It wouldn't compile on Mac without it. That matched the guidance I found online on how to resolve that a produce a more compatible binary, but the flag is not accepted on Windows.

llvm-build/llvm.cmake
swernli: Good idea, I'll add that!

src/QsCompiler/Tests.Compiler/Tests.Compiler.fsproj
cesarzc: Why isn't this copy is not required for the other platforms?

src/QsCompiler/CommandLineTool/CommandLineTool.csproj
swernli: We have an alternate mechanism that works on Windows using the `LlvmLibs` property of the compiler:
https://github.com/microsoft/qsharp-compiler/blob/738f2e4178dd4e0975d654dd7a386fdd6290cdab/src/QsCompiler/CommandLineTool/Options.cs#L157-L172
However, that mechanism doesn't work on Linux or Mac (see this comment in a related issue: https://github.com/dotnet/runtime/issues/8516#issuecomment-315151041). But importantly, while this mechanism of listing the dlls as dependencies would work across all platforms, it introduces a relationship between the two packages that causes the nuget to try and download the LlvmBindings.Native package when loading the cmdline tool, which breaks on platforms that don't have our qdk-alpha feed set up. I stumbled across this in the build pipeline, and had to rely on the fact that skipping the dependency on Windows matches the fact that we only package the nupkg on Windows.

To put it another way, this is purely a dev/test dependency and only required on Linux and Mac, and should not be part of the package we publish.

src/QsCompiler/TestTargets/Simulation/Target/QirExecution.cs
swernli: Unfortunately, yeah. I'll add some explanatory comments.

src/QsCompiler/TestTargets/Simulation/Target/Simulation.csproj
swernli: Yes, because these libraries only need to be copied on Mac, not on any other platform (took me a while to figure that out). It has to do with how the Qir.Runtime tries to load the simulator by name here: https://github.com/microsoft/qsharp-runtime/blob/8c2a77cfd1c83e2aa2e488d9be0aef45e4505170/src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp#L52-L75. Because that is native code, it does not benefit from the extra lookup logic that .NET uses. And because it goes by name only and doesn't use any relative paths, the binaries need to be in a place that the OS was going to look already. On Windows and Linux, our preloading of the binary in .NET is sufficient to cause the later native code under the pinvoke call to succeed, but on Mac it will try to find the binary again and fail if it's not in folder next to the .NET dll. This is only for test execution, not for any production code.

src/QsCompiler/Tests.Compiler/Tests.Compiler.fsproj
swernli: See this comment: https://github.com/microsoft/qsharp-compiler/pull/1304#discussion_r796186874. It comes down to the same reason, which is Mac specific behavior. There is a chance we could fix this with some `rpath` trickery like what we did for libomp, but that requires changes to the CMakeLists.txt in the runtime repo and a full build before it could propagate here.

src/QsCompiler/LlvmBindings/Extensions/LLVMMCJITCompilerOptionsExtensions.cs
swernli: The `LlvmBindings.Interop` is made up of generated code that comes from the LLVM pipeline and code that is manually authored here. Both that and the `LlvmBindings` namespace are considered an implementation detail and are not distributed externally. The `Microsoft.Quantum.LlvmBindings.Native` package doesn't even have any compiled .NET binaries; it just has native binaries for each platform and generated C#.

src/QsCompiler/CommandLineTool/CommandLineTool.csproj
swernli: For reference, this is the error that happens if those references are kept in for Windows where the package is built (from [here](https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=37744&view=logs&j=663ee715-dcc5-5b8a-8432-471131753a80&t=f52eef0a-2744-54fe-275f-cf63671504d5&l=398):
```
EXEC : error :  [D:\a\1\s\submodules\qsharp-runtime\src\Simulation\QSharpFoundation\Microsoft.Quantum.QSharp.Foundation.csproj]
    An assembly specified in the application dependencies manifest (qsc.deps.json) was not found:
      package: 'Microsoft.Quantum.LlvmBindings.Native', version: '13.0.0-CI-20220129-005156'
      path: 'runtimes/win-x64/native/libLLVM.dll'
```
That error shows up even though the .NET binaries never refer to that native dependency, because the dotnet pack logic thinks they are required.

cesarzc (approved):

src/QsCompiler/LlvmBindings/Extensions/MarshaledStringArray.cs
idavis: @swernli _N_ull_R_eference_E_xception

swernli: Ah, that makes sense. I can add a protection there.

src/QsCompiler/LlvmBindings/Interop/LLVMExecutionEngineRef.cs
swernli: Filed an issue for follow up at https://github.com/microsoft/qsharp-compiler/issues/1324.

PR #1307, opened 2022-01-21
Lambda Code Writer Support

ScottCarda-MS:
Adds support for processing lambdas to the Code Writer transformation.

### DIFF
--- a/src/QsCompiler/TextProcessor/QsKeywords.fs
+++ b/src/QsCompiler/TextProcessor/QsKeywords.fs
@@ -304,6 +304,9 @@ let qsNEGop = QsOperator.New("-", 45, false)
 let qsSetUnion = QsOperator.New("+", 10, true)
 let qsSetIntersection = QsOperator.New("*", 20, true)
 
+let qsLambdaOp = QsOperator.New("=>", 0, false);
+let qsLambdaFunc = QsOperator.New("->", 0, false);
+
 // As far as the precedence rules of Q# go,
 // there are operators (the things above, processed by an operator precedence parser),
 // modifiers (functors and unwrap, processed manually as part of certain expressions),
--- a/src/QsCompiler/TextProcessor/QsKeywords.fs
+++ b/src/QsCompiler/TextProcessor/QsKeywords.fs
@@ -304,8 +304,8 @@ let qsNEGop = QsOperator.New("-", 45, false)
 let qsSetUnion = QsOperator.New("+", 10, true)
 let qsSetIntersection = QsOperator.New("*", 20, true)
 
-let qsLambdaOp = QsOperator.New("=>", 0, false);
-let qsLambdaFunc = QsOperator.New("->", 0, false);
+let qsLambdaOp = QsOperator.New("=>", 0, false)
+let qsLambdaFunc = QsOperator.New("->", 0, false)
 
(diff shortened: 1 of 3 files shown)


src/QsCompiler/Transformations/QsharpCodeOutput.cs
```diff
@@ -528,6 +528,37 @@ private static string ReplaceInterpolatedArgs(string text, Func<int, string> rep
                 return InterpolationArg.Replace(text, ReplaceMatch);
             }
 
+            private static string QsSymbolToString(QsSymbol symbol)
+            {
+                if (symbol.Symbol is QsSymbolKind<QsSymbol>.Symbol sym)
```
ScottCarda-MS, line 533: I believe I need to add a case here for missing lambda params "`_`".

ScottCarda-MS: done

src/QsCompiler/Transformations/QsharpCodeOutput.cs
```diff
@@ -528,6 +528,49 @@ private static string ReplaceInterpolatedArgs(string text, Func<int, string> rep
...
+                        return QsSymbolToString(tup.Item.First());
+                    }
+                    else
+                    {
+                        return $"({string.Join(", ", tup.Item.Select(QsSymbolToString))})";
+                    }
+                }
+                else if (symbol.Symbol.IsOmittedSymbols)
+                {
+                    return "__omittedSymbol__";
```
bettinaheim, line 558: Omitted symbols are generally three dots, so you can replace that with "..."

src/QsCompiler/TextProcessor/QsKeywords.fs
```diff
@@ -304,6 +304,9 @@ let qsNEGop = QsOperator.New("-", 45, false)
 let qsSetUnion = QsOperator.New("+", 10, true)
 let qsSetIntersection = QsOperator.New("*", 20, true)
 
+let qsLambdaOp = QsOperator.New("=>", 0, false)
```
bettinaheim, line 307: Seeing that this has not yet been in it so far, I need to double check the parser, and whether it hardcodes the arrows. The precedence here needs to be examined carefully. Could you please add to the task list for the lambda issue to clarify the (pseudo-)precedence of the lambda operator, and remove it from here until we do that (and hard code it in the code write for now) - I can take a look as part of the PR around the data structure that I have in work.

bettinaheim (approved):
I'll approve the PR to facilitate testing, but I also expect to largely revise this particular change along with a change to use something else than QsSymbol to represent the argument tuple. Please see also my other comment on this PR.

src/QsCompiler/TextProcessor/QsKeywords.fs
ScottCarda-MS: Okay, I've commented the QsKeywords.fs ones out and added a ToDo. I've added local hardcoded the ones in code writer with another ToDo.

src/QsCompiler/Transformations/QsharpCodeOutput.cs
ScottCarda-MS: updated, thanks

PR #1316, opened 2022-01-27
Infer no characteristics for lambdas with non-unit return type

samarsha:


src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
```diff
@@ -396,19 +396,23 @@ let private characteristicsSet info =
...
-    let onCall =
-        function
+let private lambdaCharacteristics (inference: InferenceContext) (body: TypedExpression) =
+    let mutable characteristics =
+        if inference.Resolve(body.ResolvedType).Resolution = UnitType then
+            Set.ofList [ Adjointable; Controllable ]
+        else
+            Set.empty
+
+    let onCall callableType =
```
ScottCarda-MS, line 406: This function is a bit difficult to understand without any comments. The operation match case took me a bit, but I think I understand that that is used to collect up all the characteristic sets from all operation expressions in the lambda body and get the minimal characteristic set from among them. Is that right? Either way, could you please add a comment explaining this for readability?

src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
```diff
@@ -396,19 +396,23 @@ let private characteristicsSet info =
...
+        match inference.Resolve(callableType).Resolution with
         | QsTypeKind.Operation (_, info) -> characteristics <- characteristicsSet info |> Set.intersect characteristics
         | TypeParameter _ -> characteristics <- Set.empty
```
ScottCarda-MS, line 409: I've not been able to figure this match case out yet. Why do we need special handling for type parameters?

src/QsCompiler/Tests.Compiler/TestCases/TypeChecking.qs
```diff
@@ -843,6 +843,18 @@ namespace Microsoft.Quantum.Testing.TypeChecking {
         return () -> 0;
     }
```
ScottCarda-MS, line 845: Is there a test existing that has multiple operation sub-expressions in the lambda with differing characteristics to test that the lambda's characteristics are the intersection of all the operation sub-expressions?

ScottCarda-MS (approved):

src/QsCompiler/Tests.Compiler/TypeCheckingTests.fs
```diff
@@ -88,6 +99,47 @@ module TypeCheckingTests =
         expect "LambdaInvalid6" [ Error ErrorCode.LocalVariableAlreadyExists ]
```
ScottCarda-MS, line 99: Does the first line of ``supports lambda expressions`` need to be update now that there are more tests?

src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
samarsha: I added a few comments. Do they help?

src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
samarsha: I added a comment about this:

```fsharp
// When a callable type can't be resolved to an operation type based on the current knowledge of the
// inference context, pessimistically assume that it is an operation that supports no characteristics. This
// limitation exists by design to make characteristics inference easier.
```

Does it make sense?

src/QsCompiler/Tests.Compiler/TypeCheckingTests.fs
samarsha: Good idea, I updated the number.

src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
ScottCarda-MS: Yes, thanks!

src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
ScottCarda-MS: Okay so the type parameter case is if the callable is of some type parameter 'T that is a callable type? In that case, we don't have enough information about the 'T to know what characteristics it supports, so we assume it doesn't support any to be safe. Is that right? Is there some syntax in Q# that let's us know that 'T must be a callable type, and if so is there an extension of that syntax that can let us enforce that the type has certain characteristics?

samarsha: A declared type parameter can never be called, so that is actually not the case we need to prevent. The problematic case is something like this:

```qsharp
use q = Qubit();
let lambda = op => op(q);
lambda(X);
```

At the point where the type of `op => op(q)` is being inferred, `op` is some auto-generated type variable, call it `'a`. We have to choose some characteristics for `lambda` but we don't know what `'a` is without looking ahead to the next line, `lambda(X)`. For simplicity we just assume `'a` is an operation with no characteristics, even though that is not the only way to type this program.

src/QsCompiler/Tests.Compiler/TestCases/TypeChecking.qs
samarsha: No, I don't think there was test coverage of that, thanks for pointing that out. I added a few tests.

ScottCarda-MS: Looks good, thanks!

PR #1317, opened 2022-01-27
Fix crash when lambda uses non-interpolated string or empty array

samarsha:


src/QsCompiler/SyntaxProcessor/ContextVerification.fs
```diff
@@ -365,7 +365,7 @@ let rec internal freeVariables e =
     | CONDITIONAL (e1, e2, e3) -> freeVariables e1 |> merge (freeVariables e2) |> merge (freeVariables e3)
     | ValueTuple es
     | StringLiteral (_, es)
-    | ValueArray es -> es |> Seq.map freeVariables |> Seq.reduce merge
+    | ValueArray es -> es |> Seq.map freeVariables |> Seq.fold merge Map.empty
```
ScottCarda-MS, line 368: I can see how this fixes the case of an empty array, but I don't understand how this addresses the string case. Is there more to be done for this PR?

samarsha: No, the PR is complete. This match arm handles all three cases: `ValueTuple es | StringLiteral (_, es) |  ValueArray es`.

ScottCarda-MS: Oh I missed that, thanks. Was the original issue with strings something to do with empty strings `""`?

ScottCarda-MS (approved):

src/QsCompiler/SyntaxProcessor/ContextVerification.fs
samarsha: For `StringLiteral`, the second item `es` is the list of interpolated expressions, so the issue happens with both empty and non-empty strings if they are not interpolated.

ScottCarda-MS: Ah okay, it all makes sense now, thanks!

PR #1322, opened 2022-02-01
Update version forcing logic for MSBuildLocator to use .NET 6

ricardo-espinoza:


cgranade (approved):
LGTM, but it may be worth double-checking with someone closer to the codebase.

ScottCarda-MS (approved):
Thanks Ricardo!

PR #1332, opened 2022-02-03
Syntax tree transformation improvements

bettinaheim:
This PR contains the second batch of changes to improve the consistency of syntax tree transformations. The first half (included in this PR) can be found here: https://github.com/microsoft/qsharp-compiler/pull/1290 These changes are incorporated into this PR.  Additionally, this PR changes the data structures used to represent lambda expressions, and adds the common handles for dealing with argument tuples.


bettinaheim:
The repo build is failing due to an unrelated security alert.

src/QsCompiler/DataStructures/SyntaxTokens.fs
```diff
@@ -130,33 +137,61 @@ type LambdaKind =
...
 
 /// A lambda expression.
-type 'expr Lambda =
+type Lambda<'Expr, 'Type> =
     private
         {
             kind: LambdaKind
-            param: QsSymbol
-            body: 'expr
+            argTuple: QsTuple<LocalVariableDeclaration<QsLocalSymbol, 'Type>>
```
ScottCarda-MS, line 144: nit: can we you "parameter" language for definitions and "argument" language for calls? This could be called `paramTuple`.

src/QsCompiler/DataStructures/SymbolInformation.fs
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+    member this.WithType t =
+        {
+            VariableName = this.VariableName
+            Type = t
+            InferredInformation = this.InferredInformation
+            Position = this.Position
+            Range = this.Range
+        }
```
ScottCarda-MS, line 61:
```suggestion
    member this.WithType t = { this with Type = t }
```
same as above

src/QsCompiler/CompilationManager/EditorSupport/SyntaxUtils.cs
```diff
@@ -174,75 +173,35 @@ TypedExpression CondBlockExpression(Tuple<TypedExpression, QsPositionedBlock> b)
...
         }
 
-        private static IEnumerable<LocalVariableDeclaration<string>> DeclarationsInExpressionByPosition(
+        private static IEnumerable<LocalVariableDeclaration<string, ResolvedType>> DeclarationsInExpressionByPosition(
             TypedExpression expression, Position position)
         {
             return expression.ExtractAll(Declarations);
 
-            IEnumerable<LocalVariableDeclaration<string>> Declarations(TypedExpression e)
+            IEnumerable<LocalVariableDeclaration<string, ResolvedType>> Declarations(TypedExpression e)
```
ScottCarda-MS, line 181: nit: since this is only used here, I suggest using a lambda instead. GitHub won't let me add a code suggestion here, but something like
```
return expression.ExtractAll(e =>
{
    if (!e.Range.Any(r => r.Contains(position)))
    ...
});
```

src/QsCompiler/DataStructures/SymbolInformation.fs
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    member this.WithName name =
+        {
+            VariableName = name
+            Type = this.Type
+            InferredInformation = this.InferredInformation
+            Position = this.Position
+            Range = this.Range
+        }
```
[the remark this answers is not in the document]
bettinaheim: This member allows to change the type (parameterized by VariableName type and the Type type), whereas your suggested edit would not allow to do that.

src/QsCompiler/DataStructures/SymbolInformation.fs
bettinaheim: same, too. :)

src/QsCompiler/CompilationManager/EditorSupport/SyntaxUtils.cs
bettinaheim: done

src/QsCompiler/DataStructures/SymbolInformation.fs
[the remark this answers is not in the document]
ScottCarda-MS: Ah okay, too bad

src/QsCompiler/DataStructures/SyntaxTokens.fs
ScottCarda-MS: Thanks! As an aside, would we be open to having a wider change like that, changing to using "ParameterTuple" throughout the syntax tree, in a future PR? I ask because that seems mildly disruptive for 3rd party devs, but we are already being disruptive in that same way in this PR, so I thought that this might be a good place to do that, to avoid churn. What's your recommendation for these kinds of changes? Should we try to lump as many breaking changes together when we need to make a breaking change, or should we have more frequent, but less impactful breaking changes?

src/QsCompiler/Core/Transformations/SyntaxTreeTransformation.fs
```diff
@@ -73,6 +58,23 @@ type SyntaxTreeTransformation<'T> private (state: 'T, options: TransformationOpt
 
     new(state: 'T) = new SyntaxTreeTransformation<'T>(state, TransformationOptions.Default)
 
+    // These overrides are here since it is impractical to have them live in the definition of CommonTransformationItems.
+
+    override this.OnArgumentTuple argTuple =
+        this.Namespaces.OnArgumentTuple argTuple
+
+    // These overrides are only here to preserve the functionality of the now deprecated methods.
```
ScottCarda-MS, line 66: Would this be easier to track with a ToDo here?

src/QsCompiler/Optimizations/Utils/VariableRenaming.fs
```diff
@@ -91,6 +97,8 @@ and private VariableRenamingNamespaces(parent: VariableRenaming) =
         | QsTupleItem { VariableName = InvalidName } -> ()
         | QsTuple items -> Seq.iter processArgTuple items
 
+    override __.OnArgumentTuple argTuple = argTuple
```
ScottCarda-MS, line 100: Just double checking: This is to say we _don't_ want to process elements within an argument tuple?

src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
```diff
@@ -362,7 +362,7 @@ let rec internal verifyBinding (inference: InferenceContext) tryBuildDeclaration
     | Symbol name ->
         match tryBuildDeclaration (name, symbol.RangeOrDefault) rhsType with
         | Some declaration, diagnostics -> VariableName name, [| declaration |], diagnostics
-        | None, diagnostics -> VariableName name, [||], diagnostics
+        | None, diagnostics -> InvalidItem, [||], diagnostics
```
ScottCarda-MS, line 365: Just double-checking that this line is correct. We want `InvalidItem` here instead of `VariableName name`?

src/QsCompiler/Transformations/LiftLambdas.cs
```diff
@@ -116,97 +117,58 @@ public override TypedExpression OnTypedExpression(TypedExpression ex)
                     }
                 }
 
-                private ParameterTuple MakeLambdaParams(ResolvedType expressionType, QsSymbol paramNames)
+                private ParameterTuple MakeLambdaParams(ParameterTuple paramNames)
```
ScottCarda-MS, line 120: The original purpose of this function was convert from the untyped parameters to a typed ParameterTuple. Since that is redundant now, the function's purpose has changed a bit. I suggest to rename this something like `SanitizeLambdaParams` or something similar. The nested functions can be named very similarly as well, maybe just `_sanitizeLambdaParams` or `SanitizeLambdaParamsRec`.

src/QsCompiler/Transformations/LiftLambdas.cs
```diff
@@ -116,97 +117,58 @@ public override TypedExpression OnTypedExpression(TypedExpression ex)
...
-                                QsLocalSymbol.NewValidName($"__missingLambdaParam_{missingSymbolCount}__"),
-                                paramType,
-                                new InferredExpressionInformation(false, false),
-                                QsNullable<Position>.Null,
-                                paramName.Range.IsNull
-                                    ? DataTypes.Range.Zero
-                                    : paramName.Range.Item);
-                            missingSymbolCount++;
-                            return ParameterTuple.NewQsTupleItem(localVar);
+                            return param.Item.VariableName.IsInvalidName
```
ScottCarda-MS, line 128: How equivalent is an invalid variable name with the missing symbol case? Will this always catch the missing symbols `_`? Will it catch more cases than it did before?

ScottCarda-MS (commented):
I still have a bit more looking into with the changes with the base transformations.

src/QsCompiler/Core/Transformations/NamespaceTransformation.fs
```diff
@@ -45,33 +45,30 @@ type NamespaceTransformationBase internal (options: TransformationOptions, _inte
...
     abstract OnSource : Source -> Source
-
-    default this.OnSource source =
-        let file = Source.assemblyOrCodeFile source |> this.OnSourceFile
-
-        if file.EndsWith ".qs" then
-            { source with CodeFile = file }
-        else
-            { source with AssemblyFile = Value file }
+    default this.OnSource source = source
```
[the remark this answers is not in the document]
bettinaheim: Source itself already makes that distinction. The condition was just to properly recreate it after the now removed and previously deprecated OnSourceFile performed its transformation.

src/QsCompiler/Core/Transformations/SyntaxTreeTransformation.fs
```diff
@@ -73,6 +58,23 @@ type SyntaxTreeTransformation<'T> private (state: 'T, options: TransformationOpt
...
+    // They can be removed and the deprecation warning for this file can be reenabled once the deprecated methods are removed.
+
```
cesarzc, line 68: It would be great to add a `TODO` here so we don't forget to remove these overrides when the deprecated methods are removed.

src/QsCompiler/Core/Transformations/SyntaxTreeTransformation.fs
```diff
@@ -3,52 +3,37 @@
 
 namespace Microsoft.Quantum.QsCompiler.Transformations.Core
 
+#nowarn "44" // disable deprecation warnings due to the overrides pointing to deprecated methods
```
cesarzc, line 6: It would be great to add a release `TODO` here to: `TODO: RELEASE 2022-09`.

src/QsCompiler/Core/Transformations/SyntaxTreeTransformation.fs
cesarzc: It should probably be a release `TODO`:
`TODO: RELEASE 2022-09`

src/QsCompiler/Core/Transformations/NamespaceTransformation.fs
```diff
@@ -84,41 +81,49 @@ type NamespaceTransformationBase internal (options: TransformationOptions, _inte
...
 
+    // TODO: RELEASE 2022-09: Remove member.
+    [<Obsolete "Use SyntaxTreeTransformation.OnArgumentTuple instead.">]
     abstract OnArgumentTuple : QsArgumentTuple -> QsArgumentTuple
 
-    default this.OnArgumentTuple arg =
+    // TODO: RELEASE 2022-09: Make this an internal member. Keep the following comment:
+    // do not expose this - this handle is exposed as a virtual member in the SyntaxTreeTransformation itself
+    [<Obsolete "Use SyntaxTreeTransformation.OnArgumentTuple instead.">]
+    override this.OnArgumentTuple arg =
```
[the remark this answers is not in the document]
bettinaheim: How we need to go about removing obsolete functions is documented in the form of comments in the code. 
The way we go about this indeed is required to properly deprecate the old handle;
If I had the implementation in the new handle and the old handle pointed to it, then inheriting classes that override the old handle would break - thus we would not have proper deprecation. The pattern that we instead follow is that the implementation remains in the old handle until it is removed, the new handle points to the old handle, and all usages in our transformation infrastructure use the new handle. That way both overriding the old hand and overriding the new handle works properly.

src/QsCompiler/Core/Transformations/NamespaceTransformation.fs
```diff
@@ -84,41 +81,49 @@ type NamespaceTransformationBase internal (options: TransformationOptions, _inte
...
+            let t = this.Statements.Expressions.Types.OnType declInfo.Type
+            let info = this.Statements.Expressions.OnExpressionInformation declInfo.InferredInformation
+            let newDecl = LocalVariableDeclaration.New info.IsMutable (loc, name, t, info.HasLocalQuantumDependency)
+            QsTupleItem |> Node.BuildOr original newDecl
 
     abstract OnSignature : ResolvedSignature -> ResolvedSignature
 
     default this.OnSignature(s: ResolvedSignature) =
-        let typeParams = s.TypeParameters
+        let typeParams = s.TypeParameters // if this had a range is should be handled by the corresponding Common nodes
```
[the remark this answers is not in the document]
bettinaheim: There is not really a reason at this time to add a range, yet I think the comment is good to have. So I don't think it is a todo so much as a comment.

src/QsCompiler/Core/Transformations/TypeTransformation.fs
```diff
@@ -17,22 +17,46 @@ open Microsoft.Quantum.QsCompiler.Transformations.Core.Utils
...
+            let commonItems = new CommonTransformationNodes()
+            this.CommonTransformationItemsHandle <- fun _ -> commonItems
+
+    new() = TypeTransformationBase TransformationOptions.Default
 
     // supplementary type information
 
     // TODO: RELEASE 2021-10: Remove obsolete method.
-    [<Obsolete "Use OnRangeInformation(TypeRange) instead.">]
+    [<Obsolete "Use SyntaxTreeTransformation.OnTypeRange(TypeRange) instead.">]
```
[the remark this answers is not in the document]
bettinaheim: Removing these is going to be a more involved change; the reason is that when we remove these, we also need to remove the corresponding (deprecated) handle in the AST node. That would make that record have the same items as another record type, which would break a lot of the F# code. We should properly mitigate that so that we indeed don't break anything when removing these handles. This needs to be a separate PR.

src/QsCompiler/Optimizations/Utils/VariableRenaming.fs
bettinaheim: We don't want to process argument tuples, yes.

src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
bettinaheim: None here means that the variable could not be added to the symbol tracker; i.e. the variable already exists and its redeclaration is invalid. In that case we generate a diagnostic. Not replacing the AST node here (keeping it as VariableName) makes the compiler throw when you e.g. hover over invalid code (like e.g. `(a, a) -> a`). This was an oversight before, and I added a unit test.

src/QsCompiler/Transformations/LiftLambdas.cs
bettinaheim: We currently don't support underscores as parameter names (i.e. missing parameter names) in global callable declarations. This is something that may be nice to support in the future, but felt out of scope for this change. At the same time, it is nice to support it in particular for lambdas, since there the argument also doesn't require a type annotation. 
To align with how argument tuples work in general and still permit _ for parameters in lambdas, I needed to represent missing symbols as invalid variables. So the answer to your question is that missing symbols will indeed always be invalid variables, and - unless the compilation generated errors before, an invalid symbol here is indeed a missing symbol (`_`) in the source code. However, for invalid code such as e.g. `(a, a) -> a`), the second `a` here would also be represented as invalid variable name and would be replaced with an autogenerated name by this transformation. I think that is fine.

src/QsCompiler/Core/Transformations/SyntaxTreeTransformation.fs
bettinaheim: We cannot really forget to remove them, since when you remove the deprecated methods this code no longer compiles (it calls into the deprecated methods).

src/QsCompiler/SyntaxProcessor/TypeInference/InferenceContext.fs
```diff
@@ -323,7 +325,20 @@ type InferenceContext(symbolTracker: SymbolTracker) =
         |> Seq.map (fun item -> diagnostic item.Key item.Value)
         |> Seq.toList
 
-    member context.UseStatementPosition position = statementPosition <- position
+    member context.UseStatementPosition position =
```
cesarzc, line 328: It seems like `UseStatementPosition` is not really used anymore, is it kept here just because it is a public member and would break someone trying to use it? If that's the case, should it be deprecated?

src/QsCompiler/SyntaxProcessor/TypeInference/InferenceContext.fs
```diff
@@ -288,6 +288,8 @@ open Inference
 type InferenceContext(symbolTracker: SymbolTracker) =
     let variables = Dictionary()
 
+    let mutable rootNodePos = Null
+    let mutable relativePos = Null
```
cesarzc, line 292: It seems as if with the addition of these two members, `statementPostion` just became a proxy for `rootNodePosition + relativePosition`. If that is the case, should `statementPosition` be removed/deprecated?

src/QsCompiler/Core/Transformations/SyntaxTreeTransformation.fs
ScottCarda-MS: I see you've answered this [here](https://github.com/microsoft/qsharp-compiler/pull/1332#discussion_r799887933).

src/QsCompiler/Core/Transformations/NamespaceTransformation.fs
```diff
@@ -84,41 +81,49 @@ type NamespaceTransformationBase internal (options: TransformationOptions, _inte
...
+    [<Obsolete "Use SyntaxTreeTransformation.OnLocalNameDeclaration or override OnArgumentTuple instead.">]
     abstract OnArgumentName : QsLocalSymbol -> QsLocalSymbol
 
-    default this.OnArgumentName arg =
+    // TODO: RELEASE 2022-09: Remove.
+    [<Obsolete "Use SyntaxTreeTransformation.OnLocalNameDeclaration or override OnArgumentTuple instead.">]
+    override this.OnArgumentName arg =
         match arg with
-        | ValidName name -> ValidName |> Node.BuildOr arg (this.Statements.OnVariableName name)
+        | ValidName name -> this.Statements.Expressions.Common.OnLocalNameDeclaration name |> ValidName
```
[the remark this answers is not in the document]
bettinaheim: I changed it to use abbreviations everywhere (I think). The BuildOr here can be safely omitted; this change won't change the behavior of the transformation.

src/QsCompiler/Core/Transformations/NamespaceTransformation.fs
```diff
@@ -84,41 +81,49 @@ type NamespaceTransformationBase internal (options: TransformationOptions, _inte
...
             QsTuple |> Node.BuildOr original transformed
-        | QsTupleItem item as original ->
-            let loc = item.Position, item.Range
-            let name = this.OnArgumentName item.VariableName
-            let t = this.Statements.Expressions.Types.OnType item.Type
-            let info = this.Statements.Expressions.OnExpressionInformation item.InferredInformation
-
-            QsTupleItem << LocalVariableDeclaration<_>.New info.IsMutable
-            |> Node.BuildOr original (loc, name, t, info.HasLocalQuantumDependency)
+        | QsTupleItem declInfo as original ->
```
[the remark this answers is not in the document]
bettinaheim: changed to item

src/QsCompiler/Core/Transformations/StatementTransformation.fs
```diff
@@ -19,46 +19,54 @@ type StatementKindTransformationBase internal (options: TransformationOptions, _
...
     // subconstructs used within statements
 
     abstract OnSymbolTuple : SymbolTuple -> SymbolTuple
-    default this.OnSymbolTuple syms = syms
+
+    default this.OnSymbolTuple syms =
+        match syms with
+        | VariableNameTuple tuple ->
+            tuple |> Seq.map this.OnSymbolTuple |> ImmutableArray.CreateRange |> VariableNameTuple
+        | VariableName name -> this.Expressions.Common.OnLocalNameDeclaration name |> VariableName
```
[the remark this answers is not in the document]
bettinaheim: done

src/QsCompiler/Core/Transformations/StatementTransformation.fs
```diff
@@ -283,24 +289,33 @@ and StatementTransformationBase internal (options: TransformationOptions, _inter
...
     default this.OnVariableName name = name
 
     abstract OnLocalDeclarations : LocalDeclarations -> LocalDeclarations
 
     default this.OnLocalDeclarations decl =
         let onLocalVariableDeclaration (local: LocalVariableDeclaration<string>) =
-            let loc = local.Position, local.Range
-            let name = this.OnVariableName local.VariableName
+            let loc = this.Common.OnSymbolLocation (local.Position, local.Range)
+            let name = this.Expressions.Common.OnLocalName local.VariableName
```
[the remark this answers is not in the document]
bettinaheim: done

src/QsCompiler/DataStructures/SymbolInformation.fs
```diff
@@ -0,0 +1,78 @@
...
+    /// Returns the inferred expression information used for declared parameters to callables.
+    /// Note that this information is used within the declaration but not for the argument(s) in any given call-like expressions.
+    static member ParameterDeclaration = { IsMutable = false; HasLocalQuantumDependency = false }
+
+type LocalVariableDeclaration<'Name, 'Type> =
+    {
+        /// the name of the declared variable
+        VariableName: 'Name
+        /// the fully resolved type of the declared variable
+        Type: 'Type
```
cesarzc, line 33: It is not clear to me yet why the type of the `Type` member has to be a type-parameter. I see that this type-parameter propagates from the `Lambda` type so maybe the question is, why `Lambda` needs a type-parameter? Why isn't `ResolvedType` enough?

src/QsCompiler/Core/Transformations/NamespaceTransformation.fs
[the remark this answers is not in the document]
ScottCarda-MS: That seems inconsistent with this functions references to `this.Common.OnArgumentTuple`. Should it actually reference itself recursively, as it did before? With the new change, I don't see the recursion happening.

src/QsCompiler/Core/Transformations/SyntaxTreeTransformation.fs
cesarzc: I see multiple of instances of disabling this warning in this PR and since it would be worth adding a `TODO` for all of them since we could forget to re-enable them even after the deprecated methods are deleted.

src/QsCompiler/Transformations/LiftLambdas.cs
bettinaheim: Renamed.

src/QsCompiler/SyntaxProcessor/TypeInference/InferenceContext.fs
bettinaheim: I debated whether to deprecate it, but given that the inference context is part of the public API, it could be used outside the Q# compiler, and then it would likely be more convenient to work with UseStatementPosition only rather than to split that into absolute and relative location. I hence decided against deprecating it, but I could go either way.

src/QsCompiler/SyntaxProcessor/TypeInference/InferenceContext.fs
bettinaheim: statementPosition is a proxy for rootNodePosition + relativePosition if a location has been specified, but could be set and used on its own without ever specifying the location broken down into absolute and relative position (in which case the handle to get the relative location throws). See also my comment above.

src/QsCompiler/DataStructures/SymbolInformation.fs
bettinaheim: We make a type distinction between the type checked data structures and the unchecked data structures. I think this distinction between validated AST nodes vs unvalidated one is useful. At the same time, rather than a full duplication of all data structures, some structures are reused, and in particular the major DUs like ExpressionKind and TypeKind. These DUs are hence type parameterized to reflect when they are type checked and when they are not. The choice to reuse the major DUs is debatable, but there are benefits and downsides and it is the current setup. With the lambda data structure being part of such a dual-purpose DU, the same is done here for the lambda; the type here reflects and is different depending on whether the lambda has been validated or not.

src/QsCompiler/Core/Transformations/NamespaceTransformation.fs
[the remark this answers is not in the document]
bettinaheim: I am not sure I follow the statement about the inconsistency?
Following my explanation above, here is how this looks like concretely for OnArgumentTuple:

old handle: Namespaces.OnArgumentTuple
new handle: SyntaxTree.OnArgumentTuple ("paraphrasing" a bit here, to keep it concise, let me know if I need to be more precise)

The implementation is in the old handle, and (like for other usages of the handle in the transformation framework) it calls into the new handle. The new handle points back to the old handle. So the recursion is Namespaces.OnArgumentTuple -> SyntaxTree.OnArgumentTuple -> Namespaces.OnArgumentTuple -> ...

We inject a node here in order to properly deprecate the old handle; we replace one node Namespaces.OnArgumentTuple with two nodes SyntaxTree.OnArgumentTuple -> Namespaces.OnArgumentTuple, and that goes for all its uses. If the recursion was direct, then the old handle wouldn't be deprecated properly. Does that clarify things?

src/QsCompiler/DataStructures/SymbolInformation.fs
cesarzc: Thank you for the explanation! It makes sense now.

src/QsCompiler/SyntaxProcessor/TypeInference/InferenceContext.fs
cesarzc: Makes sense to not deprecate it for convenience outside the Q# compiler.

src/QsCompiler/Core/Transformations/SyntaxTreeTransformation.fs
bettinaheim: Yes, that is a good idea. I added a TODO: RELEASE 2022-09 tag here, and pointed with a `(*)` to the comment you found/mentioned in your previous comment. That should make things clearer to track.

cesarzc:
@bettinaheim after reviewing the PR, the summary of the major changes is the following:
- Unchecked vs checked lambdas are now type differentiated.
- `CommonTransformationNodes`  was introduced as a way to better organize transformations that happen across multiple kinds of nodes.
    - By introducing this, overrides affect all transformation kinds that make use a specific method.

Am I missing something?

bettinaheim:
> @bettinaheim after reviewing the PR, the summary of the major changes is the following:
> 
> * Unchecked vs checked lambdas are now type differentiated.
> * `CommonTransformationNodes`  was introduced as a way to better organize transformations that happen across multiple kinds of nodes.
>   
>   * By introducing this, overrides affect all transformation kinds that make use a specific method.
> 
> Am I missing something?

Regarding "Unchecked vs checked lambdas are now type differentiated.": 
This was also the case without my changes. The reason for my changes to the lambda data structure specifically were primarity to align how parameter tuples are handled for lambdas with how they work for global declarations. This allows to define a common handler for both in the syntax tree transformation infrastructure. Additionally, this also automatically creates room for potentially allowing to annotate a parameter type in the future (e.g. `(a : Int, b : Int) -> a + b`). 

Regarding CommonTransformationNodes: 
Yes, exactly. This class is largely not publicly visible, such that the common nodes are overridded by adding an override to the SyntaxTreeTransformation itself rather than to a subclass.

This is the summary for the change:

CommonTransformationNodes was introduces as per your comment, and the nodes in that class are overridded by overriding them in the SyntaxTreeTransformation. The common nodes extracted in that way are the following:
- The transformations now use OnLocalNameDeclaration and OnLocalName for all locally declared symbols. The distinction between the declaration of a local name and its use wasn't properly made previously.
- The change also introduces OnItemNameDeclaration and OnItemName to easily modify item named in user defined types. Previously there was no handle that would have allowed to conveniently change e.g. the name of a named item in a user defined type.
- The data structure used for parameter tuples of lambdas and globally declared callables is now the same, allowing for a common handler for dealing with argument tuples. A good future change would be to also align parameter tuples with symbol tuples in variable declarations (symbols declared in let- and mutable-statements as well as in for loops). The PR is big enough as it is, and I decided against it for this PR.
- All methods for handling positional information were moved into the CommonTransformationNodes. This includes a descriptive name for absolute locations vs relative locations, and adding a handler for symbol locations which didn't previously exist.

cesarzc (approved):

src/QsCompiler/Core/Transformations/NamespaceTransformation.fs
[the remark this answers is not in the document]
ScottCarda-MS: I see, that makes sense. Thanks for the explanation!

src/QsCompiler/Core/Transformations/ExpressionTransformation.fs
```diff
@@ -23,38 +25,40 @@ type ExpressionKindTransformationBase internal (options: TransformationOptions,
...
     let Node = if options.Rebuild then Fold else Walk
 
-    member val internal TypeTransformationHandle = missingTransformation "type" with get, set
     member val internal ExpressionTransformationHandle = missingTransformation "expression" with get, set
 
-    member this.Types = this.TypeTransformationHandle()
-    member this.Expressions = this.ExpressionTransformationHandle()
+    member this.Expressions : ExpressionTransformationBase = this.ExpressionTransformationHandle()
+    member this.Types : TypeTransformationBase = this.ExpressionTransformationHandle().Types
+    member this.Common : CommonTransformationNodes = this.ExpressionTransformationHandle().Types.Common
```
ScottCarda-MS, line 32: So there seems to be a subtle difference in how the various transformations relate to each other as a result of this PR, and I've noticed this change in all the transformation base classes, but we can use the ExpressionKindTransformationBase as an example. The PR changes to minimize the number of different transformation types being passed in. Here for example, we used to have to pass in a type transformation and an expression transformation separately, and these would each be assigned to the expression-kind transformation as fields, allowing the transformation to call into them. But now, both the expression and the type transformation fields are set up using the same transformation passed in: the expression transformation. It uses the expression transformation's Types field instead of an independent type transformation. This is what I'm observing with this change.

Does that limit the flexibility of the transformations? Before, it seems like we would have had the ability to assign one type transformation to the expressions, and a completely separate type transformation to the expression kinds, which might actually make sense in certain circumstances. It seems a shame that we are losing that flexibility with this change.

ScottCarda-MS: It seems like the major motivation for this change was to allow each of the base transformation classes to have a Common property for referencing the top-level SyntaxTreeTransformation's Common handles, without the need for us to have to pass in a separate Common transformation into each constructor. This means that all transformations in a "bundle" (if I could use that term to describe a SyntaxTreeTransformation and all of its subtransformations) must have the same implementation for, not only their Common handles, but all their handles, removing any modularity the submodules may have had before. One can no longer handle types directly under expression nodes differently from types under expression-kind nodes, for example. Maybe more critically, one can't handle argument tuples differently depending on if you are dealing with a namespace element or an expression. If a developler would need to do make this distinction, they would need to write one handler capable of handling both cases and being able to distinguish between the two based on contextual information that would need to be save and kept track of throughout the transformation "bundle". This can get quite complicated and messy to maintain.

ScottCarda-MS (approved):
Just had one open concern about the changes to how subtransformations relate to each other and its effect on transformation modularity and flexibility. Outside of that, this looks great!

src/QsCompiler/Core/Transformations/ExpressionTransformation.fs
bettinaheim: Yes, you are right - and great call out! 
Indeed it is a slight change in that now everything has to use the same subtransformation. I believe this is reasonable, since I believe it was rather fragile and error prone before, and the way the different handles where used across transformations wasn't entirely consistent. Also, there seems to be no practical use case anywhere, so I think the simplification is a good change.

PR #1334, opened 2022-02-03
Fix crash when calling operation with invalid characteristics

### LINKED ISSUE #1300: Adjointable Operation Returns Non-Unit Causes Hard Crash Instead of Error Message
**Describe the bug**

If an operation is incorrectly defined to be adjointable while also returning a non-unit type, the compiler gives a proper error message: "Adjoint specializations can only be defined for operations returning Unit." However, if that operation is defined and a call to that operation is used somewhere in the code. The type-checking logic seems to crash.

**To Reproduce**

Try to compile the following code:
```qsharp
namespace Input {
    @EntryPoint()
    operation Main() : Unit {
        Foo();
    }

    operation Foo() : Int is Adj {
        return 0;
    }
}
```

**Expected behavior**

The provided code should not compile, but it should fail gracefully. The compilation should provide the same behavior and error message as when the following code is attempted to be compiled:
```qsharp
namespace Input {
    @EntryPoint()
    operation Main() : Unit {
        //Foo();
    }

    operation Foo() : Int is Adj {
        return 0;
    }
}
```

samarsha:
Closes #1300.


src/QsCompiler/SyntaxProcessor/TypeInference/InferenceContext.fs
```diff
@@ -126,8 +126,10 @@ module private Inference =
     /// True if the characteristics of <paramref name="info1"/> are a subset of the characteristics of
```
ScottCarda-MS, line 126: The function now returns true if either of the arguments are invalid. Could you add that to the description?

ScottCarda-MS: Though it seems a bit strange for this to return true if either one of these has invalid characteristics. Same for the charsEqual function above.

ScottCarda-MS (approved):

ScottCarda-MS (approved):

src/QsCompiler/SyntaxProcessor/TypeInference/InferenceContext.fs
samarsha: This handling of invalid characteristics is consistent with invalid types, which are treated as a wildcard that match with any other type. That's to avoid too many cascading errors happening when you declare something with an invalid type/characteristics (there should already be an error at the source so we don't need to generate more).

Do you think we should handle or represent this differently?

PR #1336, opened 2022-02-03
Improve error message for implicitly ignored value

### LINKED ISSUE #1094: Improve error message for implicitly ignored result
**Describe the improvement**

The Q# compiler gives slightly unhelpful error messages when calling non-`Unit` without explicitly ignoring the result. Whenever a call is made without explicitly ignoring the result, the compiler should suggest to either use `Ignore( ... )` or change the return type of the function to `Unit`. At the moment, it only does the latter.

**To Reproduce**

To reproduce, compile following to programs:

```
namespace Example {
@EntryPoint()
operation Main() : Int
{
QuantumFunction(3);
return QuantumFunction(10);
}

operation QuantumFunction(nQubits : Int) : Int {
use qubits = Qubit[nQubits];

return 0;
}
}
```
does not compile and has following output
```
% dotnet build ConstSizeArray.csproj
Microsoft (R) Build Engine version 16.9.0+57a23d249 for .NET
Copyright (C) Microsoft Corporation. All rights reserved.

Determining projects to restore...
All projects are up-to-date for restore.
Determining projects to restore...
All projects are up-to-date for restore.
ConstSizeArray.qs(8,9): error QS0001: The type Int does not match the type Unit. [ConstSizeArray/ConstSizeArray.csproj]
Actual type: Int
Expected type: Unit

____________________________________________

Q# compilation failed: 1 error, 0 warnings

(...)
```

**Expected behavior**

```
ConstSizeArray.qs(8,9): error QS0001: Return value implicitly ignored. [ConstSizeArray/ConstSizeArray.csproj]
To fix this either user `Ignore( ... )`, use a declaration `let _ = ...` or change the return type of the function to `Unit`.
```

**Screenshots**

**System information**

- Version of the NuGet package or extension for which the problem occurs
- Your operating system and .NET Core version  
- For issues with extensions, the version of the IDE for which the problem occurs 

**Additional context**

samarsha:
Closes #1094.


ScottCarda-MS (approved):
Looks good to me!

Issue #1046, opened 2021-05-27
QIR generation fails for BigInt literals with more than 64 bit


msoeken:
**Describe the bug**

QIR generation fails when the Q# program contains a `BigInt` literal that is too large to fit an `Int`.

**To Reproduce**

```xml
<!-- project.csproj -->
<Project Sdk="Microsoft.Quantum.Sdk/0.17.2105143879">
    <PropertyGroup>
        <OutputType>Exe</OutputType>
        <TargetFramework>netcoreapp3.1</TargetFramework>
        <QirGeneration>true</QirGeneration>
        <CSharpGeneration>false</CSharpGeneration>
    </PropertyGroup>
</Project>
```

```qsharp
// Program.qs
namespace Microsoft.QDK.Test {
    @EntryPoint()
    operation RunProgram() : Unit {
        let N = 1522605027922533360535618378132637429718068114961380688657908494580122963258952897654000350692006139L;
    }
}
```

**Expected behavior**

Program should generate QIR.

**Error message**
```
  System.ArgumentException: Value does not fall within the expected range.
     at Ubiquity.NET.Llvm.Instructions.InstructionBuilder.GetValidatedGEPArgs(ITypeRef type, Value pointer, IEnumerable`1 args)
     at Ubiquity.NET.Llvm.Instructions.InstructionBuilder.GetElementPtr(ITypeRef type, Value pointer, IEnumerable`1 args)
     at Microsoft.Quantum.QsCompiler.QIR.QirExpressionKindTransformation.OnBigIntLiteral(BigInteger b)
     at Microsoft.Quantum.QsCompiler.Transformations.Core.ExpressionKindTransformationBase.OnExpressionKind(QsExpressionKind`3 kind) in D:\a\1\s\submodules\qsharp-compiler\src\QsCompiler\Core\Transformations\ExpressionTransformation.fs:line 404     
     at Microsoft.Quantum.QsCompiler.Transformations.Core.ExpressionTransformationBase.OnTypedExpression(TypedExpression ex) in D:\a\1\s\submodules\qsharp-compiler\src\QsCompiler\Core\Transformations\ExpressionTransformation.fs:line 490
     at Microsoft.Quantum.QsCompiler.QIR.QirExpressionTransformation.OnTypedExpression(TypedExpression ex)
     at Microsoft.Quantum.QsCompiler.QIR.GenerationContext.EvaluateSubexpression(TypedExpression ex)
[...]

PR #1337, opened 2022-02-04
Creating 'global constant' to fix the issue with BigInt

DmitryVasilevsky:
This code change allows large BigInts to be defined literally.
Previously only BigInts small enough to fit into int64 were supported. A code path that was supposed to support larger BigInts was calling GetElementPtr on a non-pointer type (an array type). It was causing a failure. I updated the code to be similar to handling string literals. Now a global internal variable is created to represent a large BigInt. This variable has pointer type so the code succeeds.

NOTE: C++ QIR runtime still doesn't support BigInts.

Fixes https://github.com/microsoft/qsharp-compiler/issues/1046


src/QsCompiler/QirGeneration/Context.cs
```diff
@@ -481,6 +481,17 @@ internal GlobalVariable GetOrCreateStringConstant(string str)
             return constant;
         }
 
+        /// <summary>
+        /// Creates a global constant (data array) that stores the given bytes.
+        /// </summary>
+        internal GlobalVariable CreateArrayConstant(byte[] array) {
```
bettinaheim, line 487: How about naming it CreateGlobalConstantArray, to be a bit more descriptive?

DmitryVasilevsky: Yes, this is a much better name! (I didn't like the previous name, just named it similar to the string function.)

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestBigInts.qs
```diff
@@ -16,5 +16,8 @@ namespace Microsoft.Quantum.Testing.QIR
     @EntryPoint()
     function Main() : Unit {
         let _ = TestBigInts(0L, 0L);
+        let N1 = 12345L;
```
bettinaheim, line 19: It looks like the code generation only checks the generated IR for TestBigInts. Could you hence either add these bigint instantiations to the TestBigInts function instead, or add a file to also check the generated .ll code for the Main function?

DmitryVasilevsky: Added another file to test Main. Changed qirTest to qirMultiTest.

bettinaheim (approved):
It would be good to confirm that it runs, but other than that lgtm.

Issue #1289, opened 2021-12-16
Improve usability of the Q# formatter


tcNickolas:
**Describe the solution you'd like**

* [x] Avoid overwriting files if there are no changes done to them.  
You can see this effect when the tool runs on files in format "UTF-8 with signature": the files get overwritten (in format without a signature) even if the tool doesn't update them otherwise. 
* [x] Offer a summary of the tool run upon its completion.
The tool quits without a summary, unless it exits with an error. Other msbuild-based tools, such as `dotnet build` and `dotnet test`, offer a summary of their run - whether the build succeeded, how many tests passed and failed, etc. It makes sense to say something like "X updates done in Y files" or "Y files processed, X updates done" to match that behavior.

**Describe alternatives you've considered**
I can see the changes done using `git diff`, and I can revert changes to the files if I see that there are no meaningful changes to them manually, but having this functionality in the tool itself would greatly improve its usability and ease of adoption.

samarsha:
I think that not writing unchanged files and showing a summary of number of files changed sound reasonable.

> The tool quits without a summary, unless it exits with an error. Other msbuild-based tools, such as `dotnet build` and `dotnet test`, offer a summary of their run - whether the build succeeded, how many tests passed and failed, etc. It makes sense to say something like "X updates done in Y files" or "Y files processed, X updates done" to match that behavior.

I think it would make sense to compare the summary output of QsFmt with other formatters, rather than build or test commands. For example, `dotnet format` and `fantomas` outputs a line for each file updated (but doesn't tell you how many changes were made in the file). `cargo fmt` has no output at all.

A single file could have hundreds or thousands of individual formatting changes for every syntax tree node that needs whitespace changes. I am not sure how useful counting the number of these updates is compared to using a diff between the original and updated contents instead.

tcNickolas:
I spent a bit more time with the Q# Formatter (https://github.com/microsoft/QuantumLibraries/pull/517), and to add based on that experience:
* Not overwriting unchanged files is very important for the user experience. All Q# libraries have a dependency on Standard, and it had a dozen files that were overwritten with a different encoding every time I ran the tool, that was rather inconvenient.
* Reporting the summary is mostly a convenience feature to distinguish "The tool failed" from "The tool succeeded". I don't think we need to report per-file status, maybe even just the number of files that were updated is sufficient?

bettinaheim:
It looks like we are all in agreement on not overwriting unchanged files, and I believe that is in work. 

[...]

PR #1341, opened 2022-02-08
Formatter Summary

ScottCarda-MS:
Addresses #1289.

Prevents files from being overwritten if there are no changes to the file.
Prints out a summary of the files that are changed.


src/QsFmt/App.Tests/Tests.fs
```diff
@@ -55,6 +55,23 @@ let private CleanResult =
         Error = ""
     }
 
+let private makeOutput verb files =
+    (Seq.length files |> sprintf "%s %i files:" verb)
```
kuzminrobin, line 59: (I don't know F#, I can be wrong)
I have an impression that in the fragment `sprintf "%s %i files:" verb)` the string literal `"%s %i files:"` expects two values, but only one is provided (`verb`).

kuzminrobin: Can also be applicable to the _unchanged_ fragment `(eprintfn "%s, %O" input)` in this PR.

ScottCarda-MS: This is an F# quirk. The last argument to a function is actually the result of the expression found before the `|>`. An equivalent expression might be  `(sprintf "%s %i files:" verb (Seq.length files))` though this is less idiomatic F#, and doesn't work as well with F#'s type inference.

PR #1342, opened 2022-02-08
Telemetry Library: Add support for default opt-in/out and make opted-out as default.

### LINKED ISSUE #1325: Telemetry Library: Make an option for opt-out by default.
The [Telemetry Library](https://github.com/microsoft/qsharp-compiler/tree/main/src/Telemetry) needs some work to make an option for opt-out by default.

vxfield:
Fixes #1325

- Add support for default opt-in/out and make Opted-Out as default.
- Set the QsFmt telemetry to be Opted-In by default (users can always opt-out using `QSFMT_TELEMETRY_OPT_OUT=1`)


kuzminrobin:
Just in case I would like to make sure that we still have an option of making the telemetry _Opt-Out_ for our needs (such that the users make actions, e.g. set the env var, to _disable_ the telemetry).
Previously we discussed and agreed that the telemetry needs to be _Opt-In_. But the feedback in the Sparse Simulator Spec tells me that the _Opt-Out_ may be more desirable. See feedback from Viktor Veis ("Most people won't do this") and Bettina ("What exactly is the benefit") in the Sparse Sim Spec.

src/QsFmt/.vscode/launch.json
```diff
@@ -0,0 +1,18 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
```
kuzminrobin, line 5: The syntax highlighting makes me think that it "won't compile".
See [here](https://reqbin.com/req/wtvjp1a3/json-comment-example): "JSON is not intended and does not support comments by design, which means that comments in the form // ... or / * ... * / are not allowed in JSON files. But there is a workaround"

vxfield:
> Just in case I would like to make sure that we still have an option of making the telemetry _Opt-Out_ for our needs (such that the users make actions, e.g. set the env var, to _disable_ the telemetry). Previously we discussed and agreed that the telemetry needs to be _Opt-In_. But the feedback in the Sparse Simulator Spec tells me that the _Opt-Out_ may be more desirable. See feedback from Viktor Veis ("Most people won't do this") and Bettina ("What exactly is the benefit") in the Sparse Sim Spec.

Yes, @kuzminrobin, there is still the option to have Opted-In by default (which the QsFmt will continue to use) and we can make the decision to do the same for the Sparse Simulator.

src/Telemetry/Library/TelemetryManagerConfig.cs
```diff
@@ -39,13 +55,30 @@ public string EventNamePrefix
...
         /// </summary>
         public string TelemetryOptOutVariableName { get; set; } = "QDK_TELEMETRY_OPT_OUT";
 
+        /// <summary>
+        /// The name of the environment variable to be checked at initialization to see if
+        /// the user wants to opt-in of sending telemetry data.
+        /// If the value of the environment variable is "1", data will be collected or sent.
+        /// The "TelemetryOptOut" will always take precedence.
+        /// </summary>
+        public string TelemetryOptInVariableName { get; set; } = "QDK_TELEMETRY_OPT_IN";
```
kuzminrobin, line 80: 1. I have a suspicion that "TelemetryOptOut" did not survice (mentioned in the fragment "/// The "TelemetryOptOut" will always take precedence.").
2. 
```suggestion
        /// <summary>
        /// The name of the environment variable to be checked at initialization to see if
        /// the user wants to opt-in of sending telemetry data.
        /// If the value of the environment variable is "1", data will be collected and sent.
        /// The "TelemetryOptOut" will always take precedence.
        /// </summary>
        public string TelemetryOptInVariableName { get; set; } = "QDK_TELEMETRY_OPT_IN";
```

src/Telemetry/Library/TelemetryManagerConfig.cs
```diff
@@ -39,13 +55,30 @@ public string EventNamePrefix
...
+        /// the environment variable defined in TelemetryOptOutVariableName.
+        /// </summary>
+        public ConsentKind DefaultTelemetryConsent { get; set; } = ConsentKind.OptedOut;
+
         /// <summary>
         /// The name of the environment variable to be checked at initialization to see if
         /// the user wants to opt-out of sending telemetry data.
         /// If the value of the environment variable is "1", no data will be collected or sent.
         /// </summary>
         public string TelemetryOptOutVariableName { get; set; } = "QDK_TELEMETRY_OPT_OUT";
```
ScottCarda-MS, line 72: I'm unsure why we would need both the TelemetryOptOutVariableName as well as the TelemetryOptInVariableName?

src/Telemetry/readme.md
```diff
@@ -99,6 +100,20 @@ var telemetryConfig = new TelemetryManagerConfig()
 
 **Note:** The AppId must have from 3 to 20 characters and only letters and numbers are allowed.
 
+### .DefaultTelemetryConsent
+
+The default User Consent to collect telemetry (defaults to ConsentKind.OptedOut, can also be set ConsentKind.OptedIn).
```
kuzminrobin, line 105: I have a feeling of a contradiction in the fragment 
> The default User Consent **to collect** telemetry (defaults to ConsentKind.**OptedOut**,

On the one hand it says "**to collect** by default", on the other hand it says "**OptedOut** (to NOT collect) by default".

src/Telemetry/Library/TelemetryManagerInstance.cs
```diff
@@ -417,8 +420,14 @@ private void CheckAndRunSafe(Action action, bool initializingOrTearingDown = fal
...
-        private bool GetTelemetryOptOut(TelemetryManagerConfig configuration) =>
-            Environment.GetEnvironmentVariable(configuration.TelemetryOptOutVariableName) == "1";
+        [SuppressMessage("StyleCop.CSharp.SpacingRules", "SA1009:ClosingParenthesisMustBeSpacedCorrectly", Justification = "Easier to read the logic.")]
+        private bool GetTelemetryOptIn(TelemetryManagerConfig configuration) =>
+            (
+                (configuration.DefaultTelemetryConsent == ConsentKind.OptedIn)
+                ||
+                (Environment.GetEnvironmentVariable(configuration.TelemetryOptInVariableName) == "1")
+            )
+            && !(Environment.GetEnvironmentVariable(configuration.TelemetryOptOutVariableName) == "1");
```
ScottCarda-MS, line 430: I'm not sure I understand the motivation behind the opt-out env variable having precedence. Shouldn't it be something like:
```
telemetryConsent =
    consentKind == OptIn && Environment.GetEnvironmentVariable(configuration.TelemetryOptInVariableName) == "1"
    ||
    !(Environment.GetEnvironmentVariable(configuration.TelemetryOptOutVariableName) == "1")
```

src/Telemetry/Library/TelemetryManager.cs
```diff
@@ -66,14 +66,16 @@ public static class TelemetryManager
...
         /// This property defaults to "false".
-        /// This property is automatically set to true during initialization if the environment variable
-        /// named in Configuration.HostingEnvironmentVariableName has a value of "1".
+        /// This property is automatically set during initialization based on the
+        /// Configuration.DefaultTelemetryConsent and based on the environment variables
+        /// named in Configuration.TelemetryOptOutVariableName and
+        /// Configuration.TelemetryOptInVariableName.
         /// </summary>
-        public static bool TelemetryOptOut => Instance.TelemetryOptOut;
+        public static bool TelemetryOptedIn => Instance.TelemetryOptedIn;
```
kuzminrobin, line 78: Just in case I'll mention (just to keep in mind, not to change anything in this PR):

I have a slight feeling of over-use of the concept TelemetryOptedIn/TelemetryOptedOut (and more than one variable is named `TelemetryOptedIn`). In some cases the concept (and variable name) TelemetryEnabled/TelemetryDisabled describes the reality more precisely. I'll clarify what I mean.

The telemetry (gathering) can be enabled or disabled.
It can be enabled either by default or as a result of the user action (opting-in).
In the same way it can be disabled either by default or as a result of the user action (opting-out).
For the telemetry functionality (whether to gather or not) it does not matter how the telemetry was enabled (was enabled by default or as a result of the user action), it only matters whether the telemetry is enabled or not (only the state matters, but not how that state has been achieved).

That is why I would recommend to consider using different terms (and variable names) 
for the _state_ (enabled or disabled)
and for the _user actions_ (opted-in or opted-out).

src/Telemetry/Library/TelemetryManagerConfig.cs
vxfield: Hi Scott, that's a great point but I was trying to keep consistency with how we handle telemetry opt-out for other products.
In general, either the default consent option will be opted-in and TelemetryOptOutVariableName can be used to turn it off, or the default consent option will be opted-out and TelemetryOptInVariableName can be used to to turn it on.

src/Telemetry/readme.md
vxfield: I'll make the text clearer.

src/Telemetry/Library/TelemetryManagerInstance.cs
vxfield: The opt-out env var takes precedence just in an odd situation in which both environment options are set.
Normally the user would only want (and need) to specify one of them to deviate from the default consent.

src/Telemetry/Library/TelemetryManager.cs
vxfield: Great point. I changed it to be `TelemetryEnabled`.

Issue #1025, opened 2021-05-10
Refactor compiler diagnostic data structures 


samarsha:
* Diagnostics need additional information to display a diagnostic message, but each diagnostic is an enum item, which doesn't allow additional data. The information is stored as a string sequence which is error-prone and causes confusion since each string is referred to by its index instead of a name.
* Diagnostic enums are grouped by severity, but severity should be a separate configurable property of each diagnostic, not hard-coded into its name. This also means that diagnostic codes are duplicated across severities, which is confusing.
* Diagnostic items require a range, but not all diagnostics have a range, which causes a default range to be used as a fallback throughout the compiler whenever a diagnostic is created. Diagnostics should make the range optional and use a standard default value to reduce boilerplate/code duplication.

samarsha:
An example of where the code for diagnostic messages starts to get really messy and hard to maintain is when we want to make the diagnostic message a bit smarter/adaptable to the context:

https://github.com/microsoft/qsharp-compiler/blob/29d10588c2b85a5ffe8e4243eb94cd96f1a3adbc/src/QsCompiler/DataStructures/Diagnostics.fs#L474-L481

(I don't think it's right to create separate diagnostics for the two messages, because they really are the same diagnostic, just displayed differently depending on context.)

PR #1344, opened 2022-02-08
Fix type mismatch diagnostic for types in a double negative position

### LINKED ISSUE #1335: Error message for type mismatch in a double negative position is confusing
## Example
In this example, `Foo` expects an operation of type `Int => Unit`. If you give it an operation of type `Bool => Unit`, where the mismatch occurs in a double negative (i.e. positive) position, then the error message shows the "expected" and "actual" types in the opposite way you would expect.

```qsharp
@EntryPoint()
operation Main() : Unit {
    Foo(Bar);
}

operation Foo(op : Int => Unit) : Unit {}

operation Bar(x : Bool) : Unit {}
```

```
The type Int does not match the type Bool.
Actual type:   (Int => Unit)
Expected type: (Bool => Unit)
```

This is different than when the type mismatch occurs in a negative position, e.g. `Foo("bar")`, where `Int => Unit` is now the "expected" type.

```
The type String does not match the type (Int => Unit).
Actual type:   String
Expected type: (Int => Unit)
```

## Expected behavior

The error message should better match the context of the call expression, i.e. the expected type should probably always be the parameter type specified by `Foo`. The error message should still communicate the direction of the variance somehow.

There is also a bug in the diagnostic range: it appears on the first character of the statement instead of on the argument.

samarsha:
* Add the direction of subtyping to the internal type inference API. The left-hand side is the expected type in error messages. Flip only the direction of subtyping when entering a negative position, rather than swapping the LHS and RHS. This fixes #1335.
* Refactor the way diagnostics are generated by the type inference module. Propagate context information bottom-up rather than top-down, which simplifies the private type inference API.


src/QsCompiler/SyntaxProcessor/TypeInference/InferenceContext.fs
```diff
@@ -20,100 +20,143 @@ open Microsoft.Quantum.QsCompiler.Transformations.QsCodeOutput
...
+    | TypeMismatch of TypeMismatch
+    | TypeIntersectionMismatch of ordering: Ordering * mismatch: TypeMismatch
+    | CompilerDiagnostic of QsCompilerDiagnostic
+
+module Diagnostic =
+    let withContext expected (actual: ResolvedType) =
+        let sameRange (type_: ResolvedType) : ResolvedType option -> _ =
+            Option.forall (fun context -> type_.Range = context.Range)
+
+        function
```
ScottCarda-MS, line 99: What argument is being matched here?

src/QsCompiler/SyntaxProcessor/TypeInference/InferenceContext.fs
```diff
@@ -20,100 +20,143 @@ open Microsoft.Quantum.QsCompiler.Transformations.QsCodeOutput
...
+        | TypeMismatch mismatch ->
+            if sameRange actual mismatch.ActualContext then
```
ScottCarda-MS, line 101: I don't understand this line. If the Diagnostic already has a context, then withContext is supposed to replace it, right? Here it seems, if the range of the existing context is the same as the range for the desired context, it gets replaced, or if there isn't an existing context, it gets replaced. But if a context already exists and has a different range, than NoOp?

src/QsCompiler/SyntaxProcessor/TypeInference/InferenceContext.fs
```diff
@@ -20,100 +20,143 @@ open Microsoft.Quantum.QsCompiler.Transformations.QsCodeOutput
...
-        /// The most recent relevant ancestor of the right-hand type.
-        OriginalRight: ResolvedType
+    let (.>) lhs rhs = Relation(lhs, Supertype, rhs)
+
+type TypeMismatch =
+    {
+        Expected: ResolvedType
+        ExpectedContext: ResolvedType option
+        Actual: ResolvedType
+        ActualContext: ResolvedType option
```
ScottCarda-MS, line 74: Some comments here explaining the fields, especially the context fields, would be helpful. I'm not yet understanding what the context fields are used for.

src/QsCompiler/SyntaxProcessor/TypeInference/InferenceContext.fs
```diff
@@ -20,100 +20,143 @@ open Microsoft.Quantum.QsCompiler.Transformations.QsCodeOutput
...
+            Option.forall (fun context -> type_.Range = context.Range)
+
+        function
+        | TypeMismatch mismatch ->
+            if sameRange actual mismatch.ActualContext then
+                mismatch |> TypeMismatch.withContext expected actual |> TypeMismatch
+            else
+                TypeMismatch mismatch
+        | TypeIntersectionMismatch (ordering, mismatch) ->
+            if sameRange expected mismatch.ExpectedContext && sameRange actual mismatch.ActualContext then
```
ScottCarda-MS, line 106: Is there a reason we do sameRange for both the expected and the actual contexts here, where in the TypeMismatch case we only check the actual context?

src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -465,10 +465,9 @@ type DiagnosticItem =
     static member Message(code: ErrorCode, args: IEnumerable<string>) =
         let message =
             match code with
-            | ErrorCode.TypeMismatch ->
-                "The type {1} does not match the type {0}.\nActual type:   {3}\nExpected type: {2}"
+            | ErrorCode.TypeMismatch -> "The type {0} does not match the type {1}.\nExpected: {2}\n  Actual: {3}"
```
ScottCarda-MS, line 468: Does this make TypeMismatch a special case of the TypeIntersectionMismatch, where the first argument is 'match'?

ScottCarda-MS (approved):

src/QsCompiler/SyntaxProcessor/TypeInference/InferenceContext.fs
samarsha: The third argument of `withParents`, which is a `Diagnostic`.

src/QsCompiler/SyntaxProcessor/TypeInference/InferenceContext.fs
samarsha: The range checking is designed to find the type of the expression underlined by the diagnostic, and stop updating the parent types after that point. I added a doc comment to `Diagnostic.withParents` about this.

src/QsCompiler/SyntaxProcessor/TypeInference/InferenceContext.fs
samarsha: I renamed this type and its fields to hopefully be clearer, and I added a doc comment with an example. The new type name is `TypeContext`.

src/QsCompiler/SyntaxProcessor/TypeInference/InferenceContext.fs
samarsha: It depends on whether the range of the actual type, or the ranges of both the expected and actual types, are used to create the compiler diagnostic in `Diagnostic.toCompilerDiagnostic`. I added a comment here to explain that.

src/QsCompiler/SyntaxProcessor/TypeInference/InferenceContext.fs
```diff
@@ -20,100 +20,143 @@ open Microsoft.Quantum.QsCompiler.Transformations.QsCodeOutput
...
+
+    let private typeMismatchArgs mismatch =
+        let expectedContext = mismatch.ExpectedContext |> Option.defaultValue mismatch.Expected
+        let actualContext = mismatch.ActualContext |> Option.defaultValue mismatch.Actual
+
+        [
+            describeType mismatch.Actual
+            describeType mismatch.Expected
+            SyntaxTreeToQsharp.Default.ToCode expectedContext
+            SyntaxTreeToQsharp.Default.ToCode actualContext
```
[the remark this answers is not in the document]
samarsha: `describeType` is the same as `SyntaxTreeToQsharp.Default.ToCode`, except it adds extra info if the type is a type parameter. The extra info is designed to fit into the first parent of the error message, but not the second part (which has the parents/context). Example:

```
Couldn't match type parameter 'a (bound by Foo) with type Qubit.
Expected: Qubit
  Actual: 'a
```

"Actual: parameter 'a (bound by Foo)" would look weird. (I agree that this is kind of confusing and a fragile part of the diagnostic message generation, though.)

src/QsCompiler/DataStructures/Diagnostics.fs
samarsha: Sort of, but they don't quite mean the same thing by "expected type"... for TypeMismatch, the actual type is usually a subtype of the expected type. For TypeIntersectionMismatch, the actual type may be either a subtype or a supertype of the expected type. I changed the error messages a bit based on this.

src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -465,23 +465,24 @@ type DiagnosticItem =
...
         | _ -> str // let's fail silently for now
 
-    static member Message(code: ErrorCode, args: IEnumerable<string>) =
-        let unlines = String.concat Environment.NewLine
+    static member Message(code, args) =
+        let typeMismatch summary expected actual =
+            String.concat Environment.NewLine [ summary; "Expected: " + expected; "  Actual: " + actual ]
 
         let message =
             match code with
```
ScottCarda-MS, line 473: The numbering for the arguments in these error messages seems pretty out of order. Is this does to maintain consistency with how these error codes would have been pass arguments before?

samarsha: I wanted to use the same function `Diagnostic.typeContextArgs` to create the argument list for all the type errors, and let each error message decide the best order to display them in. Unfortunately it does look pretty confusing with the argument numbers - this would be cleaner if #1025 was resolved and each diagnostic argument had a name instead of a number.

Issue #1346, opened 2022-02-09
Q# formatter doesn't add whitespaces when removing parentheses


thomashaener:
**Describe the bug**

Running the Q# formatter on a for loop with parentheses results in code that doesn't compile if there are no whitespaces between `for` and the opening `(`.

**To Reproduce**

Run the formatter on
```
operation Test() : Unit {
    for(i in 0..1){
        Message("Test.");
    }
}
```

**Expected behavior**

```
operation Test() : Unit {
    for i in 0..1 {
        Message("Test.");
    }
}
```

**Actual behavior**
```
operation Test() : Unit {
    fori in 0..1{
        Message("Test.");
    }
}
```

PR #1347, opened 2022-02-09
Ensure Spacing After Removed Parentheses

ScottCarda-MS:
Fixed a bug, #1346 where a space needed to be inserted after removing parentheses around for loops conditions and qubit declaration bindings.


kuzminrobin (approved):
Approving in case you get blocked for a long time. 
I would recommend at least one more approval, I don't know F#.

PR #1349, opened 2022-02-10
Pulling in Ryan's contribution

bettinaheim:
There is currently no support implemented for enabling debug symbols via a project property. 
We should add that before merging.


bettinaheim:
This PR should be merged via merge commit to facilitate the merge back to main.

PR #1359, opened 2022-02-15
Use correct continuation parser in symbolTuple

### LINKED ISSUE #1356: Lambdas: Associativity across commas is too greedy in lambdas.
**Describe the bug**

When a lambda is specified in a list of other arguments, it interprets the other arguments as arguments to the lambda. For example, if `foo` takes a callable as its third argument and that callable is passed as a lambda, it will fail. So this code:
```qsharp
foo(firstArg, secondArg, lambda_arg => lambda_arg + 1);
```
is interpreted as if it was:
```qsharp
foo((firstArg, secondArg, lambda_arg) => lambda_arg + 1);
```
instead of:
```qsharp
foo(firstArg, secondArg, (lambda_arg => lambda_arg + 1));
```
Manually adding the parenthesis from the last example allows it compile as expected, but those parenthesis shouldn't be necessary.

**To Reproduce**

Use the following example code:
```qsharp
namespace lambda_bugbash {

    open Microsoft.Quantum.Canon;
    open Microsoft.Quantum.Diagnostics;
    
    @EntryPoint()
    operation main() : String {
        use q = Qubit();

        within {
            X(q);
        }
        apply {
            let testfunc = qs => CNOT(q, qs[0]);
            AssertOperationsEqualReferenced(1, testfunc, qs => X(qs[0]));
        }

        return "Success!";
    }
}
```

samarsha:
On line 667 of QsExpressionParsing.fs, `continuation` is given to `buildTupleItem` when it should be `continuation'`. I renamed `continuation'` to shadow `continuation` to avoid confusing the two.

Closes #1356.

### DIFF
--- a/src/QsCompiler/Tests.Compiler/SyntaxTests.fs
+++ b/src/QsCompiler/Tests.Compiler/SyntaxTests.fs
@@ -1481,6 +1481,43 @@ let ``Lambda tests`` () =
         |> toExpr,
         []
 
+        "U(7, q => X(q))",
+        true,
+        CallLikeExpression(
+            toIdentifier "U",
+            [
+                toInt 7
+                Lambda.createUnchecked
+                    Operation
+                    (toSymbol "q")
+                    (CallLikeExpression(toIdentifier "X", toTuple [ toIdentifier "q" ]) |> toExpr)
+                |> Lambda
+                |> toExpr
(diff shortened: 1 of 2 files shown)


swernli (approved):
LGTM!

PR #1369, opened 2022-02-18
Add support for Quantinuum targets

cesarzc:
This change adds support for Quantinuum targets in the QDK.


vxfield:
I'm not familiar with the compiler infra, so I defer the review to @swernli

swernli (approved):

PR #1375, opened 2022-03-02
Make common transformation internal and remove `_internal_` constructor overloads

samarsha:
Follow up from #1290 to refactor and clean up the transformations:

- Converted public class `CommonTransformationNodes` into an internal interface `ICommonTransformation`. This type is an implementation detail and shouldn't be part of the public API. See comments here: https://github.com/microsoft/qsharp-compiler/pull/1290#discussion_r797929014
- Removed all constructor overloads with the unused `_internal_` parameter hack. This also removes the unnecessary mutable fields `_Transformations` and `*Handle` and their delayed initialization. See comments here: https://github.com/microsoft/qsharp-compiler/pull/1290#discussion_r796049677


src/QsCompiler/Core/Transformations/ExpressionTransformation.fs
```diff
@@ -17,51 +17,36 @@ open Microsoft.Quantum.QsCompiler.Transformations.Core.Utils
 
 type private ExpressionKind = QsExpressionKind<TypedExpression, Identifier, ResolvedType>
 
+type ExpressionKindTransformationBase(expressionTransformation: _ -> ExpressionTransformationBase, options) =
+    static let createExpressionTransformation kinds options =
+        let types = TypeTransformationBase options
+        let expressions = ExpressionTransformationBase((fun () -> kinds), (fun () -> types), options)
+        fun () -> expressions
```
ScottCarda-MS, line 24: So this is the actual body of the new(options) constructor, right? The reason it needs to be here instead of at the definition of the new(options) constructor is because it needs to be passed into that constructor's call to the default constructor, and F# demands the call to the default constructor happens first, is that right?

Correct me if I'm wrong, but basically we would want
```
new(options: TransformationOptions) as this =
    let typeTransformation = new TypeTransformationBase(options)    
    let expressionTransformation =
        ExpressionTransformationBase((fun _ -> this), (fun _ -> typeTransformation), options)
    ExpressionKindTransformationBase((fun () -> expressionTransformation), options)
```
But F# doesn't allow for the default constructor to be called that late in the definition. Is my understanding of why this static let is necessary correct?

ScottCarda-MS (commented):
I still have a lot of this to review, but I wanted to make sure I have clarity from you on some points before moving forward.

src/QsCompiler/Core/Transformations/ExpressionTransformation.fs
samarsha: Yes. The restriction isn't specific to F#, by the way. C# also requires the alternate constructor to be called first.

src/QsCompiler/Core/Transformations/ExpressionTransformation.fs
```diff
@@ -17,51 +17,36 @@ open Microsoft.Quantum.QsCompiler.Transformations.Core.Utils
...
-    member val internal ExpressionTransformationHandle = missingTransformation "expression" with get, set
-
-    member this.Expressions: ExpressionTransformationBase = this.ExpressionTransformationHandle()
-    member this.Types: TypeTransformationBase = this.ExpressionTransformationHandle().Types
-    member this.Common: CommonTransformationNodes = this.ExpressionTransformationHandle().Types.Common
-
-    new(expressionTransformation: unit -> ExpressionTransformationBase, options: TransformationOptions) as this =
-        new ExpressionKindTransformationBase(options, "_internal_")
-        then this.ExpressionTransformationHandle <- expressionTransformation
+    member internal this.Common = expressionTransformation().Types.Common
```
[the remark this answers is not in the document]
samarsha: External transformations should access the methods through the `SyntaxTreeTransformation` class. `ICommonTransformation` is an implementation detail to break the mutual recursion between the transformation types while they are being defined.

src/QsCompiler/Core/Transformations/ExpressionTransformation.fs
```diff
@@ -17,51 +17,36 @@ open Microsoft.Quantum.QsCompiler.Transformations.Core.Utils
...
-
-    new(options: TransformationOptions) as this =
-        new ExpressionKindTransformationBase(options, "_internal_")
-        then
-            let typeTransformation = new TypeTransformationBase(options)
-
-            let expressionTransformation =
-                new ExpressionTransformationBase((fun _ -> this), (fun _ -> typeTransformation), options)
+    new(expressionTransformation, typeTransformation: unit -> TypeTransformationBase, options: TransformationOptions) =
+        ExpressionKindTransformationBase(expressionTransformation, options)
```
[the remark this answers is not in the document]
samarsha: We should be using signature files to give explicit types to the public API of the transformations, but I don't want to add inline type annotations to the .fs file unless necessary to resolve ambiguous types, because IMO it adds clutter and is harder to read.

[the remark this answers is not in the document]
ScottCarda-MS: I really don't agree with that direction for the style, but since this is subjective, I recommend we leave the type annotations that are already present in the code, while we discuss as a broader team discussion about this style.

src/QsCompiler/Core/Transformations/ExpressionTransformation.fs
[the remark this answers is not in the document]
ScottCarda-MS: If someone were to just create an `ExpressionTransformation`, how would they get access to `OnTypeRange` for example?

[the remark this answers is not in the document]
samarsha: If `this` is of type `ExpressionTransformation`, then `this.Transformation.OnTypeRange` should work.

src/QsCompiler/Core/Transformations/ExpressionTransformation.fs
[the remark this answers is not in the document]
ScottCarda-MS: When discussing F# with other developers, they have consistently told me that they use type annotations for function parameters, as opposed to using signature files. I am not in favor of using signature files per my experience with them in the Q# formatter project. They made working in that code base significantly harder and created an even further divorce of the parameters from the types that describe them. Since we are in disagreement on this, I appreciate that I can't ask you to make a commitment to not having type annotations in general, but as a compromise I am asking you to leave the type annotations that already exist in the code as they are.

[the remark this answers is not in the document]
samarsha: We've already had internal team discussions about the use of F# signature files, and decided in favor of using them. They are recommended by the [F# component design guidelines](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/component-design-guidelines#consider-using-explicit-signature-files-fsi-for-stable-library-and-component-apis) which our F# code should follow. Conversely, the [F# coding conventions](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/conventions) do not recommend that type annotations should be used for parameter types, other than to specify the public API which is accomplished by signature files.

Please follow up with me in private if you need more clarity on our team's F# coding guidelines.

src/QsCompiler/Core/Transformations/ExpressionTransformation.fs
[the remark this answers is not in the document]
ScottCarda-MS: Oh all the non-Base transformation classes have a reference to a SyntaxTreeTransformation object with those handles. That addresses how I can _call_ the OnTypeRange from an Expression Transformation. How would I overwrite the OnTypeRange though?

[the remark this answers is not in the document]
ScottCarda-MS: With the current implementation, I can define a class that inherits from `CommonTransformationNode` and overwrite the `OnTypeRange` behavior. Then I could define my nuanced `ExpressionTransformation` class and assign it to my `CommonTransformationNode` class by assigning `this.Common = MyCommonTransformationNode;` in the constructor.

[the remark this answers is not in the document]
samarsha: I think there is a misunderstanding. This PR doesn't change the way you override the common transformation methods. In main as well as in this PR, you can only override them in SyntaxTreeTransformation.

> With the current implementation, I can define a class that inherits from `CommonTransformationNode` and overwrite the `OnTypeRange` behavior.

CommonTransformationNodes cannot be inherited from because its constructor is internal. If you try to inherit from it, you get this error: "[CS1729] 'CommonTransformationNodes' does not contain a constructor that takes 0 arguments"

> Then I could define my nuanced `ExpressionTransformation` class and assign it to my `CommonTransformationNode` class by assigning `this.Common = MyCommonTransformationNode;` in the constructor.

You cannot assign to Common because it is a read-only property. If you try to assign a different instance of CommonTransformationNode, you get this error: "[CS0200] Property or indexer 'ExpressionTransformationBase.Common' cannot be assigned to -- it is read only"

src/QsCompiler/Core/Transformations/SyntaxTreeTransformation.fs
```diff
@@ -48,270 +45,218 @@ type SyntaxTreeTransformation<'T> private (state: 'T, options: TransformationOpt
...
-            this.Expressions <- new ExpressionTransformation<'T>(this, options)
-            this.StatementKinds <- new StatementKindTransformation<'T>(this, options)
-            this.Statements <- new StatementTransformation<'T>(this, options)
-            this.Namespaces <- new NamespaceTransformation<'T>(this, options)
-
-    new(state: 'T) = new SyntaxTreeTransformation<'T>(state, TransformationOptions.Default)
+    new(state) = SyntaxTreeTransformation(state, TransformationOptions.Default)
 
-    // These overrides are here since it is impractical to have them live in the definition of CommonTransformationItems.
+    abstract OnLocalNameDeclaration: name: string -> string
```
cesarzc, line 50: Why is `OnLocalNameDeclaration` not set to be replaced with the identity function in 2022-09 if it is implemented in the identity function?

src/QsCompiler/Core/Transformations/SyntaxTreeTransformation.fs
```diff
@@ -1,39 +1,36 @@
...
 open Microsoft.Quantum.QsCompiler.SyntaxTree
-
+open Microsoft.Quantum.QsCompiler.Transformations.Core
 
 // setup for syntax tree transformations with internal state
 
-type SyntaxTreeTransformation<'T> private (state: 'T, options: TransformationOptions, _internal_: string) =
-    inherit CommonTransformationNodes()
-
+type SyntaxTreeTransformation<'T>(state, options) as this =
```
cesarzc, line 16: In this PR I have seen a couple of instances where the type of a parameter is no longer explicitly described. I personally find it useful when reading code. Is not including the type of a parameter a general practice in F#? If so, why is it favored?

cesarzc (approved):

ScottCarda-MS (commented):
I'd like to have a team meeting to go over the changes in this PR. Some of the works here are pretty complex, and it is difficult to go over it all via text and messaging. It would be really helpful for my own understanding and for the wider team if we could have a meeting to go over the finer details of the PR, please.

cesarzc (commented):
Resetting my review to go over it offline to get a bit more clarity on some of these changes.

cesarzc (changes requested):
Requesting changes so we can go over them in detail offline.

src/QsCompiler/Core/Transformations/SyntaxTreeTransformation.fs
```diff
@@ -48,270 +45,218 @@ type SyntaxTreeTransformation<'T> private (state: 'T, options: TransformationOpt
...
-        then
-            this.Types <- new TypeTransformation<'T>(this, options)
-            this.ExpressionKinds <- new ExpressionKindTransformation<'T>(this, options)
-            this.Expressions <- new ExpressionTransformation<'T>(this, options)
-            this.StatementKinds <- new StatementKindTransformation<'T>(this, options)
-            this.Statements <- new StatementTransformation<'T>(this, options)
-            this.Namespaces <- new NamespaceTransformation<'T>(this, options)
-
-    new(state: 'T) = new SyntaxTreeTransformation<'T>(state, TransformationOptions.Default)
+    new(state) = SyntaxTreeTransformation(state, TransformationOptions.Default)
```
ScottCarda-MS, line 48: Does this need to be `SyntaxTreeTransformation<'T>`?

ScottCarda-MS, line 48: Does this need to be `SyntaxTreeTransformation<'T>`?

samarsha: The `'T` is inferred.

src/QsCompiler/Core/Transformations/SyntaxTreeTransformation.fs
samarsha: Maybe the comment phrasing is confusing - what I mean is that

```fsharp
default this.OnLocalName name = this.Statements.OnVariableName name
```

should be replaced by

```fsharp
default this.OnLocalName name = name
```

OnLocalNameDeclaration already has this kind of implementation (because it doesn't have a backwards compatibility shim like OnLocalName), so a todo isn't needed.

src/QsCompiler/Core/Transformations/SyntaxTreeTransformation.fs
samarsha: Yes, F# has a strong culture of type inference that tends to prefer omitting type annotations when not needed. But also important is that there's no support in the tooling for requiring type annotations. As far as I know, there's no way to enable a compiler warning or a lint for functions that are missing parameter or return types. The standard F# way is to create a signature file (.fsi), which does enforce that every public or internal declaration in the file has an explicit type.

src/QsCompiler/Core/Transformations/SyntaxTreeTransformation.fs
ScottCarda-MS: Since we have another class that has the same name but is not generic, should we leave the `<T>` here just for code legibility, so that people don't get confused and think it is referring to the non-generic type, as I just did?

ScottCarda-MS (approved):

ScottCarda-MS:
Thanks for walking us through the mechanics here @samarsha!

src/QsCompiler/Core/Transformations/ExpressionTransformation.fs
samarsha: I just realized that F# doesn't actually have this restriction. I replaced the `static let`s with inline `let`s.

src/QsCompiler/Core/Transformations/SyntaxTreeTransformation.fs
samarsha: In this case, it can't refer to the non-generic type, because this constructor has to call one of the alternate constructors of the same class. Also, the non-generic `SyntaxTreeTransformation` constructor doesn't take a `state` argument. I understand that there's some ambiguity here due to the name `SyntaxTreeTransformation` being overloaded, but this situation feels similar to other cases of type inference, where I prefer to let F# infer as much as it can and use the language tooling for more detailed type information.

src/QsCompiler/Core/Transformations/ExpressionTransformation.fs
ScottCarda-MS: Nice catch! That makes things even cleaner looking.

PR #1381, opened 2022-03-15
Enable LiftLambdaExpressions by default in configuration

samarsha:
Resolves microsoft/iqsharp#606.


cgranade (approved):
Looks great! We should also make sure to add either per-repo tests to IQ#, or IQ# tests to the e2e build. Possibly both?

ScottCarda-MS (approved):

samarsha:
I'm not sure what you mean - I thought the tests in the IQ# repo already run automatically in the e2e build? So with microsoft/iqsharp#620, we should have enough test coverage, right?

PR #1391, opened 2022-03-23
Add "size" (used in array literals) to the list of highlighted keywords

tcNickolas:


ScottCarda-MS (approved):

src/VSCodeExtension/syntaxes/qsharp.tmLanguage.json
```diff
@@ -48,7 +48,7 @@
         },
         {
           "name": "keyword.other.qsharp",
-          "match": "\\b(new|not|and|or|w\/)\\b"
+          "match": "\\b(new|not|and|or|w\/|size)\\b"
```
samarsha, line 51: The tricky thing about `size` is that it's not a true keyword like `not`. It acts like a keyword when it occurs in a specific part of an array literal, but it's still a valid identifier:

```qsharp
let size = 123;
let xs = [size, size = 2]; // first size is a variable, second size is a "keyword"
```

So I am not sure if we want all instances of `size` highlighted.

tcNickolas: Hmm, let me see if I can make something like "a comma, optional whitespace, size, optional whitespace, equality sign" work  without also highlighting the comma and the equality...

raghavthind2005 (approved):

tcNickolas:
I didn't get to this in the past month and likely won't get to in the next month, so I've opened https://github.com/microsoft/qsharp-compiler/issues/1419 and will close this PR for now.

PR #1394, opened 2022-03-25
Evaluate repeated value in sized array expressions exactly once

samarsha:
Update C# code gen to follow the semantics for sized array expressions clarified here: microsoft/qsharp-language#129


ScottCarda-MS (approved):
LGTM

PR #1397, opened 2022-03-29
Replace QsFmt grammar with qsharp-language submodule

samarsha:
Update QsFmt to use the unified ANTLR grammar from microsoft/qsharp-language#135.


ScottCarda-MS (approved):
The code content looks good. I'd like to get this reviewed by someone who is more knowledgeable then me about the submodule linking. Adding @anpaz to the review to make sure this new linkage between the compiler repo and the language repo is correct.

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "qsharp-language"]
+	path = qsharp-language
+	url = https://github.com/microsoft/qsharp-language.git
+	branch = samarsha/unified-grammar
```
anpaz, line 4: this should to `main`, no?

anpaz: The submodule is also not pointing to `main`, so when initialized and checked out it will point to the exact commit listed in the repo. Probably not the intention.

PR #1408, opened 2022-04-12
Do not serialize syntax tree if compilation failed

### LINKED ISSUE #1401: AST should not be serialized when there are compilation errors
AST serialization doesn't currently support lambda expression nodes, and the lambda lifting rewrite step isn't run when there are compilation errors. That means every compilation error in a project that contains a lambda expression will also have a serialization error, like that shown in #1387. While AST serialization should support lambdas, there's also not really a reason for it to run when there are compilation issues.

samarsha:
Closes #1401.

### DIFF
--- a/src/QsCompiler/Compiler/CompilationLoader.cs
+++ b/src/QsCompiler/Compiler/CompilationLoader.cs
@@ -462,17 +462,17 @@ namespace Microsoft.Quantum.QsCompiler
         /// <summary>
         /// Contains the built compilation including the syntax tree after executing all configured rewrite steps.
         /// </summary>
-        public QsCompilation? CompilationOutput { get; }
+        public QsCompilation? CompilationOutput { get; private set; }
 
         /// <summary>
         /// Contains the absolute path where the binary representation of the generated syntax tree has been written to disk.
         /// </summary>
-        public string? PathToCompiledBinary { get; }
+        public string? PathToCompiledBinary { get; private set; }
 
         /// <summary>
         /// Contains the absolute path where the generated dll containing the compiled binary has been written to disk.
         /// </summary>
-        public string? DllOutputPath { get; }
+        public string? DllOutputPath { get; private set; }
 
         /// <summary>
         /// Contains the full Q# syntax tree after executing all configured rewrite steps, including the content of loaded references.
@@ -573,7 +573,7 @@ namespace Microsoft.Quantum.QsCompiler
                 thisDllUri = new Uri(Path.GetFullPath(".", "CompilationLoader.cs"));
             }
 
-            if (this.config.LoadTargetSpecificDecompositions)
+            if (this.Validation == Status.Succeeded && this.config.LoadTargetSpecificDecompositions)
             {
                 PerformanceTracking.TaskStart(PerformanceTracking.Task.ReplaceTargetSpecificImplementations);
                 this.CompilationOutput = this.ReplaceTargetSpecificImplementations(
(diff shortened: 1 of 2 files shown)


src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -719,6 +624,90 @@ public CompilationLoader(SourceLoader loadSources, IEnumerable<string> reference
         {
         }
 
+        private IEnumerable<(LoadedStep, Action<Status>)> InternalRewriteSteps(Uri executingAssembly)
+        {
+            // TODO: The dependencies for rewrite steps should be declared as part of IRewriteStep interface, and we
```
cesarzc, line 629: In order to not lose track of this, it would be great to create an issue and reference it in the comment.

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -719,6 +624,90 @@ public CompilationLoader(SourceLoader loadSources, IEnumerable<string> reference
...
+            // should query those here.
+#pragma warning disable CS0618 // Type or member is obsolete
+            var dependencies = BuiltIn.AllBuiltIns.Select(b => b.FullName);
+#pragma warning restore CS0618 // Type or member is obsolete
+
+            var status = this.compilationStatus;
+            var steps = new (IRewriteStep Step, bool Enabled, Action<Status> SetStatus)[]
+            {
+                // TODO: It would be nicer to trim unused intrinsics. Currently, this is not possible due to how the
```
cesarzc, line 638: Same here, it would be great to create an issue and reference it in the comment to not lose track of it.

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -719,6 +624,90 @@ public CompilationLoader(SourceLoader loadSources, IEnumerable<string> reference
...
+                // old setup of the C# runtime works. With the new setup (interface-based approach for target
+                // packages), it is possible to trim unused intrinsics.
+                (
```
cesarzc, line 641: nit: the formatting seems a bit weird with the opening parentheses in its own line but not the closing one.

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -719,6 +624,90 @@ public CompilationLoader(SourceLoader loadSources, IEnumerable<string> reference
...
+        {
+            PerformanceTracking.TaskStart(PerformanceTracking.Task.RewriteSteps);
+            this.config.PrepareQirGeneration = this.config.PrepareQirGeneration || this.externalRewriteSteps.Any(step => step.Name == "QIR Generation");
+
+            var externalSteps = this.externalRewriteSteps.Select((step, index) =>
+                (step, new Action<Status>(status => this.compilationStatus.LoadedRewriteSteps[index] = status)));
+
+            var steps = this.InternalRewriteSteps(executingAssembly)
+                .Concat(externalSteps)
+                .OrderByDescending(step => step.Item1.Priority);
```
cesarzc, line 670: Before this change, internal rewrite steps were executed before external rewrite steps. I think we should keep the ordering as it was before.

src/QsCompiler/Compiler/CompilationLoader.cs
samarsha: The todo was already there before, I just moved it to a new method.

src/QsCompiler/Compiler/CompilationLoader.cs
samarsha: Same as above, this is not a new todo.

src/QsCompiler/Compiler/CompilationLoader.cs
samarsha: StyleCop rules require the closing parenthesis to be on the same line as the last tuple item.

src/QsCompiler/Compiler/CompilationLoader.cs
samarsha: Could you explain in more detail? In the old code, it added all rewrite steps (internal and external) to `steps` and sorted them before executing. I think I preserved that behavior.

cesarzc: Yes, that's true, I didn't read the code correctly. Resolving this conversation.

cesarzc (approved):

raghavthind2005 (approved):

PR #1410, opened 2022-04-14
Fix several cases of incorrect diagnostic ranges for call expressions

### LINKED ISSUE #1366: Incorrect error range when calling an operation specialization in a lambda function
**Describe the bug**

The following code produces a cryptic and misplaced error:
image

**Expected behavior**

The compiler should produce a more descriptive error message with a range that is either the whole lambda expression, the lambda body, or just the functor application in the lambda body.

samarsha:
- When inferring a call expression in a function (ExpressionVerification.fs:885-886), the constructed `functionType` can end up as the representative for the call expression's type, if the expected callable type is a variable. That means it can be used for diagnostic ranges in certain cases, so it should have a non-generated range.
- In `InferenceContext.Relate`, binding a type variable can cause type constraints to be checked for types that aren't descendants of the current `expected` and `actual` types. That means that `Diagnostic.withParents` shouldn't be called on the diagnostics returned by `tryBind`.
- `Diagnostic.withParents` should default to the current types in the case of an orphan context for the purposes of checking if the ranges changed - otherwise, the `Option.forall` assumed the range did not change for an empty option.

This fixes the range of diagnostics in cases like these:

```qsharp
let f = op -> Adjoint op();
                      ~~
let g = op => op();
g(X);
  ~
```

Closes #1366.


samarsha:
In a meeting it was mentioned that it would be nice to have tests for these changes. This PR does not change which diagnostics are generated, only the range and message of each diagnostic, both of which are properties that are not checked by any of the existing diagnostic tests.

I can look into adding test utilities for checking these properties so that at least new tests could optionally take advantage of them. I know @bettinaheim has been aware of this test gap and might have some thoughts also.

src/QsCompiler/SyntaxProcessor/TypeInference/InferenceContext.fs
```diff
@@ -377,13 +387,25 @@ type InferenceContext(symbolTracker: SymbolTracker) =
...
+                | QsTypeKind.Operation _ ->
+                    QsTypeKind.Operation((input, output), CallableInformation.NoInformation) |> ResolvedType.New |> Some
+                | QsTypeKind.Function _ -> QsTypeKind.Function(input, output) |> ResolvedType.New |> Some
+                | _ -> None
+
+            // The callable constraint needs special handling for diagnostics:
+            //
+            // - If the source is here, it is assumed to be a call expression. The callable is the "expected" type. A
+            //   type mismatch is the argument's fault.
+            // - If the source is somewhere else, the callable is being provided as a value to a higher-order function.
```
bettinaheim, line 401: Does this logic (in particular the second bullet) also apply to lambdas that may be declared in the same function?

bettinaheim:
> In a meeting it was mentioned that it would be nice to have tests for these changes. This PR does not change which diagnostics are generated, only the range and message of each diagnostic, both of which are properties that are not checked by any of the existing diagnostic tests.
> 
> I can look into adding test utilities for checking these properties so that at least new tests could optionally take advantage of them. I know @bettinaheim has been aware of this test gap and might have some thoughts also.

I think it would be good to set up a test validating diagnostic ranges starting with just testing the case at hand. We won't close the test gap, but diagnostic ranges seem fragile (one of the more frequent sources of bugs), so let's take the first step. @samarsha Do you mind setting up some compiler tests for validating diagnostic ranges? You can focus on the case at hand and we will add to them over time. Let's also scope it to validating the compiler rather than any IDE feedback.

samarsha:
> Do you mind setting up some compiler tests for validating diagnostic ranges? You can focus on the case at hand and we will add to them over time. Let's also scope it to validating the compiler rather than any IDE feedback.

Added.

src/QsCompiler/SyntaxProcessor/TypeInference/InferenceContext.fs
samarsha: This special case was added for this test case involving a lambda:

```qsharp
let g = op => op();
g(X);
  ~
```

because `op` has a deferred callable constraint.

src/QsCompiler/Tests.Compiler/AccessModifierTests.fs
```diff
@@ -1,53 +1,58 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-namespace Microsoft.Quantum.QsCompiler.Testing
+module Microsoft.Quantum.QsCompiler.Testing.AccessModifierTests
```
cesarzc, line 4: Is there a reason for some of the test files being kept as namespaces and some being changed to modules?

src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
```diff
@@ -753,8 +753,8 @@ type QsExpression with
                 |> inference.Constrain
                 |> List.iter diagnose
             else
-                let diagnostics =
-                    inference.Constrain(callable.ResolvedType <. ResolvedType.New(QsTypeKind.Function(argType, output)))
+                let functionType = ResolvedType.withKind (QsTypeKind.Function(argType, output)) callable.ResolvedType
+                let diagnostics = inference.Constrain(callable.ResolvedType <. functionType)
```
cesarzc, line 757: This might be a silly question, but couldn't find the answer in F#'s documentation. What is the `<.` operator doing here?

src/QsCompiler/SyntaxProcessor/TypeInference/Diagnostic.fs
```diff
@@ -36,19 +36,17 @@ type Diagnostic =
     | CompilerDiagnostic of QsCompilerDiagnostic
 
 module Diagnostic =
-    let withParents expected actual diagnostic =
-        let hasSameRange (type1: ResolvedType) : ResolvedType option -> _ =
```
cesarzc, line 40: Was `hasSameRange` publicly accessible? If so, should we consider deprecating it now and removing it later in case there's something relying on it?

src/QsCompiler/Tests.Compiler/AccessModifierTests.fs
samarsha: Basically, there is no reason for any of these tests to be in a class, because the classes have no fields. A module is simpler and more idiomatic F#. But going from class to module is a big diff, so I only did it for small files where I was already changing most of it anyway.

src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
samarsha: It's a custom operator that we define:

https://github.com/microsoft/qsharp-compiler/blob/d21f8fa62eedbc3e1c2850f004350f0c3f67a8ca/src/QsCompiler/SyntaxProcessor/TypeInference/Constraint.fsi#L82-L85

src/QsCompiler/SyntaxProcessor/TypeInference/Diagnostic.fs
samarsha: `hasSameRange` was a local variable inside the `withParents` function. (Everything in this file is internal, also - see [Diagnostic.fsi](https://github.com/microsoft/qsharp-compiler/blob/samarsha/fix-call-diagnostic-ranges/src/QsCompiler/SyntaxProcessor/TypeInference/Diagnostic.fsi).)

src/QsCompiler/Tests.Compiler/AccessModifierTests.fs
cesarzc: Thank you for the explanation!

cesarzc (approved):

PR #1411, opened 2022-04-14
Only check type resolutions for cycles where every callable is type parameterized

bettinaheim:
It would be more efficient to just store whether a node in the call graph has type parameters along with the node.


swernli (approved):

ScottCarda-MS (approved):
This seems like a good improvement for efficiency. Could we wait to merge this in until I can chat with you all tomorrow, and get properly caught up?

swernli:
> This seems like a good improvement for efficiency. Could we wait to merge this in until I can chat with you all tomorrow, and get properly caught up?

Sure thing! Disabled the auto merge so we can sync up tomorrow.

ScottCarda-MS:
@swernli and I had a chat, and I'm all caught up. This looks ready to merge in. Thanks!

PR #1415, opened 2022-04-22
Naming Scheme for Generated Callables and Local Symbols

ScottCarda-MS:
Updates the naming scheme used for generated callables and local variables in the Q# compiler. Generated callables names will all have a GUID surrounded by double underscores prepended to the original name. Generated variable names will have a label and number surrounded by double underscores prepended to the original name, or by themselves if there is no original name. The old NameDecorator class is deprecated in favor of the new static NameGenerator class.


swernli:
@ScottCarda-MS Is this PR still relevant or can it be closed?

PR #1422, opened 2022-04-27
Update .NET SDK check in Language Server to 6.0

ricardo-espinoza:
Port of #1421 to April 2022 release branch

### DIFF
--- a/src/QsCompiler/LanguageServer/Utils.cs
+++ b/src/QsCompiler/LanguageServer/Utils.cs
@@ -224,9 +224,9 @@ namespace Microsoft.Quantum.QsLanguageServer
 
     internal static class DotNetSdkHelper
     {
-        private static readonly Regex DotNet31Regex = new Regex(@"^3\.1\.\d+", RegexOptions.Multiline | RegexOptions.Compiled);
+        private static readonly Regex DotNet31Regex = new Regex(@"^6\.0\.\d+", RegexOptions.Multiline | RegexOptions.Compiled);
 
-        public static bool? IsDotNet31Installed()
+        public static bool? IsDotNet6Installed()
         {
             var process = Process.Start(new ProcessStartInfo
             {
--- a/src/QsCompiler/LanguageServer/Utils.cs
+++ b/src/QsCompiler/LanguageServer/Utils.cs
@@ -224,7 +224,7 @@ namespace Microsoft.Quantum.QsLanguageServer
 
     internal static class DotNetSdkHelper
     {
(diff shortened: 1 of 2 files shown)


cgranade (commented):
LGTM, thanks! Not marking as Approve until @bettinaheim approves cutoff exception.

src/QsCompiler/LanguageServer/Utils.cs
```diff
@@ -224,9 +224,9 @@ public override void Initialize(IEventSource eventSource)
 
     internal static class DotNetSdkHelper
     {
-        private static readonly Regex DotNet31Regex = new Regex(@"^3\.1\.\d+", RegexOptions.Multiline | RegexOptions.Compiled);
+        private static readonly Regex DotNet31Regex = new Regex(@"^6\.0\.\d+", RegexOptions.Multiline | RegexOptions.Compiled);
```
bettinaheim, line 227: Do we really need .NET 6.0 precisely or can we make this .NET 6.0 or newer? Why exactly are we tied to a precise version?

bettinaheim (approved):

src/QsCompiler/LanguageServer/Utils.cs
ricardo-espinoza: For now, I'd like to play it safe given that during the .NET Core 3.1 days, we didn't support having only .NET 5 installations. 
Later this year, we'll need to assess the extent of our support to .NET 7, but until then, I'd like to make sure we have a well-known and supported environment.

anpaz: I think the questions is if it has to be 6.0,or can it be 6.1 or higher?

anpaz, line 227: You want to rename the variable too.

ricardo-espinoza: Oh. Of course. Sorry for missing that. :)

ricardo-espinoza: Well, we don't know what a 6.1 version would mean in the current roadmap of .NET ... We didn't have a 5.1 for example, and currently we only have 6.0.X and 7.0.X (in preview) versions. If at some point they decide to create a .NET 6.1 SDK, we will need to understand the compatibility story and make an explicit decision on the level of support we would have. I'd rather be more constrained for now.

ricardo-espinoza: @anpaz, for now it has to be 6.0.X but not higher

PR #1426, opened 2022-04-28
Check if Language Server path is an empty string

raffaeleterribile:
The extension searches in the Language Server in the path configured in the settings and then in the default folder.
But if the path in the settings is an empty string, the Language Server executable wasn't found because was checked only if this setting is undefined or null, not an empty string.
Now the default folder is used even if the setting is an empty string.


src/VSCodeExtension/src/languageServer.ts
```diff
@@ -127,7 +127,7 @@ export class LanguageServer {
 
         // If lsPath is still undefined or null, then we didn't have a manual
         // path set up above.
-        if (lsPath === undefined || lsPath === null) {
+        if (lsPath === undefined || lsPath === null || (lsPath.trim() === "")) {
```
ricardo-espinoza, line 130:
```suggestion
        if (lsPath === undefined || lsPath === null || lsPath.trim() === "") {
```

ricardo-espinoza (changes requested):

ricardo-espinoza (approved):

PR #1428, opened 2022-04-29
Add capability analyzers for QIR compilation target restrictions

### LINKED ISSUE #1429: Diagnostics for QIR submission to hardware targets
There are a number of restrictions that need to be enforced in Q# to ensure that the generated QIR can be converted into a form that quantum hardware platforms support:

- [X] Forbid recursion
- [X] Forbid while and repeat loops
- [x] Forbid using mutables of types that are not supported on hardware
- [x] Forbid conditional expressions of types that are not supported on hardware
- [X] Enforce that there is a single return per function and no fail statements
- [x] Enforce that array lengths in new arrays expressions, sized array expressions, and qubit array allocations are constants
- [x] Forbid BigInt
- [x] Allow string literals only as argument to Message
- [x] Forbid that mutable variables can be assigned to immutable things
- [x] Limit return values for entry points to the types supported by the provider or arrays/tuples of supported types
- [x] Forbid default-initialized array constructor
- [x] Forbid for loops over mutable iteration lengths
- [x] Forbid mutables in array item access and the indices for copy-and-update
- [x] Allow entry point arguments only of type Int and Bool
- [x] Forbid first-class callable values

samarsha:
- Refactor the previous result capability analyzer to be part of a more modular design.
- Add new capability analyzers for the new target restrictions listed in #1429.
- Refactor the capability tests to be simpler and have better coverage of each test case.

Closes #1429.


samarsha:
I rewrote history so that each commit could be reviewed separately, if that helps.

src/QsCompiler/DataStructures/RuntimeCapability.fs
```diff
@@ -3,76 +3,139 @@
...
+
+    static member AdaptiveExecution = { resultOpacity = Transparent; classical = Integral }
+
+    static member BasicQuantumFunctionality = { resultOpacity = Opaque; classical = Full }
+
+    static member BasicMeasurementFeedback = { resultOpacity = Controlled; classical = Full }
+
+    static member FullComputation = { resultOpacity = Transparent; classical = Full }
+
+module RuntimeCapability =
```
bettinaheim, line 99: Discussed during PR walkthrough: Since we make a bunch of breaking changes, we should just take the opportunity to align terminology across the board and call this TargetCapability.

src/QsCompiler/DataStructures/RuntimeCapability.fs
```diff
@@ -3,76 +3,139 @@
...
+            | :? CaseInsensitive as other ->
+                String.Compare(ci.Original, other.Original, StringComparison.InvariantCultureIgnoreCase)
+            | _ -> ArgumentException("Type mismatch.", nameof obj) |> raise
+
+module CaseInsensitive =
+    let original (CaseInsensitive s) = s
+
+type ResultOpacity =
+    | Opaque
+    | Controlled
```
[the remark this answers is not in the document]
samarsha: I think controlled result opacity correlates with what is possible to lift into the [classical control operations](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.classicalcontrol)? That was my idea for the name.

src/QsCompiler/DataStructures/RuntimeCapability.fs
samarsha: I renamed it, but there are still leftovers:

- The qsc command line option is called `--runtime` but should probably be `--target-capability` or `--capability`.
- The project property is still called `ResolvedRuntimeCapabilities` - this needs an SDK change.

src/QsCompiler/DataStructures/RuntimeCapability.fs
```diff
@@ -3,76 +3,139 @@
...
+        Map.tryFindKey (fun _ -> (=) capability) names |> Option.map CaseInsensitive.original
+
+    [<CompiledName "FromName">]
+    let ofName name =
+        Map.tryFind (CaseInsensitive name) names
+
+type RuntimeCapability with
+    member capability.Name = RuntimeCapability.name capability |> Option.toObj
+
+    static member Parse name =
```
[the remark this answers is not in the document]
samarsha: It never throws. I renamed it back to `TryParse`.

src/QsCompiler/DataStructures/RuntimeCapability.fs
```diff
@@ -3,76 +3,139 @@
...
+
+    let ofName name =
+        Map.tryFind (CaseInsensitive name) names
+
+type ClassicalCapability =
+    | Empty
+    | Integral
+    | Full
+
+module ClassicalCapability =
```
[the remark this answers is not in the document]
samarsha: What about `ClassicalCompute`?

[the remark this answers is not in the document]
cesarzc: I like `ClassicalCompute`.

src/QsCompiler/DataStructures/RuntimeCapability.fs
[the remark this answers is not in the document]
samarsha: For example here you see that `ConvertClassicalControl` is true if the result opacity is `Controlled`:

https://github.com/microsoft/qsharp-compiler/blob/e81e3845a3fef4ae52624d763b42b6cef42d6a83/src/QsCompiler/Compiler/CompilationLoader.cs#L205-L206

I like the symmetry and would prefer to avoid using a different word.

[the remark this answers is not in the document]
bettinaheim: It does indeed. I am fine keeping it as is.

src/QsCompiler/Core/SymbolResolution.fs
```diff
@@ -322,26 +322,39 @@ module SymbolResolution =
...
+                // Backwards compatibility with (Name : String, Reason : String).
+                getString args[0] |> Option.bind TargetCapability.ofName
+            | ValueTuple args when args.Length = 3 ->
+                // Parse (ResultOpacity : String, Classical : String, Reason : String).
+                Seq.fold2 applyBuilder TargetCapability.bottom args builders |> Some
+            | _ -> None
+
+        Seq.filter BuiltIn.MarksRequiredCapability attributes
+        |> Seq.choose getCapability
+        |> Seq.fold (fun acc c -> Option.fold TargetCapability.merge c acc |> Some) None
```
bettinaheim, line 356: What does merge do; does merge mean I could have several capability attributes and it would resolve to a capability that satisfies both?

src/QsCompiler/SyntaxProcessor/CapabilityInference/TypeAnalyzer.fs
```diff
@@ -0,0 +1,210 @@
...
+
+let shallowTypeName (ty: ResolvedType) =
+    match ty.Resolution with
+    | ArrayType _ -> "array"
+    | TupleType _ -> "tuple"
+    | QsTypeKind.Operation _ -> "operation"
+    | QsTypeKind.Function _ -> "function"
+    | _ -> SyntaxTreeToQsharp.Default.ToCode ty
+
+let describeScenario =
```
bettinaheim, line 85: This seems like the error string is not split between the diagnostics file and here. I'd prefer to absorb it all into the diagnostics file.

src/QsCompiler/SyntaxProcessor/CapabilityInference/TypeAnalyzer.fs
```diff
@@ -0,0 +1,210 @@
...
+                QsCompilerDiagnostic.Error(ErrorCode.UnsupportedClassicalCapability, args) range |> Some
+
+        Some
+            {
+                Capability = capability
+                Diagnose = diagnose
+                Properties = ()
+            }
+
+let rec flattenTuple =
```
bettinaheim, line 121: This one keeps being redefined on every other PR. The SyntaxGenerator contains a method for this, and I know there has been a desire to have a C# extension to flatten argument tuples. Maybe it would be a good time to add that?

src/QsCompiler/SyntaxProcessor/CapabilityInference/TypeAnalyzer.fs
```diff
@@ -0,0 +1,210 @@
...
+                returnPattern context.Value callable |> Option.iter patterns.Add
+                callable
+        }
+
+    transformation.Statements <-
+        { new StatementTransformation(transformation, TransformationOptions.NoRebuild) with
+            override _.OnStatement statement =
+                match statement.Statement with
+                | QsFailStatement _ ->
+                    use _ = local { context.Value with StringLiteralsOk = true } context
```
bettinaheim, line 161: So we are saying stings are ok in messages and fail statements, though we are currently forbidding fail statements. I think that will work, and we may want to reenable failstatements.

src/QsCompiler/SyntaxProcessor/CapabilityInference/ResultAnalyzer.fs
```diff
@@ -0,0 +1,186 @@
...
+            let code =
+                if target.Capability.ResultOpacity = ResultOpacity.opaque then
+                    ErrorCode.UnsupportedResultComparison
+                else
+                    ErrorCode.ResultComparisonNotInOperationIf
+
+            Some(code, [ target.Name ])
+        | ReturnInDependentBranch ->
+            if target.Capability.ResultOpacity = ResultOpacity.opaque then
+                None
```
bettinaheim, line 61: I am reading this as:
If we have a return in a branch whose execution depends on a measurement result, and the target doesn't support result comparisons anywhere, then give no error. Is that because in this case we already gave an error for the branching in the first place?

src/QsCompiler/SyntaxProcessor/CapabilityInference/ResultAnalyzer.fs
```diff
@@ -0,0 +1,186 @@
...
+let isResultEquality expression =
+    let validType =
+        function
+        | InvalidType -> None
+        | kind -> Some kind
+
+    let binaryType lhs rhs =
+        validType lhs.ResolvedType.Resolution |> Option.defaultValue rhs.ResolvedType.Resolution
+
+    // This assumes that:
```
bettinaheim, line 87: Very important and helpful comment.

src/QsCompiler/SyntaxProcessor/CapabilityInference/ResultAnalyzer.fs
```diff
@@ -0,0 +1,186 @@
...
+    // - Result has no subtype that supports equality comparisons.
+    // - Compound types containing Result (e.g., tuples or arrays of results) do not support equality comparison.
+    match expression.Expression with
+    | EQ (lhs, rhs)
+    | NEQ (lhs, rhs) -> binaryType lhs rhs = Result
+    | _ -> false
+
+let freezeContext dependsOnResult (block: QsPositionedBlock) (context: _ ref) =
+    let knownVars = Seq.map (fun v -> v.VariableName) block.Body.KnownSymbols.Variables |> Set.ofSeq
+    let frozenVars = if dependsOnResult then knownVars else context.Value.FrozenVars
```
bettinaheim, line 97: What are frozen variables?

src/QsCompiler/SyntaxProcessor/CapabilityInference/FeatureAnalyzer.fs
```diff
@@ -0,0 +1,94 @@
...
+let createPattern range feature =
+    let capability = TargetCapability.withClassical ClassicalCapability.full TargetCapability.bottom
+
+    let diagnose (target: Target) =
+        let range = QsNullable.defaultValue Range.Zero range
+
+        if TargetCapability.subsumes target.Capability capability then
+            None
+        else
+            let description =
```
bettinaheim, line 30: Same comment regarding the split of the diagnostic text here as on the other file.

src/QsCompiler/SyntaxProcessor/CapabilityInference/FeatureAnalyzer.fs
```diff
@@ -0,0 +1,94 @@
...
+
+                match expression.Expression with
+                | CallLikeExpression (callee, _) ->
+                    let isTrivial =
+                        function
+                        | InvalidExpr
+                        | Identifier (GlobalCallable _, _)
+                        | Identifier (InvalidIdentifier, _)
+                        | AdjointApplication _
+                        | ControlledApplication _ -> true
```
bettinaheim, line 83: This is now actually too permissive; You would need to look into whether the inner item is a local callable or not.

src/QsCompiler/SyntaxProcessor/CapabilityInference/ConstAnalyzer.fs
```diff
@@ -0,0 +1,181 @@
...
+let createPattern range =
+    let capability = TargetCapability.withClassical ClassicalCapability.full TargetCapability.bottom
+
+    let diagnose (target: Target) =
+        let range = QsNullable.defaultValue Range.Zero range
+
+        if TargetCapability.subsumes target.Capability capability then
+            None
+        else
+            let description = "conditional expression or mutable variable in a constant context"
```
bettinaheim, line 26: I am a bit worried about users not knowing what they need to do. What do you think about something very explicit like "mutable variables and conditional expressions cannot be used in a constant context"? I also think we need to somewhere define what a constant context is - maybe even spelling that out in the error message itself.

src/QsCompiler/Core/SymbolResolution.fs
samarsha: Yes. I don't really have a strong preference on how multiple `RequiresCapability` attributes are handled. Maybe it should be invalid.

src/QsCompiler/SyntaxProcessor/CapabilityInference/TypeAnalyzer.fs
```diff
@@ -0,0 +1,210 @@
...
+            match ty.Resolution with
+            | TupleType _ -> ClassicalCapability.empty
+            | Bool
+            | Int -> ClassicalCapability.integral
+            | _ -> ClassicalCapability.full
+        | EntryPointReturn ty ->
+            match ty.Resolution with
+            | UnitType
+            | Result
+            | TupleType _
```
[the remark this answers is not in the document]
samarsha: No, since the `findScenarios` function already traverses the entire type, including type arguments to tuples and arrays.

src/QsCompiler/SyntaxProcessor/CapabilityInference/TypeAnalyzer.fs
samarsha: Yes, but I didn't want to create a ton of different error codes for each scenario. Since the scenarios we are looking for are very fluid and will likely change over time, it seemed better to have one code for an unsupported classical capability, with the error message describing the specific problem.

I would prefer not to use strings here, too, but that's difficult to do unless microsoft/qsharp-compiler#1025 is implemented.

src/QsCompiler/SyntaxProcessor/CapabilityInference/TypeAnalyzer.fs
samarsha: I did this to avoid a double error for `fail` (one for the `fail` statement itself, and one for the message string). Are `fail` statements supported?

src/QsCompiler/DataStructures/RuntimeCapability.fs
bettinaheim: I can go an update that if you merge the changes here into the hardware targeting branch. 
Is this PR on its own fully functional? If so, I suggest I just cover those as part of the second half (getting the hardware_targeting branch merged) that contains the Sdk changes.

src/QsCompiler/DataStructures/RuntimeCapability.fs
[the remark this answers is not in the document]
bettinaheim: Sounds good.

src/QsCompiler/Core/SymbolResolution.fs
bettinaheim: For other attributes I was giving warnings when they were redundant and errors when they were redefining. But actually, I think just merging them in this case works great. I was just curious.

src/QsCompiler/SyntaxProcessor/CapabilityInference/TypeAnalyzer.fs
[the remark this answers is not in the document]
bettinaheim: Ah, I overlooked that. A comment here in the code would be helpful. Are there tests that cover this?

src/QsCompiler/SyntaxProcessor/CapabilityInference/TypeAnalyzer.fs
bettinaheim: Alright, I see your point about an expected number of changes in the next couple of weeks. Let's defer that then to be processed with 1025. Could you please add a comment though to the strings in each of the analyzers pointing that out and linking issue 1025?

src/QsCompiler/SyntaxProcessor/CapabilityInference/TypeAnalyzer.fs
bettinaheim: No, they are definitively not supported on hardware, but it is a question of whether we should allow them at the language level and fail after qat validation if they have not been optimized away (i.e. fail statements cannot be supported by the runtime, but plenty of fails check compile time information). Writing this down, I think I am coming to the conclusion that we should not allow them, and instead think about where to improve the language/libraries such that these fail statements are not needed.

src/QsCompiler/SyntaxProcessor/CapabilityInference/FeatureAnalyzer.fs
bettinaheim: Resolving as per the discussion above.

src/QsCompiler/SyntaxProcessor/CapabilityInference/FeatureAnalyzer.fs
samarsha: This function does not need to recurse on the expression, since it is used on the next line with `callee.Exists(isTrivial >> not)`, which does the recursion implicitly. There is an existing test case that checks this:

```qsharp
operation CallFunctorOfExpression() : Unit {
    use q = Qubit();
    Adjoint [Unitary][0](q);
}
```

but I'll add your test case as well.

bettinaheim (approved):
Looks great!

src/QsCompiler/DataStructures/RuntimeCapability.fs
samarsha: Yeah, everything should work with this PR, just the names will be inconsistent. I'm fine with doing it in another PR.

src/QsCompiler/SyntaxProcessor/CapabilityInference/TypeAnalyzer.fs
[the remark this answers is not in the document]
samarsha: There are these tests:

```qsharp
@EntryPoint()
operation EntryPointReturnDoubleArray() : Double[] {
    return [];
}

@EntryPoint()
operation EntryPointReturnResultDoubleTuple() : (Result, Double) {
    return (Zero, 0.0);
}
```

that assert that the `Double` type produces an error when it's inside of an array or a tuple.

src/QsCompiler/SyntaxProcessor/CapabilityInference/ConstAnalyzer.fs
samarsha: > What do you think about something very explicit like "mutable variables and conditional expressions cannot be used in a constant context"?

The full message is:

> This construct requires a classical runtime capability that is not supported by the target {targetName}: conditional expression or mutable variable in a constant context.

You would like to change it to:

> This construct requires a classical runtime capability that is not supported by the target {targetName}: mutable variables and conditional expressions cannot be used in a constant context.

This is a bit inconsistent from the other error messages, where the description after the colon is a noun phrase describing what you did (e.g., "recursion" or "fail statement"), rather than a complete sentence. Since the first half of the message says "a ... capability that is not supported," IMO it's already clear that what is being described is not allowed?

> I also think we need to somewhere define what a constant context is - maybe even spelling that out in the error message itself.

A constant context is one of:

- The iterator in a for loop.
- The length `n` of an array in `[x, size = n]` or `Qubit[n]`.
- The value of an immutable variable binding.
- An array index.
- A call argument.

It's hard to describe all of that succinctly in an error message - I would prefer to link to documentation, but we don't have any right now.

src/QsCompiler/SyntaxProcessor/CapabilityInference/ResultAnalyzer.fs
samarsha: I added a comment to the field in `Context` describing it. It's the term I use for variables declared outside of an result conditional branch while you are inside that branch. For example:

```qsharp
mutable x = 0;
if M(q) == One {
    // x is frozen
}
```

src/QsCompiler/SyntaxProcessor/CapabilityInference/TypeAnalyzer.fs
samarsha: I see, I didn't know about QsTuple.Items. I removed `flattenTuple`. I don't think I need to add anything new right now.

src/QsCompiler/SyntaxProcessor/CapabilityInference/ResultAnalyzer.fs
bettinaheim: Got it, thanks.

src/QsCompiler/SyntaxProcessor/CapabilityInference/ConstAnalyzer.fs
bettinaheim: Alright, let's leave it as is, and we should think about documentation (not sure if we need to formalize further first).

PR #1427, opened 2022-04-29
Added target capability to entry point diriver.

kuzminrobin:
3-repo change: [Q#RT](https://github.com/microsoft/qsharp-runtime/pull/1002) + [Q#Compiler](https://github.com/microsoft/qsharp-compiler/pull/1427) + Internal Repo.
The CI build is expected to fail, but the qdk.release with full validation is expected to succeed.

PR #1437, opened 2022-05-16
Repeated PR: Added target capability to entry point driver

kuzminrobin:
Repeated #1427 (after revert).


kuzminrobin:
Force-merging since the overall qdk.release succeeded, and this is a repeated PR after a revert.

PR #1446, opened 2022-05-24
Edits to C# generation to make intrinsics in target packages work

bettinaheim:


src/QsCompiler/CSharpGeneration/SimulationCode.fs
```diff
@@ -1032,7 +1032,7 @@ module SimulationCode =
                         let undecorated = nameDecorator.Undecorate name
                         if String.IsNullOrWhiteSpace undecorated then name else undecorate undecorated
 
-                    "IIntrinsic" + undecorate n.Name
```
swernli, line 1035: This would seem to break (or perhaps supersede?) the `GenerateConcreteIntrinsics` feature that enables the use of target packages with C# generation. If this is removed, do the interfaces defined in https://github.com/microsoft/qsharp-runtime/tree/main/src/Simulation/TargetDefinitions/Interfaces no longer apply?

swernli:
I think that to get target packages and C# regeneration to work, all that needs to be added is logic that ensures regeneration is performed as if `GenerateConcreteIntrinsics` is specified (essentially inheriting the setting from the target packages instead of using the one from the current project). Normal C# generation already works with target packages.

src/QsCompiler/CSharpGeneration/SimulationCode.fs
bettinaheim: This is outdated and not part of the working e2e. If the changes to the C# generation are removed and left as it is on main, then the e2e using GenerateConcreteIntrinsics is failing with the failure linked on the runtime PR.

bettinaheim:
> I think that to get target packages and C# regeneration to work, all that needs to be added is logic that ensures regeneration is performed as if `GenerateConcreteIntrinsics` is specified (essentially inheriting the setting from the target packages instead of using the one from the current project). Normal C# generation already works with target packages.

I don't believe this is correct. I had both GenerateConcreteIntrinsics and the regeneration of C# code enabled by the Sdk, and that resulted in the linked build failure. No additional modifications. You are welcome to confirm yourself.

PR #1455, opened 2022-06-03
Removed .NET Core 2.1 and .NET Core 3.0 from TFMs recognized by the compiler as valid

filipw:
As per the official [.NET and .NET Core Support Policy](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core), support for .NET Core 3.0 ended on March 3, 2020 and for .NET Core 2.1 on August 21, 2021.

Recognizing these TFMs as valid for Q# programs not only produces the following MSBuild warnings:

```
The target framework 'netcoreapp2.1' is out of support and will not receive security updates in the future. 
Please refer to https://aka.ms/dotnet-core-support for more information about the support policy.
```

but can also be viewed as security risk.


swernli (approved):
Thanks for updating this!

anpaz:
@filipw, as you probably figured out we've been having some problems with our build pipeline that we had to escalate with Azure Devops support. Looks like the problem is fixed and we should be able to merge this soon.

PR #1458, opened 2022-06-07
QIR generation support for hardware targets

bettinaheim:


src/QuantumSdk/DefaultItems/DefaultItems.targets
```diff
@@ -28,42 +28,49 @@
...
+    <TargetCapability Condition="'$(TargetCapability)' != '' And '$(ResolvedProcessorArchitecture)' == 'IonQProcessor'">Invalid</TargetCapability>
+    <TargetCapability Condition="'$(TargetCapability)' != '' And '$(TargetCapability)' != 'AdaptiveExecution' And '$(TargetCapability)' != 'BasicExecution' And '$(ResolvedProcessorArchitecture)' == 'QCIProcessor'">Invalid</TargetCapability>
+    <TargetCapability Condition="'$(TargetCapability)' != '' And '$(TargetCapability)' != 'BasicExecution' And '$(ResolvedProcessorArchitecture)' == 'RigettiProcessor'">Invalid</TargetCapability>
+    <TargetCapability Condition="'$(TargetCapability)' != '' And '$(ResolvedProcessorArchitecture)' == 'MicrosoftSimulator'">Invalid</TargetCapability>
+    <TargetCapability Condition="'$(TargetCapability)' != '' And '$(TargetCapability)' != 'AdaptiveExecution' And '$(TargetCapability)' != 'BasicExecution' And '$(ResolvedProcessorArchitecture)' == 'Unspecified'">Invalid</TargetCapability>
```
swernli, line 63: I'm not sure how much we care about this, but it would appear from this line that a developer cannot use the target "Any" and explicity choose a capability like "FullComputation" or "BasicMeasurementFeedback" for their generated code. Is it intentional that the user can't pick these? While there aren't any scenarios for it at the moment, it seems like that could cause headaches for testing where the code would need to use artificial execution targets just to trigger the desired capability.

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -167,7 +162,7 @@ public NamespaceTransformation(SyntaxTreeTransformation<TransformationState> par
                 public override QsCallable OnCallableDeclaration(QsCallable c)
                 {
                     var relevantAccessModifiers = this.SharedState.GetAccessModifiers.Apply(this.SharedState.TypeParams.Values)
-                        .Append(c.Access);
+                        .Append(c.Access).Append(Access.Internal); // monomorphized callables should never be public
```
swernli, line 165: Couldn't this result in the internall modifier appearing more than once if the generic was already internal? Does that matter or is later code resilient to duplicate modifiers?

src/QsCompiler/Transformations/Monomorphization.cs
```diff
@@ -31,15 +31,20 @@ namespace Microsoft.Quantum.QsCompiler.Transformations.Monomorphization
...
         /// is set to false, then intrinsics will not be monomorphized.
         /// </summary>
         public static QsCompilation Apply(QsCompilation compilation, bool monomorphizeIntrinsics = false)
         {
+            // If this compilation is for a library project, there are no defined entry points. Instead,
+            // treat every public, non-generic callable as a possible entry point into the library.
+            var hasEntryPoints = compilation.EntryPoints.Length > 0;
+            if (!hasEntryPoints)
+            {
+                compilation = new QsCompilation(compilation.Namespaces, compilation.InteroperableSurface(includeReferences: false).ToImmutableArray());
```
swernli, line 45: What happens here if the library has a non-internal generic? Does it throw? Leave it generic? It seems like that should be a failure case, but perhaps that doesn't trigger a failure until it gets to a generation step that cares about monomorphization having completed.

src/QsCompiler/CSharpGeneration/SimulationCode.fs
```diff
@@ -1122,14 +1111,14 @@ module SimulationCode =
     let buildOperationInfoProperty (globalContext: CodegenContext) operationInput operationOutput operationName =
         let propertyType =
             match globalContext.ProcessorArchitecture with
-            | target when target = AssemblyConstants.HoneywellProcessor ->
-                sprintf "HoneywellEntryPointInfo<%s, %s>" operationInput operationOutput
```
swernli, line 1126: Removing Honeywell here does make sense, however will it appear as a breaking change for anyone who is still using Honeywell targets? I think the answer should be no based on how the QDK configures things, but much of the system still uses the old name (there is only a Honeywell QDK template, for example).

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -631,11 +638,7 @@ private IEnumerable<(LoadedStep, Action<Status>)> InternalRewriteSteps(Uri execu
                 // TODO: It would be nicer to trim unused intrinsics. Currently, this is not possible due to how the
                 // old setup of the C# runtime works. With the new setup (interface-based approach for target
                 // packages), it is possible to trim unused intrinsics.
-                (
-                    new SyntaxTreeTrimming(keepAllIntrinsics: true, dependencies),
-                    this.config.IsExecutable && !this.config.SkipSyntaxTreeTrimming,
-                    s => status.TreeTrimming = s),
-
+                (new SyntaxTreeTrimming(keepAllIntrinsics: true, dependencies), this.config.TrimTree, s => status.TreeTrimming = s),
```
swernli, line 641: Food for thought in the future: per the comment, the unused intrinsics are only needed for C# that includes generated abstract intrinsics. In other words, an intrinsic can be safely trimmed if it is from a target package and is unused. Not needed for this change, but an interesting optimization to consider in the future (perhaps an update to the comment with our latest understandings).

src/QsCompiler/Core/SyntaxGenerator.fs
```diff
@@ -204,13 +216,13 @@ module SyntaxGenerator =
     let CallNonGeneric (lhs: TypedExpression, rhs: TypedExpression) =
         let kind = CallLikeExpression(lhs, rhs)
 
-        let quantumDep =
+        let quantumDep = // NOTE: for an operation call we always need to set the quantum dependency to true
```
swernli, line 219: Just for my own understanding, does this mean that every operation call is assumed to use quantum resources? That seems like it would make any use of the quantum dependency info rather pessimistic, but perhaps that is the itent.

src/QsCompiler/DataStructures/ReservedKeywords.fs
```diff
@@ -275,7 +279,10 @@ module AssemblyConstants =
     let PerfDataOutputPath = "PerfDataOutputPath"
     let DocsOutputPath = "DocsOutputPath"
     let DocsPackageId = "DocsPackageId"
+    [<Obsolete("Replaced by IsTargetPackage.")>]
```
swernli, line 282: I understand the intent here is to avoid a breaking change for any compiler consumers who may be using the old `GenerateConcreteIntrinsic` property, but since the support for this property is gone, it means any consumers will compile successfully with a warning and then get different behavior. From https://docs.microsoft.com/en-us/dotnet/api/system.obsoleteattribute?view=net-6.0, it appears properties marked as Obsolete only generate warnings instead of errors. I think we should either keep it like this for the warning but then use that value to set the new `IsTargetPackage` property or just remove it such that consumers will get a compilation failure instead.

src/QsCompiler/CSharpGeneration/CSharpGeneration.fsproj
```diff
@@ -24,4 +25,11 @@
   <ItemGroup>
     <ProjectReference Include="..\RoslynWrapper\RoslynWrapper.fsproj" />
   </ItemGroup>
+
+  <ItemGroup>
+    <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleTo">
+      <_Parameter1>Tests.Microsoft.Quantum.CSharpGeneration$(PublicKeySuffix)</_Parameter1>
```
kuzminrobin, line 31: (personal curiosity) If needed then where can one see the description of identifiers like this one - `<_Parameter1>`?

src/QsCompiler/QirGeneration/AddOutputRecording.cs
```diff
@@ -0,0 +1,572 @@
...
+                definedElements.Item2.Concat(definedCallables.Values.Select(QsNamespaceElement.NewQsCallable))
+                .ToImmutableArray());
+            return namespaces.Where(x => x.Name != apiNamespace.Name).Append(apiNamespace).ToImmutableArray();
+        }
+
+        private static (SymbolTuple, ImmutableDictionary<string, LocalVariableDeclaration<string, ResolvedType>>) CreateDeconstruction(ResolvedType returnType, int enumerationStart = 0)
+        {
+            var newVars = ImmutableDictionary.CreateBuilder<string, LocalVariableDeclaration<string, ResolvedType>>();
+            static string MakeVariableName(int enumeration) =>
+                $"__rtrnVal{enumeration}__";
```
ScottCarda-MS, line 336: I know it is only used here, but "rtrnVal" as a label should be defined somewhere more accessible, like as a static readonly string near the top of the file.

src/QsCompiler/QirGeneration/AddOutputRecording.cs
```diff
@@ -0,0 +1,572 @@
...
+
+                        var paramDecl = SyntaxGenerator.ExtractItems(c.ArgumentTuple).ValidDeclarations();
+                        var localDeclarations = variableDeclarations.Values.Concat(paramDecl)
+                            .ToImmutableDictionary(decl => decl.VariableName);
+                        return new QsScope(
+                            this.record(outputTuple, localDeclarations).Prepend(outputDeconstruction).ToImmutableArray(),
+                            new LocalDeclarations(paramDecl));
+                    }
+
+                    var wrapperName = new QsQualifiedName(c.FullName.Namespace, c.FullName.Name + "__main");
```
ScottCarda-MS, line 495: Please use the `NameDecorator` class to generate names for callables and variables. If the behavior you are looking for is not in the `NameDecorator`, then please add it to the `NameDecorator`. That way, when we go to update our naming standards, this generated name doesn't accidentally get left behind, and if you add new behavior to the `NameDecorator`, this can be considered and incorporated into new standards we come up with for generated names.

src/QsCompiler/QirGeneration/QIR/DataStructures.cs
```diff
@@ -386,85 +444,253 @@ private Value AllocateTuple(bool registerWithScopeManager)
...
+        public Value OpaquePointer =>
+            Types.IsArray(this.LlvmType)
+            ? this.Value
+            : throw new InvalidOperationException("cannot get opaque pointer for a constant array allocated on the stack");
 
         public Value Length => this.length.Load();
 
         /// <summary>
         /// Creates a new array value.
+        /// The array will be stack allocated if and only if maxCount is set.
```
swernli, line 475: It looks like this comment might be out of date, since the params are now `count` and `allocOnStack`

src/QsCompiler/QirGeneration/QIR/Values.cs
```diff
@@ -23,10 +24,112 @@ internal class QirValues
...
+                return this.sharedState.Values.CreateCustomType(udt.Item.GetFullName(), values, allocOnStack: this.sharedState.TargetQirProfile);
+            }
+
+            if (type.Resolution is ResolvedTypeKind.ArrayType itemType)
+            {
+                return this.sharedState.Values.CreateArray(itemType.Item, Array.Empty<IValue>(), allocOnStack: this.sharedState.TargetQirProfile);
+            }
+            else if (type.Resolution.IsFunction || type.Resolution.IsOperation)
+            {
+                // We can't simply set this to null, unless the reference and alias counting functions
```
swernli, line 93: Both the QIR spec and the current QIR Runtime implementation treat null callables for the reference and alias counting functions as no-op, so perhaps this can indeed be simplified?

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestReferenceCounts2.ll
```diff
@@ -49,20 +49,5 @@ else__1:                                          ; preds = %entry
   ret { %String*, %Array* }* %s
 
 continue__1:                                      ; No predecessors!
```
swernli, line 51: I know it's not part of this change per-se, but why are we generating a block with no predecessors? This would get trimmed by CFG simplification anyway, so it's not a big deal, but it strikes me as odd. Perhaps an investigation for later?

swernli (approved):
Overall changes look good. I have a few comments that would be nice to resolve before merge but nothing blocking.

src/QuantumSdk/DefaultItems/DefaultItems.targets
bettinaheim: I edited the Sdk, and I believe these changes permit the scenario you describe. Please take another look at the Sdk.

src/QsCompiler/Transformations/Monomorphization.cs
bettinaheim: This is already an enumerable collecting all access modifiers for types that are used. A couple of lines down .Min is used to determine the least permissible access. Hence, adding one to the bucket is fine matter.

src/QsCompiler/Transformations/Monomorphization.cs
bettinaheim: Basically the public generic will be stripped if trimming is enabled (it is by default). What my changes did is precisely making sure that if trimming is disabled for monomorphization, the original generic should just be preserved, and the monomorphized versions will be added with (at most) internal access only (i.e. monormophization no longer silently trims only a part of the trimmable things, which previously resulted in the C# generation bug).

src/QsCompiler/CSharpGeneration/SimulationCode.fs
bettinaheim: Well, if you manually use/invoke a newer version of the C# generation with an older version of the Sdk, then yes, this will technically be breaking. However, we have a couple of breaking APIs with this month anyway, so I think using a newer version of the C# generation with an older Sdk version likely will not work for other reasons. The Sdk translates both an ExecutionTarget that contains honeywell or quantinuum to a ResolvedProcessorArchitecture of QuantinuumProcessor. I think it is reasonable to rely on that change here.

src/QsCompiler/Compiler/CompilationLoader.cs
bettinaheim: Hm, that was meant to be captured by the second half of the comment: "With the new setup (interface-based approach for target packages), it is possible to trim unused intrinsics." I edited the second half in an attempt to make it clearer: Intrinsics that come from packages that use the new interface-based approach (which is the case for target packages) can be trimmed if they are unused. Honestly, I was hoping at some point we would get to a place where everything uses interfaces...

src/QsCompiler/Core/SyntaxGenerator.fs
bettinaheim: Yes, quantum dependency is pessimistic. It is used for generating adjoints, and in that sense the info merely captures if something can be dependent on side effects. So, this is basically the definition of quantum dependency - it has nothing to do with qubits, and instead merely tracks if something has a direct or indirect dependency on an operation call. I think the reason this was never an issue is likely that the syntax generator is primarily used for code for which functor support has already been generated (e.g. this is used for the generated functor specializations themselves).

src/QsCompiler/DataStructures/ReservedKeywords.fs
bettinaheim: Ah, you are right - good catch. Since this is an undocumented feature used as part of our internal infrastructure and we already have other breaking changes, I am inclined to just remove it.

src/QsCompiler/CSharpGeneration/CSharpGeneration.fsproj
bettinaheim: That's a good question. I generally haven't found a really good answer for where to look for msbuild features (there is some good docs, but rarely for the things I was looking for). On this particular topic, I would search online for defining InternalsVisibleTo (or any assembly attribute) in the project file, and might e.g. find the [blog post](https://www.meziantou.net/declaring-internalsvisibleto-in-the-csproj.htm)
or the [StackOverflow answer](https://stackoverflow.com/questions/42810705/visual-studio-2017-and-the-new-csproj-internalsvisibleto), which by the way has a comment "Not sure this approach is documented since it is rather customization than common practice."

src/QsCompiler/QirGeneration/AddOutputRecording.cs
bettinaheim: I move it out of the function and made it a private static member instead.

src/QsCompiler/QirGeneration/AddOutputRecording.cs
bettinaheim: As discussed offline, I can't. This name has to be predictable due to being an entry point. I don't like it as is, and I'll think if I can come up with something cleaner (hence the todo in the code).

src/QsCompiler/QirGeneration/QIR/DataStructures.cs
bettinaheim: Removed that line in particular, but I think I meant to make a general pass to update the doc comment to reflect my changes in the QIR generation. I'll go do that.

src/QsCompiler/QirGeneration/QIR/Values.cs
bettinaheim: I meant to also either completely disable any memory management for QIR targets, or (which I prefer since it is cleaner and needed for using the new data structures independently of the target capability) update the memory management to actually properly work even when somethings are stack allocated.

src/QsCompiler/QirGeneration/AddOutputRecording.cs
bettinaheim: Mimicking the SpecializationSuffix in InferTargetInstructions, I decided to make this an argument to the AddOutputRecording that is defined by the QIR generation, cleaning up (meaning removing) the hardcoded dependency.

src/QsCompiler/QirGeneration/AddOutputRecording.cs
```diff
@@ -0,0 +1,572 @@
...
+    using ParameterTuple = QsTuple<LocalVariableDeclaration<QsLocalSymbol, ResolvedType>>;
+    using ResolvedTypeKind = QsTypeKind<ResolvedType, UserDefinedType, QsTypeParameter, CallableInformation>;
+
+    /// <summary>
+    /// This transformation replaces callables with type parameters with concrete
+    /// instances of the same callables. The concrete values for the type parameters
+    /// are found from uses of the callables.
+    /// This transformation also removes all callables that are not used directly or
+    /// indirectly from any of the marked entry point.
+    /// Monomorphizing intrinsic callables is optional and intrinsics can be prevented
```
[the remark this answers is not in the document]
bettinaheim: I added the proper summary.

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestReferenceCounts2.ll
bettinaheim: This is an artifact of how if statements are handled. Basically we add the block before we know it is needed - now I can't remember exactly the details, but if I remember correctly this was contributed by someone else as part of a bug fix, and it indeed seems the simplest and most robust solution.

src/QsCompiler/QirGeneration/Context.cs
```diff
@@ -177,18 +178,19 @@ internal void EndLoop()
         /// </summary>
         /// <param name="syntaxTree">The syntax tree for which QIR is generated.</param>
         /// <param name="isLibrary">Whether the current compilation is being performed for a library.</param>
-        internal GenerationContext(IEnumerable<QsNamespace> syntaxTree, bool isLibrary)
+        internal GenerationContext(IEnumerable<QsNamespace> syntaxTree, bool isLibrary, TargetCapability capability)
```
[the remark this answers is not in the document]
bettinaheim: Added.

src/QuantumSdk/DefaultItems/DefaultItems.targets
bettinaheim: I edited the Sdk and I believe with the latest changes the scenario you are describing would now indeed be supported.

PR #1462, opened 2022-06-10
Update repository to .NET 6.0.30x and address build issues

ricardo-espinoza:


build/ci.yml
```diff
@@ -48,9 +48,9 @@ jobs:
 - job: Style
   steps:
   - task: UseDotNet@2
-    displayName: Use .NET SDK 6.0.105
+    displayName: Use .NET SDK 6.0.300
     inputs:
-      version: 6.0.105
+      useGlobalJson: true
```
swernli, line 53: I'm a little confused here, this indicates that the version should be taken from the global json but that file is not updated in this PR and still has version `6.0.100` in main. Should that file also be updated here?

ricardo-espinoza: @swernli, the global.json is updated as part of this PR to 6.0.300

swernli: oh, I see it now! Apologies for the noise, somehow that file was minimized in my web view :).

cgranade (approved):

PR #1466, opened 2022-06-15
Update Fantomas version to 4.7.9

samarsha:
Update Fantomas and re-format all F# files. I fixed a mistake in .editorconfig where our style settings weren't applying to .fsi files, which is why some .fsi files have a lot of changes here.


swernli (approved):
Review the configuration changes, assuming that the rest of the changes were purely fantomas format updates.

PR #1468, opened 2022-06-16
Lint F# projects with FSharpLint

### LINKED ISSUE #1310: Add F# linter to build pipeline
While we use Fantomas for formatting F# code, this doesn't enforce all F# style conventions (for example, naming conventions, `new` keywords, etc.). We should consider adding an F# linter to the pipeline to enforce a consistent style and improve code readability.

samarsha:
Resolves #1310.

The important files changed are:

- .config/dotnet-tools.json
- build/ci.yml
- fsharplint.json

The rest of the files are only changed to fix lints. This is mostly renaming things to match the naming conventions, and some minor refactoring. I excluded projects that I consider to be legacy (not expected to be updated often, possibly removed soon) from the linter.

FSharpLint has a slightly weird config file, where [you have to copy the complete default config into your repo](https://fsprojects.github.io/FSharpLint/how-tos/rule-configuration.html). I only changed a few things from the defaults.


ScottCarda-MS:
@samarsha Could you add a link to the complete list of F# style rules enforced by the new linter?

samarsha:
All of the rules are here: https://fsprojects.github.io/FSharpLint/how-tos/rule-configuration.html

But many of them are disabled by default. Only rules with `"enabled": true` in [fsharplint.json](https://github.com/microsoft/qsharp-compiler/blob/samarsha/fsharplint/fsharplint.json) are enforced.

cesarzc:
In the local development workflow, we would run `dotnet tool run dotnet-fsharplint lint QsCompiler.sln` to check that linting runs successfully and to fix any lint issues we might have introduced?

If that is the case, it would be good to add a short paragraph in the [Building From Source](https://github.com/microsoft/qsharp-compiler/blob/main/README.md#building-from-source) section of the README explaining this.

src/QsCompiler/Core/SyntaxGenerator.fs
```diff
@@ -35,17 +35,15 @@ and private StripPositionInfoFromNamespace(parent: StripPositionInfo) =
     inherit NamespaceTransformation(parent)
     override this.OnLocation _ = Null
 
-and public StripPositionInfo private (_internal_) =
+and public StripPositionInfo() as this =
```
cesarzc, line 38: With this change, `StripPositionInfo()` now becomes a function that returns an instance of the class type `StripPositionInfo` and the `do` bindings are used to initialize the members inherited from `SyntaxTreeTransformation`? Am I understanding this change correctly?

src/QsCompiler/Tests.Compiler/RegexTests.fs
```diff
@@ -77,24 +77,20 @@ let ``Strip unique variable name resolution`` () =
             "'TName"
         ]
 
-    origNames
-    |> List.map (fun var -> var, UniqueVariableNames.StripUniqueName var)
-    |> List.iter Assert.Equal
+    for var in origNames do
```
cesarzc, line 80: Liked this change! The code is easier to read now :)

cesarzc (approved):

src/QsCompiler/Tests.Compiler/RegexTests.fs
samarsha: Yeah, the linter complained about repeated calls to List.map that could be combined, but I realized it was simpler to rewrite it as a for loop.

src/QsCompiler/Core/SyntaxGenerator.fs
samarsha: The API of StripPositionInfo is the same before and after the change - it's a class with a public parameterless constructor. I removed the additional private constructor that took a parameter named `_internal_`, because this constructor actually isn't necessary (I'm not sure why it was there to begin with).

samarsha:
> In the local development workflow, we would run `dotnet tool run dotnet-fsharplint lint QsCompiler.sln` to check that linting runs successfully and to fix any lint issues we might have introduced?

Yes - although linting the whole solution can take a while, so you can also run it on specific projects or files.

> If that is the case, it would be good to add a short paragraph in the [Building From Source](https://github.com/microsoft/qsharp-compiler/blob/main/README.md#building-from-source) section of the README explaining this.

There are some instructions for how to use Fantomas in CONTRIBUTING.md, so I added instructions for FSharpLint to there as well.

PR #1472, opened 2022-06-20
Use logical shift right instead of arithmetic

swernli:
This changes the use of bit shifts to be logical instead of arithemtic, such that they will no longer preserve the sign bit. It does not yet update the C# generation to do the same.


src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
```diff
@@ -2260,7 +2260,7 @@ public override ResolvedExpressionKind OnRightShift(TypedExpression lhsEx, Typed
             IValue value;
             if (exType.Resolution.IsInt)
             {
-                var res = this.SharedState.CurrentBuilder.ArithmeticShiftRight(lhs.Value, rhs.Value);
+                var res = this.SharedState.CurrentBuilder.LogicalShiftRight(lhs.Value, rhs.Value);
```
bettinaheim, line 2263: I see above that for the left shift there is only the direct use of left shift, which I believe should give us the right behavior. Could you confirm?

src/QsCompiler/RoslynWrapper/ExpressionStatement.fs
```diff
@@ -164,6 +164,9 @@ module Expressions =
 
     // left >>> right
     let (.>>>.) left right =
+        // TODO: Should be replaced with a use of SyntaxFactor.Token to generate an unsigned shift right. For reference,
```
bettinaheim, line 167: Ech, what is the current behavior of the C# code here - arithmetic shift in both directions?

bettinaheim, line 167:
```suggestion
        // TODO: Should be replaced with a use of SyntaxFactory.Token to generate an unsigned shift right. For reference,
```

src/QsCompiler/QirGeneration/Subtransformations/ExpressionKindTransformation.cs
swernli: Yes, that's correct. The shift left instruction does not have a variant, only `shl` that we are already using.

src/QsCompiler/RoslynWrapper/ExpressionStatement.fs
swernli: C# uses shift left (which is always effectively unsigned) and arithmetic shift right. They introduced a new operator for logical shift right (`>>>` for `lshr` and `>>` for `ashr`). There is an open issue against Roslyn to add support for this (see https://github.com/dotnet/roslyn/issues/61854).

swernli: I'll fix this and add a link to the Rosyln issue.

swernli: While it's not required for the QIR fix, we should try to resolve this soon. Without fixing the C# gen part, a user might observe difference behavior between local simulation via C# and QIR simulation (though only when right shifting negative numbers). This example from the C# docs captures it well:
```csharp
int x = -8;
Console.WriteLine($"Before:    {x,11}, hex: {x,8:x}, binary: {Convert.ToString(x, toBase: 2),33}");

int y = x >> 2;
Console.WriteLine($"After  >>: {y,11}, hex: {y,8:x}, binary: {Convert.ToString(y, toBase: 2),33}");

int z = x >>> 2;
Console.WriteLine($"After >>>: {z,11}, hex: {z,8:x}, binary: {Convert.ToString(z, toBase: 2),33}");
// Output:
// Before:             -8, hex: fffffff8, binary:  11111111111111111111111111111000
// After  >>:          -2, hex: fffffffe, binary:  11111111111111111111111111111110
// After >>>:  1073741822, hex: 3ffffffe, binary:    111111111111111111111111111110
```

bettinaheim: Yes, agreed. It is for me plain odd to have the left shift be effectively logical and the right shift be arithmetic.

PR #1477, opened 2022-06-27
Fixed Issue with Range Slicing Bug

### LINKED ISSUE #1430: Open ended array slices are running into a type inference issue
The following code results in a compiler exception from [this location](https://github.com/microsoft/qsharp-compiler/blob/f92292d6892132378cabbbb0c5113f39c76a1b1d/src/QsCompiler/Core/SyntaxGenerator.fs#L332).

```qsharp
        let arr1 = [1,2,3];
        let arr2 = [6, size = 3];

        let concatenated = arr1 + arr2;
        let slice = concatenated[arr1[0]..arr1[1]...];
```
A similar issue may exist for other open ended ranges (as part of slicing or as part of copy-and-update).

Edit: Capturing another case that seems to silently fail here since it is also an issue with open ended ranges; this may or may not be the same/a similar issue. This is the case that throws during building the compilation:
```qsharp
    operation LengthCaching (vals : Int[]) : Int[] {
        return vals[...vals[0]...];
    }
```

ScottCarda-MS:
Fixes #1430

### DIFF
--- a/src/VSCodeExtension/package-lock.json.v.template
+++ b/src/VSCodeExtension/package-lock.json.v.template
@@ -10,6 +10,7 @@
             "license": "MIT",
             "dependencies": {
                 "@types/fs-extra": "^8.0.0",
+                "@vscode/extension-telemetry": "0.6.2",
                 "decompress-zip": "^0.2.2",
                 "fs-extra": "^8.1.0",
                 "lodash": "^4.17.21",
@@ -17,7 +18,6 @@
                 "semver": "^6.0.0",
                 "tmp": "0.0.33",
                 "ts-optchain": "^0.1.7",
-                "vscode-extension-telemetry": "0.0.18",
                 "vscode-languageclient": "7.0.0",
                 "which": "1.3.1",
                 "yeoman-environment": "^3.4.1",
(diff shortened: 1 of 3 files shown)


src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
```diff
@@ -434,13 +434,15 @@ type QsExpression with
             let openStartInSlicing =
                 function
                 | Some step when Some step |> validSlicing ->
-                    conditionalIntExpr (IsNegative step) (LengthMinusOne array) (SyntaxGenerator.IntLiteral 0L)
+                    let inferredStep = { step with ResolvedType = Int |> ResolvedType.New } // validSlicing ensures that the type must be an Int
```
samarsha, line 437: I would recommend moving the call to `inference.Resolve` out of `validSlicing` and instead have `integerExpr` return:

```fsharp
{ ex with ResolvedType = inference.Resolve ex.ResolvedType }
```

Then `validSlicing`, `openStartInSlicing`, and `openEndInSlicing` don't need to worry about resolving the type.

ScottCarda-MS: I'm not sure what all that line affects, so let me know if it can be removed or not, please.

src/QsCompiler/SyntaxProcessor/ExpressionVerification.fs
```diff
@@ -446,8 +445,7 @@ type QsExpression with
             let resolveSlicingRange start step end_ =
                 let integerExpr ex =
                     let ex = resolve context ex
-                    inference.Constrain(ResolvedType.New Int .> ex.ResolvedType) |> List.iter diagnose
-                    ex
+                    { ex with ResolvedType = inference.Resolve ex.ResolvedType }
```
samarsha, line 448: Can you rename this function since it isn't related to integers anymore?

ScottCarda-MS: Oh yah I guess this function's purpose changed. Renamed to `toResolvedExpr`.

samarsha (approved):

PR #1479, opened 2022-07-07
Job submission

jmonas:
job submission, listing, and pulling of results


global.json
```diff
@@ -1,6 +1,6 @@
 {
     "sdk": {
-        "version": "6.0.100",
+        "version": "6.0.300",
```
ricardo-espinoza, line 3: We're in the process of taking this change in another change. Do you need 6.0.300 in order for the changes to work? In any case, you should exclude this from the PR, even if locally you're using 6.0.300 for testing.

src/VSCodeExtension/package.json.v.template
```diff
@@ -1,6 +1,6 @@
 {
     "name": "quantum-devkit-vscode",
-    "version": "#VSCODEVSIX_VERSION#",
+    "version": "0.9999.22060152",
```
ricardo-espinoza, line 3: Make sure that your v.template file keeps the #VSCODEVSIX_VERSION#. We can chat offline if you have questions about how this works.

src/VSCodeExtension/package.json.v.template
```diff
@@ -74,7 +119,7 @@
                         "null"
                     ],
                     "description": "Path to the language server assembly for Q#.",
-                    "default": null
+                    "default": "C:/Users/t-jackmonas/work/qsharp-compiler/src/QsCompiler/LanguageServer/bin/Debug/net6.0/win10-x64/publish/Microsoft.Quantum.QsLanguageServer.exe"
```
ricardo-espinoza, line 122: Don't include this in the PR. It makes sense to change locally while you're developing, but it will help us avoid an accident if this never makes it to the repo.

src/VSCodeExtension/src/utils/excludedProviders.json
```diff
@@ -0,0 +1,3 @@
+{
+    "optimizationProviders":["Microsoft"]
```
ricardo-espinoza, line 2: It would be better to exclude targets than providers, since Microsoft could later on add a QC target, or another provider could add an optimization one.

ricardo-espinoza (changes requested):

global.json
jmonas: I forgot to change back when troubleshooting job submission issues. Changes will work with 6.0.100

Issue #1470, opened 2022-06-17
Unable to load projects on .NET SDK 6.0.301 due to NuGet.Frameworks load errors


baronfel:
**Describe the bug**

The language server extension bundles a version of NuGet dlls that is incompatible with the ones shipped in the SDK. Because MSbuild Locator is used to locate and run MSBuild DLLs from the SDK, there are version mismatches that lead to errors running targets in the build.

This was initially reported at https://github.com/dotnet/sdk/issues/26065, with detailed resolution steps [here](https://github.com/dotnet/sdk/issues/26065#issuecomment-1159051054).

**To Reproduce**

Please see the linked issue for reproduction steps.

**Expected behavior**

The language server should be able to load projects.

**System information**

- Version of the NuGet package or extension for which the problem occurs: v0.24.210930
- Your operating system and .NET Core version: Ubuntu 22.04, .NET SDK 6.0.301
- For issues with extensions, the version of the IDE for which the problem occurs:  Visual Studio Code version 1.68.1

**Additional context**

The short form of the resolution is that the NuGet.Frameworks PackageReference in the LanguageServer should be marked `ExcludeAssets="runtime"`.

bettinaheim:
@ricardo-espinoza Could you take a look at this?

ausbin:
I just tested and #1467 seems to resolve this. With the current Quantum SDK release (0.24.210930), I see the same behavior reported in dotnet/sdk#26065. But testing on the branch `release/v0.25.2206` by building the QsCompiler solution locally and using the `<QscExe>` setting in the project file and `quantumDevKit.languageServerPath` in my VSCode settings, I don't see the error spam in VSCode or when running `dotnet build`.

baronfel:
That PR looks like it would resolve the issue for now, but I would expect it to come back again with other SDK version updates. In my opinion (and that of other MSBuild team members), the correct fix long term is to update your PackageReference with `ExcludeAssets="runtime"`.

ausbin:
Thanks for that suggestion, @baronfel. Could you share what that means to folks such as me who aren't msbuild gods?

bettinaheim:
@baronfel I believe we are already setting ExcludeAssets="runtime" for all relevant ms build or nuget dependencies. We don't currently have an explicit reference to NuGet.Frameworks. Could you clarify your suggestion?

baronfel:
[...]

PR #1481, opened 2022-07-12
Update syntax checking to require no namespace{}s in notebooks

ausbin:
This is a smaller PR for syntax checking (no semantic checking) for `namespace{}` blocks in notebooks, because #1457 was getting out of control.

Magic commands are broken in this code, but that is out of scope for this PR. The goal here is to get this code for syntax checking for `namespace{}` blocks (which hasn't changed too much in several weeks) reviewed and merged into the feature branch.

## Testing
* Ran unit tests, which includes some tests I added for both the compiler and the language server
* Did manual test locally with Azure Notebooks (with [this notebook](https://github.com/microsoft/Quantum/blob/feature/samples-gallery/samples/azure-quantum/parallel-qrng/ParallelQrng.ipynb)). As expected, the language server returns angry diagnostics for magic commands, but other syntax errors are caught

### DIFF
--- a/src/QsCompiler/CompilationManager/CompilationUnitManager.cs
+++ b/src/QsCompiler/CompilationManager/CompilationUnitManager.cs
@@ -31,6 +31,8 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
     {
         internal bool EnableVerification { get; private set; }
 
+        public BuildConfiguration BuildConfiguration { get; }
+
         internal ProjectProperties BuildProperties => this.compilationUnit.BuildProperties;
 
         /// <summary>
@@ -86,7 +88,8 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
             Action<Exception>? exceptionLogger = null,
             Action<string, MessageType>? log = null,
             Action<PublishDiagnosticParams>? publishDiagnostics = null,
-            bool syntaxCheckOnly = false)
+            bool syntaxCheckOnly = false,
+            BuildConfiguration? buildConfiguration = null)
         {
             this.EnableVerification = !syntaxCheckOnly;
             this.compilationUnit = new CompilationUnit(buildProperties);
@@ -95,6 +98,7 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
             this.PublishDiagnostics = publishDiagnostics ?? (_ => { });
             this.Log = log ?? ((_, __) => { });
             this.LogException = exceptionLogger ?? Console.Error.WriteLine;
+            this.BuildConfiguration = buildConfiguration ?? BuildConfiguration.DefaultConfiguration();
             this.Processing = new ProcessingQueue(this.LogException);
             this.waitForTypeCheck = new CancellationTokenSource();
         }
@@ -257,9 +261,10 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
             Uri uri,
             string fileContent,
             Action<PublishDiagnosticParams>? publishDiagnostics = null,
-            Action<Exception>? onException = null)
+            Action<Exception>? onException = null,
+            BuildConfiguration? buildConfiguration = null)
         {
-            var file = new FileContentManager(uri, GetFileId(uri));
+            var file = new FileContentManager(uri, GetFileId(uri), buildConfiguration);
             try
             {
                 file.ReplaceFileContent(fileContent);
@@ -283,7 +288,8 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
         public static ImmutableHashSet<FileContentManager> InitializeFileManagers(
             IDictionary<Uri, string> files,
             Action<PublishDiagnosticParams>? publishDiagnostics = null,
-            Action<Exception>? onException = null)
+            Action<Exception>? onException = null,
+            BuildConfiguration? buildConfiguration = null)
         {
             if (files.Any(item => !item.Key.IsAbsoluteUri || !item.Key.IsFile))
             {
@@ -293,7 +299,7 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
             return files.AsParallel()
                 .WithDegreeOfParallelism(Environment.ProcessorCount > 1 ? Environment.ProcessorCount - 1 : Environment.ProcessorCount)
                 .WithExecutionMode(ParallelExecutionMode.ForceParallelism) // we are fine with a slower performance if the work is trivial
-                .Select(entry => InitializeFileManager(entry.Key, entry.Value, publishDiagnostics, onException))
+                .Select(entry => InitializeFileManager(entry.Key, entry.Value, publishDiagnostics, onException, buildConfiguration))
                 .ToImmutableHashSet();
         }
 
--- a/src/QsCompiler/CompilationManager/CompilationUnitManager.cs
+++ b/src/QsCompiler/CompilationManager/CompilationUnitManager.cs
@@ -31,8 +31,6 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
     {
         internal bool EnableVerification { get; private set; }
 
-        public BuildConfiguration BuildConfiguration { get; }
-
         internal ProjectProperties BuildProperties => this.compilationUnit.BuildProperties;
 
         /// <summary>
@@ -88,8 +86,7 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
             Action<Exception>? exceptionLogger = null,
             Action<string, MessageType>? log = null,
             Action<PublishDiagnosticParams>? publishDiagnostics = null,
-            bool syntaxCheckOnly = false,
-            BuildConfiguration? buildConfiguration = null)
+            bool syntaxCheckOnly = false)
         {
             this.EnableVerification = !syntaxCheckOnly;
             this.compilationUnit = new CompilationUnit(buildProperties);
@@ -98,7 +95,6 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
             this.PublishDiagnostics = publishDiagnostics ?? (_ => { });
             this.Log = log ?? ((_, __) => { });
             this.LogException = exceptionLogger ?? Console.Error.WriteLine;
-            this.BuildConfiguration = buildConfiguration ?? BuildConfiguration.DefaultConfiguration();
             this.Processing = new ProcessingQueue(this.LogException);
             this.waitForTypeCheck = new CancellationTokenSource();
         }
@@ -262,9 +258,9 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
             string fileContent,
             Action<PublishDiagnosticParams>? publishDiagnostics = null,
             Action<Exception>? onException = null,
-            BuildConfiguration? buildConfiguration = null)
+            bool isNotebook = false)
         {
-            var file = new FileContentManager(uri, GetFileId(uri), buildConfiguration);
+            var file = new FileContentManager(uri, GetFileId(uri), isNotebook);
             try
             {
                 file.ReplaceFileContent(fileContent);
@@ -288,8 +284,7 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
         public static ImmutableHashSet<FileContentManager> InitializeFileManagers(
             IDictionary<Uri, string> files,
             Action<PublishDiagnosticParams>? publishDiagnostics = null,
-            Action<Exception>? onException = null,
-            BuildConfiguration? buildConfiguration = null)
+            Action<Exception>? onException = null)
         {
             if (files.Any(item => !item.Key.IsAbsoluteUri || !item.Key.IsFile))
             {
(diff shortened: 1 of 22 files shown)


src/QsCompiler/DataStructures/ReservedKeywords.fs
```diff
@@ -198,6 +198,9 @@ module InternalUse =
     /// to be used as name for all namespaces that do not have a valid name
     let UnknownNamespace = "__UnknownNamespaceName__"
 
+    /// the implicit namespace in notebooks
+    let NotebookNamespace = "__NotebookNamespace__"
+
```
ausbin, line 203: Carrying this comment from @bettinaheim over so it doesn't get forgotten: https://github.com/microsoft/qsharp-compiler/pull/1457#discussion_r907866459

src/QsCompiler/CompilationManager/ProjectManager.cs
```diff
@@ -793,11 +826,12 @@ public bool ManagerTask(Uri file, Action<CompilationUnitManager> executeTask, ID
...
             Action<string, MessageType>? log = null,
             Action<PublishDiagnosticParams>? publishDiagnostics = null,
-            SendTelemetryHandler? sendTelemetry = null)
+            SendTelemetryHandler? sendTelemetry = null,
+            BuildConfiguration? buildConfiguration = null)
         {
             this.load = new ProcessingQueue(exceptionLogger);
             this.projects = new ConcurrentDictionary<Uri, Project>();
-            this.defaultManager = new CompilationUnitManager(ProjectProperties.Empty, exceptionLogger, log, publishDiagnostics, syntaxCheckOnly: true);
+            this.defaultManager = new CompilationUnitManager(ProjectProperties.Empty, exceptionLogger, log, publishDiagnostics, syntaxCheckOnly: true, buildConfiguration);
```
ausbin, line 834: Carrying this comment over too: https://github.com/microsoft/qsharp-compiler/pull/1457#discussion_r907862698

I'm not sure about it, though, since `BuildConfiguration` is maintained by the `EditorState` and passed to the `ProjectManager`. So would the `ProjectManager` create a new `BuildConfiguration` with `syntaxCheckOnly=true`? It's possible but I am struggling to see the benefit

src/QsCompiler/Tests.Compiler/TestCases/StringParsingTests/StringParsing.qs
```diff
@@ -32,6 +32,14 @@ namespace Microsoft.Quantum.Testing.LocalVerification {
         let str = "//";
     }
 
+    operation NotebookStringParsingTest1 () : Unit {
+        let str = "hi"; %
+    }
+
+    operation NotebookStringParsingTest2 () : Unit {
+%simulate SampleRandomNumber nQubits=3
+    }
```
swernli, line 41: These two test cases seem like they should work as-is, but they are also pretty tied to the magic command parsing that is delayed now. Would it be worth pulling them out for now in case the nature or strategy of the test changes once the feature is pulled back in?

ausbin: Thanks, I'll nuke this for now

src/QsCompiler/LanguageServer/LanguageServer.cs
```diff
@@ -764,6 +764,7 @@ public object OnCodeAction(JToken arg)
                 return
                     param.Command == CommandIds.ApplyEdit ? CastAndExecute<ApplyWorkspaceEditParams>(edit =>
                         this.rpc.InvokeWithParameterObjectAsync<ApplyWorkspaceEditResponse>(Methods.WorkspaceApplyEditName, edit)) :
+                    param.Command == CommandIds.FileIsNotebookCell ? CastAndExecute<TextDocumentIdentifier>(this.editorState.FileIsNotebookCell) :
```
ausbin, line 767: I can't decide if this is goofy or not. The code to get this to work is pretty ugly thanks to [C# limitations][1], and I'm not sure it contributes much to the unit tests where it's used. But it took me long enough to get it to work that I included it anyway

[1]: https://devblogs.microsoft.com/dotnet/try-out-nullable-reference-types/#the-issue-with-t

src/QsCompiler/LanguageServer/EditorState.cs
```diff
@@ -261,11 +261,15 @@ void AddProperty(IDictionary<string, string?> props, string property, params str
                     return;
                 }
 
-                var newManager = CompilationUnitManager.InitializeFileManager(textDocument.Uri, textDocument.Text, this.publish, ex =>
+                bool isNotebook = textDocument.LanguageId.Contains("notebook");
```
ausbin, line 264: I feel like this string should be a constant, but not sure where to put it. Also, I'm not sure if this check is robust enough. I'll send y'all the accompanying code on the Azure Notebooks side (it's very simple)

ausbin:
Posting another revision based on what we talked about today. Tested again manually with Azure Notebooks and running `dotnet test`.

I am hitting #1470 again locally with .NET SDK 6.0.302, so I had to manually lock my local `global.json`(s) at 6.0.301. Oh me.

swernli (approved):
Changes look good! Glad you worked out the streamlined approach.

src/QsCompiler/CompilationManager/CompilationUnitManager.cs
```diff
@@ -941,6 +942,23 @@ private void TypeCheckFile(string fileId, CancellationToken cancellationToken)
...
+        /// </returns>
+        /// <remarks>
+        /// Waits for all currently running or queued tasks to finish
+        /// before getting the file content by calling <see cref="FlushAndExecute"/>.
+        /// </remarks>
+        public object? FileIsNotebookCell(TextDocumentIdentifier textDocument) =>
+
+            // Boxing needed here because FileQuery() is generic:
+            // https://devblogs.microsoft.com/dotnet/try-out-nullable-reference-types/#the-issue-with-t
+            this.FlushAndExecute(() =>
```
samarsha, line 959: I would recommend unboxing before returning to keep the public API clean. Alternatively, if you update the project to use C# 10, you can change `FlushAndExecute` and `FileQuery` to use `T?` which would allow you to use `bool` without boxing.

src/QsCompiler/CompilationManager/ContextBuilder.cs
```diff
@@ -534,8 +535,12 @@ internal static IEnumerable<(Range, QsQualifiedName)> CallablesWithContentModifi
...
             {
                 var ns = file.TryGetNamespaceAt(tuple.Item2.Start);
-                QsCompilerError.Verify(ns != null, "namespace for callable declaration should not be null"); // invalid namespace names default to an unknown namespace name, but remain included in the compilation
-                return (tuple.Item2.Start, new QsQualifiedName(ns, tuple.Item1));
+
+                // invalid namespace names default to an unknown namespace name, but remain included in the compilation
+                QsCompilerError.Verify(
+                    file.IsNotebook == (ns == null),
+                    "namespace for callable declaration should be null if and only if this is a notebook");
+                return (tuple.Item2.Start, new QsQualifiedName(ns ?? InternalUse.NotebookNamespace, tuple.Item1));
```
samarsha, line 543: Is it possible to insert the fake namespace in an earlier stage? It seems like it would be a little nicer if "invalid namespace name" and "implicit notebook namespace" were handled in the same place, instead of notebook namespaces propagating up here.

src/QsCompiler/DataStructures/ReservedKeywords.fs
```diff
@@ -198,6 +198,9 @@ module InternalUse =
     /// to be used as name for all namespaces that do not have a valid name
     let UnknownNamespace = "__UnknownNamespaceName__"
 
+    /// the implicit namespace in notebooks
+    let NotebookNamespace = "__NotebookNamespace__"
```
samarsha, line 202: nit: I kind of prefer just `__Notebook__` and `__Unknown__`, since by context we know they are namespaces.

src/QsCompiler/CompilationManager/ProjectManager.cs
```diff
@@ -1431,6 +1431,35 @@ editCollection switch
...
+        /// Returns a boxed bool (or null) representing if <paramref name="textDocument"/>
+        /// is a notebook cell or not.
+        /// </summary>
+        /// <remarks>
+        /// Returns null if the given file is null.
+        /// <para/>
+        /// This method waits for all currently running or queued tasks to finish
+        /// before getting the file content.
+        /// </remarks>
+        public object? FileIsNotebookCell(TextDocumentIdentifier textDocument)
```
samarsha, line 1444: I think this boxing will also be removed if you change `FileIsNotebookCell` like in my previous comment.

src/QsCompiler/LanguageServer/EditorState.cs
samarsha: Even though we don't check the language ID anywhere else, I think I'd still feel more comfortable with a strict equality check when we do check it.

src/QsCompiler/DataStructures/ReservedKeywords.fs
samarsha: We could have only one namespace name, but I'd want to rename it to something more neutral than "Unknown," like "Default" or "Main." But I think "Notebook" is probably least confusing...

src/QsCompiler/SyntaxProcessor/ContextVerification.fs
```diff
@@ -257,15 +259,15 @@ let private followedByApply context =
     | _ -> false, [| (ErrorCode.MissingContinuationApply |> Error, context.Range) |]
 
 
-type ContextVerification = delegate of SyntaxTokenContext -> (bool * QsCompilerDiagnostic [])
+type ContextVerification = delegate of SyntaxTokenContext * bool -> (bool * QsCompilerDiagnostic [])
```
samarsha, line 262: I think there is some boolean blindness, maybe a new type would be clearer. Something like:

```fsharp
type DocumentKind = File | Notebook
```

src/QsCompiler/CompilationManager/CompilationUnitManager.cs
ausbin: I couldn't get it to cooperate on C# 10, but casting to `bool?` is less nasty, thanks

src/QsCompiler/Core/Transformations/SyntaxTreeTransformation.fs
```diff
@@ -262,6 +263,7 @@ type SyntaxTreeTransformation(options) as this =
 
     /// Invokes the transformation for all namespaces in the given compilation.
     abstract OnCompilation: compilation: QsCompilation -> QsCompilation
+
```
ausbin, line 266: This is from me running Fantomas, for the record. (I am not losing my mind and adding empty lines to random files)

samarsha: Usually that means your Fantomas version is different from what the repo specifies? There are only a few extra changes though so it's not a big deal.

src/QsCompiler/LanguageServer/LanguageServer.cs
samarsha: I think it's ok to have.

src/QsCompiler/Core/Transformations/SyntaxTreeTransformation.fs
ausbin: After I run `dotnet tool restore`, I get the following version of Fantomas:

```
$ dotnet tool run fantomas --version
Fantomas v4.7.9+cbaa31cc6977666ab1103b1ee49aff31b85e39d8
```

That seems to match the 4.7.9 version in `.config/dotnet-tools.json`, but I am just a humble Boy Scout

ausbin:
Rebased this against `main` and #1462 has left the language server tests in a broken state. To test this locally, I applied the following hack:

```
diff --git a/src/QsCompiler/LanguageServer/global.language-server.json b/src/QsCompiler/LanguageServer/global.language-server.json
index 3b31a29a..2a1043b5 100644
--- a/src/QsCompiler/LanguageServer/global.language-server.json
+++ b/src/QsCompiler/LanguageServer/global.language-server.json
@@ -1,6 +1,5 @@
 {
   "sdk": {
-    "version": "6.0.302",
-    "rollForward": "latestPatch"
+    "version": "6.0.301"
   }
 }
```

src/QsCompiler/LanguageServer/LanguageServer.cs
ausbin: After re-plumbing DocumentKind everywhere, it feels less disturbing thankfully

ausbin:
@filipw I would call this a hack at the moment, specifically to make Q# IntelliSense work on Azure Notebooks.

We have discussed adding this to the Q# spec as "REPL mode", but that's out of my pay grade. The language changes would be:

* Treating magic commands (e.g., `%simulate`) as comments, or maybe even parsing them (undecided)
* Banning `namespace{}` blocks in favor of an implicit namespace
* `open` directives scoping to this entire implicit namespace instead of just the cell where they're executed

src/QsCompiler/CompilationManager/CompilationUnitManager.cs
```diff
@@ -941,6 +943,25 @@ private void TypeCheckFile(string fileId, CancellationToken cancellationToken)
...
+        /// </returns>
+        /// <remarks>
+        /// Waits for all currently running or queued tasks to finish
+        /// before getting the file content by calling <see cref="FlushAndExecute"/>.
+        /// </remarks>
+        public DocumentKind? FileDocumentKind(TextDocumentIdentifier textDocument) =>
+            this.FlushAndExecute(() =>
+
+                // Boxing needed here because FileQuery() is generic:
+                // https://devblogs.microsoft.com/dotnet/try-out-nullable-reference-types/#the-issue-with-t
```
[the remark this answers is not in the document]
ausbin: Oops, thanks

src/QsCompiler/CompilationManager/ProjectManager.cs
samarsha: Is this comment moot now? It doesn't look like this line is changed in the current version anymore.

samarsha (approved):

src/QsCompiler/CompilationManager/ProjectManager.cs
ausbin: Yeah, when changing isNotebook to be per-file, I sent `BuildConfiguration` to the farm upstate

PR #1483, opened 2022-07-12
Remove `waitForInit` from LanguageServer

swernli:
This change simplifies the synchronization of the LanguageServer by removing the `waitForInit` ManualResetEvent. Since LSP clients should not send further messages until they have received the response to init and the change adjusts the rpc listener to not connect until after the object is constructed, the synchronization should no longer be needed.


src/QsCompiler/LanguageServer/LanguageServer.cs
```diff
@@ -356,11 +344,6 @@ public void OnInitialized(JToken token)
         [JsonRpcMethod(Methods.TextDocumentDidOpenName)]
         public Task OnTextDocumentDidOpenAsync(JToken arg)
         {
-            if (this.waitForInit != null)
```
ausbin, line 359: Does removing these checks change the behavior?

In `main` right now, it seems like waitForInit can represent 3 states:

1. The constructor has not finished
2. The constructor has finished
3. We got the `initialize` message and finished processing it

By moving `rpc.StartListening()` to the bottom of the constructor, it makes sense to me that we no longer need to distinguish between (1) and (2). But we still need to check for (3), right? If the client is sending other messages before it sends `initialize` (violating the LSP spec)

swernli: That's a good question. Looking throw the handlers for the other messages, we have appropriate handling for members that are set during initialize and are nullable, like the `workspaceFolder` and the `clientCapabilities`, while other members like `editorState` are configured during the constructor and can be assumed to be non-null/fully initialized. I think this means we are safe to remove the checks, but it does mean that a misbehaving client might get default behavior instead of successes for messages sent out of order (see line 534 for an example where a file will get treated as plain text if the client capabilities are not yet set).

swernli: As it turns out, the LSP spec does specify that error messages should be used, so I'll revert the part that removed the if-statements and corresponding tests.

samarsha (approved):
This seems like a reasonable change to reduce the number of states `QsLanguageServer` can be in. I'd want to double check that it still works with VS and VS Code, but the increased time taken to respond to the initialize message should be very small, so it seems like it shouldn't be an issue.

PR #1484, opened 2022-07-12
Output Recording will run for Microsoft Simulator

ScottCarda-MS:
Updates the LLVM's BitcodeModule to add hooks that are used by the Runtime to get EntryPoint information from the bit-code.


src/QsCompiler/QirGeneration/CompilationSteps.cs
```diff
@@ -46,9 +46,17 @@ internal static bool VerifyGenerationPrecondition(QsCompilation compilation, Lis
...
         {
+            var targetCapability = assemblyConstants.TryGetValue(ReservedKeywords.AssemblyConstants.TargetCapability, out var capability) && !string.IsNullOrWhiteSpace(capability)
+                ? TargetCapability.TryParse(capability) // null if parsing fails
+                : null;
+            var isMicrosoftSimulator =
+                assemblyConstants.TryGetValue(ReservedKeywords.AssemblyConstants.ProcessorArchitecture, out var architecture)
+                && architecture == ReservedKeywords.AssemblyConstants.MicrosoftSimulator;
+
             compilation =
+                isMicrosoftSimulator ||
```
cesarzc, line 59: I believe we might now be effectively applying `AddOutputRecording` for all cases where we generate QIR. This condition could go away and the transformation could always be applied.

ScottCarda-MS: Should I also get rid of the part of the condition pertaining to the target capabilities?

cesarzc: Yes, it seems like you already did that :)

cesarzc (approved):

PR #1489, opened 2022-07-19
Re-enable Language Server tests

ricardo-espinoza:

PR #1298, opened 2022-01-06
Fixing an issue with that certain updates to the project file were not detected by the language server

bettinaheim:
This PR fixes https://github.com/microsoft/qsharp-compiler/issues/1297.

PR #1492, opened 2022-07-21
Enable semantic validation for notebooks

ausbin:
Add a stand-in project to enable type checking to run for notebook cells

## Testing
* Manual testing with Azure Notebooks locally ([notebook](https://github.com/microsoft/Quantum/blob/feature/samples-gallery/samples/azure-quantum/parallel-qrng/ParallelQrng.ipynb))
* Ran unit tests. Unfortunately still using this patch to work around broken language server tests: https://github.com/microsoft/qsharp-compiler/pull/1481#issuecomment-1188490904 (relevant PR: #1489)


src/QsCompiler/TestProjects/test18/Bell.qs
```diff
@@ -1,11 +1,19 @@
-﻿open Microsoft.Quantum.Intrinsic;
-open Microsoft.Quantum.Canon;
-
-@EntryPoint()
```
ausbin, line 4: This is kind of awkward. Because the stand-in project doesn't have references for these standard library namespaces, we get a bunch of diagnostics for that. That wasn't the point of that test (it's about namespace blocks), so I'm commenting this until we add the references soon

src/QsCompiler/CompilationManager/ProjectManager.cs
```diff
@@ -226,6 +228,9 @@ private class Project : IDisposable
             internal bool ContainsAnySourceFiles(Func<Uri, bool>? filter = null) =>
                 this.specifiedSourceFiles?.Any(filter ?? (_ => true)) ?? false; // keep this as specified, *not* loaded!
 
+            internal bool IsNotebookProject() =>
+                this.ProjectFile == NotebookProjectUri;
```
ausbin, line 232: Seems like it could be a lot of string comparisons, but maybe it's a drop in the bucket

src/QsCompiler/Core/SymbolTable/NamespaceManager.fs
```diff
@@ -1226,6 +1226,20 @@ type NamespaceManager
         finally
             syncRoot.ExitWriteLock()
 
+    /// Add an empty namespace named nsName if it does not already exist. Useful for adding the
+    /// stand-in namespace for notebooks.
+    member this.AddEmptyNamespaceIfNotExists(nsName) =
+        syncRoot.EnterWriteLock()
```
ausbin, line 1232: I'm not sure how to write this without acquiring the write lock, I won't lie. (Like if it's possible to acquire the read lock and then only upgrade to the write lock if needed)

ausbin: Okay, it turns out that the code in TypeChecking can just call `CopyForExtension` so this isn't necessary. Will remove this in the next revision

ausbin: Removed

src/QsCompiler/LanguageServer/Program.cs
```diff
@@ -177,13 +177,13 @@ private static void Log(object msg, string? logFile = null, bool stdout = false)
 
         internal static QsLanguageServer ConnectViaStdInOut(string? logFile = null)
         {
-            Log($"Connecting via stdin and stdout.", logFile, stdout: true);
+            Log($"Connecting via stdin and stdout.", logFile);
```
ausbin, line 180: I couldn't figure out why #1298 started logging to stdout instead of stderr, and it confuses Azure Notebooks (which expects JSON-RPC messages on stdout, not log messages), so I'd like to remove it

src/QsCompiler/CompilationManager/ProjectManager.cs
samarsha: I think it's probably fine. (It might even be doing a faster pointer comparison first before it tries a string comparison, since we only have one instance of the URI.)

src/QsCompiler/LanguageServer/LanguageServer.cs
```diff
@@ -298,6 +298,8 @@ public object Initialize(JToken arg)
             this.LogToWindow($"workspace folder: {this.workspaceFolder ?? "(Null)"}", MessageType.Info);
             this.fileWatcher.ListenAsync(this.workspaceFolder, true, dict => this.InitializeWorkspaceAsync(dict), "*.csproj", "*.dll", "*.qs").Wait(); // not strictly necessary to wait but less confusing
 
+            this.editorState.CreateNotebookProjectAsync().Wait();
```
samarsha, line 301: Do we want to always create a notebook project or wait until the first notebook cell?

src/QsCompiler/CompilationManager/ProjectManager.cs
```diff
@@ -1042,6 +1071,15 @@ public bool ManagerTask(Uri file, Action<CompilationUnitManager> executeTask, ID
                 : this.defaultManager;
         }
 
+        /// <summary>
+        /// Add the file <paramref name="file"/> to the stand-in notebook project
+        /// </summary>
+        public Task RegisterNotebookCellAsync(Uri file) =>
+            this.load.QueueForExecutionAsync(() =>
+            {
+                _ = this.projects.Values.Where(project => project.IsNotebookProject()).Single().AddNotebookCellAsync(file);
```
samarsha, line 1080: This could fail if there isn't a notebook project.

src/QsCompiler/LanguageServer/Program.cs
samarsha: I agree, I think all logging should go to stderr.

src/QsCompiler/CompilationManager/TypeChecking.cs
```diff
@@ -205,12 +204,13 @@ internal static ImmutableArray<string> DocumentingComments(this FileContentManag
         /// </summary>
         /// <param name="items">A collection of positioned items.</param>
         /// <exception cref="ArgumentException">No item precedes <paramref name="pos"/>.</exception>
-        private static T ContainingParent<T>(Position pos, IReadOnlyCollection<(Position, T)> items)
+        private static T ContainingParent<T>(Position pos, IReadOnlyCollection<(Position, T)> items, T? fallback = null)
+            where T : class
```
samarsha, line 208: I think it would be cleaner to return `T?` instead of taking in a default value.

src/QsCompiler/CompilationManager/ProjectManager.cs
```diff
@@ -226,6 +228,13 @@ private class Project : IDisposable
             internal bool ContainsAnySourceFiles(Func<Uri, bool>? filter = null) =>
                 this.specifiedSourceFiles?.Any(filter ?? (_ => true)) ?? false; // keep this as specified, *not* loaded!
 
+            // TODO: Going forward, instead of having the notebook project masquerade as a msbuild
+            //       project, it would be better to have e.g. an abstract Project base class with
+            //       subclasses NotebookProject and MSBuildProject. There is some work on this in
+            //       the aadams/notebook-project-refactor-wip branch
+            internal bool IsNotebookProject() =>
```
samarsha, line 235:
```suggestion
            internal bool IsNotebookProject =>
```

src/QsCompiler/CompilationManager/TypeChecking.cs
```diff
@@ -281,6 +278,16 @@ private static T ContainingParent<T>(Position pos, IReadOnlyCollection<(Position
                     .ToImmutableDictionary(ns => ns.Name); // making sure the namespaces are extended even if they occur multiple times in the same file
                 var namespaces = namespaceHeaders.Select(tuple => (tuple.Item2.Position, distinctNamespaces[tuple.Item2.SymbolName])).ToList();
 
+                Namespace? notebookNamespace =
+                    file.DocumentKind == DocumentKind.NotebookCell
+                    ? compilation.GlobalSymbols.CopyForExtension(ReservedKeywords.InternalUse.NotebookNamespace, file.FileName)
+                    : null;
+
+                Namespace ContainingNamespace(Position pos, IReadOnlyCollection<(Position, Namespace)> items)
```
samarsha, line 286:
```suggestion
                Namespace ContainingNamespace(Position pos)
```

src/QsCompiler/CompilationManager/TypeChecking.cs
```diff
@@ -491,9 +505,16 @@ internal static void ImportGlobalSymbols(this FileContentManager file, Compilati
             try
             {
                 var namespaces = file.GetNamespaceHeaderItems().Select(tuple => (tuple.Item2.Position, tuple.Item2.SymbolName)).ToList();
+
+                string ContainingNamespaceName(Position pos, IReadOnlyCollection<(Position, string)> items)
```
samarsha, line 509:
```suggestion
                string ContainingNamespaceName(Position pos)
```

samarsha (approved):

PR #1493, opened 2022-07-22
Always send telemetry events from Visual Studio and Language Server

ricardo-espinoza:
The QsCompiler and VisualStudioExtension solutions are being built with the `TELEMETRY` flag undefined. This is preventing telemetry events from being published. The flag should be set as a consequence of Assembly Constants containing `SIGNED:TELEMETRY`, which is set, but in those projects it's not.

For this release, we're directly enabling the telemetry code (on the release branch) while a proper solution is done in 'main' for the next release.

### DIFF
--- a/bootstrap.ps1
+++ b/bootstrap.ps1
@@ -49,7 +49,7 @@ if ("$VsCodeVsixVersion".Trim().Length -eq 0) {
 }
 
 
-$Telemetry = "true";
+$Telemetry = "$($Env:ASSEMBLY_CONSTANTS)".Contains("TELEMETRY").ToString().ToLower();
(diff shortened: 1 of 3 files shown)


samarsha:
I assume this is for testing rather than to be merged, right?

ricardo-espinoza:
> I assume this is for testing rather than to be merged, right?

At this point mostly for debugging. However, we're considering workarounds for the release build of this month (which could have Telemetry enabled by force in the officially built VSIX), but this won't be merged back to main. We would address it properly, so the flag gets populated correctly before the next release.

cgranade (approved):

PR #1496, opened 2022-07-27
Updating QDK version to 0.25.222597

ricardo-espinoza:


src/QsCompiler/TestProjects/test12/test12.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.Quantum.Sdk/0.24.217480-beta">
```
ricardo-espinoza, line 1: Normally we don't update those projects, but in this case, this beta build is very close to the current one, so it's a good opportunity to migrate to a regular release build for this test.

examples/QIR/Development/Development.csproj
```diff
@@ -9,8 +9,8 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Quantum.LlvmBindings.Native" Version="13.0.0-CI-20220129-005156" PrivateAssets="All" GeneratePathProperty="true"/>
-    <PackageReference Include="Microsoft.Quantum.Qir.Runtime" Version="0.24.217480-alpha" GeneratePathProperty="true" />
-    <PackageReference Include="Microsoft.Quantum.Simulators" Version="0.24.217480-beta" GeneratePathProperty="true" />
+    <PackageReference Include="Microsoft.Quantum.Qir.Runtime" Version="0.25.222597-alpha" GeneratePathProperty="true" />
```
ScottCarda-MS, line 12: Should this be non-alpha?

ricardo-espinoza: There is a special case for 'Microsoft.Quantum.Qir.Runtime' that the package is published as -alpha, so we need to treat it differently. See the version in NuGet:

https://www.nuget.org/packages/Microsoft.Quantum.Qir.Runtime

ScottCarda-MS (approved):

cesarzc (approved):

PR #1497, opened 2022-07-28
Clean up and caching for callables during QIR generation

bettinaheim:
I started some clean up a while back as a prep for potentially using stack allocations more extensively. I didn't really have time to get as far as I wanted, but let me put what I have up.


src/QsCompiler/QirGeneration/QIR/DataStructures.cs
```diff
@@ -383,40 +421,46 @@ internal TupleValue(Value tuple, ImmutableArray<ResolvedType> elementTypes, Gene
...
+            new(pointer is null || this.IsRuntimeManaged
+                    ? null
+                    : this.CreateTupleElementPointer(index, pointer.CurrentCache()),
+                this.sharedState,
+                () => this.CreateTupleElementPointer(index));
+
+        private IValue.Cached<PointerValue>[] CreateTupleElementPointersCaches() =>
+            Enumerable.ToArray(Enumerable.Range(0, this.ElementTypes.Length).Select(index => this.CreateTupleElementPointerCache(index)));
+
+        private Value AllocateTuple(bool allocOnStack) =>
```
swernli, line 455: `AllocateArray` can only be used for opaque, runtime managed arrays, while `AllocateTuple` takes a boolean to determine if it will stack allocated instead. For arrays, stack allocation appears to be handled by separate functions. While the implementation seems fine, it's a little odd to have this inconsistency. Should tuples also be split out into two different functions rather than use `AllocateTuple` for both?

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestBuiltInIntrinsics1.ll
```diff
@@ -49,24 +49,24 @@ entry:
   %27 = getelementptr inbounds { %String* }, { %String* }* %26, i32 0, i32 0
   store %String* %24, %String** %27, align 8
   call void @__quantum__rt__callable_invoke(%Callable* %19, %Tuple* %25, %Tuple* null)
```
swernli, line 51: I'm not sure if this was the incomplete part you referred to in the description or if I'm missing something about these changes, but I would've expected all of the calls to `__quantum__rt__callable_invoke` to have been replaced by the calls to the global functions the callable's correlate to, since none of these callables utilize capture tuples or escape their scope. But for some reason only the last invocation was replaced. Is this known unfinished work or a bug?

swernli: Oh, never mind, I can see it now. The `%options` value is a UDT tuple that is created via `__quantum__rt__tuple_copy` so all of that value's cached state before the last copy-update that defines it is lost, so only the last callable value is known by the compiler. The logic as written would work for all three callables if callabes were completely local and stack allocated, which would happen more reliably if this were targeted QIR that went through the inlining pass and stack allocated tuples. Fascinating.

swernli: if I skip the UDT and just use a tuple like this:
```qsharp
        let options = (Message, DumpMachine, DumpMachine);
        let (a, b, c) = options;
        a("foo");
        b("bar");
        c();
```
Then I can see all three functions being replaced!

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestLocalCallables1.ll
```diff
@@ -28,84 +30,77 @@ exit__1:                                          ; preds = %header__1
...
-  %29 = getelementptr inbounds { %String*, double }, { %String*, double }* %27, i32 0, i32 1
-  store %String* %str, %String** %28, align 8
-  store double %val, double* %29, align 8
+  %17 = call %Tuple* @__quantum__rt__tuple_create(i64 ptrtoint ({ %String*, double }* getelementptr ({ %String*, double }, { %String*, double }* null, i32 1) to i64))
+  %18 = bitcast %Tuple* %17 to { %String*, double }*
+  %19 = getelementptr inbounds { %String*, double }, { %String*, double }* %18, i32 0, i32 0
+  %20 = getelementptr inbounds { %String*, double }, { %String*, double }* %18, i32 0, i32 1
+  store %String* %str, %String** %19, align 8
+  store double %val, double* %20, align 8
+  call void @__quantum__rt__callable_make_adjoint(%Callable* %9)
```
swernli, line 60: It is unfortunate the callable management code is still generated even though the callables are now never invoked. Hopefully we can find a future optimization that recoginizes the callables are never really used and skip outputting any of this code, because LLVM doesn't know these functions are side effect free and therefore won't optimize them away (from the LLVM pass perspective, the callabes are used because they are manipulated by the ref counting and function runtime functions).

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestOpArgument.ll
```diff
@@ -22,6 +24,8 @@ entry:
   call void @__quantum__qis__cnot__body(%Qubit* %q1, %Qubit* %q2)
   call void @__quantum__qis__swap(%Qubit* %q1, %Qubit* %q2)
   %8 = call %Callable* @__quantum__rt__callable_create([4 x void (%Tuple*, %Tuple*, %Tuple*)*]* @Microsoft__Quantum__Intrinsic__CNOT__FunctionTable, [2 x void (%Tuple*, i32)*]* null, %Tuple* null)
```
swernli, line 26: This doesn't appear to be a new problem, but some of these callabes look like they are getting leaked. The callable `%8` is created (ref count: 1), new code inserts an increment because it gets stored into a tuple (ref count: 2), then new code tracks when the tuple references go away (ref count: 1). Nothing takes away that final ref count. In the old QIR from this diff, the ref count on `%8` was never managed at all, so the the leak seems to be preexisting.

Compare this to `%40` below which gets created (ref count: 1), passed as an argument, and then decremented (ref count: 0), so that it is not leaked.

swernli (approved):
Some feedback and questions above, but nothing that I think should be blocking. The callable memory leak is a little worrisome, but does not appear to be a new problem due to these changes.

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestBuiltInIntrinsics1.ll
bettinaheim: Yes, exactly; this change really is primarily intended for targeted QIR - but it is enabled in general (didn't see a reason why it shouldn't be).

bettinaheim:
> Some feedback and questions above, but nothing that I think should be blocking. The callable memory leak is a little worrisome, but does not appear to be a new problem due to these changes.

@swernli Nothing should get leaked (actually I think the clean up should also clean up the memory management for targeted QIR, even though that is largely irrelevant since it is going to get stripped by subsequent compilation in any case). Could you give me more specifics regarding where something looks like it is leaked? Is it for targeted or untargeted IR?

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestLocalCallables1.ll
bettinaheim: I think there are a couple of LLVM attributes we should look into that may make it clear that these functions are side effect free. That said, the caching is already avoiding to apply functors when possible - just in some cases it is not clear that it is possible. The functors are applied lazily and/but accessing the .Value property of the CallableValue triggers their application (because I have to assume that I need to apply them at this point).

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestOpArgument.ll
bettinaheim: There is line 33 
```llvm
store %Callable* %8, %Callable** %11, align 8
```
that stores the callable, so I believe a ref count should be correct. We do validate that there are no memory leaks as part of the e2e (Q# samples), and while I don't know if this exact case is covered, it would be rather surprising if it wasn't.

src/QsCompiler/QirGeneration/QIR/DataStructures.cs
bettinaheim: updated so that this function doesn't take an argument

src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestOpArgument.ll
swernli: I did get the chance to run this case through the evaluator and saw that no memory leaks are detected, so you are right, I just can't see how. I'll get into detail below, but again, it looks like the pattern than concerns me exists before this change, so I don't think it's any reason to worry about this change (hence the sign off). Here's my logic for future reference though...

Looking at the file _before_ the changes, I can see that a callable is created on line 24 that is stored in value `%8`:
https://github.com/microsoft/qsharp-compiler/blob/7714168fda4379fb7e6a6c616f680ec039c482f4/src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestOpArgument.ll#L24
At this point, it has reference count of 1 by definition of being newly created. It is then immediately stored in the tuple created as value `%9` and accessed through `%11` (tuple also has ref count of 1):
https://github.com/microsoft/qsharp-compiler/blob/7714168fda4379fb7e6a6c616f680ec039c482f4/src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestOpArgument.ll#L25-L28
After that, `%8` and `%11` are never used again, so any access would have to come through the tuple `%9` or it's bitcast value `%10`. However, `%10` is also never used again outside of these lines, and `%9` is only used one more time as an argument to the creation of another callable value on line 31:
https://github.com/microsoft/qsharp-compiler/blob/7714168fda4379fb7e6a6c616f680ec039c482f4/src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestOpArgument.ll#L31-L32
This callable value `%13` is used as an argument to a function call right below that, and then later cleaned up by decrementing the alias and reference count at the end of the function:
https://github.com/microsoft/qsharp-compiler/blob/7714168fda4379fb7e6a6c616f680ec039c482f4/src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestOpArgument.ll#L81-L82
But the original callable value in `%8` and tuple value in `%9` are _never_ interacted with again, so I can't see how/where the original reference count would be decremented to zero and the two items cleaned up.

The only difference in this specific flow after the changes made here is an extra increment and decrement. So instead of appearing to have a value of 1 throughout execution, `%8` is immediately incremented to ref count of 2 after creation:
https://github.com/microsoft/qsharp-compiler/blob/6745f31d11b1fa8d2ceb400524256bc6ea35800c/src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestOpArgument.ll#L26-L28
and then decremented back to 1 at the end of the function:
https://github.com/microsoft/qsharp-compiler/blob/6745f31d11b1fa8d2ceb400524256bc6ea35800c/src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestOpArgument.ll#L95-L96
But there still doesn't appear to be anything that would decrement the final ref count and clean up the value. Likewise, the access pattern for the tuple in `%9` hasn't changed at all; it is created, stored into, passed as an argument to another callable creation, and then never accessed again, so appears to reach the end of the function with a ref count of 1.

It's very likely I'm missing something here that makes this all work, given that taking this exact code and putting into an execution tests passes the leak checks, but the pattern in the IR itself seems suspicious.

PR #1522, opened 2022-08-04
Add References for Stand-In Notebook Project

ausbin:
Read reference dlls from disk for stand-in notebook project. As we discussed, the code here does not download the dlls themselves — to avoid circular dependencies in the build pipeline, that is the responsibility of a docker image or the Azure Notebooks folks.

## Testing
* Manual testing locally for [an example notebook][1], with references downloaded by powershell script I added in the AzureNotebook repo (will send link offline)
* Ran unit tests for the whole compiler and for the language server, which includes a new unit test for references

[1]: https://github.com/microsoft/Quantum/blob/feature/samples-gallery/samples/azure-quantum/parallel-qrng/ParallelQrng.ipynb

### DIFF
--- a/src/QsCompiler/LanguageServer/Program.cs
+++ b/src/QsCompiler/LanguageServer/Program.cs
@@ -85,6 +85,8 @@ namespace Microsoft.Quantum.QsLanguageServer
             return (int)code;
         }
 
+        private static string? NotebookReferencesDir => Environment.GetEnvironmentVariable("NOTEBOOK_REFERENCES_DIR");
+
         public static string? Version { get; set; } =
             typeof(Server).Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion
             ?? typeof(Server).Assembly.GetName().Version?.ToString();
@@ -99,13 +101,13 @@ namespace Microsoft.Quantum.QsLanguageServer
             var parser = new Parser(parser => parser.HelpWriter = null); // we want our own custom format for the version info
             var options = parser.ParseArguments<Options>(args);
             return options.MapResult(
-                (Options opts) => Run(opts),
+                (Options opts) => Run(opts, NotebookReferencesDir),
                 errs => errs.IsVersion()
                     ? LogAndExit(ReturnCode.SUCCESS, message: Version, stdout: true)
                     : LogAndExit(ReturnCode.INVALID_ARGUMENTS, message: HelpText.AutoBuild(options)));
         }
 
-        private static int Run(Options options)
+        private static int Run(Options options, string? notebookReferencesDir = null)
         {
             if (options == null)
             {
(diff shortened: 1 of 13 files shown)


src/QsCompiler/LanguageServer/Program.cs
```diff
@@ -181,7 +181,7 @@ internal static QsLanguageServer ConnectViaStdInOut(string? logFile = null)
             return new QsLanguageServer(Console.OpenStandardOutput(), Console.OpenStandardInput());
         }
 
-        internal static QsLanguageServer ConnectViaNamedPipe(string writerName, string readerName, string? logFile = null)
+        internal static QsLanguageServer ConnectViaNamedPipe(string writerName, string readerName, string? logFile = null, string? notebookReferencesDir = null)
```
ausbin, line 184: Having to pass this all the way through everything is ugly, but I'm not sure what else to do. It's tempting to just have a hacky command (like the get diagnostics command) to set this, but that's also not very pretty

src/QsCompiler/TestProjects/test18/Bell.qs
```diff
@@ -1,15 +1,5 @@
-﻿// TODO: Uncomment this once references are fixed for notebooks
```
ausbin, line 1: Removing this since I don't think we'll have the reference dlls on-disk when the unit tests run

src/QsCompiler/Tests.LanguageServer/Tests.LanguageServer.csproj
```diff
@@ -27,6 +27,9 @@
     <ProjectReference Include="..\LanguageServer\LanguageServer.csproj">
       <Private>true</Private>
     </ProjectReference>
+    <ProjectReference Include="..\Compiler\Compiler.csproj">
+      <Private>true</Private>
+    </ProjectReference>
```
ausbin, line 32: Using CompilationLoader for `TestUtils.BuildDll()` means we need to have the language server tests have a reference to the compiler project. I'm not sure how big of a deal this is

samarsha: I don't see an issue with it.

src/QsCompiler/LanguageServer/Program.cs
samarsha: Looks like #1491 runs the server in a new process, so we might need to do something else there, too (maybe a command line argument). Right now I can't think of a better solution.

ausbin: That's a good point, thanks. What do you think about a `workspace/executeCommand` command for setting this? (Similar to the "get diagnostics" one, for example.) That would make this explicitly for tests/debugging and not part of the public api

samarsha: I think to me it feels like something that only the program launching the LSP server should be able to set, rather than the LSP client. For example if the server is running on a different host than the client, it's weird for the client to be able to tell the server where to look for binaries locally.

How about an environment variable? It seems like it's kind of analagous to looking up something in `$PATH` or `$LD_LIBRARY_PATH`.

ausbin: The analogy to `$LD_LIBRARY_PATH` is pretty convincing, thanks. I'll try that

ausbin: Pushed up a little commit (e7ba09da69b6), is that what you were thinking?

samarsha: I was thinking that the environment variable would replace the QsLanguageServer constructor parameter and all of the fields storing the path; in NotebookProjectLoader we would read the environment variable directly or fall back to the assembly path. I think that minimizes the number of places we need to pass the path around.

src/QsCompiler/LanguageServer/Program.cs
```diff
@@ -85,6 +85,8 @@ private static int LogAndExit(ReturnCode code, string? logFile = null, string? m
             return (int)code;
         }
 
+        private static string? NotebookReferencesDir => Environment.GetEnvironmentVariable("NOTEBOOK_REFERENCES_DIR");
```
samarsha, line 88: Just some bikeshedding on the name - environment variables share a global namespace so a vendor prefix seems like a good idea Maybe `QSHARP_NOTEBOOK_REFERENCES`?

samarsha (approved):

PR #1523, opened 2022-08-09
Optimize conditions with compile time Booleans

swernli:
This change optimizes handling of conditions where some of the values are known compile time Boolean values. In particular, it expands handling of if-elif-else sequences by compressing known true branches into the current block and skipping any other branches. Likewise, for compound conditional evaluation it adds handling for special cases where part of a logical `and` or logical `or` statement is known to be constant, removing the need for a phi-node and propagating the appropriate constant or dynamic value.

New and updated test cases were manually verified for correct output.


src/QsCompiler/QirGeneration/Subtransformations/StatementKindTransformation.cs
```diff
@@ -268,7 +268,14 @@ public override QsStatementKind OnConditionalStatement(QsConditionalStatement st
                 var testValue = this.SharedState.EvaluateSubexpression(clauses[n].Item1).Value;
                 this.SharedState.ScopeMgr.CloseScope(false);
 
-                if (!(QirValues.AsConstantUInt32(testValue) == 0))
+                if (QirValues.AsConstantUInt32(testValue) == 1)
+                {
+                    // Because this condition is known to be true, add the corresponding clauses to the current block without
+                    // condition and skip the rest of the conditional clauses evaluation.
+                    contBlockUsed = this.ProcessBlock(this.SharedState.CurrentBlock, clauses[n].Item2.Body, contBlock) || contBlockUsed;
```
bettinaheim, line 275: I am reviewing this on the side, so my apologies if I am misreading something. Shouldn't this go into its own block in any case? I am thinking of the case where the condition of the first block may or may not be true, the condition of the second block is certainly true, and then there is some more code. If the condition of the first block is met, then the second block must not be executed. Hence I think the second block needs to be its on block in any case.

src/QsCompiler/QirGeneration/Context.cs
```diff
@@ -1268,6 +1268,30 @@ Value ProcessConditionalBlock(BasicBlock block, Func<IValue>? evaluate)
                 }
 
                 this.SetCurrentBlock(contBlock);
+
+                // Handle special cases with constant values where the phi node is not necessary.
+                if (QirValues.AsConstantUInt32(evaluatedOnTrue) == 0 && evaluatedOnFalse == condition)
+                {
+                    // This case is the equivalent of: `val = cond ? false : cond;`
+                    // which is the same as: `val = false;` so we can just return the `false` value
```
bettinaheim, line 1276: But you may not have executed the onTrue block... So no, I don't think you can return the evaluatedOnTrue value.

src/QsCompiler/QirGeneration/Context.cs
```diff
@@ -1268,6 +1268,30 @@ Value ProcessConditionalBlock(BasicBlock block, Func<IValue>? evaluate)
...
+                    // from the evaluated true block.
+                    return evaluatedOnTrue!;
+                }
+                else if (QirValues.AsConstantUInt32(evaluatedOnFalse) == 1 && evaluatedOnTrue == condition)
+                {
+                    // This case is the equivalent of: `val = cond ? cond : true;`
+                    // which is the same as: `val = true;` so we can just return the `true` value
```
bettinaheim, line 1283: Same here.

bettinaheim (changes requested):

src/QsCompiler/QirGeneration/Subtransformations/StatementKindTransformation.cs
swernli: Actually, at this point, the `CurrentBlock` is already set to the "false" block created by the previous condition, if there was one. This happens on line 296 (after changes), seen at the end here: https://github.com/microsoft/qsharp-compiler/blob/51e1fe7afe598bccc4fb589d6b23f4c91e82e5a1/src/QsCompiler/QirGeneration/Subtransformations/StatementKindTransformation.cs#L287-L296
That is how the loop updates the shared state to be pointing to the right place for the evaluation of the next conditional clause. Here, because we know the next conditional clause is a branch on the constant true, we can instead just fill in the contents of the block that would have been jumped to and drop the false block that is never executed.

This is verified by one of the new test cases I added: https://github.com/microsoft/qsharp-compiler/blob/51e1fe7afe598bccc4fb589d6b23f4c91e82e5a1/src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestTargetingProfile.qs#L272-L280
This code is invoked with a measurement result that ensures the first condition is not known at compile time, and by passing a static 1 and a static 0 for the second parameter to verify both the known true and known false cases for the else-if clause.

Looking at the generated IR for these test cases, it verifies the correct output is produced both for second clause known false and second clause known true. In the known false case, we see the dump machine with `15` getting dropped, because we know that block will never get executed, instead turning into an empty else block: https://github.com/microsoft/qsharp-compiler/blob/51e1fe7afe598bccc4fb589d6b23f4c91e82e5a1/src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestTargetingProfile.ll#L373-L389
For the known true case, we see the dump machine with `15` being simplified into the else case because we know it will always be executed (eliminating the `br i1 true ...` that used to occur in that case):  https://github.com/microsoft/qsharp-compiler/blob/51e1fe7afe598bccc4fb589d6b23f4c91e82e5a1/src/QsCompiler/Tests.Compiler/TestCases/QirTests/TestTargetingProfile.ll#L423-L448

src/QsCompiler/QirGeneration/Context.cs
swernli: Do you mean executed at runtime or generated at compile time?
For both `evaluatedOnTrue` and `evaluatedOnFalse` they are are always processed by generation if a lambda was provided for them, otherwise they just get the default value. The value they represent doesn't come from the true/false block, but from the block that produces the conditional value itself. That's why I can know at this point in the generation code if they are static Boolean values and optimize appropriately.

The more generic notion of this optimization is one other languages warn about: using an if-statement to produce a Boolean value is inefficient and can be replaced by the condition in the if-statement itself or a constant. So what I'm doing here is the equivalent of optimizing:
```
bool val;
if cond {
    val = false;
} else {
    val = cond;
}
```
into:
```
bool val = false;
```

bettinaheim: Alright that works - it is a bit confusing to read for me (so I would have instead explicitly created a new constant for the sake of readability), but I get what you are saying.

src/QsCompiler/QirGeneration/Subtransformations/StatementKindTransformation.cs
bettinaheim: Alright - the block naming is somewhat confusing if we inline the block into the condition instead, and I am a bit worried that it might be easy to introduce bugs if the code is changed. But I am happy with the test coverage (wouldn't hurt to add an execution test for this, though).

bettinaheim (approved):

PR #1524, opened 2022-08-09
Allow feature rollforward for .NET, use 6.0.400

swernli:


swernli:
This doesn't work with F# for some reason, so the approach will be fixing the e2e builds a different way.

PR #1528, opened 2022-08-10
Allow opening a namespace multiple times and with multiple aliases

ausbin:
This diff got much bigger than I intended. I think reviewing it commit-by-commit would help. The commits are, respectively, and in increasing order of diff size:

1. The heart of it: allow opening namespaces multiple times, including with different aliases
2. Change code to use `""` to represent no alias instead of `null`. This diff ended up being bigger than (1), largely because I traced back the null's birth to the "root of all evil" aka the definition of an `OpenDirective` fragment
3. Surprisingly, when I removed a ton of error diagnostics definitions in (1), that didn't break any unit tests. In fact, I couldn't find any unit tests at all referencing those namespace error diagnostics, so I tried to add some. In the process, I had to cook up some little infrastructure for compiler unit tests that aren't testing code written in a Q# callable

Note: this does not do the alias merging we discussed, Sarah. That is, it does not allow opening two namespaces with the same alias. My only reason is lack of time.

## Testing
* Ran unit tests, including new tests
* Did manual testing in VSCode
* Fantomas is happy

### DIFF
--- a/src/QsCompiler/Core/SymbolTable/Namespace.fs
+++ b/src/QsCompiler/Core/SymbolTable/Namespace.fs
@@ -458,8 +458,8 @@ type Namespace
     /// or if the given alias is already in use for a different namespace.
     /// </summary>
     /// <exception cref="SymbolNotFoundException">The source file does not contain this namespace.</exception>
-    member internal this.TryAddOpenDirective source (openedNS, nsRange) (alias, aliasRange) =
-        let alias = if String.IsNullOrWhiteSpace alias then "" else alias.Trim()
+    member internal this.TryAddOpenDirective source (openedNS, nsRange) (alias: string, aliasRange) =
+        let alias = alias.Trim()
 
         match parts.TryGetValue source with
         | true, partial ->
--- a/src/QsCompiler/Core/SymbolTable/Namespace.fs
+++ b/src/QsCompiler/Core/SymbolTable/Namespace.fs
@@ -461,35 +461,23 @@ type Namespace
     member internal this.TryAddOpenDirective source (openedNS, nsRange) (alias, aliasRange) =
         let alias = if String.IsNullOrWhiteSpace alias then null else alias.Trim()
 
-        let aliasIsSameAs str =
-            (isNull str && isNull alias) || (not (isNull str) && not (isNull alias) && str = alias)
-
         match parts.TryGetValue source with
         | true, partial ->
             let imported = partial.ImportedNamespaces
 
-            match imported.TryGetValue openedNS with
-            | true, existing when aliasIsSameAs existing && isNull existing ->
-                [|
-                    nsRange |> QsCompilerDiagnostic.Warning(WarningCode.NamespaceAleadyOpen, [])
-                |]
-            | true, existing when aliasIsSameAs existing ->
-                [|
-                    nsRange |> QsCompilerDiagnostic.Warning(WarningCode.NamespaceAliasIsAlreadyDefined, [])
-                |]
-            | true, existing when not (isNull existing) ->
-                [|
-                    nsRange |> QsCompilerDiagnostic.Error(ErrorCode.AliasForNamespaceAlreadyExists, [ existing ])
-                |]
-            | true, _ ->
-                [|
-                    nsRange |> QsCompilerDiagnostic.Error(ErrorCode.AliasForOpenedNamespace, [])
-                |]
-            | false, _ when not (isNull alias) && imported.ContainsValue alias ->
+            // TODO: Ideally, we would allow opening multiple namespaces with the same alias. There
+            // is some preliminary work for this in the aadams/overlapping-alias-wip branch, but
+            // it would take a nontrivial refactor since it means some qualified names of
+            // types/callables would no longer fully qualified. It requires merging the logic for
+            // unqualified and qualified symbol lookups, basically.
+            if
+                not (isNull alias)
+                && imported |> Seq.exists (fun kv -> kv.Key <> openedNS && kv.Value.Contains(alias))
+            then
                 [|
                     aliasRange |> QsCompilerDiagnostic.Error(ErrorCode.InvalidNamespaceAliasName, [ alias ])
                 |]
-            | false, _ ->
+            else
                 typesDefinedInAllSourcesCache <- null
                 callablesDefinedInAllSourcesCache <- null
                 partial.AddOpenDirective(openedNS, alias)
--- a/src/QsCompiler/Core/SymbolTable/Namespace.fs
+++ b/src/QsCompiler/Core/SymbolTable/Namespace.fs
@@ -459,7 +459,7 @@ type Namespace
     /// </summary>
     /// <exception cref="SymbolNotFoundException">The source file does not contain this namespace.</exception>
     member internal this.TryAddOpenDirective source (openedNS, nsRange) (alias, aliasRange) =
-        let alias = if String.IsNullOrWhiteSpace alias then null else alias.Trim()
+        let alias = if String.IsNullOrWhiteSpace alias then "" else alias.Trim()
 
         match parts.TryGetValue source with
         | true, partial ->
@@ -470,10 +470,7 @@ type Namespace
             // it would take a nontrivial refactor since it means some qualified names of
             // types/callables would no longer fully qualified. It requires merging the logic for
             // unqualified and qualified symbol lookups, basically.
-            if
-                not (isNull alias)
-                && imported |> Seq.exists (fun kv -> kv.Key <> openedNS && kv.Value.Contains(alias))
-            then
+            if alias <> "" && imported |> Seq.exists (fun kv -> kv.Key <> openedNS && kv.Value.Contains(alias)) then
                 [|
                     aliasRange |> QsCompilerDiagnostic.Error(ErrorCode.InvalidNamespaceAliasName, [ alias ])
                 |]
--- a/src/QsCompiler/Core/SymbolTable/NamespaceManager.fs
+++ b/src/QsCompiler/Core/SymbolTable/NamespaceManager.fs
@@ -76,7 +76,7 @@ type NamespaceManager
     /// </exception>
     let openNamespaces (nsName, source) =
         let isKnown (kv: KeyValuePair<string, ImmutableHashSet<string>>) =
-            if not (kv.Value.Contains(null)) then
+            if not (kv.Value.Contains("")) then
                 None
             else
                 match namespaces.TryGetValue kv.Key with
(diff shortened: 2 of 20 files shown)


src/QsCompiler/Core/SymbolTable/NamespaceManager.fs
```diff
@@ -76,8 +76,8 @@ type NamespaceManager
     /// A namespace with the given name was not found, or the source file does not contain the namespace.
     /// </exception>
     let openNamespaces (nsName, source) =
-        let isKnownAndNotAliased (kv: KeyValuePair<_, _>) =
-            if not (isNull kv.Value) then
+        let isKnownAndNotAliased (kv: KeyValuePair<string, ImmutableHashSet<string>>) =
+            if not (kv.Value.Contains("")) then
```
ausbin, line 80: Pretty sure this is wrong. Even a namespace has an alias, we shouldn't skip it when looking for unqualified symbols

ausbin: Oops, I misread, this is legit

src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -335,8 +333,6 @@ type WarningCode =
     | SetInResultConditionedBlock = 5026
     | UnsupportedCallableCapability = 5027
 
-    | NamespaceAleadyOpen = 6003
```
swernli, line 338: Interestingly, removing this enum value breaks IQ#, since it uses this value internally. Rather than require this to be a cross-repo change that also updates the uses of this enum there, I would recommend keeping this file unchanged and just removing the uses in this repo.

src/QsCompiler/Core/SymbolTable/Namespace.fs
```diff
@@ -459,37 +459,22 @@ type Namespace
...
-            | true, _ ->
-                [|
-                    nsRange |> QsCompilerDiagnostic.Error(ErrorCode.AliasForOpenedNamespace, [])
-                |]
-            | false, _ when not (isNull alias) && imported.ContainsValue alias ->
+            // TODO: Ideally, we would allow opening multiple namespaces with the same alias. There
+            // is some preliminary work for this in the aadams/overlapping-alias-wip branch, but
+            // it would take a nontrivial refactor since it means some qualified names of
+            // types/callables would no longer fully qualified. It requires merging the logic for
+            // unqualified and qualified symbol lookups, basically.
```
swernli, line 472: Thanks for leaving a pointer here to the WIP branch!

src/QsCompiler/DataStructures/Diagnostics.fs
ausbin: Thanks, I didn't even consider that I was breaking a public API here

src/QsCompiler/Core/SymbolTable/PartialNamespace.fs
```diff
@@ -23,7 +23,7 @@ type private PartialNamespace
         name: string,
         source: string,
         documentation: IEnumerable<ImmutableArray<string>>,
-        openNS: IEnumerable<KeyValuePair<string, string>>,
+        openNS: IEnumerable<KeyValuePair<string, HashSet<string>>>,
```
swernli, line 26: Unfortunately, it looks like this change also affects the public API surface... this line in IQ# expects to be run on a single string but won't work properly against a hashset: https://github.com/microsoft/iqsharp/blob/cad319493cd00934a40e1c37e2856fa18d3a1543/src/Core/Compiler/CompilerService.cs#L119
Since this is pretty fundamental to the feature, I think we'll need to keep this one here and make it a cross-repo change coordinated with IQ#.

src/QsCompiler/Tests.Compiler/TestUtils/SetupWholeFileTests.fs
```diff
@@ -0,0 +1,94 @@
...
+open Microsoft.Quantum.QsCompiler.SyntaxTree
+open Microsoft.VisualStudio.LanguageServer.Protocol
+open System.Collections.Generic
+open System.Collections.Immutable
+open System
+open System.IO
+open System.Linq
+open Xunit
+
+type WholeFileTests() =
```
samarsha, line 18: This class has no fields. I think it should be a module.

src/QsCompiler/DataStructures/SyntaxTokens.fs
```diff
@@ -47,6 +47,7 @@ type QsSymbolKind<'Symbol> =
     /// Used for the arguments of the original method omitted upon functor gen declaration.
     | OmittedSymbols
     /// Used to allow destructs of the form let (_,a) = ...
+    /// Also used for when an alias is omitted in open directives.
```
samarsha, line 50: It feels wrong to use `MissingSymbol` in this way, since it otherwise corresponds to a physical `_` in the file. Would it be possible to keep using `QsNullable<QsSymbol>` in the open directive, but just convert the absence of a symbol to empty string later down the line?

src/QsCompiler/Core/SymbolTable/PartialNamespace.fs
swernli: Actually, never mind... I have the wrong spot in the code. I'll close this comment and put one on the right spot.

src/QsCompiler/DataStructures/SyntaxTokens.fs
```diff
@@ -447,7 +448,7 @@ type QsFragmentKind =
     | FunctionDeclaration of CallableDeclaration
     | TypeDefinition of TypeDefinition
     | DeclarationAttribute of QsSymbol * QsExpression
-    | OpenDirective of QsSymbol * QsNullable<QsSymbol>
+    | OpenDirective of QsSymbol * QsSymbol // second symbol will be MissingSymbol if no alias provided
```
swernli, line 451: This is the correct spot that results in a break in IQ# (here https://github.com/microsoft/iqsharp/blob/cad319493cd00934a40e1c37e2856fa18d3a1543/src/Core/Compiler/CompilerService.cs#L119). `ValueOr` exists on `QsNullable` but not on `QsSymbol`. If we follow Sarah's recommendation above, this goes back to `QsNullable` and the break goes away.

ausbin: Wow, I had no idea that fragments were a public API. I can partially revert commit (2), which will simplify this PR as well. I was hoping to get rid of the idea of a null alias entirely but it's not worth breaking things, yeah

src/QsCompiler/DataStructures/SyntaxTokens.fs
ausbin: Dang, I don't like the QsNullable but I'll do whatever is necessary. I will partially revert the second commit

samarsha: I was bringing this up from a design standpoint, by the way, rather than considering whether it's a breaking change for IQ#. I think every case of QsSymbol corresponds to something physically in the file:

- Symbol -> `foo`
- QualifiedSymbol -> `foo.bar`
- SymbolTuple -> `(foo, bar)`
- OmittedSymbols -> `...`
- MissingSymbol -> `_`
- InvalidSymbol -> something unparseable

But for `open foo as bar;` vs `open foo;`, there's nothing to point to, not even a placeholder like `_`. So in that sense it really is an optional symbol (I don't like `QsNullable` much either, but the same argument works for `QsSymbol option`). Does that make sense, or do you disagree?

src/QsCompiler/DataStructures/SyntaxTokens.fs
ausbin: Done

src/QsCompiler/DataStructures/SyntaxTokens.fs
ausbin: > (I don't like QsNullable much either, but the same argument works for QsSymbol option).

Yeah, that was my problem with it, that we have this rich type system in F# yet we just use a null pointer when you specify an alias of `""`.

In any case, I pushed up a commit that restores the QsNullable. I don't want to take up a lot of y'all's time dying on unnecessary hills

samarsha: QsNullable isn't a null pointer - it has the same type safety as `option`. It's a sum of two cases, `Null` and `Value of 'T`, which is isomorphic to `None` and `Some of 'T` from `option`. The naming is confusing though because `null` (lowercase) really is a null pointer, which we were using in a few places for strings in F#, and I agree that we shouldn't be doing that.

ausbin: I'm not saying it's equivalent to a null pointer in terms of type safety, sorry. I agree that `null` is different from `Null`. I mean that instead of having some conception in an `OpenDirective` fragment of opening a namespace with alias `""` (which is what writing `open X;` should hopefully mean someday), we instead represent that case as `("X", Null)` and downstream we have IsValue/IsNull checks in different places manually rewriting that to `("X", "")`. So yes, you have type safety, but the end result is the same as a null pointer for the alias in that you're checking for a state called "null" every time you want the alias name (three examples: [1][1], [2][2], [3][3]). Those checks don't need to exist if you bake that logic into the QsSymbol type like I originally did.

[1]: https://github.com/microsoft/qsharp-compiler/blob/4c176e5f07e9ae75c368b9a0fb4c632a17e65885/src/QsCompiler/SyntaxProcessor/SymbolOccurrence.fs#L140
[2]: https://github.com/microsoft/qsharp-compiler/blob/4c176e5f07e9ae75c368b9a0fb4c632a17e65885/src/QsCompiler/CompilationManager/EditorSupport/CodeActions.cs#L196
[3]: https://github.com/microsoft/qsharp-compiler/blob/4c176e5f07e9ae75c368b9a0fb4c632a17e65885/src/QsCompiler/CompilationManager/TypeChecking.cs#L136

When I was originally writing this code change, I spent a while trying to trace back where the conception of a null alias was coming from, and found it went all the way back to the fragment kind definition, which is why I changed the definition of an OpenDirective fragment kind. But I understand your point that a fragment is supposed to represent the literal source of the program, and if `""` isn't in the literal source, it shouldn't be a QsSymbol by convention. I'm guessing that despite the name, QsNullable exists as an alternative to having to deal with F# Options in C# and not as a way to represent something called "null"

src/QsCompiler/DataStructures/SyntaxTokens.fs
swernli: Great, thanks for addressing that! With this change and the change putting back the diagnostics the e2e builds are now passing again 👍

swernli (approved):

PR #1531, opened 2022-08-12
Fix assembly constants not getting applied at build time

anpaz:
As discovered during the July 2022 release, the assembly constant 'SIGNED:TELEMETRY' defined in the QDK build ([here](https://github.com/microsoft/qdk/blob/main/build/qdk.template.yml#L97)) is not propagating as a build time flag 'TELEMETRY' in the [QSharpVsix](https://github.com/microsoft/qsharp-compiler/blob/main/src/VisualStudioExtension/QSharpVsix/Telemetry.cs#L49) and [LanguageServer](https://github.com/microsoft/qsharp-compiler/blob/main/src/QsCompiler/LanguageServer/LanguageServer.cs#L180) projects.

I have find the root cause for this. Powershell now enforces `$args` is read-only. Our script was using this as a variable to provide extra arguments to the build command, but now the value is not longer applied so the flags were not properly set.
Fix simply consists on changing the name of the variable.

### DIFF
--- a/build/build.ps1
+++ b/build/build.ps1
@@ -22,14 +22,14 @@ function Build-One {
 
     Write-Host "##[info]Building $project ..."
     if ("" -ne "$Env:ASSEMBLY_CONSTANTS") {
-        $extra_args = @("/property:DefineConstants=$Env:ASSEMBLY_CONSTANTS");
+        $extraArgs = @("/property:DefineConstants=$Env:ASSEMBLY_CONSTANTS");
     }  else {
-        $extra_args = @();
+        $extraArgs = @();
     }
     dotnet build (Join-Path $PSScriptRoot $project) `
         -c $Env:BUILD_CONFIGURATION `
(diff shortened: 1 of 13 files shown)


build/build.ps1
```diff
@@ -22,14 +22,14 @@ function Build-One {
 
     Write-Host "##[info]Building $project ..."
     if ("" -ne "$Env:ASSEMBLY_CONSTANTS") {
-        $args = @("/property:DefineConstants=$Env:ASSEMBLY_CONSTANTS");
+        $extra_args = @("/property:DefineConstants=$Env:ASSEMBLY_CONSTANTS");
```
cgranade, line 25: [nit, please feel free to ignore] Generally, PowerShell style for variables is to use `$camelCase`, such that I would suggest `$extraArgs` here.

cgranade (approved):

ricardo-espinoza (approved):

PR #1532, opened 2022-08-12
Treat magic commands as comments in notebooks

ausbin:
This is a rebase of the magic command code from #1457 against the most recent `feature/langserver-notebooks` branch. It treats magic commands as comments _if_ they are the first non-whitespace character in the cell. That requires some state tracking which isn't too beautiful

This may or may not be a good idea to merge, but I'm posting this for discussion/for reference at least

## Testing
* Ran unit tests for all of `Compiler.sln` and also the language server tests (both of which include new tests)
* Did manual test locally with Azure Notebooks


swernli (approved):
I think this looks good for merge to the feature branch!

PR #1541, opened 2022-08-18
Auto-open Canon and Intrinsic in notebooks

ausbin:
Pretty simple change to match IQ# behavior of auto-opening Canon and Intrinsic until we implement preludes. I think this has been rejected but posting it anyway

## Testing
* Ran unit tests locally for compiler and language server
* Tested locally with Azure Notebooks


swernli:
See proposed approach in https://github.com/microsoft/qsharp-compiler/issues/1550

PR #1552, opened 2022-09-01
Q#-Compiler - Fullstate Simulator V1 Output Format

ScottCarda-MS:
1. Adds the `LoadFrom` public function to allow the Runtime to load a `BitcodeModule` from qir bitcode.
2. Fixes issue where OutputRecording definitions' parameter construction had mixed up the case where there was no parameters with the case where there were non-zero parameters.
3. Replaces TargetCapability with assembly constants dictionary in the `GenerateBitcode` and `GenerateLlvmIR` API parameters so that we can use both the target capability and the process architecture.
4. When targeting the Microsoft Full State Simulator (specified by process architecture), runs the Add Output Recording rewrite step.
5. Adds the Replace Message Calls transformation to the Add Output Recording rewrite step. This transformation replaces Message references to the API `RecordMessage`, which gets translated into the `__quantum__rt__message_record_output` in QIR.


src/QsCompiler/Tests.Compiler/ExecutionTests.fs
```diff
@@ -252,7 +252,7 @@ type ExecutionTests(output: ITestOutputHelper) =
         Assert.Equal(0, exitCode)
 
 
-    [<Fact>]
+    [<Fact(Skip = "ToDo: Test fails due to corrupted memory issue.")>]
```
samarsha, line 255: This test passes on main, why does it need to be skipped?

ScottCarda-MS: If you have some ideas as to why the test is failing, that would be very helpful here.

samarsha: I'm not sure. But it does seem to be caused by this PR, so I think it should be investigated instead of skipping the test.

`Microsoft__Quantum__Testing__ExecutionTests__TestNativeTypeHandling` is segmentation faulting when there's a call to `__quantum__rt__message_record_output`. Here's an MRE (removing the record output call fixes the segfault):

```llvm
; ModuleID = 'src/QsCompiler/Tests.Compiler/bin/Debug/net6.0/outputFolder/TargetedExecutionTests.bc'
source_filename = "src/QsCompiler/Tests.Compiler/bin/Debug/net6.0/outputFolder/TargetedExecutionTests.ll"

%String = type opaque

@0 = internal constant [3 x i8] c", \00"

define void @Microsoft__Quantum__Testing__ExecutionTests__TestNativeTypeHandling() #0 {
entry:
  %0 = call %String* @__quantum__rt__string_create(i8* getelementptr inbounds ([3 x i8], [3 x i8]* @0, i32 0, i32 0))
  call void @__quantum__rt__message_record_output(%String* %0)
  ret void
}

declare %String* @__quantum__rt__string_create(i8*)

declare void @__quantum__rt__message_record_output(%String*)

attributes #0 = { "EntryPoint" }
```

Does the JIT context used by the execution tests define `__quantum__rt__message_record_output`?

samarsha (dismissed):
I want to understand why the "QIR native llvm type handling" execution test is failing.

src/QsCompiler/Tests.Compiler/ExecutionTests.fs
samarsha: > Does the JIT context used by the execution tests define `__quantum__rt__message_record_output`?

Seems like no:

```
(gdb) info symbol __quantum__rt__string_create
__quantum__rt__string_create in section .text of /home/samarsha/Code/microsoft/qsharp-compiler/src/QsCompiler/TestTargets/Simulation/Target/bin/Debug/net6.0/runtimes/linux-x64/native/libMicrosoft.Quantum.Qir.Runtime.so
(gdb) info symbol __quantum__rt__message_record_output
No symbol "__quantum__rt__message_record_output" in current context.
```

If this is a new symbol that was added to the runtime, the library version used by the execution tests probably needs to be updated?

ScottCarda-MS: Hmm, I tried updating the Microsoft.Quantum.Simulators version, but that didn't seem to address the issue. Am I looking the the wrong place?

samarsha: I recommend first checking which library is supposed to define `__quantum__rt__message_record_output` and which version it was added in. Then finding out how that's being loaded into the testing JIT context and updating it.

ScottCarda-MS: I've added that to the runtime in this PR: https://github.com/microsoft/qsharp-runtime/pull/1079. It is not yet merged into main as it relies on this compiler PR to be complete.

samarsha: The options are to either bring in a pre-release package of that qsharp-runtime branch into qsharp-compiler, or to avoid testing the runtime behavior of `__quantum__rt__message_record_output` in qsharp-compiler.

Someone more familiar with this feature can probably talk about what the right testing strategy is better than I can. @swernli owns the QIR codegen files.

ScottCarda-MS: After talking with @swernli, he suggested that the issue should be resolved by creating a fresh build of the beta packages using the changes in both the runtime and compiler PRs and referencing those packages here.

ScottCarda-MS: Updating the version to use a beta package fixed the issues with this test.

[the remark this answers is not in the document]
ScottCarda-MS: The test was fixed.

src/QsCompiler/Tests.Compiler/ExecutionTests.fs
```diff
@@ -236,7 +236,7 @@ type ExecutionTests(output: ITestOutputHelper) =
         assertEqual expected out
 
 
-    [<Fact>]
+    [<Fact(Skip = "Temporary failure.")>]
     member this.``QIR memory management``() =
```
samarsha, line 240: Can this skip be removed too?

ScottCarda-MS: Woops, I removed that locally and forgot to push that. It is removed now.

PR #1554, opened 2022-09-09
Hardware capabilities as warning.

anpaz:
This changes the compiler to report missing hardware capabilities as warnings instead of errors.
Sometimes we still want compilation to succeeded for running new tests or as the hardware capabilities expand.
On a next iteration, we'll add to the compiler the ability to treat warnings as errors (similar to https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/errors-warnings), to finally, make these warnings by treated as errors by default.

### DIFF
--- a/src/QsCompiler/SyntaxProcessor/CapabilityInference/Analyzer.fs
+++ b/src/QsCompiler/SyntaxProcessor/CapabilityInference/Analyzer.fs
@@ -24,11 +24,10 @@ module Target =
     let create name capability =
         { _Name = name; _Capability = capability }
 
-type 'Props Pattern =
+type Pattern =
     {
         Capability: TargetCapability
         Diagnose: Target -> QsCompilerDiagnostic option
-        Properties: 'Props
     }
 
 module Pattern =
@@ -36,7 +35,6 @@ module Pattern =
         {
             Capability = pattern.Capability
             Diagnose = pattern.Diagnose
-            Properties = ()
         }
 
     let concat patterns =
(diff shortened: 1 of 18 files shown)


src/QsCompiler/SyntaxProcessor/CapabilityInference/Capabilities.fs
```diff
@@ -18,57 +16,11 @@ let syntaxAnalyzer callableKind =
                       ResultAnalyzer.analyzer callableKind
                       TypeAnalyzer.analyzer ]
 
-// TODO: Remove this function as part of https://github.com/microsoft/qsharp-compiler/issues/1448.
```
samarsha, line 21: Could you also update Pattern to remove the 'Props type parameter and the Properties field? It shouldn't be needed anymore:

https://github.com/microsoft/qsharp-compiler/blob/f5f492db61baed935e9d1c2bcb5662f06856c6b0/src/QsCompiler/SyntaxProcessor/CapabilityInference/Analyzer.fsi#L23-L29

anpaz:
> So, the overview of this is that hardware capabilities errors will now be warnings? And in the future, the compiler will have a property to treat these warnings as errors, but they'll keep being warnings by default?

@cesarzc, they'll always be reported as warnings by the compiler; however the Quantum SDK will mark them as errors by default.

src/QsCompiler/SyntaxProcessor/CapabilityInference/Analyzer.fs
```diff
@@ -24,31 +24,22 @@ module Target =
...
+type Pattern = { Capability: TargetCapability; Diagnose: Target -> QsCompilerDiagnostic option }
 
 module Pattern =
     let discard pattern =
-        {
-            Capability = pattern.Capability
-            Diagnose = pattern.Diagnose
-            Properties = ()
-        }
+        { Capability = pattern.Capability; Diagnose = pattern.Diagnose }
```
samarsha, line 31: This function can be removed too; it only acted on the Properties field, so it doesn't do anything anymore.

cesarzc (approved):

PR #1561, opened 2022-09-20
Add support for microsoft-qc provider

xinyi-joffre:
Add support for new microsoft-qc provider.

This new provider, called Microsoft Quantum Computing will contain all future targets for Microsoft's first party quantum offerings. With targets such as:

- microsoft.estimator
- microsoft.sim.*
- microsoft.qpu.*

For now, maintain support for older provider called Microsoft.Simulator (soon to be deprecated) and older targets such as:

- microsoft.simulator.*

### DIFF
--- a/src/QsCompiler/DataStructures/ReservedKeywords.fs
+++ b/src/QsCompiler/DataStructures/ReservedKeywords.fs
@@ -267,6 +267,7 @@ module AssemblyConstants =
     let QCIProcessor = "QCIProcessor"
     let QuantinuumProcessor = "QuantinuumProcessor"
     let RigettiProcessor = "RigettiProcessor"
+    let MicrosoftQuantum = "MicrosoftQuantum"
     let MicrosoftSimulator = "MicrosoftSimulator"
     let ExecutionTarget = "ExecutionTarget"
     let TargetCapability = "TargetCapability"
--- a/src/QsCompiler/DataStructures/ReservedKeywords.fs
+++ b/src/QsCompiler/DataStructures/ReservedKeywords.fs
@@ -268,7 +268,6 @@ module AssemblyConstants =
     let QuantinuumProcessor = "QuantinuumProcessor"
     let RigettiProcessor = "RigettiProcessor"
     let MicrosoftQuantum = "MicrosoftQuantum"
-    let MicrosoftSimulator = "MicrosoftSimulator"
     let ExecutionTarget = "ExecutionTarget"
     let TargetCapability = "TargetCapability"
     let DefaultSimulator = "DefaultSimulator"
--- a/src/QuantumSdk/DefaultItems/DefaultItems.targets
+++ b/src/QuantumSdk/DefaultItems/DefaultItems.targets
@@ -28,7 +28,7 @@
     <ValidOutputTypes>Possible values are 'Exe', or 'Library'.</ValidOutputTypes>
   </PropertyGroup>
 
-  <!-- Initializing the ResolvedProcessorArchitecture to either IonQProcessor, QCIProcessor, QuantinuumProcessor, RigettiProcessor, MicrosoftSimulator or Unspecified.-->
+  <!-- Initializing the ResolvedProcessorArchitecture to either IonQProcessor, QCIProcessor, QuantinuumProcessor, RigettiProcessor, MicrosoftQuantum, MicrosoftSimulator or Unspecified.-->
   <PropertyGroup>
     <!-- These architectures are in production and available for public use. -->
     <ResolvedProcessorArchitecture Condition="$(ExecutionTarget.ToLowerInvariant().Contains('ionq'))">IonQProcessor</ResolvedProcessorArchitecture>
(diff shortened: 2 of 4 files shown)


src/QsCompiler/DataStructures/ReservedKeywords.fs
```diff
@@ -267,6 +267,7 @@ module AssemblyConstants =
     let QCIProcessor = "QCIProcessor"
     let QuantinuumProcessor = "QuantinuumProcessor"
     let RigettiProcessor = "RigettiProcessor"
+    let MicrosoftQuantum = "MicrosoftQuantum"
```
cesarzc, line 270: I don't think there's a need to introduce a new processor architecture. `MicrosoftQuantum` as a target might encompass multiple processor architectures but for now there's no need to introduce a new one.

src/QuantumSdk/DefaultItems/DefaultItems.targets
```diff
@@ -28,18 +28,19 @@
...
     <ResolvedProcessorArchitecture Condition="$(ExecutionTarget.ToLowerInvariant().Contains('ionq'))">IonQProcessor</ResolvedProcessorArchitecture>
     <ResolvedProcessorArchitecture Condition="$(ExecutionTarget.ToLowerInvariant().Contains('qci'))">QCIProcessor</ResolvedProcessorArchitecture>
     <ResolvedProcessorArchitecture Condition="$(ExecutionTarget.ToLowerInvariant().Contains('quantinuum'))">QuantinuumProcessor</ResolvedProcessorArchitecture>
     <ResolvedProcessorArchitecture Condition="$(ExecutionTarget.ToLowerInvariant().Contains('honeywell'))">QuantinuumProcessor</ResolvedProcessorArchitecture>
     <ResolvedProcessorArchitecture Condition="$(ExecutionTarget.ToLowerInvariant().Contains('rigetti'))">RigettiProcessor</ResolvedProcessorArchitecture>
+    <ResolvedProcessorArchitecture Condition="$(ExecutionTarget.ToLowerInvariant().Contains('microsoft')) And !$(ExecutionTarget.ToLowerInvariant().Contains('microsoft.simulator'))">MicrosoftQuantum</ResolvedProcessorArchitecture>
     <ResolvedProcessorArchitecture Condition="$(ExecutionTarget.ToLowerInvariant().Contains('microsoft.simulator'))">MicrosoftSimulator</ResolvedProcessorArchitecture>
```
cesarzc, line 40:
```suggestion
    <ResolvedProcessorArchitecture Condition="$(ExecutionTarget.ToLowerInvariant().Contains('microsoft.estimator')) Or $(ExecutionTarget.ToLowerInvariant().Contains('microsoft.sim')) Or $(ExecutionTarget.ToLowerInvariant().Contains('microsoft.simulator'))">MicrosoftSimulator</ResolvedProcessorArchitecture
```

bettinaheim (changes requested):
What exactly is this change for? 
I don't think there is currently a processor architecture that would fall in this category and the chosen naming convention is also unclear to me. I see that the instruction set and the profile matches the MicrosoftSimulator definition. Why not resolve the processor architecture to MicrosoftSimulator?

src/QuantumSdk/DefaultItems/DefaultItems.targets
cesarzc: Execution target can be handled this way when not introducing a new processor architecture.

cesarzc:
> What exactly is this change for? I don't think there is currently a processor architecture that would fall in this category and the chosen naming convention is also unclear to me. I see that the instruction set and the profile matches the MicrosoftSimulator definition. Why not resolve the processor architecture to MicrosoftSimulator?

This work is to introduce a new target naming convention.

I made a similar [comment](https://github.com/microsoft/qsharp-compiler/pull/1561#discussion_r974877604) and suggested this could be handled by resolving the new target names to the `MicrosoftSimulator` architecture (see [here](https://github.com/microsoft/qsharp-compiler/pull/1561#discussion_r974878863)).

bettinaheim:
> > What exactly is this change for? I don't think there is currently a processor architecture that would fall in this category and the chosen naming convention is also unclear to me. I see that the instruction set and the profile matches the MicrosoftSimulator definition. Why not resolve the processor architecture to MicrosoftSimulator?
> 
> This work is to introduce a new target naming convention.
> 
> I made a similar [comment](https://github.com/microsoft/qsharp-compiler/pull/1561#discussion_r974877604) and suggested this could be handled by resolving the new target names to the `MicrosoftSimulator` architecture (see [here](https://github.com/microsoft/qsharp-compiler/pull/1561#discussion_r974878863)).

Got it. If this is only about naming conventions for targets, then the only change should be to check if the ExecutionTarget matches the new naming convention and set the ResolvedProcessorArchitecture to the same value as the current naming is set to. There is no need to add anything else.

src/QsCompiler/DataStructures/ReservedKeywords.fs
xinyi-joffre: Renamed all instances of MicrosoftSimulator to MicrosoftQuantum so that there is no leakage of old architecture with new provider

cesarzc (approved):
I liked how it turned out!

PR #1564, opened 2022-10-13
Introduce new Quantinuum proj/app templates and remove Honeywell

vxfield:
### Context
In November 2021, Honeywell became Quantinuum: Press release: [Quantinuum is a result of the combination of two global leaders in quantum computing: Honeywell Quantum Solutions and Cambridge Quantum](https://www.quantinuum.com/pressrelease/introducing-quantinuum)

In February/March 2022, Azure Quantum started supporting the Quantinuum provider and targets.

### This PR
This PR adds new Quantinuum .NET/C#/Q# project templates, removes the project templates and other obsolete code related to Honeywell.
Also uses the latest target names from https://learn.microsoft.com/en-us/azure/quantum/provider-quantinuum


src/ProjectTemplates/Quantum.Quantinuum.App1/.template.config/template.json.v.template
```diff
@@ -0,0 +1,15 @@
+{
```
kuzminrobin, line 1: Looks like inside of the dir "src/ProjectTemplates" you renamed the dir 
"Quantum.**Honeywell**.App1" to 
"Quantum.**Quantinuum**.App1" _manually_ or using an IDE (rather than using `git mv`). That's why the file ".template.config/template.json.v.template" inside of the renamed dir is shown in the diff as fully removed and then new file added. As a result the history of the file changes (and line changes in it) gets lost.

If situation above is the case then consider using the `git mv` to rename the dir. Then commit (without making any other edits). Then make changes to the file. Then commit. To summarize, keep renaming in a commit separate from file changes.

kuzminrobin (approved):
Approving to unblock. 

I would recommend more approvals for the renaming that is not evident for me.
```diff
-   <ExecutionTarget>honeywell.hqs-lt-1.0</ExecutionTarget>
---
+   <ExecutionTarget>quantinuum.qpu.h1</ExecutionTarget>
```

src/ProjectTemplates/Quantum.Quantinuum.App1/.template.config/template.json.v.template
vxfield: Thanks, Great point, I'll revert the changes and do as suggest.

vxfield:
> Approving to unblock.
> 
> I would recommend more approvals for the renaming that is not evident for me.
> 
> ```diff
> -   <ExecutionTarget>honeywell.hqs-lt-1.0</ExecutionTarget>
> ---
> +   <ExecutionTarget>quantinuum.qpu.h1</ExecutionTarget>
> ```

Thanks. These are the new Quantinuum targets that just got renamed.
@warren-jones worked on them and the documentation is here: https://learn.microsoft.com/en-us/azure/quantum/provider-quantinuum

vxfield:
Closed in favor of #1565

PR #1570, opened 2022-10-28
2022.10 Hotfix: Updated to 0.27.238334.

kuzminrobin:


examples/QIR/Emission/Emission.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.Quantum.Sdk/0.25.228311">
+<Project Sdk="Microsoft.Quantum.Sdk/0.25.228311">   <!-- TODO(#1569): Requires refactoring for Rust QIR RT. -->
```
billti, line 1: All these TODO comments that got added are interesting. I assume they have no runtime impact?

billti (approved):

examples/QIR/Emission/Emission.csproj
kuzminrobin: They should not have runtime impact. At least since mid-September (when the Rust QIR RT was merged) no any impact was observed.

PR #1574, opened 2022-11-05
Replace generate Antlr files by msbuild task

rstm-sf:
Try using https://github.com/kaby76/Antlr4BuildTasks instead of explicit java and bump Antlr4.Runtime.Standard version

And bump Antlr version to 4.11.1


rstm-sf:
We can use JavaDownloadDirectory parameter

samarsha:
Hi - we've moved all development to a new version of the Q# compiler and QDK, which is in another repo: https://github.com/microsoft/qsharp. This repo is in "maintenance mode" and we're trying to avoid making unnecessary changes. It looks like this change isn't related to fixing a bug, so we can't accept it at this time, but thanks for your contribution!

PR #1588, opened 2022-12-15
Conditionally treat hardware capability errors as warnings

DmitryVasilevsky:
Adding option TreatCapabilityErrorsAsWarnings (command line argument treat-capability-errors-as-warnings). When set, it will instruct compiler to emit warnings instead of errors for target capabilities. Affected diagnostics:

UnsupportedResultComparison = 5023
ResultComparisonNotInOperationIf = 5024
ReturnInResultConditionedBlock = 5025
SetInResultConditionedBlock = 5026
UnsupportedCallableCapability = 5027
UnsupportedClassicalCapability = 5028

Current approach propagates this flag into F# type Target in module Analyzer. Then the flag is used to issue errors or warnings.
(Alternative approach is to convert specific errors to warnings in C# code after they are collected).


src/QsCompiler/CompilationManager/ProjectManager.cs
```diff
@@ -75,6 +75,25 @@ public class ProjectProperties
...
+
+                if (!bool.TryParse(errorsAsWarnings, out bool result))
+                {
+                    return false;
+                }
+
+                return result;
+            }
+        }
+
```
ScottCarda-MS, line 96: TryParse is guaranteed to return `false` when given `null` so the logic can be simplified a bit. Consider:
```suggestion
        public bool TreatCapabilityErrorsAsWarnings =>
			this.BuildProperties.TryGetValue(MSBuildProperties.TreatCapabilityErrorsAsWarnings, out var errorsAsWarnings)
			&& bool.TryParse(errorsAsWarnings, out var result)
			&& result;

```

ScottCarda-MS (commented):
Nice work! Just have a couple of comments.

src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -637,6 +647,28 @@ type DiagnosticItem =
             | ErrorCode.ExpectingCallableExpr ->
                 "The type of the expression must be a function or operation type. The given expression is of type {0}."
             | ErrorCode.UnknownIdentifier -> "No identifier with the name \"{0}\" exists."
+
+            // TODO: When the names of the runtime capabilities are finalized, they can be included in the result
```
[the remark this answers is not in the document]
DmitryVasilevsky: This is a good question. Maybe this TODO is no longer needed. It was in the original code before the change to change capability errors into warnings unconditionally. It is in the main branch now (https://github.com/microsoft/qsharp-compiler/blob/4a5aafab2b4c28e0c9218680f366cb90415f1412/src/QsCompiler/DataStructures/Diagnostics.fs#L645). So maybe this change, which I'm making now, actually resolves it.

src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -874,25 +906,21 @@ type DiagnosticItem =
                 "This expression may be short-circuited, and operation calls may not be executed."
             | WarningCode.DeprecationWithRedirect -> "{0} has been deprecated. Please use {1} instead."
             | WarningCode.DeprecationWithoutRedirect -> "{0} has been deprecated."
+
+            // Reuse error messages and hope that the expanded message doesn't contain {}.
             | WarningCode.UnsupportedResultComparison ->
-                "The target {0} does not support comparing measurement results."
+                DiagnosticItem.Message(ErrorCode.UnsupportedResultComparison, args)
```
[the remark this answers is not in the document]
DmitryVasilevsky: This is similar to how it was done in the original change and is also what I'm stumbled upon. The goal here is to reuse corresponding error message for a warning. We don't have a separate function to get a string without parameters applied. Therefore, we get a string with parameters applied. Later no parameters will be applied because expanded string doesn't have {} (hopefully). I don't like this, just followed the same approach as in the original change that I got as a reference. Since you also raised this question, I will change the code and add a function that would produce unexpanded error strings.

src/QsCompiler/SyntaxProcessor/CapabilityInference/Analyzer.fs
```diff
@@ -13,16 +13,23 @@ type Target =
...
         _Name: string
         _Capability: TargetCapability
+        _UseWarnings: bool
     }
 
     member target.Name = target._Name
 
     member target.Capability = target._Capability
 
+    member target.UseWarnings = target._UseWarnings
```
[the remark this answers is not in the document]
DmitryVasilevsky: Renamed to UseCapabilityWarnings. Not sure if this is public facing. We only use it in CapabilityDiagnostics in TypeChecking.cs. It is controlled by other public facing properties.

src/QsCompiler/DataStructures/Diagnostics.fs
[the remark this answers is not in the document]
DmitryVasilevsky: Added member to reuse capability error strings.

src/QsCompiler/CommandLineTool/Options.cs
```diff
@@ -93,6 +93,13 @@ public class CompilationOptions : Options
             HelpText = "The capability of the compilation target. This determines which QIR profile to compile to.")]
         public string? TargetCapabilityName { get; set; }
 
+        [Option(
+            "treat-capability-errors-as-warnings",
+            Required = false,
+            Default = false,
+            HelpText = "Specifies whether hardware capabilities should be treated as warnings.")]
```
kuzminrobin, line 100:
```suggestion
            HelpText = "Specifies whether hardware capability errors should be treated as warnings.")]
```

src/QsCompiler/SyntaxProcessor/CapabilityInference/ResultAnalyzer.fs
```diff
@@ -42,30 +42,55 @@ let createPattern kind range =
...
+                if target.Capability.ResultOpacity = ResultOpacity.opaque then
+                    None
+                else
+                    Some(WarningCode.SetInResultConditionedBlock, [ name; target.Name ])
+            |> Option.map (fun (code, args) -> QsCompilerDiagnostic.Warning (code, args) range)
+
+        else
+            let range = QsNullable.defaultValue Range.Zero range
+
+            match kind with
```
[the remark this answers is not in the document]
DmitryVasilevsky: We've discussed an alternative approach with Sarah. This alternative would keep F# code unchanged generating errors unconditionally. Then it will conditionally change the severity of some errors to warnings if necessary.
Each approach has merits and drawbacks. I tried this approach in another (private) branch and I didn't like it much. First, I didn't find a good way to add tests. Second, the F# code would always generate errors so it would provide different results if used by itself. Third, if some action is taken before the translation, it would use errors rather than warnings.
So I'm in favor of creating errors or warnings conditionally from the start.

src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -410,6 +420,30 @@ type DiagnosticItem =
         with
         | _ -> str // let's fail silently for now
 
+    // TODO: When the names of the runtime capabilities are finalized, they can be included in the result
```
kuzminrobin, line 423: Consider providing either an alias or an item number or both in TODO:
```suggestion
    // TODO (DmitryVasilevsky, #<ItemNumber>): When the names of the runtime capabilities are finalized, they can be included in the result
```

src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -410,6 +420,30 @@ type DiagnosticItem =
...
+    //       comparison error messages.
+    static member private CapabilityErrorFormatString(code) =
+        match code with
+        | ErrorCode.UnsupportedResultComparison -> "The target {0} does not support comparing measurement results."
```
kuzminrobin, line 427: (nit)
At certain stage of my career I came to an impression that it is not always a good idea to provide a dot (full stop) `.` in the end of a message, because some subsequent code may want to surround the message with quotes `"` or single quotes `'` or parentheses `()`, etc. E.g. the subsequent code may want to print:
> The process failed (The target MyTestTarget does not support comparing measurement results.). 

In the fragment above the presence of a dot `.` in front of the closing parenthesis `)` is not what some users would expect to see.

Consider not providing the dot `.` in the end of the messages (but in the middle of the multi-sentence messages seems reasonable).

src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -410,6 +420,30 @@ type DiagnosticItem =
...
+            "The variable \"{0}\" cannot be reassigned here. "
+            + "In conditional blocks that depend on a measurement result, the target {1} only supports reassigning variables that were declared within the block."
+        | ErrorCode.UnsupportedCallableCapability ->
+            "The callable {0} requires runtime capabilities which are not supported by the target {1}."
+            + Environment.NewLine
+            + "Result Opacity: {2}"
+            + Environment.NewLine
+            + "Classical Capability: {3}"
+        | ErrorCode.UnsupportedClassicalCapability ->
+            "This construct requires a classical runtime capability that is not supported by the target {0}: {1}."
```
kuzminrobin, line 444: (nit) Consider
```suggestion
        | ErrorCode.UnsupportedClassicalCapability ->
            "This construct requires a classical runtime capability {1} that is not supported by the target {0}."
```

src/QsCompiler/DataStructures/Diagnostics.fs
```diff
@@ -874,25 +923,21 @@ type DiagnosticItem =
                 "This expression may be short-circuited, and operation calls may not be executed."
             | WarningCode.DeprecationWithRedirect -> "{0} has been deprecated. Please use {1} instead."
             | WarningCode.DeprecationWithoutRedirect -> "{0} has been deprecated."
+
+            // Capability errors
             | WarningCode.UnsupportedResultComparison ->
-                "The target {0} does not support comparing measurement results."
+                DiagnosticItem.CapabilityErrorFormatString(ErrorCode.UnsupportedResultComparison)
```
kuzminrobin, line 929: Consider
```suggestion
                DiagnosticItem.CapabilityErrorFormatString(WarningCode.UnsupportedResultComparison)
```

I have an impression that currently the `ErrorCode.UnsupportedResultComparison` and the `WarningCode.UnsupportedResultComparison` have the same value. But theoretically in the future they can get different values.

src/QsCompiler/SyntaxProcessor/CapabilityInference/ResultAnalyzer.fs
```diff
@@ -42,30 +42,55 @@ let createPattern kind range =
...
+                    Some(WarningCode.ReturnInResultConditionedBlock, [ target.Name ])
+            | SetInDependentBranch name ->
+                if target.Capability.ResultOpacity = ResultOpacity.opaque then
+                    None
+                else
+                    Some(WarningCode.SetInResultConditionedBlock, [ name; target.Name ])
+            |> Option.map (fun (code, args) -> QsCompilerDiagnostic.Warning (code, args) range)
+
+        else
+            let range = QsNullable.defaultValue Range.Zero range
```
kuzminrobin, line 71: Just curiosity.
Is it possible to extract the line `let range = QsNullable.defaultValue Range.Zero range` (L71 and L46) to before the `if` (before the L45)?

kuzminrobin (approved):
Approving to unblock. But nearly all the code is unfamiliar to me, I would recommend at least one more approval.

src/QsCompiler/DataStructures/Diagnostics.fs
DmitryVasilevsky: Interesting observation. However, there're about other 500 error messages in this file and all of them have period in the end. I'd say it's better to be uniform in this case.

src/QsCompiler/DataStructures/Diagnostics.fs
DmitryVasilevsky: The whole point of this block it to reuse the same string for errors and warnings.

src/QsCompiler/SyntaxProcessor/CapabilityInference/ResultAnalyzer.fs
DmitryVasilevsky: Definitely possible. Updated code.

src/QsCompiler/DataStructures/Diagnostics.fs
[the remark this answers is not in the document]
DmitryVasilevsky: Maybe. However, arguments are present in numerical order in all messages. I'm not sure if changing this would make it better.

DmitryVasilevsky: It would be interesting to see @samarsha opinion on this. These messages were already present in the code, they were already approved by other reviewers, I just moved them to a different place in the file. The proposed change makes it sound better, but there could be other reasons to phrase the way it is.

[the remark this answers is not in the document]
kuzminrobin: Yeah, I later noticed that the sequences `target {0}`, `target {1}` rely on the already existing code. It is better to leave as is. Agreed.

samarsha (changes requested):
I still have concerns with this approach. In my opinion, warning vs. error is something that only the surface (UI) layer of the compiler should know about. The internals shouldn't need to be concerned with it, because it doesn't affect how the static analysis is done - the only difference is whether ErrorCode.Foo or WarningCode.Foo is emitted.

However, with this design, the knowledge of whether the top layer is requesting warnings or errors is propogated deep into the compiler. I think this adds complexity that's proportional to the number of capability diagnostics and the number of places that either generate or test a specific diagnostic code. The changes are also specific to the capability diagnostics and don't generalize.

For example, an alternative design could be:

- The compiler always emits capability diagnostics as warnings.
- There's a compiler option to turn any diagnostic code from a warning to an error. (Most compilers support this.)

In this case, the changes would be orthogonal to the capability diagnostics - neither side needs to know about the other, keeping a separation of concerns. Also, the changes could be useful in more scenarios (where warnings from something other than the capability analyzers can be turned into errors) without any changes.

swernli:
> * There's a compiler option to turn any diagnostic code from a warning to an error. (Most compilers support this.)

To that end, see for example the [WarningsAsErrors](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/errors-warnings#warningsaserrors-and-warningsnotaserrors) feature of the C# compiler or the [Lint Levels](https://doc.rust-lang.org/rustc/lints/levels.html#via-compiler-flag) feature of the Rust compiler. It's a common compiler pattern to allow someone to set all warnings, warnings of a specific class, or individual warnings identified by code to be treated as errors for a given compilation. I think that pattern is more standard and flexible than duplicating the codes among both the error and warning enum.

DmitryVasilevsky:
An alternative approach has been implemented.
https://github.com/microsoft/qsharp-compiler/pull/1594

PR #1563, opened 2022-10-02
Changing the representation of stack allocated arrays to allow for array valued command line arguments

bettinaheim:
This PR changes the representation of stack allocated arrays. The new representation allows to retroactively define the correct array length as part of a subsequent compilation step. For more details, read the comments in the code. 
Existing tests should cover existing functionality. Adding tests for the new functionality requires merging https://github.com/microsoft/qsharp-compiler/pull/1554; I'll add the tests once that PR has merged.

PR #1591, opened 2023-01-10
Changing the representation of stack allocated arrays to allow for array valued command line arguments

swernli:
This PR changes the representation of stack allocated arrays. The new representation allows to retroactively define the correct array length as part of a subsequent compilation step. Supersedes #1563.


swernli:
Leaving as draft for now... while I expect this to pass the tests, it still has some issues that need to be addressed in terms of extra load and store instructions that are added to the output QIR that should not be present.

PR #1594, opened 2023-01-18
Conditionally treat capability warnings as errors (alternative approach)

DmitryVasilevsky:
Adding optional --error command line argument to treat specified capability errors as warnings. Accepts a list of warning numbers (space separated).

Language server accepts build property WarningsAsErrors - a list of semicolon separated warning numbers.

Diagnostics is generated as Warnings in F# code, but when the corresponding diagnostic in C# is constructed, its severity is controlled by the list of integers passed as the --error option. If the warning number is among the integers passed, it is generated as error.

### DIFF
--- a/src/QsCompiler/CommandLineTool/Options.cs
+++ b/src/QsCompiler/CommandLineTool/Options.cs
@@ -103,7 +103,7 @@ namespace Microsoft.Quantum.QsCompiler.CommandLineCompiler
             "error",
             Required = false,
             Default = null,
-            HelpText = "Comma-separated warning numbers that should be treated as errors.")]
+            HelpText = "Warning numbers that should be treated as errors.")]
         public IEnumerable<int>? WarningsAsErrors { get; set; }
 
         [Option(
--- a/src/QsCompiler/CommandLineTool/Options.cs
+++ b/src/QsCompiler/CommandLineTool/Options.cs
@@ -104,7 +104,7 @@ namespace Microsoft.Quantum.QsCompiler.CommandLineCompiler
             Required = false,
             Default = null,
             HelpText = "Comma-separated warning numbers that should be treated as errors.")]
-        public string? WarningAsErrorNumbers { get; set; }
+        public IEnumerable<int>? WarningsAsErrors { get; set; }
 
         [Option(
             "build-exe",
--- a/src/QsCompiler/CommandLineTool/Options.cs
+++ b/src/QsCompiler/CommandLineTool/Options.cs
@@ -100,6 +100,13 @@ namespace Microsoft.Quantum.QsCompiler.CommandLineCompiler
             TargetCapability.TryParse(this.TargetCapabilityName) ?? TargetCapabilityModule.Top;
 
         [Option(
+            "error",
+            Required = false,
+            Default = null,
+            HelpText = "Comma-separated warning numbers that should be treated as errors.")]
+        public string? WarningAsError { get; set; }
+
+        [Option(
             "build-exe",
             Required = false,
             Default = false,
--- a/src/QsCompiler/CommandLineTool/Options.cs
+++ b/src/QsCompiler/CommandLineTool/Options.cs
@@ -104,7 +104,7 @@ namespace Microsoft.Quantum.QsCompiler.CommandLineCompiler
             Required = false,
             Default = null,
             HelpText = "Comma-separated warning numbers that should be treated as errors.")]
-        public string? WarningAsError { get; set; }
+        public string? WarningAsErrorNumbers { get; set; }
 
         [Option(
             "build-exe",
--- a/src/QsCompiler/CompilationManager/Diagnostics.cs
+++ b/src/QsCompiler/CompilationManager/Diagnostics.cs
@@ -27,7 +27,7 @@ namespace Microsoft.Quantum.QsCompiler.CompilationBuilder
                 _ => throw new ArgumentException("Unrecognized fragment ending."),
             };
 
-        private static DiagnosticSeverity Severity(QsCompilerDiagnostic msg, HashSet<int>? warningAsErrorNumbers)
+        private static DiagnosticSeverity Severity(QsCompilerDiagnostic msg, HashSet<int>? warningsAsErrors)
         {
             if (msg.Diagnostic.IsError)
             {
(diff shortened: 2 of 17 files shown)


src/QsCompiler/CommandLineTool/Options.cs
```diff
@@ -99,6 +99,13 @@ public class CompilationOptions : Options
         internal TargetCapability TargetCapability =>
             TargetCapability.TryParse(this.TargetCapabilityName) ?? TargetCapabilityModule.Top;
 
+        [Option(
+            "error",
+            Required = false,
+            Default = null,
+            HelpText = "Comma-separated warning numbers that should be treated as errors.")]
+        public string? WarningAsErrorNumbers { get; set; }
```
samarsha, line 107: Why not `IEnumerable<int>`? I think that would avoid needing to manually parse the string?

src/QsCompiler/DataStructures/ReservedKeywords.fs
```diff
@@ -248,6 +248,7 @@ module MSBuildProperties =
     [<Obsolete("Replaced by ResolvedTargetCapability for Microsoft.Quantum.Sdk version 0.25 and newer.")>]
     let ResolvedRuntimeCapabilities = "ResolvedRuntimeCapabilities"
     let ResolvedTargetCapability = "ResolvedTargetCapability"
+    let WarningAsErrorNumbers = "WarningAsErrorNumbers"
```
samarsha, line 251: What about "WarningsAsErrors"?

src/QuantumSdk/Sdk/Sdk.targets
```diff
@@ -97,6 +97,7 @@
       <_QscCommandSkipTargetingFlag Condition="'$(ResolvedQSharpOutputType)' == 'QSharpLibrary' And '$(NoEntryPoint)' != 'true'">$(_NewLine)--skip-targeting</_QscCommandSkipTargetingFlag>
       <_QscCommandSkipMonomorphization Condition="'$(ResolvedTargetCapability)' == 'FullComputation' And $(CSharpGeneration) And !$(EnableQirSubmission)">$(_NewLine)--skip-monomorphization</_QscCommandSkipMonomorphization>
       <_QscCommandPerfDataGenerationFlag Condition="$(PerfDataGeneration)">$(_NewLine)--perf$(_NewLineIndent)$(PerfDataOutputPath)</_QscCommandPerfDataGenerationFlag>
+      <_QscCommandCapabilityWarningsFlag Condition="'$(TreatCapabilityWarningsAsErrors)' == 'true'">$(_NewLine)--error 5023,5024,5025,5026,5027,5028</_QscCommandCapabilityWarningsFlag>
```
samarsha, line 100: I don't see where `TreatCapabilityWarningsAsErrors` is set, how is it intended to be used?

src/QsCompiler/CommandLineTool/Options.cs
DmitryVasilevsky: CompilationLoader constructs CompilationUnitManager with build properties represented as a string-string map. Other items are added via string representation to this collection. Therefore, I'm keeping the string representation until ProjectProperties are constructed.
If I use a collection of ins early on, I'd have to use a different way to construct ProjectProperties. Let's see if it works.

src/QuantumSdk/Sdk/Sdk.targets
DmitryVasilevsky: I expect TreatCapabilityWarningsAsErrors to be set by the user in a project so that capability warnings are treated as errors.

src/QsCompiler/CompilationManager/Diagnostics.cs
```diff
@@ -27,15 +27,22 @@ invalidFragmentEnding switch
...
+        private static DiagnosticSeverity Severity(QsCompilerDiagnostic msg, HashSet<int>? warningsAsErrors)
         {
             if (msg.Diagnostic.IsError)
             {
                 return DiagnosticSeverity.Error;
             }
             else if (msg.Diagnostic.IsWarning)
             {
-                return DiagnosticSeverity.Warning;
+                if (warningsAsErrors != null && warningsAsErrors.Contains(msg.Code))
```
MattMcL4475, line 38: Could use the null conditional operator here:

`if (warningsAsErrors?.Contains(msg.Code) == true)`

src/QsCompiler/Compiler/CompilationLoader.cs
```diff
@@ -549,7 +554,7 @@ public CompilationLoader(SourceLoader loadSources, ReferenceLoader loadReference
             buildProperties.Add(MSBuildProperties.ResolvedProcessorArchitecture, processorArchitecture);
 
             var compilationManager = new CompilationUnitManager(
-                new ProjectProperties(buildProperties),
+                new ProjectProperties(buildProperties, this.config.WarningsAsErrors),
```
samarsha, line 557: There are two scenarios you have to handle: when the compiler is used as a command line program and when it's running as a language server. This handles the CLI case, but I don't think WarningsAsErrors will be initialized properly in the language server because the extra argument is omitted here:

https://github.com/microsoft/qsharp-compiler/blob/2227253bf5b0791d616d3e6b60f592b5db062bc0/src/QsCompiler/CompilationManager/ProjectManager.cs#L159

src/QsCompiler/CommandLineTool/Options.cs
```diff
@@ -99,6 +99,13 @@ public class CompilationOptions : Options
         internal TargetCapability TargetCapability =>
             TargetCapability.TryParse(this.TargetCapabilityName) ?? TargetCapabilityModule.Top;
 
+        [Option(
+            "error",
+            Required = false,
+            Default = null,
+            HelpText = "Comma-separated warning numbers that should be treated as errors.")]
```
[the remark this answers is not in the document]
DmitryVasilevsky: Correct. Fixed description.

src/QsCompiler/Compiler/CompilationLoader.cs
DmitryVasilevsky: OK. I accept it in build properties. Command line options are also written to the build properties.

samarsha (commented):
Everything looks good now - for tests you could consider:

- Tests.Compiler/CommandLineTests.fs for the CLI `--error` option
- Tests.LanguageServer/Tests.cs for language server reading the project property

samarsha (approved):

cesarzc (approved):

PR #1598, opened 2023-01-25
Change decomposition package for IonQ depending on whether QIR generation is enabled

cesarzc:


src/QuantumSdk/DefaultItems/DefaultItems.targets
```diff
@@ -97,7 +87,15 @@
     <QSharpDocsGeneration Condition="'$(QSharpDocsOutputPath)' != '' And '$(QSharpDocsGeneration)' == 'default'">true</QSharpDocsGeneration>
     <QSharpDocsGeneration Condition="'$(QSharpDocsOutputPath)' == '' And '$(QSharpDocsGeneration)' == 'default'">false</QSharpDocsGeneration>
     <CSharpGeneration Condition="$(QirGeneration)">false</CSharpGeneration>
+    <!-- Resolving the QuantumInstructionSet to either Type1, Type2, Type3, Type4 or Default.-->
     <EnableQirSubmission Condition="'$(ResolvedProcessorArchitecture)' == 'MicrosoftQuantum' Or '$(ResolvedTargetCapability)' == 'AdaptiveExecution' Or '$(ResolvedTargetCapability)' == 'BasicExecution'">true</EnableQirSubmission>
+    <ResolvedQuantumInstructionSet Condition="'$(ResolvedProcessorArchitecture)' == 'QuantinuumProcessor'">Type1</ResolvedQuantumInstructionSet>
+    <ResolvedQuantumInstructionSet Condition="'$(ResolvedProcessorArchitecture)' == 'IonQProcessor' And !$(EnableQirSubmission)">Type2</ResolvedQuantumInstructionSet>
```
swernli, line 93: Technically this is not necessary, as no process ever used the Type2 decompositions. You could simplify this by just picking Type4 for IonQ always. At that point, Type2 would be unused and could be removed.

cesarzc: The behavior before these changes was that for IonQ targets Type2 decompositions were used. I left it this way to not break the current default behavior when submitting to the Azure Quantum service. Do you think not including this reference would have no effect on the current behavior?

swernli: I believe it wouldn't affect any current behavior, but if you'd prefer keeping things consistent then I think that's fine. The code as is definitely accomplishes the goal!

swernli (approved):

PR #1607, opened 2023-02-11
Upgraded Newtonsoft.Json dependency.

kuzminrobin:
**Multi-repo PR:**
[Q#C](https://github.com/microsoft/qsharp-compiler/pull/1607), [Q#RT](https://github.com/microsoft/qsharp-runtime/pull/1128), [iQ#](https://github.com/microsoft/iqsharp/pull/770), [Katas](https://github.com/microsoft/QuantumKatas/pull/873), [QLib](https://github.com/microsoft/QuantumLibraries/pull/657), [Q](https://github.com/microsoft/Quantum/pull/781), [Q-NC](https://github.com/microsoft/Quantum-NC/pull/85), private repo, QDK.
DevOps items 48477, 48480, 48478.


ScottCarda-MS (approved):

src/QsCompiler/TestProjects/test10/Driver.cs
```diff
@@ -1,13 +0,0 @@
...
-namespace test3
-{
-    class Driver
-    {
-        static void Main(string[] args)
-        {
-
-        }
-    }
-}
```
billti, line 13: Is this file intentionally deleted as part of the version changes?

src/QsCompiler/TestProjects/test11/test11.csproj
```diff
@@ -1,14 +1,8 @@
...
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp3.1</TargetFramework>
-    <PlatformTarget>x64</PlatformTarget>
   </PropertyGroup>
 
-  <ItemGroup>
-    <PackageReference Include="Microsoft.Quantum.Standard" Version="0.6.1905.301" />
-    <PackageReference Include="Microsoft.Quantum.Development.Kit" Version="0.6.1905.301" />
-  </ItemGroup>
```
billti, line 12: Why are these projects getting changed from the .NET SDK to the Quantum SDK and the packages removed as part of upgrading Newtonsoft?

src/QsCompiler/Tests.LanguageServer/ProjectLoaderTests.cs
```diff
@@ -316,7 +315,6 @@ public async Task LoadQSharpConsoleAppAsync()
                 Path.Combine(projDir, "Operation11.qs"),
             };
 
-            Assert.IsTrue(context.UsesIntrinsics());
```
billti, line 319: Why did these asserts need to be removed?

src/QsCompiler/TestProjects/test15/test15.csproj
```diff
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.Quantum.Sdk/0.19.2109165653">
+﻿<Project Sdk="Microsoft.Quantum.Sdk/0.27.244707">
```
billti, line 1: The above projects are being changed to Quantum SDK "0.27.253010". Why is this being changed to "0.27.244707"?

src/QsCompiler/TestProjects/test10/Driver.cs
ScottCarda-MS: I believe this is intentional as part of clean up. I don't think this Driver.cs is needed anymore. @kuzminrobin if this is a mistake.

src/QsCompiler/TestProjects/test11/test11.csproj
ScottCarda-MS: Removing unnecessary dependencies in the tests seems like a positive.

src/QsCompiler/TestProjects/test15/test15.csproj
ScottCarda-MS: Isn't this just an update to a more up-to-date sdk version?

billti: Yes, but per the comment, why not to the same version as the other projects got updated to?

src/QsCompiler/TestProjects/test10/Driver.cs
kuzminrobin: This file is needed if you try to open the .csproj and compile the project. 
But the test does not copy this file to the test working dir (see the [copying fragment](https://github.com/microsoft/qsharp-compiler/blob/e09f8d7db23a07a3f896e9903d0b06c0452d2b94/src/QsCompiler/Tests.LanguageServer/Tests.LanguageServer.csproj#L33-L40)), the test never tries to compile the test project. This file is never used.

src/QsCompiler/TestProjects/test11/test11.csproj
kuzminrobin: This particular fragment that is being deleted refers to the (very) outdated QDK (that refers the vulnerable external dependencies). I believe this was the way the quantum programs were written before the `Microsoft.Quantum.Sdk` was developed. I upgraded the project according to the [migration guide](https://learn.microsoft.com/en-us/azure/quantum/install-update-qdk).

src/QsCompiler/Tests.LanguageServer/ProjectLoaderTests.cs
kuzminrobin: Because the updated version does not use some `*Intrinsic*.dll` (I don't remember the exact name, hence the wildcard). And that DLL does not exist any more in the later packages.

src/QsCompiler/TestProjects/test15/test15.csproj
kuzminrobin: There was a specific commit regarding this particular change. I don't remember exact details. 
Let me try to unify the version to 0.27.253010. If it works then I'll update the PR.
Thanks for catching.

It was something related to the test that intentionally tests some older functionality. Presumably in [this line](https://github.com/microsoft/qsharp-compiler/blob/9bfde74871358d273198d3752fe466e79c779a0e/src/QsCompiler/Tests.LanguageServer/Tests.cs#L403) and [this fragment](https://github.com/microsoft/qsharp-compiler/blob/9bfde74871358d273198d3752fe466e79c779a0e/src/QsCompiler/Tests.LanguageServer/Tests.cs#L377-L400).

kuzminrobin: Fixed.

src/QsCompiler/TestProjects/test10/Driver.cs
billti: Look at the 'test10' project, the .qs file contains barely anything it anyway. I'm not sure the value of this project/test. Fine I guess.

src/QsCompiler/TestProjects/test11/test11.csproj
billti: Right. But I thought some of the value of these tests was to ensure old projects could still open in the latest builds. Maybe I'm mistaken.

src/QsCompiler/Tests.LanguageServer/ProjectLoaderTests.cs
billti: Whether the DLL is present or not doesn't seem relevant here. Are you saying this method on the context no longer exists, and intrinsics are always used by default now?

billti (approved):

src/QsCompiler/Tests.LanguageServer/ProjectLoaderTests.cs
kuzminrobin: The `UsesIntrinsics()` is defined like [this](https://github.com/microsoft/qsharp-compiler/blob/fa13bf673983247c5f3b570eb78cb0e692e9b859/src/QsCompiler/Tests.LanguageServer/ProjectLoaderTests.cs#L418). 

During debugging I got an impression that this check verifies if the test project still has either the "Microsoft.Quantum.Intrinsic.dll" or "Microsoft.Quantum.Primitives.dll" as a dependency (I believe those are the parts of the very old QDK). After an update of test11 to the new version of QDK (that has no those DLLs any more) the check has started failing. So I decided to remove that check.

If we leave the test11 as is (revert in this PR the test11 change and the change being discussed) then I believe one particular Component Governance issue will continue being reported (the one that you can navigate to from the DevOps item 48477, then search for "test11" in the Component Governance issue).

src/QsCompiler/TestProjects/test11/test11.csproj
kuzminrobin: I'm inclined to agree.

But if we leave the test11 as is (revert in this PR the test11 change) then I believe one particular Component Governance issue will continue being reported (the one that you can navigate to from the DevOps item 48477, then search for "test11" in the Component Governance issue).

src/QsCompiler/TestProjects/test10/Driver.cs
kuzminrobin: Those test projects, I believe, are tested against _loadability_, not _compilability_ or _runnability_. Their main files are .csproj ones (the .qs ones are just for completeness/consistency).

PR #1608, opened 2023-03-02
Update to 0.27.258160

ScottCarda-MS:


src/QsCompiler/TestProjects/test10/test10.csproj
```diff
@@ -1,8 +1,8 @@
-﻿<Project Sdk="Microsoft.Quantum.Sdk/0.27.253010">
+﻿<Project Sdk="Microsoft.Quantum.Sdk/0.27.258160">
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
```
kuzminrobin, line 5: 1. Is this change required for this PR?
2. I remember the `src/QsCompiler/TestProjects` have certain specifics. They test that the old projects can be loaded to the IDE. Some of them must stay old, IIRC.

ScottCarda-MS: The QDK versions associated with these tests were up-to-date prior to the release, so I thought they should be updated to the new release version. The new release version is "not compatible" with netcoreapp3.1 and recommended I update to net6.0. I didn't see anything in the tests that indicated they needed to be frozen at an outdated version. Then again, they are seemingly the exact same test (test10 and test11).
For the future, it would be much more helpful if the tests were named by what they were testing instead of just numbers.

kuzminrobin: OK.

FYI, These are just the test projects, to be more precise they are the test input data for the actual tests. More info/logic about these test projects is in the sources of "src/QsCompiler/Tests.LanguageServer/ProjectLoaderTests.cs" below (but I didn't dive deep into that info/logic either).

kuzminrobin (approved):

PR #1613, opened 2023-03-21
Cleanup after removal of the resources estimator

DmitryVasilevsky:
1. Removed parameter that is no longer used (and switched to a beta package to do that)
2. Removed test that is no longer tests anything.

### DIFF
--- a/src/QsCompiler/Tests.Compiler/TestCases/LocalVerification.qs
+++ b/src/QsCompiler/Tests.Compiler/TestCases/LocalVerification.qs
@@ -1380,8 +1380,6 @@ namespace Microsoft.Quantum.Testing.LocalVerification {
     @Test("QuantumSimulator")
     function ValidTestAttribute10 () : ((Unit)) {}
 
-    function ValidTestAttribute11 (arg : Unit) : Unit { }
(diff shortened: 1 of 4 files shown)


src/QsCompiler/Tests.Compiler/TestCases/LocalVerification.qs
```diff
@@ -1380,8 +1380,6 @@ namespace Microsoft.Quantum.Testing.LocalVerification {
     @Test("QuantumSimulator")
     function ValidTestAttribute10 () : ((Unit)) {}
 
-    function ValidTestAttribute11 (arg : Unit) : Unit { }
```
cesarzc, line 1383: Was this missing the `@Test` attribute before?

cesarzc (approved):

src/QsCompiler/Tests.Compiler/TestCases/LocalVerification.qs
DmitryVasilevsky: It was testing local resource estimation only. I removed the attribute, but left the function in my previous PR.

PR #1620, opened 2023-07-20
Fix metadata warnings for doc builds

Bradben:
Add an ms.author value, and format the date correctly. This should fix the thousands of warnings in the API doc builds every month.


Bradben:
@kuzminrobin - This is a simple fix that should take care of all those build warnings from the docs build every month. I may not have my build environment set up correctly for this repo, though - can you take a look or just add the updates yourself in a different PR? Thanks.

kuzminrobin:
Closing in favor of #1623

PR #1623, opened 2023-07-31
Fix metadata warnings for doc builds.

kuzminrobin:
Fix metadata warnings for doc builds (authored by @Bradben).
Replaces #1620.

### DIFF
--- a/src/Documentation/DocumentationGenerator/DocumentationWriter.cs
+++ b/src/Documentation/DocumentationGenerator/DocumentationWriter.cs
@@ -145,8 +145,10 @@ namespace Microsoft.Quantum.Documentation
                 ["title"] = title,
 
                 // docs.ms metadata
-                ["ms.date"] = DateTime.Today.ToString(),
+                ["ms.date"] = DateTime.Today.ToString("MM/dd/yyyy"),
                 ["ms.topic"] = "managed-reference",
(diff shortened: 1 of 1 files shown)


src/Documentation/DocumentationGenerator/DocumentationWriter.cs
```diff
@@ -145,8 +145,10 @@ public async Task WriteOutput(QsNamespace ns, DocComment docComment)
                 ["title"] = title,
 
                 // docs.ms metadata
-                ["ms.date"] = DateTime.Today.ToString(),
+                ["ms.date"] = DateTime.Today.ToString("MM/dd/yyyy"),
```
ScottCarda-MS, line 148: Does this format need localization?

kuzminrobin: I don't think so. I believe this change is to resolve the Doc warnings like [here](https://github.com/MicrosoftDocs/quantum-api/pull/98), search for "Invalid date format for".

@bradben, please correct me if I'm wrong.

Bradben: Correct. Format can't include the timestamp, only mm/dd/yyyy.

Bradben (approved):

billti (approved):

